PR #34, opened 2019-09-06
Removed compatibility check that Snowflake table has only 2 columns

snagafritz:
We wanted to add a cluster key to our Snowflake table. We were able to accomplish this by pre-creating the Snowflake table with an extra "virtual" column that is computed from the contents of the `RECORD_CONTENT` column, and the changes in this PR which remove the constraint that the table must have only the two `RECORD_METADATA` and `RECORD_CONTENT` columns.

Tested this locally and everything still works. Don't see any reason to require the target table to have only 2 columns.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .idea/
 .settings/
 target/
+*.iml
 *.json
 .project
 .classpath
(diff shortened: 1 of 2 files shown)


binglihub:
fixed in https://github.com/snowflakedb/snowflake-kafka-connector/commit/f42008fc7eacac53e087542d05145f4fb30b0f96

PR #51, opened 2019-10-01
Move Bouncy Castle dependencies to compile scope

ivanyu:
I propose to move the Bouncy Castle dependencies to the default `compile` scope unless there's a rationale behind having them in `provided`. Otherwise, it requires additional effort to set up the connector.


yarmiganosca:
Just ran into this blocking our upgrade to 0.5.0. Hope it gets merged soon!

binglihub:
Since the BC lib conflicts with the one which Confluent package built-in, we are not going to include BC in the JAR file. 
For OSS Kafka user, we will include an instruction in our Doc later.

Issue #54, opened 2019-10-01
Production ready status?


arkady-emelyanov:
Hey! Thanks for supporting kafka-connect platform.

Just wondering, is there any version that I can deploy to production?
After quick evaluation, v0.4.0 fails at some point with non-recoverable error (`failed to drop stage`).

I've tried to use v0.5.0, but, got a `NullPointerException`, was not able to solve problem.
We are using topic2table mapping, so, according to latest changes - it seems there was a bug introduced.

So, going back to original question: is there any version (maybe v0.5.1?) that can be considered stable to be used in production?

sabarigk:
We just tool the 0.5.0 upgrade and seeing a null pointer error. I believe this is the same error mentioned above. This is the stack trace.
java.lang.NullPointerException
        at com.snowflake.kafka.connector.SnowflakeSinkTask.lambda$preCommit$1(SnowflakeSinkTask.java:189)
        at java.util.HashMap.forEach(HashMap.java:1288)
        at com.snowflake.kafka.connector.SnowflakeSinkTask.preCommit(SnowflakeSinkTask.java:186)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.commitOffsets(WorkerSinkTask.java:378)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.closePartitions(WorkerSinkTask.java:590)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.execute(WorkerSinkTask.java:196)
        at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:175)
        at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:219)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)

arkady-emelyanov:
@sabarigk yep, that was our case indeed both on v0.5.0 tag and for master branch:
```java.lang.NullPointerException
at com.snowflake.kafka.connector.SnowflakeSinkTask.lambda$preCommit$1(SnowflakeSinkTask.java:189)
```

PR #57, opened 2019-10-04
Throw ConfigException in parseTopicToTableMap

ivanyu:
`parseTopicToTableMap` currently returns `null` in case of issues with
parsing. This behaviour is non-intuitive and potentially causes bugs
like https://github.com/snowflakedb/snowflake-kafka-connector/issues/54.

`ConfigException` in contrast is special Kafka exception designed for
reporting invalid configuration.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/Utils.java
+++ b/src/main/java/com/snowflake/kafka/connector/Utils.java
@@ -30,6 +30,8 @@ import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.apache.kafka.common.config.ConfigException;
+
 /**
  * Various arbitrary helper functions
  */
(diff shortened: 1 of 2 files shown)


binglihub:
Could you provide a test case which can produce error?
getTopicToTableMap method in https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java handles NULL value.

ivanyu:
Wouldn't it be better for the connector to fail if it sees the provided configuration as incorrect? IMHO ignoring incorrect configuration parts (even with logging) is a bit counter-intuitive and probably users would rather see it failing and fix the config quickly. It might save them hours of debugging.

binglihub:
@ivanyu yes, you are right. We need check it.

binglihub:
fixed in https://github.com/snowflakedb/snowflake-kafka-connector/commit/1d4469ea4fd2a8f16f7b935a9f62f7362b2f605d

PR #97, opened 2020-02-05
Separate PREFIX and FILE NAME when uploading data to support GCP

Mingli-Rui:
SNOW-115446


binglihub:
need double check if the new method caches the data on disk

PR #102, opened 2020-02-28
SNOW-121816 added additional flags to control the amount of metadata

lizihan021:


lizihan021:
How can I do a pre-build check?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -214,6 +224,42 @@ static ConfigDef newConfigDef()
...
+        "Flag to control whether kafka topic name is collected in snowflake metadata",
+        SNOWFLAKE_METADATA_FLAGS,
+        2,
+        ConfigDef.Width.NONE,
+        SNOWFLAKE_METADATA_TOPIC)
+      .define(SNOWFLAKE_METADATA_OFFSET_AND_PARTITION,
+        Type.STRING,
+        SNOWFLAKE_METADATA_DEFAULT,
+        Importance.LOW,
+        "Flag to control whether kafka partition and offset is collected in snowflake metadata",
```
binglihub: kafka partition and offset **are** ....

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -214,6 +224,42 @@ static ConfigDef newConfigDef()
         3,
         ConfigDef.Width.NONE,
         BUFFER_FLUSH_TIME_SEC)
+      .define(SNOWFLAKE_METADATA_ALL,
+        Type.STRING,
+        SNOWFLAKE_METADATA_DEFAULT,
+        Importance.LOW,
+        "Flag to control whether there is metadata collected",
```
binglihub: a little bit confuse.
You can add one more sentence to high it will disable all metadata when turning off.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -135,6 +135,7 @@ public void start(final Map<String, String> parsedConfig)
       .setRecordNumber(bufferCountRecords)
       .setFlushTime(bufferFlushTime)
       .setTopic2TableMap(topic2table)
+      .setMetadataConfig(parsedConfig)
```
binglihub: Do we really need pass all config to Sink? topic2table is extracted from parsedConfig, can we do the same thing on metadata config?

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -68,6 +73,29 @@ public RecordService()
...
+  {
+    if (!metadataConfig.containsKey(key))
+    {
+      return true;
+    }
+    else
+    {
+      return metadataConfig.get(key).equals(SnowflakeSinkConnectorConfig.SNOWFLAKE_METADATA_DEFAULT);
+    }
+  }
```
binglihub: Same thing, if we create a Metadata Config object, we can check it in SinkTask, and we don't need check it here.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
lizihan021: ack

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
lizihan021: ack

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
lizihan021: Yes, that makes sense. I'll change that

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
lizihan021: will change that

src/main/java/com/snowflake/kafka/connector/records/SnowflakeMetadataConfig.java
```diff
@@ -0,0 +1,58 @@
...
+import com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig;
+
+import java.util.Map;
+
+public class SnowflakeMetadataConfig {
+  boolean createtimeFlag = true;
+  boolean topicFlag = true;
+  boolean offsetAndPartitionFlag = true;
+  boolean allFlag = true;
+
```
binglihub: Since all the variables will be initialized when object created and will not be changed, we can make all of them to be **final**. 
You may need two constructors instead of parseMetadataConfig function.
The default constructor doesn't allow any arguments, and sets all the variables to be true.
The second constructor has a Map<String, String> parameter, and initializes all variables according to this config.

src/main/java/com/snowflake/kafka/connector/records/SnowflakeMetadataConfig.java
```diff
@@ -0,0 +1,58 @@
...
+      !config.get(SnowflakeSinkConnectorConfig.SNOWFLAKE_METADATA_ALL)
+        .equals(SnowflakeSinkConnectorConfig.SNOWFLAKE_METADATA_DEFAULT))
+    {
+      allFlag = false;
+    }
+  }
+
+  public String toString()
+  {
+    return
```
binglihub: Why does the toString method return a JSON format record? Will we use it in logging or test only?

src/main/java/com/snowflake/kafka/connector/records/SnowflakeMetadataConfig.java
lizihan021: ack

src/main/java/com/snowflake/kafka/connector/records/SnowflakeMetadataConfig.java
lizihan021: I used that for testing. In the committed code this function is not called, so user won't see this.

binglihub: I see, and it is not a JSON format actually.

PR #115, opened 2020-03-27
SNOW-143637 support hyphen on application name

lizihan021:


src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -404,13 +416,40 @@ static String validateConfig(Map<String, String> config)
...
+   * @return valid application name
+   */
+  public static String convertAppName(Map<String, String> config)
+  {
+    String appName =
+      config.getOrDefault(SnowflakeSinkConnectorConfig.NAME, "");
+    // If appName is empty the following call will throw error
+    String validAppName = generateValidName(appName, new HashMap<String, String>());
+
+    config.put(SnowflakeSinkConnectorConfig.NAME, validAppName);
```
binglihub: If config map was updated, do we need return app name again?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
```diff
@@ -128,13 +128,17 @@ public void start(final Map<String, String> parsedConfig)
...
+
+    // modify invalid connector name
+    connectorName = Utils.convertAppName(config);
 
     // create a persisted connection, and validate snowflake connection
     // config as a side effect
     conn = SnowflakeConnectionServiceFactory
       .builder()
       .setProperties(parsedConfig)
+      .setConnectorName(connectorName)
```
binglihub: if replace _parsedConfig_ by _config_ here, may not need to call setConenctorName() method.

lizihan021: ack

src/main/java/com/snowflake/kafka/connector/Utils.java
lizihan021: ack

PR #120, opened 2020-04-07
SNOW-119851 SMT working with JSON and AVRO input

lizihan021:


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -396,14 +397,24 @@ private void insert(final SinkRecord record)
...
+            newSFContent, record.kafkaOffset(), record.timestamp(), record.timestampType(), record.headers());
         }
+        else
+        {
+          snowflakeRecord = record;
+        }
+
         //broken record
-        else if (((SnowflakeRecordContent) record.value()).isBroken())
+        if (((SnowflakeRecordContent) snowflakeRecord.value()).isBroken())
```
sfc-gh-bli, line 415: if converter can process record, the record should not be broken.

PR #127, opened 2020-04-16
SNOW-152647 Single field config validation

sfc-gh-zli:
Major changes:
- removed configuration related with schema registry, as it is not necessary
- added new versions in helm values
- added single field config validation

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
@@ -21,6 +21,7 @@ import com.snowflake.kafka.connector.internal.SnowflakeConnectionService;
 import com.snowflake.kafka.connector.internal.SnowflakeConnectionServiceFactory;
 import com.snowflake.kafka.connector.internal.SnowflakeErrors;
 import com.snowflake.kafka.connector.internal.SnowflakeTelemetryService;
+import org.apache.kafka.common.config.Config;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.connect.connector.Task;
 import org.apache.kafka.connect.sink.SinkConnector;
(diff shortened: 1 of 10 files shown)


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -218,14 +226,15 @@ static ConfigDef newConfigDef()
...
+        ConfigDef.Range.atLeast(BUFFER_FLUSH_TIME_SEC_MIN),
         Importance.LOW,
         "The time in seconds to flush cached data",
         CONNECTOR_CONFIG,
         3,
         ConfigDef.Width.NONE,
         BUFFER_FLUSH_TIME_SEC)
       .define(SNOWFLAKE_METADATA_ALL,
-        Type.STRING,
+        Type.BOOLEAN,
```
sfc-gh-bli, line 237: is this a behavior change? What does the config file look like after this change?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -234,7 +243,7 @@ static ConfigDef newConfigDef()
         ConfigDef.Width.NONE,
         SNOWFLAKE_METADATA_ALL)
       .define(SNOWFLAKE_METADATA_CREATETIME,
-        Type.STRING,
+        Type.BOOLEAN,
```
sfc-gh-bli, line 246: Same thing here.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -243,7 +252,7 @@ static ConfigDef newConfigDef()
         ConfigDef.Width.NONE,
         SNOWFLAKE_METADATA_CREATETIME)
       .define(SNOWFLAKE_METADATA_TOPIC,
-        Type.STRING,
+        Type.BOOLEAN,
```
sfc-gh-bli, line 255: Same

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -252,7 +261,7 @@ static ConfigDef newConfigDef()
         ConfigDef.Width.NONE,
         SNOWFLAKE_METADATA_TOPIC)
       .define(SNOWFLAKE_METADATA_OFFSET_AND_PARTITION,
-        Type.STRING,
+        Type.BOOLEAN,
```
sfc-gh-bli, line 264: Same

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-zli: I didn't modify the test case for this config, so there shouldn't be behavior changes. But I still don't full understand whether this config type actually has any influence on the config file we get. It seems to me that it is only used for UI stuff... Anyway I'll put this task as low priority for now. To finish this PR, I still have to add cross-field validation and more test cases.

sfc-gh-bli: config from REST is a JSON record. Json has boolean type. Not sure if it is an impact

sfc-gh-zli: I'll add test about this a let you know the result.

sfc-gh-zli: Both string type and boolean type would work in user's config, I verified that with the modification in travis_correct_json_json.json.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
```diff
@@ -249,6 +214,12 @@ public ConfigDef config()
     return SnowflakeSinkConnectorConfig.newConfigDef();
   }
 
+  @Override
+  public Config validate(Map<String, String> connectorConfigs) {
+    // TODO cross-fields validation here
+    return super.validate(connectorConfigs);
```
sfc-gh-bli, line 220: I see, so SinkConnector has a validate function right?

src/test/java/com/snowflake/kafka/connector/ConnectorTest.java
```diff
@@ -0,0 +1,36 @@
...
+      SnowflakeSinkConnectorConfig.BUFFER_SIZE_BYTES_DEFAULT + "");
+    config.put(SnowflakeSinkConnectorConfig.BUFFER_FLUSH_TIME_SEC,
+      SnowflakeSinkConnectorConfig.BUFFER_FLUSH_TIME_SEC_DEFAULT + "");
+    return config;
+  }
+
+  @Test
+  public void testValidate()
+  {
+
```
sfc-gh-bli, line 34: Empty test?

src/test/java/com/snowflake/kafka/connector/SecurityTest.java
```diff
@@ -35,21 +35,6 @@ public void testRSAPasswordOutput() throws IOException,
...
-    Map<String, String> testConf = ConnectorConfigTest.getConfig();
-    testConf.put(SnowflakeSinkConnectorConfig.SCHEMA_REGISTRY_AUTH_CREDENTIALS_SOURCE, "USER_INFO");
-    testConf.put(SnowflakeSinkConnectorConfig.SCHEMA_REGISTRY_AUTH_USER_INFO,
-      testUserName + ":" + testPasswd);
-    Utils.validateConfig(testConf);
-    TestUtils.assertError(SnowflakeErrors.ERROR_0012,
-      () -> new SnowflakeAvroConverter().configure(testConf, false));
-    assert ! searchInLogFile(testPasswd);
-  }
-
```
sfc-gh-bli, line 52: why remove?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
sfc-gh-zli: Yes, correct. We can do cross-fields validation here. I haven't finished this part yet, it is a placeholder.

src/test/java/com/snowflake/kafka/connector/ConnectorTest.java
sfc-gh-zli: I haven't finished implement cross-field validation yet. This test function is a placeholder to test that. I will fill this part shortly with next PR.

src/test/java/com/snowflake/kafka/connector/SecurityTest.java
sfc-gh-zli: This test is not necessary, since everything related with schema registry is removed from connector. "value.converter.basic.auth.credentials.source", "value.converter.basic.auth.user.info" and many configurations defined [here](https://docs.confluent.io/current/schema-registry/connect.html) that starts with "value.converter" or "key.converter" will be directly sent to converters. Connectors don't need to handle those properties. 
Also since SnowflakeSinkConnectorConfig.SCHEMA_REGISTRY_AUTH_USER_INFO is removed, the connector will never have a chance to print the password out.

PR #129, opened 2020-04-17
SNOW-145919 Run perf test only once

sfc-gh-zli:


sfc-gh-zli:
Schedule perf test to be run every day at 6:00 UTC (23:00 pst)

Issue #122, opened 2020-04-15
String Format Decimals Converting To Unknown Codes


naga-kondapaturi-zefr:
We are using SnowflakeAvroConverter with defined schema registry. The connector is decoding these values to unknown string formats when storing data in Snowflake. We are using 1.0.0 (in prod) but the latest on local seems to give the same error. We tried available sql functions within snowflake to decode these values but no use. The data is in an unusable state.

For e.g.
`"0.2700"` in Kafka stored in Snowflake DB as `"\n"`
`"0.3500"` in Kafka stored in Snowflake DB as `"\r¬"`

Schema registry definition:

    {
      "name": "percent_field",
      "type": [
        "null",
        {
          "type": "bytes",
          "scale": 4,
          "precision": 64,
          "connect.version": 1,
          "connect.parameters": {
            "scale": "4"
          },
          "connect.name": "org.apache.kafka.connect.data.Decimal",
          "logicalType": "decimal"
        }
      ],
      "default": null
    },

sfc-gh-zli:
please try to upgrade to 1.4.0, this should have been fixed in https://github.com/snowflakedb/snowflake-kafka-connector/pull/131

mjreiss:
@sfc-gh-zli Is this issue still ongoing? We are currently on version 1.4.1, and all messages are going into snowflake with this unusable format.

drobert:
Looking at the referenced PR (#131) I commented on a few things that look like they may be contributing: https://github.com/snowflakedb/snowflake-kafka-connector/pull/131/files#r463259581

Notably:
- the conversion code casts the byte array to `BigDecimal` without specifying scale or precision
- json Number types don't support arbitrary precision. I would expect only a String representation to be able to represent a BigDecimal accurately

sfc-gh-zli:
This is a bug inside SnowflakeAvroConverter. We internally used `org.apache.avro.generic.GenericDatumReader` to convert AVRO bytes to JSON in SnowflakeAvroConverter, which does not handle byte array and BigDecimal well. I have a walk around:

1. change your `value.converter` from com.snowflake.kafka.connector.records.SnowflakeAvroConverter to io.confluent.connect.avro.AvroConverter.

2. Make sure your Snowflake connector version >= 1.4.2

[...]

PR #131, opened 2020-04-19
SNOW-119851 SMT native converter parse Struct into json

sfc-gh-zli:


src/main/java/com/snowflake/kafka/connector/records/SnowflakeRecordContent.java
```diff
@@ -30,19 +52,161 @@
...
+        case FLOAT64:
+          return JsonNodeFactory.instance.numberNode((Double) value);
+        case BOOLEAN:
+          return JsonNodeFactory.instance.booleanNode((Boolean) value);
+        case STRING:
+          CharSequence charSeq = (CharSequence) value;
+          return JsonNodeFactory.instance.textNode(charSeq.toString());
+        case BYTES:
+          if (Decimal.LOGICAL_NAME.equals(schema.name())) {
+            return JsonNodeFactory.instance.numberNode((BigDecimal) value);
```
drobert, line 133: Very late to the party here, but my team at work is having trouble with avro logical decimal types being stored in snowflake properly (we're seeing what's reported in https://github.com/snowflakedb/snowflake-kafka-connector/issues/122 ) and this looks odd to me. 

Two things look odd:
1. this does a cast from a byte array but does not specify the precision or scale of the logical data type to the resulting BigDecimal
2. json doesn't support arbitrary precision in its numeric type (it's basically a `double`). The only safe way to represent a java BigDecimal in json is to use a string

PR #144, opened 2020-05-12
Fix thread leak

sfc-gh-zli:


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -326,13 +326,14 @@ private void startCleaner()
             {
               Thread.sleep(CLEAN_TIME);
               checkStatus();
+              if (System.currentTimeMillis() - startTime > ONE_HOUR)
+              {
+                sendUsageReport();
+              }
             } catch (InterruptedException e)
             {
               logError("cleaner error:\n{}", e.getMessage());
```
sfc-gh-bli, line 335: change to logInfo or logWarn
Clear terminated by ..., because ...

PR #149, opened 2020-05-18
SNOW-161134 Batch Snowpipe in recovery and backoff when important http call is throttled 

sfc-gh-zli:
Changes:
- SNOW-161134 backoff when important http call is throttled
- SNOW-161149 batch call Snowpipe in recovery
- SNOW-161148 backoff telemetry

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
@@ -14,10 +14,14 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.Map;
 import java.util.Properties;
 import java.util.TimeZone;
+import java.util.function.Function;
+
+import static com.sun.corba.se.impl.activation.ServerMain.logError;
 
 class InternalUtils
 {
@@ -36,6 +40,9 @@ class InternalUtils
 
   private static final Logger LOGGER =
     LoggerFactory.getLogger(InternalUtils.class.getName());
+
+  // backoff with 1, 2, 4, 8 seconds
+  public static final int backoffSec[] = {0, 1, 2, 4, 8};
 
   /**
    * count the size of result set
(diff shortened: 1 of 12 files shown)


src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
```diff
@@ -37,6 +38,9 @@
   private static final Logger LOGGER =
     LoggerFactory.getLogger(InternalUtils.class.getName());
 
+  // backoff with 1, 2, 4, 8 seconds
+  public static final int backoffSec[] = {1, 2, 4, 8};
```
sfc-gh-zli, line 42: backoff time interval is set to 1,2,4,8 seconds, which cumulates to 15 seconds. Kafka has a default poll interval of 30 seconds. So our function should not block longer than 30 seconds, otherwise it will trigger rebalance.

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
```diff
@@ -237,4 +241,63 @@ static IngestedFileStatus convertIngestStatus(IngestStatus status)
...
+        }
+      }
+      Thread.sleep(backoffSec[times] * 1000);
+      times = times + 1;
+    }
+  }
+  public static Object backoffAndRetry(SnowflakeTelemetryService telemetry, backoffFunctionWithParameter runnable,
+                                       Object input) throws Exception
+  {
+    int times = 0;
```
sfc-gh-bli, line 284: We can combine these two functions, a lot duplicate code.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
```diff
@@ -160,13 +188,14 @@ public String getStageName()
...
     while (!startTimeInclusive.equals(endTimeExclusive))
     {
       try
       {
-        response = ingestManager.getHistoryRange(null,
-          startTimeInclusive, endTimeExclusive);
+        response = (HistoryRangeResponse) InternalUtils.backoffAndRetry(telemetry,
+            (Object startTime) ->
+                ingestManager.getHistoryRange(null, (String) startTime, endTimeExclusive), startTimeInclusive
+        );
```
sfc-gh-bli, line 198: Why we need a argument startTime here? Can we change it to
```java
() -> ingestManager.getHistoryRange(null, startTimeInclusive, endTimeExclusive) 
```
?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryServiceV1.java
```diff
@@ -226,7 +237,8 @@ private String getAppName()
     KAFKA_CREATE_STAGE("kafka_create_stage"),
     KAFKA_REUSE_STAGE("kafka_reuse_stage"),
     KAFKA_CREATE_PIPE("kafka_create_pipe"),
-    KAFKA_FILE_FAILED("kafka_file_failed");
+    KAFKA_FILE_FAILED("kafka_file_failed"),
+    SNOWFLAKE_THROTTLE("snowflake_throttle");
```
sfc-gh-bli, line 241: change this to `kafka_snowflake_throttle`

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryServiceV1.java
```diff
@@ -186,6 +186,17 @@ public void reportKafkaFileFailure(final String tableName,
...
   }
 
+  @Override
+  public void reportSnowflakeThrottle(final String errorDetail)
+  {
+    ObjectNode msg = MAPPER.createObjectNode();
+    msg.put(ERROR_NUMBER, errorDetail);
+    msg.put(TIME, System.currentTimeMillis());
+    msg.put(APP_NAME, getAppName());
+
```
sfc-gh-bli, line 196: I think we need include throttle time in this message.

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
sfc-gh-zli: ack

PR #152, opened 2020-05-28
SNOW-163247 remove flusher

sfc-gh-zli:
All changes of v1.4 should be included in this PR.
Major modifications:
- Removed flusher thread and added flush triggers (record number trigger, buffer size trigger, time trigger) to put()
- Moved offset commitment from cleaner to preCommit()
- Removed re-ingest in the recover process, now recover only checks whether table, stage and pipe exist.
- Pressure test and integration test minor modifications.


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkService.java
```diff
@@ -23,11 +23,19 @@
...
+   */
+  void insert(final Collection<SinkRecord> records);
+
   /**
    * call pipe to insert a JSON record
+   * will not trigger time based flush
    * @param record record content
    */
-  void insert(SinkRecord record);
+  void insert(final SinkRecord record);
```
sfc-gh-zli, line 38: Kept two insert method exposed in the interface, the first one will trigger time based flush, the second one won't. Did this for testing.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -79,19 +79,35 @@ public void startTask(final String tableName, final String topic,
   }
 
   @Override
-  public void insert(final SinkRecord record)
+  public void insert(final Collection<SinkRecord> records)
```
sfc-gh-zli, line 82: This function will be triggered periodically with empty collection. Every time that happens, we will check whether we should flush a pipe based on current time and the last time it was flushed.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -253,6 +269,7 @@ private static String getNameIndex(String topic, int partition)
     private final String prefix;
     private long committedOffset; // loaded offset + 1
     private long processedOffset; // processed offset
+    private long previousFlushTimeStamp;
```
sfc-gh-zli, line 272: record the time of previous flush.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -458,6 +435,26 @@ private void insert(final SinkRecord record)
 
     }
 
+    private boolean shouldFlush()
+    {
+      return (System.currentTimeMillis() - this.previousFlushTimeStamp) >= (getFlushTime() * 1000);
```
sfc-gh-zli, line 440: only check time here. I didn't move the buffer size and record number triggered flush to here because this function will be called frequently. And only insert can trigger buffer size and record number based flush.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -473,27 +470,33 @@ private long getOffset()
 
     private void flush(PartitionBuffer buff)
     {
-      if (buff == null || buff.isEmpty())
-      {
-        return;
-      }
+      flusherExecutor.submit(
```
sfc-gh-zli, line 473: detach this thread. Note that this thread does not contain while loop so it won't leak. Async here because in the insert we check all pipes whether they need to flush or not. If it happens that all of them need to be flushed, synced flush will block insert for a long time. Therefore do an async flush.

src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
```diff
@@ -111,11 +112,11 @@ public void testIngestion() throws Exception
       , "test", input.schema(), input.value(), offset);
 
     service.insert(record1);
-    List<String> files = conn.listStage(stage,
-      FileNameUtils.filePrefix(TestUtils.TEST_CONNECTOR_NAME, table,
-        partition));
-
-    assert files.size() == 1;
+    TestUtils.assertWithRetry(() -> conn.listStage(stage, FileNameUtils.filePrefix(TestUtils.TEST_CONNECTOR_NAME,
```
sfc-gh-zli, line 115: Since insert does not synchronically trigger flush anymore, file will not appear on stage immediately after insert return. Therefore added the retry function here.

src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
```diff
@@ -239,24 +243,28 @@ public void testRecordNumber() throws Exception
...
-      TestUtils.tableSize(table) == numOfRecord + numOfRecord1, 30, 6);
+    TestUtils.assertWithRetry(() -> {
+        service.insert(new ArrayList<>()); // trigger time based flush
+        return TestUtils.tableSize(table) == numOfRecord + numOfRecord1;
+      }, 30, 8);
 
     service.closeAll();
   }
 
-  private Future<Integer> insert(SnowflakeSinkService sink, int partition,
```
sfc-gh-zli, line 254: this insert function does not check file number on stage anymore, because flush is async now.

Issue #167, opened 2020-06-05
Connector fails when insertFiles(..) method is invoked with empty set


MThomassen:
In `SnowflakeSinkServiceV1::recoverFileStatues` [loadHistory HTTP method](https://docs.snowflake.com/en/user-guide/data-load-snowpipe-rest-apis.html#endpoint-loadhistoryscan) is invoked and subsequently, a list of to be submitted files in the [insertFiles HTTP method](https://docs.snowflake.com/en/user-guide/data-load-snowpipe-rest-apis.html#endpoint-insertfiles).

This list of files (`filesForIngestion`) could be empty, resulting in an error and irrecoverable Connector Task crash.

HTTP capture:
[failure-http-stream-cleaned.log](https://github.com/snowflakedb/snowflake-kafka-connector/files/4738800/failure-http-stream-cleaned.log)

sfc-gh-zli:
Fixed in https://github.com/snowflakedb/snowflake-kafka-connector/pull/152

PR #168, opened 2020-06-05
Fix (#167); Connector fails when inserting empty list of files

MThomassen:
Checks whether any file should be ingested before invoking insertFiles-HTTP  method

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
@@ -736,9 +736,12 @@ class SnowflakeSinkServiceV1 extends Logging implements SnowflakeSinkService
           }
         }
       );
-      // batch call Snowpipe to ingest file
-      ingestionService.ingestFiles(filesForIngestion);
-      
+
+      if (!filesForIngestion.isEmpty())
+      {
+        // batch call Snowpipe to ingest file
+        ingestionService.ingestFiles(filesForIngestion);
+      }
       if (!loadedFiles.isEmpty())
       {
         purge(loadedFiles);
--- a/src/main/java/com/snowflake/kafka/connector/records/RecordService.java
+++ b/src/main/java/com/snowflake/kafka/connector/records/RecordService.java
@@ -299,7 +299,7 @@ public class RecordService extends Logging
               }
             }
           } else {
-            objectMode = schema.keySchema().type() == Schema.Type.STRING;
+            objectMode = schema.keySchema() == null || schema.keySchema().type() == Schema.Type.STRING;
           }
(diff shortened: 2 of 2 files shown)


src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -299,7 +299,7 @@ else if (value instanceof ByteBuffer)
               }
             }
           } else {
-            objectMode = schema.keySchema().type() == Schema.Type.STRING;
+            objectMode = schema.keySchema() == null || schema.keySchema().type() == Schema.Type.STRING;
```
sfc-gh-zli, line 302: what does it mean to have key schema as null here?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -736,9 +736,12 @@ private void recover()
           }
         }
       );
-      // batch call Snowpipe to ingest file
-      ingestionService.ingestFiles(filesForIngestion);
-      
+
+      if (!filesForIngestion.isEmpty())
```
sfc-gh-zli, line 740: Can you put this check inside ingestionService.ingestFiles?

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
MThomassen: The keySchema here, apparently, is sometimes null in our environment (reason unknown for now..). However, I've reverted this change for now, since it's unrelated to the intended bugfix #167 

Will investigate further and might make a new PR

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
MThomassen: Thanks for the feedback. However, I'm not sure I can agree with your suggestion as I feel this would break a behavioural contract with the *Service API.

_If_ the `SnowflakeIngestionServiceV1` should handle faulty input, it would imho make sense to raise expections on invalid input instead (empty-Set is invalid input here), and subsequently, any invoking client would need to make sure to call the service in a valid way (either by checking themself, as is done here, or in some other form, such as wrapping the invocation with `try..catch..`.

It should be the responsibility of orchestrating code to either invoke or not invoke the service, instead of silent-dropping any such request.

What do you think?

sfc-gh-zli: That makes sense to me. Thanks a lot for your PR. However, we are having a major restructure going on right now, which completely removed this recoverFileStatus function. I'll notify you once that is done.

PR #171, opened 2020-06-13
Finalize smt

sfc-gh-zli:
Keys and Values are now treated as the same. All native connectors are supported for both key and value. Most SMT are supported. The only SMT tested to be not compatible is "RegexRouter". 
Major changes:
- Handle the case when key is not SnowflakeContext.
- When a record is broken, send both key and value as bytes to table stage
- Added test of non-SnowflakeContext key and value 
- Added end2end test of SMT
- Added end2end test for protobuf input


.github/workflows/End2EndFullTest.yml
```diff
@@ -35,6 +35,7 @@ jobs:
         pip3 install --upgrade setuptools
         pip3 install "confluent-kafka[avro]"
         pip3 install avro-python3
+        pip3 install protobuf
```
sfc-gh-bli, line 38: Does this PR include Protebuf converter?

pom_confluent.xml
```diff
@@ -405,7 +405,7 @@
         <dependency>
             <groupId>io.confluent</groupId>
             <artifactId>kafka-schema-registry-client</artifactId>
-            <version>5.3.0</version>
+            <version>5.4.0</version>
```
sfc-gh-bli, line 408: Upgrade to Kafka 5.4?

src/main/java/com/snowflake/kafka/connector/internal/FileNameUtils.java
```diff
@@ -58,23 +59,26 @@ static String fileName(String prefix, long start, long end)
    * @param table table name
    * @param partition partition id
    * @param offset record offset
+   * @param isKey is the broken record a key or a value
    * @return file name
    */
-  static String brokenRecordFileName(String appName, String table, int partition, long offset)
+  static String brokenRecordFileName(String appName, String table, int partition, long offset, boolean isKey)
```
sfc-gh-bli, line 65: If key is broken or value is broken?

sfc-gh-zli: Yes, currently if the record is broken, we only send value as raw bytes. Here I want to send both key and value as raw bytes to table stage, if either key or value is broken. This isKey will tell the generated filename to indicate whether the file contains key bytes or value bytes.

PR #173, opened 2020-06-16
modified to support scanning on feature branch and handle exception

sfc-gh-jfan:


sfc-gh-bli:
@sfc-gh-zli In this PR, we disable the release action, both to Maven and S3 for Whitesource test. Please change it back before next release.

PR #185, opened 2020-07-10
SNOW-174168 buffer size issue

sfc-gh-zli:


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -802,7 +803,7 @@ private void insert(SinkRecord record)
 
         stringBuilder.append(data);
         numOfRecord++;
-        bufferSize += data.length() * 2; //1 char = 2 bytes
+        bufferSize += data.getBytes(StandardCharsets.UTF_8).length; // get buffer size
```
sfc-gh-bli, line 806: Java char is UTF-16 format. That means each char uses at 2 to 4 bytes.

In all UTF formats, character can use at most 4 bytes, the difference is the minimum size.  


|Name|Minimum Size|Maximum size|
|--|--|--|
|UTF-8|1 byte|4 bytes|
|UTF-16|2 bytes|4 bytes|
|UTF-32|4 bytes|4 bytes|

PR #195, opened 2020-07-29
wrap checkStatus and sendUsageReport in try/catch(Throwable t)

deweyjose:
I am seeing cases where the KConnect logs have no "purge {} files from stage: {}" for a number of stages. We also see stranded files in those stages. The number of files can grow to a size where "list <stage>" takes > 10 minutes. This is an issue at KConnect cluster startup time as it triggers max.poll.interval.ms breach in the consumer, which triggers a never ending loop of rebalances.

It looks like it is possible to lose the checkStatus thread which would align with our observations above. There are functions called within checkStatus and sendUsageReports that can throw SnowflakeKafkaConnectorExceptions which are unchecked.

This PR attempts to catch those cases, log and continue running.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
@@ -364,10 +364,18 @@ class SnowflakeSinkServiceV1 extends Logging implements SnowflakeSinkService
             try
             {
               Thread.sleep(CLEAN_TIME);
-              checkStatus();
+              try {
+                checkStatus();
+              } catch (Throwable t) {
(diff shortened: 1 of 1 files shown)


sfc-gh-zli:
Thanks a lot for the PR. When exception happens while cleaning, we don't know how many files are successfully purged. Thus I think we need to scan the stage again for files to be deleted. I'd rather go with this PR: https://github.com/snowflakedb/snowflake-kafka-connector/pull/197 .

deweyjose:
> Thanks a lot for the PR. When exception happens while cleaning, we don't know how many files are successfully purged. Thus I think we need to scan the stage again for files to be deleted. I'd rather go with this PR: #197 .

That makes sense. Thanks again for jumping on this.

PR #196, opened 2020-07-29
SNOW-178013 Fix data losing issue

sfc-gh-zli:
The most important modification is in **SnowflakeSinkTask.java stop() and preCommit()**.
stop():
- do nothing now

preCommit():
- when there is error, return empty map to indicate that offset commitment is not desired

Customer reported that when updating the connector's configuration, there is data loss. Modifications on the e2e test:
- Added a test that will update the connector's configuration every minute (this test reproduced the customer's bug). 
- Moved connector creation from shell script to python
- Added another round of testing to allow not only parallel testing, but also serial testing

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
@@ -155,13 +155,8 @@ public class SnowflakeSinkTask extends SinkTask
   @Override
   public void stop()
   {
-    // long startTime = System.currentTimeMillis();
     LOGGER.info(Logging.logMessage("SnowflakeSinkTask[ID:{}]:stop", this.id));
-
     // do nothing
-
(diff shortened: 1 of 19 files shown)


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -155,15 +155,13 @@ public void start(final Map<String, String> parsedConfig)
...
-    {
-      this.sink.closeAll();
-    }
 
-    LOGGER.info(Logging.logMessage("SnowflakeSinkTask[ID:{}]:stop. Time: {} seconds", this.id,
-      (System.currentTimeMillis() - startTime) / 1000));
+    // do nothing
+
+    // LOGGER.info(Logging.logMessage("SnowflakeSinkTask[ID:{}]:stop. Time: {} seconds", this.id,
+    //   (System.currentTimeMillis() - startTime) / 1000));
```
sfc-gh-bli, line 164: we can remove those commented off codes

PR #197, opened 2020-07-29
SNOW-179311 Fix cleaner being killed by uncaught exceptions

sfc-gh-zli:


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -373,6 +374,22 @@ private void startCleaner()
             {
               logInfo("Cleaner terminated by an interrupt:\n{}", e.getMessage());
               break;
+            } catch (Exception e)
+            {
+              logWarn("Cleaner encountered an exception {}:\n{}\n{}",
+                e.getClass(), e.getMessage(), e.getStackTrace());
+              // list stage again and try to clean the files leaked on stage
+              List<String> tmpCleanerFileNames = conn.listStage(stageName, prefix);
```
deweyjose, line 382: @sfc-gh-zli what happens if conn.listStage throws here? Looks like there's still an opportunity to lose the cleaner thread.

PR #220, opened 2020-09-16
Do not lose precision in RecordService#convertToJson for numbers, that exceed Snowflake's limit of 38 significant digits.

avocader:
Whenever Snowflake connector sinks a record with a header, that has a numeric value, with more than 38 significant numbers, it loses its precision because of `NUMBER` Snowflake datatype. Here is a limitation, described in doc - https://docs.snowflake.com/en/sql-reference/data-types-numeric.html#data-types-for-fixed-point-numbers.

Since the user doesn't want to lose precision, we want to be able to preserve it, even at a cost of sending these values as strings.

This PR updates method `convertToJson` to check, whether the decimal value exceeds Snowflake's limit of 38 significant digits and convert it to text node if it does.


sfc-gh-zli:
merged with #225

PR #235, opened 2020-10-09
SNOW-205602 Added pipeStatus Telemetry Object

sfc-gh-zli:
Added a telemetry objects
**pipeStatus** is send every minute and contains:

1. Lag on Kafka side, on file commitment, and on ingestion
2. File count of files on stage, files being ingesting, files that are purged, and files that are moved to table stage
3. Offset of records that are processed, uploaded, committed and purged.
4. Total record count and data size count
5. Times that the cleaner restarted
6. Start and end time of the recording period


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryServiceV1.java
```diff
@@ -126,10 +136,7 @@ private void send(TelemetryType type, JsonNode data, boolean flush)
     {
       telemetry.addLogToBatch(TelemetryUtil.buildJobData(msg));
       logDebug("sending telemetry data: {}", data.toString());
-      if (flush)
-      {
-        telemetry.sendBatchAsync();
-      }
+      telemetry.sendBatchAsync();
```
sfc-gh-zli, line 139: Removed this flush parameter since we always flush.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -819,6 +854,7 @@ private void close()
         logWarn("Failed to terminate Cleaner or Flusher");
       }
       ingestionService.close();
+      telemetryService.reportKafkaPipeUsage(pipeStatus, true);
```
sfc-gh-zli, line 857: send pipe usage before connector closing

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -405,6 +408,7 @@ private void startCleaner(long recordOffset, SnowflakeTelemetryPipeCreation pipe
           {
             try
             {
+              telemetryService.reportKafkaPipeUsage(pipeStatus, false);
```
sfc-gh-zli, line 411: send pipeStatus every minute

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -287,9 +286,6 @@ private static String getNameIndex(String topic, int partition)
     private List<String> cleanerFileNames;
     private PartitionBuffer buffer;
     private final String prefix;
-    private final AtomicLong committedOffset; // loaded offset + 1
```
sfc-gh-bli, line 290: why removed?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatus.java
```diff
@@ -0,0 +1,181 @@
...
+            this.cleanerRestartCount.get() == 0 &&
+            this.averageKafkaLag.get() == 0 &&
+            this.averageKafkaLagRecordCount.get() == 0 &&
+            this.averageIngestionLag.get() == 0 &&
+            this.averageIngestionLagFileCount.get() == 0 &&
+            this.averageCommitLag.get() == 0 &&
+            this.averageCommitLagFileCount.get() == 0;
+  }
+
+  void dumpTo(ObjectNode msg)
```
sfc-gh-bli, line 146: Seems like `dumpTo` function is a common function of all implementations of `SnowflakeTelemetryBasicInfo`
Can you 
1. change `SnowflakeTelemetryBasicInfo` to `abstract class`
2. move this function to `SnowflakeTelemetryBasicInfo`
3. override `dumpTo` function here

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatus.java
```diff
@@ -0,0 +1,181 @@
...
+import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.node.ObjectNode;
+
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+public class SnowflakeTelemetryPipeStatus extends SnowflakeTelemetryBasicInfo {
+  // Offset info
+  AtomicLong processedOffset;           // processed offset + 1 (offset that is most recent in buffer)
```
sfc-gh-bli, line 12: Do these variables have to be package private?  Can we change them to private?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatus.java
```diff
@@ -0,0 +1,181 @@
...
+      long lagFileCount = averageFileCount.getAndIncrement();
+      // Weighted average to calculate the cumulative average lag
+      averageLag.updateAndGet(value -> (value * lagFileCount + lag) / (lagFileCount + 1));
+    }
+    finally {
+      lagLock.unlock();
+    }
+  }
+
+  boolean empty()
```
sfc-gh-bli, line 125: Is `empty` a common function? if yes, add to `SnowflakeTelemetryBasicInfo`

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatus.java
```diff
@@ -0,0 +1,181 @@
...
+  {
+    return this.processedOffset.get() == 0 &&
```
sfc-gh-bli, line 127: Do we need a lock here?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatus.java
```diff
@@ -0,0 +1,181 @@
...
+  // a lock is required to protect the access
+  private final Lock lagLock;
+
+  AtomicLong startTime;             // start time of the status recording period
+  static final String START_TIME    = "start_time";
+  static final String END_TIME      = "end_time";
+
+  SnowflakeTelemetryPipeStatus(final String tableName, final String stageName, final String pipeName)
+  {
+    this.tableName = tableName;
```
sfc-gh-bli, line 70: seems like some variables can be constant. could you change them to `final`?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryService.java
```diff
@@ -38,6 +38,14 @@
    */
   void reportKafkaFatalError(String errorDetail);
 
+  /**
+   * report connector usage
```
sfc-gh-bli, line 42: If it contains all connector usage, we may need to rethink about the name of `SnowflakeTelemetryPipeStatus`. if it only related to `pipe`, then we need update this comments

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryServiceV1.java
```diff
@@ -100,6 +100,21 @@ public void reportKafkaFatalError(final String errorDetail)
     send(TelemetryType.KAFKA_FATAL_ERROR, msg);
   }
 
+  @Override
+  public void reportKafkaPipeUsage(final SnowflakeTelemetryPipeStatus pipeStatus, boolean isClosing)
+  {
+    if (pipeStatus.empty())
```
sfc-gh-bli, line 106: what if `pipeStatus` is not empty but hasn't changed since last report?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatus.java
sfc-gh-zli: nice catch, will do that

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatus.java
sfc-gh-zli: ack, will change that

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatus.java
sfc-gh-zli: We are directly accessing the variables in other classes, so if changed to private, we need a function to set the value and a function to get the value, which complicate things.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatus.java
sfc-gh-zli: This is only reading from the values and thus does not need a lock.

sfc-gh-bli: what if during reading, some one modifies those values?

sfc-gh-zli: So this function can tolerate data racing issue. It is used to prevent too much empty telemetry sending. 

- If data racing happens, let's say after empty() returns true, then someone changed those values to be none zero. This situation is fine because after one minute we will send telemetry again which will cover the newly change values. 

- The other case is that after empty() returns false, then someone changed those values to be all zeros. This is also fine because we are just sending an all-zero telemetry. Since the pipeStatus object always accurately reflects the status of the pipe in real time, the telemetry, although only contains zero, is accurate as well. 

Thus I think this function does not have to be atomic. We only need to ensure the access to each individual value is atomic.

sfc-gh-bli (approved):

sfc-gh-bli:
approval

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -482,7 +486,7 @@ private void insert(final SinkRecord record)
       }
 
       //ignore ingested files
-      if (record.kafkaOffset() > processedOffset.get())
+      if (record.kafkaOffset() >= pipeStatus.processedOffset.get())
```
sfc-gh-sshankar, line 489: This seems like an unrelated change, how did you catch this ?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -503,11 +507,17 @@ private void insert(final SinkRecord record)
         }
         else
         {
+          // lag telemetry, note that sink record timestamp might be null
+          if (snowflakeRecord.timestamp() != null &&
```
sfc-gh-sshankar, line 511: Can snowflake record be null ?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-zli: When I was writing the pipeStatus comments, I found that processedOffset stores the offset of the most recent processed record. However, flushedOffset and commitOffset stores the offset of the most recent flushed/committed record offset + 1. So I think it's better to keep that consistent and let them all store last offset + 1.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-zli: No. snowflakeRecord is guarantied to be not null.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatus.java
```diff
@@ -0,0 +1,185 @@
...
+
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+public class SnowflakeTelemetryPipeStatus extends SnowflakeTelemetryBasicInfo {
+  // Offset info
+  AtomicLong processedOffset;           // processed offset + 1 (offset that is most recent in buffer)
+  AtomicLong flushedOffset;             // flushed offset + 1 (files on stage)
```
sfc-gh-sshankar, line 13: Why are we storing this with a +1 ? It's confusing. What exactly is committedOffset ?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatus.java
```diff
@@ -0,0 +1,185 @@
...
+import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.node.ObjectNode;
+
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+public class SnowflakeTelemetryPipeStatus extends SnowflakeTelemetryBasicInfo {
+  // Offset info
+  AtomicLong processedOffset;           // processed offset + 1 (offset that is most recent in buffer)
```
sfc-gh-sshankar, line 12: Should this be called nextOffset ?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-sshankar: Also, this means that pipestatus is not just used for telemetry, right ? It contains actual state used during processing ?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -316,14 +315,13 @@ private ServiceContext(String tableName, String stageName,
       this.buffer = new PartitionBuffer();
       this.ingestionService = conn.buildIngestService(stageName, pipeName);
       this.prefix = FileNameUtils.filePrefix(conn.getConnectorName(), tableName, partition);
-      this.processedOffset = new AtomicLong(-1);
```
sfc-gh-sshankar, line 319: I am wondering if it is better to leave this stuff as is and pass these in to pipestatus when you need to ? THoughts @sfc-gh-bli

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-zli: Correct, I moved processedOffset, flushedOffset, and committedOffset to the telemetry object.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatus.java
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  void updateCommitLag(final long lag)
+  {
+    updateLag(lag, averageCommitLagFileCount, averageCommitLag);
+  }
+
+  private void updateLag(final long lag, AtomicLong averageFileCount, AtomicLong averageLag)
+  {
+    lagLock.lock();
```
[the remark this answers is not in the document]
sfc-gh-zli: I think it makes no difference according to this post: https://stackoverflow.com/questions/10868423/lock-lock-before-try 
In Java document's example about ReentrantLock, the lock is outside the try block so we followed that style.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatus.java
sfc-gh-zli: So if we flushed a file that contains offset from 0-99, then flushed offset will be 100. I think you are right, it's better to either change the name of the variable to "next"someOffset or remove the +1 from all the variables to avoid confusion. 
I think I'll go with removing +1 from all offsets, WDYT?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -634,7 +654,7 @@ private long getOffset()
     {
       if (fileNames.isEmpty())
       {
-        return committedOffset.get();
+        return committedOffset.get() + 1;
```
sfc-gh-zli, line 657: This function needs to return the offset that is going to be sent for next round. For example, if we already committed record 4, we should return 5 here, so that the next record we receive will be record  5.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryServiceV1.java
sfc-gh-sshankar: Then why have the parameter ?

sfc-gh-sshankar: Can you remove the parameter ?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-zli: Previously committed and flushed offset stores a + 1. I think we discussed that either we change the naming to nextCommittedOffset or we change committedOffset to store "a" instead of "a + 1". I chose the second option. But this function requires retuning "a+1", that's why I have this plus one added in return.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryServiceV1.java
sfc-gh-zli: Yes, this flush parameter is already removed.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -672,7 +699,10 @@ private void flush(final PartitionBuffer buff)
       String content = buff.getData();
       conn.putWithCache(stageName, fileName, content);
 
-      flushedOffset.set(Math.max(buff.getLastOffset() + 1, flushedOffset.get()));
+      // This is safe and atomic
+      flushedOffset.updateAndGet((value) -> Math.max(buff.getLastOffset() + 1, value));
+      pipeStatus.flushedOffset.set(flushedOffset.get() - 1);
```
sfc-gh-zli, line 704: To make it clear:
processedOffset: a
flushedOffset: a + 1
committedOffset: a + 1

pipeStatus.processedOffset: a
pipeStatus.flushedOffset: a
pipeStatus.committedOffset: a
pipeStatus.purgedOffset: a

sfc-gh-sshankar (approved):
Approve

PR #238, opened 2020-10-14
Tweak Confluent Hub description to be less restrictive in input formats

C0urante:
We had someone ask about whether this connector supports reading Protobuf data since the Confluent Hub description only mentions Avro and JSON. Since it's a Kafka Connect connector it should play nicely with any format as long as the right converter is provided, so we'd like to tweak the language a little to be more permissive and avoid giving the impression that this connector only works with two different types of input format.

### DIFF
--- a/pom_confluent.xml
+++ b/pom_confluent.xml
@@ -96,8 +96,9 @@
                             <description>
                                 Snowflake is the data warehouse built for the
                                 cloud. The Snowflake Kafka connector lets you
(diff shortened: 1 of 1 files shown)


sfc-gh-zli (approved):
LGTM

sfc-gh-zli:
duplicated by https://github.com/snowflakedb/snowflake-kafka-connector/pull/239

PR #240, opened 2020-10-16
SNOW-201517 Added SSL test to reproduce

sfc-gh-zli:
Added folder apache_properties_ssl, which contains all config files necessary to spin up a SASL enabled Kafka cluster.
Added script generate_ssl_key.sh, which generate SSL key on the fly for testing.
Currently this test is only run locally to reproduce SNOW-201517.

### DIFF
--- a/test/apache_properties_ssl/kafka_server_jaas.conf
+++ b/test/apache_properties_ssl/kafka_server_jaas.conf
@@ -1,0 +1,20 @@
+KafkaServer {
+   org.apache.kafka.common.security.plain.PlainLoginModule required
+   username="kafkabroker"
+   password="kafkabroker-secret"
+   user_kafkabroker="kafkabroker-secret"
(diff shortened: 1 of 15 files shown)


test/apache_properties_ssl/kafka_server_jaas.conf
```diff
@@ -0,0 +1,20 @@
...
+Server {
+   org.apache.zookeeper.server.auth.DigestLoginModule required
+   user_super="admin-secret"
+   user_kafka="kafka-secret";
+};
+
+Client {
+   org.apache.zookeeper.server.auth.DigestLoginModule required
+   username="kafka"
+   password="kafka-secret";
```
sfc-gh-bli, line 19: Is it the credential of test server in Docker?

sfc-gh-zli: Test server

sfc-gh-bli (approved):

PR #241, opened 2020-10-17
SNOW-201517 Fixed converter class loader error by upgrading versions

sfc-gh-zli:
Bump up dependency version of the converter. Implemented new APIs in the mock client.


sfc-gh-bli (approved):
LGTM

PR #244, opened 2020-10-21
Fix a minor bug in telemetry

sfc-gh-zli:
Initial value of processed/flushed/committed/purged offset should be set to -1, because the offset stands for the last offset of the record that are at the status. When record with offset 0 is processed, processedOffset will be updated to 0. Therefore initial value should be set to -1 to indicate that no record have been processed.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatus.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatus.java
@@ -71,6 +71,10 @@ public class SnowflakeTelemetryPipeStatus extends SnowflakeTelemetryBasicInfo {
   {
     super(tableName, stageName, pipeName);
 
+    // Initial value of processed/flushed/committed/purged offset should be set to -1,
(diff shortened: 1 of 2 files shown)


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatus.java
```diff
@@ -125,10 +125,10 @@ private void updateLag(final long lag, AtomicLong averageFileCount, AtomicLong a
 
   boolean empty()
   {
-    return this.processedOffset.get() == 0 &&
-            this.flushedOffset.get() == 0 &&
-            this.committedOffset.get() == 0 &&
-            this.purgedOffset.get() == 0 &&
+    return this.processedOffset.get() == -1 &&
```
sfc-gh-bli, line 128: please add a comment here, to explain why it is `-1` but not `0`

sfc-gh-bli (approved):

sfc-gh-sshankar (approved):
Approval

PR #247, opened 2020-10-26
SNOW-197027 Add test that covers SnowflakeSinkTask and SnowflakeSinkConnector

sfc-gh-zli:
Added test that covers the top level object SnowflakeSinkTask and SnowflakeSinkConnector


sfc-gh-bli (approved):
LGTM

PR #249, opened 2020-10-27
SNOW-213553 Fixed flaky test on code coverage

sfc-gh-zli:
Increased the time that the mocked api throws error to make sure cleaner is restarted. Therefor covering code that are in the cleaner restart process.

### DIFF
--- a/src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
+++ b/src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
@@ -769,7 +769,7 @@ public class SinkServiceIT
     System.out.println("recover connection");
     doCallRealMethod().when(spyConn).purgeStage(anyString(), anyList());
 
-    // Total sleep time is 10 minutes to test read ingestHistory
+    // Sleep 4 minutes. Total sleep time is 10 minutes to test read ingestHistory
     Thread.sleep(4 * 60 * 1000);
 
     TestUtils.assertWithRetry(() -> spyConn.listStage(stage, FileNameUtils.filePrefix(TestUtils.TEST_CONNECTOR_NAME,
--- a/src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
+++ b/src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
@@ -763,14 +763,14 @@ public class SinkServiceIT
 
     System.out.println("break connection");
     doThrow(SnowflakeErrors.ERROR_2001.getException()).when(spyConn).purgeStage(anyString(), anyList());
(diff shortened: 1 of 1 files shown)


sfc-gh-bli (approved):
LGTM

src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
```diff
@@ -763,14 +763,14 @@ public void testCleanerRecover() throws Exception
...
+    // Sleep 6 minutes so that cleaner encounters 6 exceptions. Just to make sure cleaner restart is triggered
+    Thread.sleep(6 * 60 * 1000);
 
     System.out.println("recover connection");
     doCallRealMethod().when(spyConn).purgeStage(anyString(), anyList());
 
-    // read ingestHistory
-    Thread.sleep(480000);
+    // Total sleep time is 10 minutes to test read ingestHistory
+    Thread.sleep(4 * 60 * 1000);
```
sfc-gh-sshankar, line 773: Ia thi sten minutes or four minutes

src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
```diff
@@ -763,14 +763,14 @@ public void testCleanerRecover() throws Exception
 
     System.out.println("break connection");
     doThrow(SnowflakeErrors.ERROR_2001.getException()).when(spyConn).purgeStage(anyString(), anyList());
-    // Sleep two minutes so that cleaner encounters two exceptions.
-    Thread.sleep(SnowflakeSinkServiceV1.CLEAN_TIME * 2);
+    // Sleep 6 minutes so that cleaner encounters 6 exceptions. Just to make sure cleaner restart is triggered
+    Thread.sleep(6 * 60 * 1000);
```
sfc-gh-sshankar, line 767: Why are you hardcoding this instead of using CLEAN_TIME

src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
sfc-gh-zli: Here we first sleep 6 minutes with purge() mocked to throw error, then we recover purge() and wait for 4 minutes. So the total sleep time is 10 minutes.

src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
sfc-gh-zli: CLEAN_TIME is currently set to 1 minutes. I found the sleep time here this is not really associated with CLEAN_TIME. We just want to sleep long enough so that mocked broken api is triggered and the total sleep time should be 10 minutes.

PR #253, opened 2020-10-29
SNOW-220064 Modified pom

sfc-gh-zli:
Currently our test is distributed at the following maven life cycle: 
- `validate` 
- `compile`
- `test` 
- `package` - unit test
- `verify` - integrations test
- `install`
- `deploy`

We only need GPG signing when we push the build to Maven central, which uses command `mvn deploy`. Therefore, moving GPG signing from `verify` stage to `install` stage would remove GPG signing during integration test, but reserve GPG signing during `mvn deploy`. 

This modification is used to remove GPG signing error on Jenkins test.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -218,7 +218,7 @@
                 <executions>
                     <execution>
                         <id>sign-artifacts</id>
-                        <phase>verify</phase>
+                        <phase>install</phase>
                         <goals>
                             <goal>sign</goal>
                         </goals>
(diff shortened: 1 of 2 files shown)


sfc-gh-bli (commented):
Does it also sign on the jar when we run
```bash
mvn package
```
?
we also need to sign the artifact sent to Confluent

sfc-gh-zli:
No. I'll change the GPG signing in Confluent POM from current "verify" stage to "package" stage to trigger signing during deployment. Nice catch!

sfc-gh-bli (approved):
LGTM

PR #258, opened 2020-11-10
Split end to end and integration test into different GitHub Action workflow

sfc-gh-zli:
Split end to end and integration test, so that they can run in parallel. This speeds up the checkin process as tests are taking longer to run.


sfc-gh-bli (approved):
LGTM

PR #267, opened 2020-12-17
SNOW-223109 read from secret files

sfc-gh-zli:
If secrets in the config are read from a ConfigFileProvider, we cannot validate the config. Reading file path and do the validation is too complex and might result in permission error. So the simple solution would be to skip validation when passwords or passphrase is from file.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
@@ -249,8 +249,8 @@ public class SnowflakeSinkConnector extends SinkConnector
       }
     }
     // If private key or private key passphrase is provided through file, skip validation
-    if (connectorConfigs.get(Utils.SF_PRIVATE_KEY).contains("${file:") ||
-      connectorConfigs.get(Utils.PRIVATE_KEY_PASSPHRASE).contains("${file:"))
+    if (connectorConfigs.getOrDefault(Utils.SF_PRIVATE_KEY, "").contains("${file:") ||
+      connectorConfigs.getOrDefault(Utils.PRIVATE_KEY_PASSPHRASE, "").contains("${file:"))
       return result;
 
     // We don't validate name, since it is not included in the return value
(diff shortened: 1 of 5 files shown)


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
```diff
@@ -252,6 +248,10 @@ public Config validate(Map<String, String> connectorConfigs)
             ": proxy username and password must be provided together");
       }
     }
+    // If private key or private key passphrase is provided through file, skip validation
+    if (connectorConfigs.getOrDefault(Utils.SF_PRIVATE_KEY, "").contains("${file:") ||
+      connectorConfigs.getOrDefault(Utils.PRIVATE_KEY_PASSPHRASE, "").contains("${file:"))
+      return result;
```
sfc-gh-bli, line 254: I am ok on skipping validation in this case. please make sure, reading from file works.

sfc-gh-bli (approved):

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
sfc-gh-zli: Reading from file is handled by the Kafka Connect framework. I think they should send us the "parsed" configuration in this validation api, but they are not. File references in the config are not properly read & replaces with content in the file. 
But when they start the connector, the config is replaced with the content of the file.

test/rest_request_template/travis_correct_string_json.json
```diff
@@ -6,6 +6,7 @@
     "snowflake.url.name":"SNOWFLAKE_HOST",
     "snowflake.user.name":"SNOWFLAKE_USER",
     "snowflake.private.key":"SNOWFLAKE_PRIVATE_KEY",
+    "snowflake.private.key.passphrase":"${file:./apache_properties/file-secrets.txt:PASSPHRASE}",
```
sfc-gh-zli, line 9: This test demos that reading from file works. If snowflake.private.key.passphrase is not replaced with empty string in the file, then the connector will throw error of not being able to parse the private key with the passphrase.

PR #273, opened 2021-01-26
[SNOW-242554] Calling JDBC's put api with proxy parameters

sfc-gh-japatel:
Passing only proxy parameters in putwith cache instead of in JDBC connection

- Also added a test in internalStageIT
- Decouple SinkTaskIT and proxy tests
- Added few more helper method to reset the JVM parameters after every test
- Added exec maven goal to wipe out the proxy setting. 

### Testing
- Added test in internalStageIT
- Decouple Proxy test in SinkTask to SinkTaskProxy
- Added exec-maven-goal which is used inside proxy end to end test to wipe out the JVM and JDBC config after the test has completed. (Inside clean method)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -251,6 +251,15 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>exec-maven-plugin</artifactId>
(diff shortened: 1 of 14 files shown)


sfc-gh-zli (approved):
LGTM nice job

PR #276, opened 2021-02-03
Add Support for dates without time

mkeshav:


mkeshav:
Can some one pls have a look. Currently the avro converter cant handle dates without time in them.

sfc-gh-japatel:
Do you mind adding a test for it?

PR #282, opened 2021-03-25
Fix handling of schemaless Date objects

C0urante:
The default header converter for Connect is the `SimpleHeaderConverter`, which tries to parse byte array headers into structured data and, when possible, provide a schema for that data.

When parsing maps, it will sometimes be unable to provide a schema for their keys or values if disjoint types are detected. For example, if a map contains the values `34` and `true`, there is no one Connect schema that can cover both of those types (int and boolean). When this happens, it still parses the map and all of its keys and values, but will leave out schema information for the key and/or value type.

If one of those map keys or values gets parsed as a temporal type (date, time, timestamp), this ends up causing issues when the Snowflake connector tries to convert these headers to JSON, since it encounters a value with no schema and for which the `ConnectSchema::schemaType` method returns `null`. See https://github.com/snowflakedb/snowflake-kafka-connector/blob/52bfb720d3e43eeed944e92fbb302752a5c4ae0f/src/main/java/com/snowflake/kafka/connector/records/RecordService.java#L238-L242

This change adds some logic to handle those schemaless `Date` objects as timestamps, and includes a unit test that leverages the `SimpleHeaderConverter` to provide a minimal case where the issue arises.

The `connect-api` dependency is bumped to 2.7.0 in order to leverage improvements in the `SimpleHeaderConverter`'s parsing logic that, unfortunately, make this issue possible in the Snowflake connector.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -287,7 +287,7 @@
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>connect-api</artifactId>
-            <version>2.3.0</version>
+            <version>2.7.0</version>
             <scope>provided</scope>
             <exclusions>
                 <exclusion>
--- a/src/main/java/com/snowflake/kafka/connector/records/RecordService.java
+++ b/src/main/java/com/snowflake/kafka/connector/records/RecordService.java
@@ -235,11 +235,18 @@ public class RecordService extends Logging
     try
     {
       final Schema.Type schemaType;
-      if (schema == null)
-      {
-        schemaType = ConnectSchema.schemaType(value.getClass());
-        if (schemaType == null)
-          throw SnowflakeErrors.ERROR_5015.getException("Java class " + value.getClass() + " does not have corresponding schema type.");
+      if (schema == null) {
+        Schema.Type primitiveType = ConnectSchema.schemaType(value.getClass());
+        if (primitiveType != null) {
+          schemaType = primitiveType;
+        } else {
+          if (value instanceof java.util.Date) {
+            schema = Timestamp.SCHEMA;
+            schemaType = Schema.Type.INT64;
+          } else {
+            throw SnowflakeErrors.ERROR_5015.getException("Java class " + value.getClass() + " does not have corresponding schema type.");
+          }
+        }
       } else {
         schemaType = schema.type();
       }
--- a/src/test/java/com/snowflake/kafka/connector/records/ConverterTest.java
+++ b/src/test/java/com/snowflake/kafka/connector/records/ConverterTest.java
@@ -42,6 +42,7 @@ import org.apache.kafka.connect.data.SchemaAndValue;
 import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.ObjectMapper;
 import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.node.ObjectNode;
 import org.apache.kafka.connect.json.JsonConverter;
+import org.apache.kafka.connect.storage.SimpleHeaderConverter;
 import org.junit.Test;
 
 import java.io.IOException;
@@ -256,6 +257,19 @@ public class ConverterTest
     expected.put("test", new BigDecimal("999999999999999999999999999999999999999"));
     //TODO: uncomment it once KAFKA-10457 is merged
     //assert expected.toString().equals(result.toString());
+  }
+
+  @Test
+  public void testConnectSimpleHeaderConverter_MapDateAndOtherTypes() throws JsonProcessingException {
+    SimpleHeaderConverter headerConverter = new SimpleHeaderConverter();
+    String rawHeader = "{\"f1\": \"1970-03-22\", \"f2\": true}";
+    SchemaAndValue schemaAndValue = headerConverter.toConnectHeader("test", "h1", rawHeader.getBytes(StandardCharsets.US_ASCII));
+    JsonNode result = RecordService.convertToJson(schemaAndValue.schema(), schemaAndValue.value());
+
+    ObjectNode expected = mapper.createObjectNode();
+    expected.put("f1", 6930000000L);
+    expected.put("f2", true);
+    assert expected.toString().equals(result.toString());
   }
 
   @Test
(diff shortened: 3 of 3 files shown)


pom.xml
```diff
@@ -287,7 +287,7 @@
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>connect-api</artifactId>
-            <version>2.3.0</version>
+            <version>2.7.0</version>
```
gharris1727, line 290: Is this api upgrade necessary for this fix? I'm not catching any obvious new functionality that would make a difference here.

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -235,11 +235,18 @@ public static JsonNode convertToJson(Schema schema, Object logicalValue) {
...
-        schemaType = ConnectSchema.schemaType(value.getClass());
-        if (schemaType == null)
-          throw SnowflakeErrors.ERROR_5015.getException("Java class " + value.getClass() + " does not have corresponding schema type.");
+      if (schema == null) {
+        Schema.Type primitiveType = ConnectSchema.schemaType(value.getClass());
+        if (primitiveType != null) {
+          schemaType = primitiveType;
+        } else {
+          if (value instanceof java.util.Date) {
+            schema = Timestamp.SCHEMA;
```
gharris1727, line 244: We also use `java.util.Date` to represent `Date.SCHEMA` and `Time.SCHEMA` values as well.
A mix of schema'd `Time.SCHEMA/Date.SCHEMA` objects with schemaless `java.util.Date` objects would cause this to switch between strings formatted with `TIME_FORMAT/ISO_DATE_FORMAT` and timestamps written as numbers. But without knowing the schema assigned to a different record, the connector will just need to go along with it.

I think this is probably the best option given the circumstances, and is just a caveat for anyone running in the strange mixed schema/schemaless environment.

src/test/java/com/snowflake/kafka/connector/records/ConverterTest.java
```diff
@@ -258,6 +259,19 @@ public void testConnectJsonConverter_MapBigDecimalExceedsMaxPrecision() throws J
     //assert expected.toString().equals(result.toString());
   }
 
+  @Test
+  public void testConnectSimpleHeaderConverter_MapDateAndOtherTypes() throws JsonProcessingException {
+    SimpleHeaderConverter headerConverter = new SimpleHeaderConverter();
+    String rawHeader = "{\"f1\": \"1970-03-22\", \"f2\": true}";
+    SchemaAndValue schemaAndValue = headerConverter.toConnectHeader("test", "h1", rawHeader.getBytes(StandardCharsets.US_ASCII));
+    JsonNode result = RecordService.convertToJson(schemaAndValue.schema(), schemaAndValue.value());
```
gharris1727, line 267: Is this the right place to add a RecordService test? every other test in this file seems to be testing custom avro/json converters, while this is using a stock simple header converter.

gharris1727 (approved):
only nits, LGTM.

pom.xml
C0urante: I tried to call this out in the description:
> The connect-api dependency is bumped to 2.7.0 in order to leverage improvements in the SimpleHeaderConverter's parsing logic that, unfortunately, make this issue possible in the Snowflake connector.

src/test/java/com/snowflake/kafka/connector/records/ConverterTest.java
C0urante: The converters in this test aren't all custom; there are also cases including the `JsonConverter`, which comes out of the box with Connect. And the `SimpleHeaderConverter` is, at this point, anything but simple; the parsing logic in it is at least as complex as the stuff in the `JsonConverter`.

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
C0urante: Yeah--since this is only triggered at all when there isn't any schema associated with the given value, there's no way to distinguish between date, time, and timestamp; I opted to use timestamp in order to avoid losing information.

C0urante:
@sfc-gh-bli @sfc-gh-japatel @sfc-gh-zli is this a fix you'd be interested in?

C0urante:
@sfc-gh-bli @sfc-gh-japatel @sfc-gh-zli ping :)

sfc-gh-zli:
@sfc-gh-japatel Could you please take a look?

sfc-gh-japatel:
> @sfc-gh-bli @sfc-gh-japatel @sfc-gh-zli ping :)

Apologies for the delay, @C0urante will take a look today.

C0urante:
Thanks!

src/test/java/com/snowflake/kafka/connector/records/ConverterTest.java
```diff
@@ -258,6 +259,19 @@ public void testConnectJsonConverter_MapBigDecimalExceedsMaxPrecision() throws J
     //assert expected.toString().equals(result.toString());
   }
 
+  @Test
+  public void testConnectSimpleHeaderConverter_MapDateAndOtherTypes() throws JsonProcessingException {
```
sfc-gh-japatel, line 263: Thank you for adding the test. Overall looks good. 
Can this be a breaking change? 

What does the current  json node look like with `"{\"f1\": \"1970-03-22\", \"f2\": true}";`?

sfc-gh-japatel: @C0urante upon running the test, looks like it never executes the new code path. And the json node doesnt match - failing the test, can you verify?

C0urante: I'm sorry, are you referring to a run of the test with or without the functional changes here?

sfc-gh-japatel: Run of the test. 
I opened one more PR #288 which will run the automated tests.

C0urante: So a run of the test _with_ the functional changes?

C0urante:
I've pushed an update that should hopefully fix the test regardless of which timezone it's in.

sfc-gh-japatel (approved):
Alright, this looks good to me. I tested the UT without the change and it fails with the error. The new tests seems agnostic to machine TZ, so should pass on all machines. 

I will approve this PR and wait for my tests to run on #288 

Once they passes, I will push #288

Thank you much!

PR #287, opened 2021-04-09
Added github actions which runs against GCS storage in Snowflake

sfc-gh-japatel:
- Tested using github actions on local machine. (Using github act)
  - `act -W .github/workflows/End2EndTestConfluentGCS.yml -v -P ubuntu-18.04=nektos/act-environments-ubuntu:18.04`
- Added SECRET passphrase in github.
- This is required to verify performance of future put API changes

Future:
- Will add Stress test after all these new tests passes on PR.


sfc-gh-rramachandran (commented):
Looks like the GCS test run failed, will wait for it to be green before signing off.

sfc-gh-japatel:
> Looks like the GCS test run failed, will wait for it to be green before signing off.

Yes, looks like it is a valid failure. I will need to make a code change to not use putwith cache api if gcs is used.

sfc-gh-japatel:
> > Looks like the GCS test run failed, will wait for it to be green before signing off.
> 
> Yes, looks like it is a valid failure. I will need to make a code change to not use putwith cache api if gcs is used.

I have modified the failed test to match the current scenario of code. i.e use putWithCache for AWS and Azure and use plain put api with connection for GCP. 

IngestStageIT test is the one to look for.

sfc-gh-rramachandran (approved):
LGTM!

sfc-gh-zli (approved):

PR #288, opened 2021-04-15
Fix handling of schemaless Date objects

sfc-gh-japatel:
For running automated Tests 
#282 
CC: @C0urante

### DIFF
--- a/src/test/java/com/snowflake/kafka/connector/records/ConverterTest.java
+++ b/src/test/java/com/snowflake/kafka/connector/records/ConverterTest.java
@@ -45,6 +45,7 @@ import org.apache.kafka.connect.data.SchemaAndValue;
 import org.apache.kafka.connect.data.SchemaBuilder;
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.json.JsonConverter;
+import org.apache.kafka.connect.storage.SimpleHeaderConverter;
 import org.junit.Test;
 
 public class ConverterTest {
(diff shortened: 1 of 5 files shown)


src/test/java/com/snowflake/kafka/connector/records/ConverterTest.java
```diff
@@ -258,6 +259,19 @@ public void testConnectJsonConverter_MapBigDecimalExceedsMaxPrecision() throws J
     //assert expected.toString().equals(result.toString());
   }
 
+  @Test
+  public void testConnectSimpleHeaderConverter_MapDateAndOtherTypes() throws JsonProcessingException {
```
sfc-gh-japatel, line 263: The merge gates are failing because the new code path is not executed.

C0urante: It's executed when I run locally, and when I revert the changes to `RecordService`, the test fails. There may be a difference between my local environment and the testing environment that's contributing to this.

C0urante: Is the testing environment using the newer version of the `connect-api` artifact?

sfc-gh-japatel: Yes, it is using newer version, I think issue is with timestamp conversion. Hence the expected result and actual results dont match. (6940800000 vs 6930000000L)

Should we put some date format?

sfc-gh-japatel: `Timestamp.fromLogical(schema, (java.util.Date) value));`
^ is the place where is it converting the value. May be in test we should explicitly provide Time Zone to avoid this issue or being dependent on the timeformat where the test runs.

sfc-gh-japatel:
Tests have passed, we can push this after one approval @sfc-gh-zli

sfc-gh-zli (approved):
lgtm

PR #291, opened 2021-04-16
Remove travis test

sfc-gh-zli:


sfc-gh-japatel (approved):
lgtm

sfc-gh-bli (approved):

sfc-gh-japatel:
Should we also remove the profile.json.enc?

sfc-gh-zli:
We need to update profile.json.enc as pk rotates. To re-enable all test. Removing the yaml file should be sufficient for stoping Travis test.

sfc-gh-japatel:
> We need to update profile.json.enc as pk rotates. To re-enable all test. Removing the yaml file should be sufficient for stoping Travis test.

I was under impression that profile.json.enc is only used for travis test. We use profile.json.gpg for GH Action tests.

sfc-gh-japatel:
So for us to re-enable all failed tests, we will need to update the gpp in .github/script folder. I will do that in upcoming PR.

PR #292, opened 2021-04-19
Modified profile encrypted files after pub-private keys are changed

sfc-gh-japatel:
- I will also removed the GCS key passphrase and use the same pwd. 
- Azure tests will fail since I dont have admin account. 
- I will change the perf_json gpg in another commit if I get perf_test.json format from either of @sfc-gh-zli or @sfc-gh-bli


sfc-gh-bli:
@sfc-gh-japatel Do we invoke `codcov` from Github Action?

sfc-gh-japatel:
> @sfc-gh-japatel Do we invoke `codcov` from Github Action?

Yes we do, we use the 

- name: Code Coverage
        uses: codecov/codecov-action@v1

PR #294, opened 2021-04-20
Modify profile.json after changing the public key

sfc-gh-japatel:
- Removed the GCS key passphrase and use the same pwd for all three deployments(Cloud). 
- Modified a bit of tests which uses encrypted private keys as JWT authentication for snowpipe and JDBC. 
- Downloading bcfips jars and putting it in local directories as they are required for encrypted keys.


sfc-gh-zli:
We should disable codecov. Otherwise this new key might have been leaked as well..

sfc-gh-japatel:
> We should disable codecov. Otherwise this new key might have been leaked as well..

I dont think we need to, the email the security folks sent us mentioned that if we used bash, we need to stop using it which we dont. And we already have refreshed our credentials.

test/test_verify.py
```diff
@@ -40,7 +40,7 @@ def __init__(self, kafkaAddress, schemaRegistryAddress, kafkaConnectAddress, cre
 
         self.SEND_INTERVAL = 0.01  # send a record every 10 ms
         self.VERIFY_INTERVAL = 60  # verify every 60 secs
-        self.MAX_RETRY = 120  # max wait time 120 mins
+        self.MAX_RETRY = 5  # max wait time 5 mins
```
sfc-gh-zli, line 43: Changing this to 5 minutes might not pass the 3 hours test. We can fix this in later PRs.

sfc-gh-zli (approved):

test/test_verify.py
sfc-gh-japatel: Ahh okay, I can change it back and see if it runs successfully.

PR #300, opened 2021-04-28
GCS Put API Improvements + Refactoring

sfc-gh-japatel:
- Added Java doc wherever I could to make code more readable.
- Added few log statements which will help for debugging.
- Refactoring a bit of code. (Sorry for mixing two, but it was too late when I decided I should have decoupled them)
- To make it easy, here are the changes:
  - SnowflakeMetadataWithExpiration - Added `stageType` and made all fields `private final`. 
  - `getStageType` - Make use of cache since cache has that information
  - putWithCache API had a nested `try catch` -> Decoupled the mono-function to sub-functions based on the jobs and also added more logs which will help debug. `isValidCredential`, `refreshCredential`. 
  - Removed `catch throwable` + `throw t`  which eventually `caught exception`. -> reduced it to just caught exception because Errors would have still be uncaught before. -> No change in behavior 
- Using uploadWithoutConnection API for GCS
  - This will return error if API throws error. 
  - Even if we are calling GS for every file, new uploadWithoutConnection is faster
  - Test 1
    - `[SF_KAFKA_CONNECTOR] Time for putting 50 files in GCS is:16719 ms`
    - `[SF_KAFKA_CONNECTOR] Time for putting 50 files in GCS(UploadWithoutConnection) is:10978 ms`
  - Test 2
    - `[SF_KAFKA_CONNECTOR] Time for putting 50 files in GCS is:16719 ms`
    - `[SF_KAFKA_CONNECTOR] Time for putting 50 files in GCS(UploadWithoutConnection) is:10978 ms`
- Need an ugly way to manipulate put in GCS since there is bug in JDBC:
  - PR in JDBC: https://github.com/snowflakedb/snowflake-jdbc/pull/486/files
- Tested on GCS, AWS, Azure - Especially `InternalStageIT.java`
- I haven't run any End to End test but relying on Github Actions. 

### Reviewers:
Required: @sfc-gh-zli 
Just FYI for @sfc-gh-rramachandran and @sfc-gh-tjones

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
@@ -188,13 +188,16 @@ public interface SnowflakeConnectionService {
    * @param content file content
    * @param stageName stage name
    */
+  @Deprecated
   void put(String stageName, String fileName, String content);
 
   /**
-   * put a file to stage. Cache credential for AWS and Azure storage. Don't cache for GCS.
+   * put a file to stage. Cache credential for AWS, Azure and GCS storage.
    *
-   * @param fileName file name
-   * @param content file content
+   * @param fileName  file name which is of format (stageName/connectorName/tableName/partition/offsetStart_offsetEnd_timestamp)
+   *                  Check {@link FileNameUtils to find out fileName is determined}
+   *
+   * @param content   file content
    * @param stageName stage name
    */
   void putWithCache(final String stageName, final String fileName, final String content);
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
@@ -194,10 +194,10 @@ public interface SnowflakeConnectionService {
   /**
    * put a file to stage. Cache credential for AWS, Azure and GCS storage.
    *
-   * @param fileName  file name which is of format (stageName/connectorName/tableName/partition/offsetStart_offsetEnd_timestamp)
-   *                  Check {@link FileNameUtils to find out fileName is determined}
-   *
-   * @param content   file content
+   * @param fileName file name which is of format
+   *     (stageName/connectorName/tableName/partition/offsetStart_offsetEnd_timestamp)
+   *     <p>Check {@link FileNameUtils to find out how fileName is determined}
+   * @param content file content
    * @param stageName stage name
    */
   void putWithCache(final String stageName, final String fileName, final String content);
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
@@ -11,6 +11,7 @@ import java.sql.SQLException;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Properties;
+import java.util.concurrent.TimeUnit;
 import net.snowflake.client.jdbc.SnowflakeConnectionV1;
 import net.snowflake.client.jdbc.SnowflakeDriver;
 import net.snowflake.client.jdbc.cloud.storage.StageInfo;
@@ -32,6 +33,8 @@ public class SnowflakeConnectionServiceV1 extends Logging implements SnowflakeCo
   private final SnowflakeInternalStage internalStage;
   private StageInfo.StageType stageType;
 
+  private static final long CREDENTIAL_EXPIRY_TIMEOUT_MILLIS = TimeUnit.MINUTES.toMillis(30);
+
   SnowflakeConnectionServiceV1(
       Properties prop,
       SnowflakeURL url,
@@ -49,7 +52,7 @@ public class SnowflakeConnectionServiceV1 extends Logging implements SnowflakeCo
     } catch (SQLException e) {
       throw SnowflakeErrors.ERROR_1001.getException(e);
     }
-    long credentialExpireTimeMillis = 30 * 60 * 1000L;
+    long credentialExpireTimeMillis = CREDENTIAL_EXPIRY_TIMEOUT_MILLIS;
     this.internalStage =
         new SnowflakeInternalStage(
             (SnowflakeConnectionV1) this.conn, credentialExpireTimeMillis, proxyProperties);
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
@@ -1,9 +1,5 @@
 package com.snowflake.kafka.connector.internal;
 
-import net.snowflake.client.jdbc.SnowflakeConnectionV1;
-import net.snowflake.client.jdbc.SnowflakeDriver;
-import net.snowflake.client.jdbc.cloud.storage.StageInfo;
-
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
@@ -15,6 +11,9 @@ import java.sql.SQLException;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Properties;
+import net.snowflake.client.jdbc.SnowflakeConnectionV1;
+import net.snowflake.client.jdbc.SnowflakeDriver;
+import net.snowflake.client.jdbc.cloud.storage.StageInfo;
 
 /**
  * Implementation of Snowflake Connection Service interface which includes all handshake between KC
@@ -50,10 +49,10 @@ public class SnowflakeConnectionServiceV1 extends Logging implements SnowflakeCo
     } catch (SQLException e) {
       throw SnowflakeErrors.ERROR_1001.getException(e);
     }
-    long credentialExpireTime = 30 * 60 * 1000L;
+    long credentialExpireTimeMillis = 30 * 60 * 1000L;
     this.internalStage =
         new SnowflakeInternalStage(
-            (SnowflakeConnectionV1) this.conn, credentialExpireTime, proxyProperties);
+            (SnowflakeConnectionV1) this.conn, credentialExpireTimeMillis, proxyProperties);
     this.telemetry =
         SnowflakeTelemetryServiceFactory.builder(conn)
             .setAppName(this.connectorName)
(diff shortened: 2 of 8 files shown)


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeInternalStage.java
```diff
@@ -12,50 +14,89 @@
...
   // Any operation on the map should be atomic
   private final ConcurrentMap<String, SnowflakeMetadataWithExpiration> storageInfoCache =
       new ConcurrentHashMap<>();
 
-  public static String dummyPutCommandTemplate =
-      "PUT file:///tmp/dummy_location_kakfa_connector_tmp/ @";
+  // GCS Put version requires the dummy command to have filename and entire filePath including
+  // stageName after "@"
+  // For example: PUT file:///fileName @stageName/app/table/partition
+  public static String dummyPutCommandTemplateGCS = "PUT file:///%s @%s/%s";
```
sfc-gh-rramachandran, line 62: Template being different is very tricky and nuanced. Can we add a little bit more details in comments for future readers of the code?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeInternalStage.java
```diff
@@ -65,72 +106,139 @@ public SnowflakeInternalStage(
...
 
-  private boolean isCredentialValid(SnowflakeMetadataWithExpiration credential) {
-    // Kay is cached and not expired
-    return credential != null && System.currentTimeMillis() - credential.timestamp < expirationTime;
+  @VisibleForTesting
+  protected void refreshCredentials(
+      final String stageName, final StageInfo.StageType stageType, final String fullFilePath)
+      throws SnowflakeSQLException {
+    String putCommandToFetchMetadata;
+    if (stageType == StageInfo.StageType.GCS) {
```
sfc-gh-rramachandran, line 211: Can this be a helper method?  We can make this code can be free of stage type/cloud details

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeInternalStage.java
```diff
@@ -65,72 +106,139 @@ public SnowflakeInternalStage(
...
+      throw SnowflakeErrors.ERROR_5018.getException(e.getMessage());
+    }
+
+    // Lets fetch the file transfer metadata from cache again. (Because we might have refreshed
+    // them)
+    SnowflakeFileTransferMetadataV1 fileTransferMetadata =
+        storageInfoCache.get(stageName).fileTransferMetadata;
+    // Set filename to be uploaded
+    // This set is not useful in GCS since there is a bug in JDBC which doesnt use destFileName.
+    fileTransferMetadata.setPresignedUrlFileName(fullFilePath);
```
sfc-gh-rramachandran, line 156: Is there a JIRA to track the JDBC bug? Worth adding that in code

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeInternalStage.java
```diff
@@ -65,72 +106,139 @@ public SnowflakeInternalStage(
...
 
+    byte[] dataBytes = data.getBytes(StandardCharsets.UTF_8);
+    InputStream inStream = new ByteArrayInputStream(dataBytes);
+
+    // This uploadWithoutConnection api cannot handle expired credentials very well.
```
sfc-gh-rramachandran, line 161: What happens if we pass an expired credential? Currently expiration buffer is 30 seconds, should we increase it to 5 minutes to be sure that this doesn't happen?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeInternalStage.java
```diff
@@ -65,72 +106,139 @@ public SnowflakeInternalStage(
...
+    if (fileTransferMetadata.getStageInfo().getStageType() == StageInfo.StageType.LOCAL_FS) {
+      logError(
+          "StageName:{} is not a valid stageType:{}",
+          stageName,
+          fileTransferMetadata.getStageInfo().getStageType());
+      throw SnowflakeErrors.ERROR_5017.getException();
+    } else {
+      // Overwrite the credential to be used
+      SnowflakeMetadataWithExpiration credential =
+          new SnowflakeMetadataWithExpiration(fileTransferMetadata, System.currentTimeMillis());
```
sfc-gh-rramachandran, line 240: Should we skip this block for GCS as we are not using the cache?

src/test/java/com/snowflake/kafka/connector/internal/InternalStageIT.java
```diff
@@ -12,86 +12,123 @@
...
+
+    System.out.println("Same test but with uploadWithoutConnection API");
+    startTime = System.currentTimeMillis();
+    for (int i = 0; i < numberOfFiles; i++) {
+      service.putWithCache(
+          stageNameGCSPutWithCache,
+          "appName/tableName/partition/testCacheFileName" + i,
+          "Any data");
     }
+    System.out.println(
```
sfc-gh-rramachandran, line 116: How long does putWithCache take vs put for GCS? How does it stack up with other clouds? Can we add the expected range as a comment?

sfc-gh-rramachandran (commented):
Left some comments, changes mostly look good.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
```diff
@@ -188,12 +188,15 @@
    * @param content file content
    * @param stageName stage name
    */
+  @Deprecated
   void put(String stageName, String fileName, String content);
```
sfc-gh-tjones, line 192: Is it possible to remove this entirely?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -49,10 +49,10 @@
     } catch (SQLException e) {
       throw SnowflakeErrors.ERROR_1001.getException(e);
     }
-    long credentialExpireTime = 30 * 60 * 1000L;
+    long credentialExpireTimeMillis = 30 * 60 * 1000L;
```
sfc-gh-tjones, line 52: Nit, might be able to use `TimeUnit` here to make this more readable.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
sfc-gh-japatel: Yes, we wont need this. Will remove. I kept it to see how old and new one performed in terms of latency.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeInternalStage.java
sfc-gh-japatel: Will add more details. (I have an example already but I see how it can be confusing)

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeInternalStage.java
sfc-gh-japatel: We don't have a jira but have a bug assigned (Linked in Description) Will create a JIRA.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeInternalStage.java
sfc-gh-japatel: Technically here it how I see this affecting customers:
- If we pass in expired credentials, we will get expired credentials error from cloud. 
- JDBC itself should renew the token but looks like that part of the code is not working. https://github.com/snowflakedb/snowflake-jdbc/blob/master/src/main/java/net/snowflake/client/jdbc/cloud/storage/SnowflakeS3Client.java#L738
- We should be doing a retry for this failure and renew the credentials from our end. 

Worst case we would put this file inside a table stage. (Which is considered as data loss)

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeInternalStage.java
sfc-gh-japatel: yes, makes sense. lemme decouple.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeInternalStage.java
sfc-gh-japatel: I dont think we can skip this, after refresh credentials call above in putWithCache, we fetch the credentials from this `storageInfoCache` so as to make a call `uploadWithoutConnection`

src/test/java/com/snowflake/kafka/connector/internal/InternalStageIT.java
sfc-gh-japatel: I found uploadWithoutConnection to be overall faster compared to put. Will comments. For other clouds it will be faster since we cache the creds. (Only one call in 30 mins + expired credentials exception)

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
sfc-gh-japatel: Will remove in future changes.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeInternalStage.java
sfc-gh-japatel: https://snowflakecomputing.atlassian.net/browse/SNOW-350676

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeInternalStage.java
sfc-gh-rramachandran: Makes sense

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeInternalStage.java
sfc-gh-rramachandran: Thanks for the explanation, this makes sense. Can you add this in comment for future reference?
Also are we currently retrying for this failure? If not, we can decouple it from this work but we should track it

sfc-gh-rramachandran (approved):
LGTM!

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeInternalStage.java
sfc-gh-japatel: Yes, lemme copy paste this in docs and we are retrying it. But we dont have enough logs to understand if retry is successful or not. (So added retry logging in this PR as well)

Also, logger level in this exceptions are warn and after max retries, it is error.

sfc-gh-zli (approved):
lgtm

PR #302, opened 2021-04-30
Add thread safety around simple data format

sfc-gh-japatel:
Added test

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/records/RecordService.java
+++ b/src/main/java/com/snowflake/kafka/connector/records/RecordService.java
@@ -50,14 +50,20 @@ public class RecordService extends Logging {
   private static final String KEY_SCHEMA_ID = "key_schema_id";
   static final String HEADERS = "headers";
 
-  public static final SimpleDateFormat ISO_DATE_FORMAT =
-      new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
-  public static final SimpleDateFormat TIME_FORMAT = new SimpleDateFormat("HH:mm:ss.SSSZ");
-  static final int MAX_SNOWFLAKE_NUMBER_PRECISION = 38;
+  // For each task, we require a separate instance of SimpleDataFormat, since they are not
+  // inherently thread safe
+  static final ThreadLocal<SimpleDateFormat> ISO_DATE_TIME_FORMAT =
+      ThreadLocal.withInitial(
+          () -> {
+            SimpleDateFormat simpleDateFormat =
+                new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+            simpleDateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+            return simpleDateFormat;
+          });
 
-  static {
-    ISO_DATE_FORMAT.setTimeZone(TimeZone.getTimeZone("UTC"));
-  }
+  public static final ThreadLocal<SimpleDateFormat> TIME_FORMAT =
+      ThreadLocal.withInitial(() -> new SimpleDateFormat("HH:mm:ss.SSSZ"));
+  static final int MAX_SNOWFLAKE_NUMBER_PRECISION = 38;
 
   // This class is designed to work with empty metadata config map
   private SnowflakeMetadataConfig metadataConfig = new SnowflakeMetadataConfig();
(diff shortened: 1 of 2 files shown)


src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -50,14 +50,20 @@
...
+                new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+            simpleDateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+            return simpleDateFormat;
+          });
 
-  static {
-    ISO_DATE_FORMAT.setTimeZone(TimeZone.getTimeZone("UTC"));
-  }
+  public static final ThreadLocal<SimpleDateFormat> TIME_FORMAT =
+      ThreadLocal.withInitial(() -> new SimpleDateFormat("HH:mm:ss.SSSZ"));
```
sfc-gh-japatel, line 65: Time_format doesnt have timezone set. Should we do this behavior change and be uniform?

sfc-gh-japatel: CC: @sfc-gh-zli

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -50,14 +50,20 @@
   private static final String KEY_SCHEMA_ID = "key_schema_id";
   static final String HEADERS = "headers";
 
-  public static final SimpleDateFormat ISO_DATE_FORMAT =
-      new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
-  public static final SimpleDateFormat TIME_FORMAT = new SimpleDateFormat("HH:mm:ss.SSSZ");
-  static final int MAX_SNOWFLAKE_NUMBER_PRECISION = 38;
+  // For each task, we require a separate instance of SimpleDataFormat, since they are not
+  // inherently thread safe
+  private static final ThreadLocal<SimpleDateFormat> ISO_DATE_FORMAT =
```
sfc-gh-rramachandran, line 55: Can we call this ISO_DATE_TIME_FORMAT as it also includes time information?

src/test/java/com/snowflake/kafka/connector/records/ConverterTest.java
```diff
@@ -57,6 +68,27 @@
 
   private static final String TEST_TOPIC = "test";
 
+  // For each task, we require a separate instance of SimpleDataFormat, since they are not
+  // inherently thread safe
+  private static final ThreadLocal<SimpleDateFormat> ISO_DATE_FORMAT_THREAD_SAFE =
```
sfc-gh-rramachandran, line 73: Instead of re-declaring, can you make the record service member public static (looks like it was public static before) and use it in the test?

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-rramachandran: Can you explain the customer facing impact of this behavior change?

src/test/java/com/snowflake/kafka/connector/records/ConverterTest.java
```diff
@@ -232,6 +264,47 @@ public void testConnectJsonConverter_MapInt64() throws JsonProcessingException {
...
+    String value =
+        "{ \"schema\": { \"type\": \"int32\", \"name\": \"org.apache.kafka.connect.data.Date\","
+            + " \"version\": 1 }, \"payload\": 10000 }";
+    SchemaAndValue schemaInputValue = jsonConverter.toConnectData("test", value.getBytes());
+
+    JsonNode result =
+        RecordService.convertToJson(schemaInputValue.schema(), schemaInputValue.value());
+    System.out.println("Record Service result:" + result + " Thread :" + Thread.currentThread());
+
+    String a = ISO_DATE_FORMAT_THREAD_SAFE.get().format(CALENDAR_THREAD_SAFE.get().getTime());
```
sfc-gh-rramachandran, line 304: a more descriptive variable name please :)

sfc-gh-rramachandran (approved):
LGTM! Left some minor comments

src/test/java/com/snowflake/kafka/connector/records/ConverterTest.java
sfc-gh-japatel: Oops, will do!

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-japatel: Will do!

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-japatel: My understanding is that, this `time_format` has the timezone which is dependent on worker node timezone which usually is UTC but it is not explicit.

sfc-gh-japatel: (We don't have to do this behavior change, I am just starting this conversation here as it feels more related)

src/test/java/com/snowflake/kafka/connector/records/ConverterTest.java
sfc-gh-japatel: Ahh yes. Will do..

PR #304, opened 2021-05-04
Added test to verify behavior of tombstone records

sfc-gh-japatel:
SNOW-298877

Adding a test to repro behavior of tombstone records. 

Tombstone records are those records whose key is non null but value(Payload) is null. 

- We have decided that we would do a breaking change and **not** put an empty JSON(`{}`) inside table. 
- We can use SMT to filter empty records. 
  - https://kafka.apache.org/documentation/#connect_transforms

### DIFF
--- a/src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
+++ b/src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
@@ -13,8 +13,10 @@ import io.confluent.kafka.schemaregistry.client.MockSchemaRegistryClient;
 import io.confluent.kafka.schemaregistry.client.SchemaRegistryClient;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
+import java.sql.ResultSet;
 import java.util.*;
 import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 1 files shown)


sfc-gh-zli (approved):
lgtm

src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
```diff
@@ -170,6 +172,68 @@ public void testIngestion() throws Exception {
...
+    LinkedList<String> contentResult = new LinkedList<>();
+    while (resultSet.next()) {
+      contentResult.add(resultSet.getString("RECORD_CONTENT"));
+    }
+    resultSet.close();
+
+    assert contentResult.size() == 1;
+
+    ObjectNode emptyNode = MAPPER.createObjectNode();
+    assert contentResult.get(0).equalsIgnoreCase(emptyNode.toString());
```
sfc-gh-rramachandran, line 227: assert that there is only one result too maybe?

sfc-gh-rramachandran (approved):
LGTM!

src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
sfc-gh-japatel: line 224

PR #307, opened 2021-05-11
Bumping up jdbc version to which fixes uploadwithoutconnection respon…

sfc-gh-japatel:
…ses in case of errors

Based on:
- https://github.com/snowflakedb/snowflake-jdbc/blob/master/CHANGELOG.rst there are no significant changes which stops us from not upgrading to latest version.
- We are interested in changes brought in by JDBC Driver 3.12.16
  - Which throws a `SnowflakeSQLLoggedException` in case of RestRequest which didnt return a successful response. 

This PR will also be used to test whether new JDBC passes existing KC tests. 

More testing on local machine:
- I tried to setup datagen connector from confluent and loaded continuous data. 
- Manually created an exception while puting file in GCS but it didnt warn Kafka connector. 
- This gave out Connector a false indication that put succeeded and hence we continued our snowpipe call. Naturally, snowpipe failed since JDBC was never successful. (404 file not found exception)
[jdbc_log_upload_failure.log](https://github.com/snowflakedb/snowflake-kafka-connector/files/6455828/jdbc_log_upload_failure.log)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -399,7 +399,7 @@
         <dependency>
             <groupId>net.snowflake</groupId>
             <artifactId>snowflake-jdbc</artifactId>
-            <version>3.12.12</version>
+            <version>3.13.3</version>
         </dependency>
 
         <!--junit for unit test-->


sfc-gh-rramachandran (approved):
LGTM

PR #309, opened 2021-05-14
Fix perf test

sfc-gh-japatel:
- Also temporarily modified perf yaml to run on all PRs and pushes to
  master. (Will fix when I see them passing)
- Fix JDBC version
- As recommended here: https://stackoverflow.com/questions/43488853/confluent-maven-repository-not-working

Test:
Check the result here inside sfctest0 account for evaluating perf.
```
select to_timestamp(kafka_perf_test_result:time::string) as testTime, kafka_perf_test_result:test_suites[2] as avro_with_schema_registry,avro_with_schema_registry:test_cases[0].time as time_sec, * from DO_NOT_DELETE_KAFKA_PERF_TEST_RESULT_TABLE order by testTime desc;

show tables like 'KAFKA_PERF_TEST';

### CODE REVIEW

File: .github/workflows/PerfTest.yml [hunk 1]
```diff
@@ -1,8 +1,10 @@
 name: Kafka Connector Performance Test
 
 on:
-  schedule:
-    - cron:  '0 6 * * *'
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: '**'
```
sfc-gh-rramachandran, line 7: How long does the test take? I am wondering if we can keep this for good if it takes < 30 minutes

test/perf_test/scripts/run.sh
```diff
@@ -34,7 +34,7 @@ popd
 
 export CONFLUENT_HOME="$c_dir/confluent"
 #run test
-mvn clean compile
+mvn clean compile -U
```
sfc-gh-rramachandran, line 37: Out of curiosity: What does -U do?

sfc-gh-rramachandran (approved):
LGTM!

.github/workflows/PerfTest.yml
sfc-gh-japatel: (Honestly, I am running this for first time.)
Creating PRs is my goto way to run GH Actions and figure out if code is working or not. (Honestly, that is the fastest way right now to test anything)

sfc-gh-japatel: But, I will keep this temporarily and once they start running, I will revert this back to cron expression

test/perf_test/scripts/run.sh
sfc-gh-japatel: -U is for force downloading the repositories instead of using it from cache. https://stackoverflow.com/questions/43488853/confluent-maven-repository-not-working (Added in PR description as well)

.github/workflows/PerfTest.yml
sfc-gh-rramachandran: That is sad :(. Maybe we can add a job for this in Jenkins so that we can schedule this adhoc? Can you file a JIRA, I might have bandwidth in the next couple of weeks to pick it up

sfc-gh-japatel: I did dig a bit this morning, but need to spend more time. There seem to exist Tests repo already which has this perf_test code. https://github.com/snowflakedb/Tests
And there is a Jenkins associated but it seems it is dormant. 
https://jenkins.int.snowflakecomputing.com/job/CLIENTS_PERF/

Also, I think you can run this add hoc from github action.

sfc-gh-japatel: Example: https://github.com/snowflakedb/snowflake-kafka-connector/actions/runs/843190933
Just click re-run.

sfc-gh-japatel: The test have passed after some modifications and play around. I will push this and send another PR to revert to cron job. (It takes around 2 hours to complete it)

sfc-gh-zli: Yes please change it to a scheduled daily run before checking in. Blocking PRs for 2 hours is not good.

sfc-gh-japatel: Changed

Issue #290, opened 2021-04-16
Avro converter does not support reader schema


anekdoti:
It is a very common pattern that a Kafka topic contains messages with different Avro schemas. A consumer chooses an Avro schema that is compatible to each of the schemas occuring in the topic and that is used for deserialization. This way, the consumer consumes each message in the topic in the schema that the consumer expects.

This is a strategy that is supported (and even the default) for the Avro deserializers provided by Confluent (see [here](https://github.com/confluentinc/schema-registry/blob/master/avro-serializer/src/main/java/io/confluent/kafka/serializers/KafkaAvroDeserializer.java)).

The Avro converter of the Snowflake Kafka connector does not support this. It outputs each message value as a variant record using the schema that was used upon serialization. This leaves the responsibility for converting messages with different schemas to a common schema to a later processing stage, e.g., a task consuming from a change data stream on the sink table. 

An option to provide a reader schema to the Avro converter (either as a string or as the id of a schema in the schema registry), would make it possible to do this projection from writer to reader schema in the deserializer (where it is done in other Avro deserializers as well using the rules for schema compatibility defined for Avro), instead of having to replicate this conversion using SQL later on.

anekdoti:
Thanks four your interest!

In the [Snowflake Avro deserializer](https://github.com/snowflakedb/snowflake-kafka-connector/blob/28fdc3779c145aa6e078017d7ee6c2b8086ecefc/src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java#L180) a GenericDatumReader is instantiated, and it is provided with the writer schema and a reader schema. In the current implementation, both are chosen to be the same, i.e., the writer schema as obtained from the schema registry. But looking at this line, I think it would be quite straigthforward to choose another reader schema - it would be mostly a decision about how to provide it to the deserializer - by a schema id or as a string?

Would it be helpful if I would create a PR for such an extension?

PR #312, opened 2021-05-21
Avro converter supports a fixed reader schema

anekdoti:
This pull requests addresses #290. The issue addresses the problem that schema evolution in a Kafka topic leads to `RECORD_CONTENT` elements in the staging table with different schemas. 

While in many cases  this might be desirable (e.g., in a data lake of semi-structured data), it makes downstream processing of the staging table difficult, e.g., when populating another table from it using Snowflake streams and tasks. 

For such cases, it would be great to have an option to interpret all messages in the Kafka topic as having the same schema, so that the `RECORD_CONTENT` elements in the staging table have the same structure. 

The Avro deserializer provided by the Avro project, which is also used in the `SnowflakeAvroConverter` internally, supports a writer schema (the schema in which a message got serialized), and a reader schema (a schema that has to be compatible to the writer schema according to [certain rules](https://avro.apache.org/docs/1.10.2/spec.html#Schema+Resolution), which does exactly this. Until now, the Snowflake Kafka connector does not allow to set a reader schema.

The PR adds a configuration property `reader.schema` to the `SnowflakeAvroConverter.` Using this reader schema, all messages in a Kafka topic, independent of their individual writer schema, can be interpreted as having the same schema (provided the reader schema is compatible to the writer schemas). 

If the property is not set, the `SnowflakeAvroConverter` works as usual, i.e., by deserializing each consumed message (value or key) with the writer schema that is referenced in the message. 

If the property is set and can be parsed as an Avro schema, the message is deserialized into the provided reader schema. If the reader schema is not compatible to the writer schema, an error is raised on deserialization.

If the property is set but can not be parsed as an Avro schema, an error is signalled on startup of the connector.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java
+++ b/src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java
@@ -41,11 +41,12 @@ public class SnowflakeAvroConverter extends SnowflakeConverter {
   private SchemaRegistryClient schemaRegistry = null;
 
   public static final String BREAK_ON_SCHEMA_REGISTRY_ERROR = "break.on.schema.registry.error";
-  // By default, we don't break when schema registry is not found
-  private boolean breakOnSchemaRegistryError = false;
-
   public static final String READER_SCHEMA = "reader.schema";
 
+  // By default, we don't break when schema registry is not found
+  private boolean breakOnSchemaRegistryError = false;
+  /* By default, no reader schema is set. In this case, the writer schema of each item is also used
+  as the reader schema. See https://avro.apache.org/docs/1.9.2/spec.html#Schema+Resolution */
   private Schema readerSchema = null;
 
   @Override
@@ -75,6 +76,11 @@ public class SnowflakeAvroConverter extends SnowflakeConverter {
     }
   }
 
+  /**
+   * get reader schema from config if provided
+   *
+   * @param configs configuration for converter
+   */
   void readReaderSchema(final Map<String, ?> configs)
   {
     Object readerSchemaFromConfig = configs.get(READER_SCHEMA);
@@ -92,11 +98,13 @@ public class SnowflakeAvroConverter extends SnowflakeConverter {
       }
       catch (SchemaParseException e)
       {
+        LOGGER.error(Logging.logMessage("the string provided for reader.schema is no valid Avro schema: " + e.getMessage()));
         throw SnowflakeErrors.ERROR_0024.getException(e);
       }
     }
     else
     {
+      LOGGER.error(Logging.logMessage("reader.schema has to be a string"));
       throw SnowflakeErrors.ERROR_0024.getException();
     }
   }
@@ -139,11 +147,11 @@ public class SnowflakeAvroConverter extends SnowflakeConverter {
       return logErrorAndReturnBrokenRecord(e, bytes);
     }
 
-    // If there is any error while getting schema from schema registry,
+    // If there is any error while getting writer schema from schema registry,
     // throw error and break the connector
-    Schema schema;
+    Schema writerSchema;
     try {
-      schema = schemaRegistry.getById(id);
+      writerSchema = schemaRegistry.getById(id);
     } catch (Exception e) {
       if (breakOnSchemaRegistryError) {
         throw SnowflakeErrors.ERROR_0011.getException(e);
@@ -159,7 +167,12 @@ public class SnowflakeAvroConverter extends SnowflakeConverter {
 
       return new SchemaAndValue(
           new SnowflakeJsonSchema(),
-          new SnowflakeRecordContent(parseAvroWithSchema(data, schema), id));
+          new SnowflakeRecordContent(
+                  parseAvroWithSchema(
+                    data,
+                    writerSchema,
+                    readerSchema == null ? writerSchema : readerSchema),
+                  id));
     } catch (Exception e) {
       return logErrorAndReturnBrokenRecord(e, bytes);
     }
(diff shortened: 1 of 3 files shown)


src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java
```diff
@@ -69,6 +75,32 @@ void readBreakOnSchemaRegistryError(final Map<String, ?> configs) {
...
+
+    if (readerSchemaFromConfig instanceof String)
+    {
+      try
+      {
+        readerSchema = new Schema.Parser().parse((String) readerSchemaFromConfig);
+      }
+      catch (SchemaParseException e)
+      {
+        throw SnowflakeErrors.ERROR_0024.getException(e);
```
sfc-gh-japatel, line 95: Can we add log line here?

src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java
```diff
@@ -69,6 +75,32 @@ void readBreakOnSchemaRegistryError(final Map<String, ?> configs) {
...
+      }
+    }
+    else
+    {
+      throw SnowflakeErrors.ERROR_0024.getException();
```
sfc-gh-japatel, line 100: Log line here.

src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java
```diff
@@ -153,7 +185,10 @@ private JsonNode parseAvroWithSchema(final byte[] data, Schema schema) {
 
     InputStream is = new ByteArrayInputStream(data);
     Decoder decoder = DecoderFactory.get().binaryDecoder(is, null);
-    DatumReader<GenericRecord> reader = new GenericDatumReader<>(schema, schema, genericData);
+    DatumReader<GenericRecord> reader = new GenericDatumReader<>(
+            schema,
```
sfc-gh-japatel, line 189: Can we name this writer schema?

src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java
```diff
@@ -43,9 +44,14 @@
   // By default, we don't break when schema registry is not found
   private boolean breakOnSchemaRegistryError = false;
 
+  public static final String READER_SCHEMA = "reader.schema";
```
sfc-gh-japatel, line 47: Let's club this together with other public static.

sfc-gh-japatel: Also, doc comments please.

src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java
```diff
@@ -69,6 +75,32 @@ void readBreakOnSchemaRegistryError(final Map<String, ?> configs) {
     }
   }
 
+  void readReaderSchema(final Map<String, ?> configs)
```
sfc-gh-japatel, line 78: Can we add comment here? (I know it is self explanatory but going through docstrings is much faster than through code.

src/test/java/com/snowflake/kafka/connector/records/ConverterTest.java
```diff
@@ -152,6 +154,32 @@ public void testAvroWithSchemaRegistry() throws IOException {
...
+    converter.setSchemaRegistry(client);
+
+    SchemaAndValue input = converter.toConnectData("test", client.getData());
+    SnowflakeRecordContent content = (SnowflakeRecordContent) input.value();
+    assert content.getData().length == 1;
+    assert content.getData()[0].asText().equals(mapper.readTree("{\"int" +
+            "\":1234, \"newfield\":1}").asText());
+
+    //null value
+    input = converter.toConnectData("test",null);
```
sfc-gh-japatel, line 179: I suppose this is the current behavior?

sfc-gh-japatel (changes requested):
Very minor changes. 

At the moment we don't allow merging codepaths outside Snowflake, so I will clone this and then push it once I approve this. I will link this PR as well. 

Overall looks good, thanks for doing this.

src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java
anekdoti: Sure!

src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java
anekdoti: Ok!

src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java
anekdoti: Done

src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java
anekdoti: Done

src/test/java/com/snowflake/kafka/connector/records/ConverterTest.java
anekdoti: Yes, I followed the same structure in the test as in the test without a reader schema. In principle, the null case could be removed here. Should I?

src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java
```diff
@@ -92,11 +98,13 @@ void readReaderSchema(final Map<String, ?> configs)
       }
       catch (SchemaParseException e)
       {
+        LOGGER.error(Logging.logMessage("the string provided for reader.schema is no valid Avro schema: " + e.getMessage()));
```
sfc-gh-japatel, line 101: thank you!

src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java
```diff
@@ -172,23 +185,22 @@ private SchemaAndValue logErrorAndReturnBrokenRecord(final Exception e, final by
   }
 
   /**
-   * Parse Avro record with schema
+   * Parse Avro record with a writer schema and a reader schema. The writer and the reader schema
+   * have to be compatible as described in https://avro.apache.org/docs/1.9.2/spec.html#Schema+Resolution
```
sfc-gh-japatel, line 189: Thank you, looks good!

sfc-gh-japatel:
@cla-bot

sfc-gh-japatel:
Closing this merged #315 since the CLA was signed. Thank you for the patience @anekdoti

PR #313, opened 2021-05-27
Added provider config which is fed into userAgentSuffix for snowpipe for monitoring purpose

sfc-gh-japatel:
Testing:
- I have locally tested this by overriding the snowpipe sdk jar. (Which is not yet released) Reason being I want to get an approval on this one so that I can release the SDK. 
- Here is what the userAgent looks like: 
  - `SnowpipeJavaSDK/0.10.3 (Mac OS X 10.16 x86_64) JAVA/13 SFKafkaConnector/1.5.4 provider/SELF_HOSTED`
- I have added comments in the code which are self explanatory. 
- There are three allowed values to this configuration - `null/empty(equivalent to not provided)= UNKNOWN, SELF_HOSTED, CONFLUENT

PS:
- The test will fail now since I have not added the pom xml file. 
- Plan of action: Release snowpipe sdk, and push this with new version of sdk and see if the test passes. 

Here is how I have tested the local jar"
```
<repository>
            <id>local-maven-repo</id>
            <url>file:///Users/japatel/Desktop/Kafka-Connector/snowflake-kafka-connector/local-maven-repo</url>
</repository>

-- snowpipe sdk
<version>0.10.3</version>

-- maven command to install a local jar
mvn deploy:deploy-file -DgroupId=net.snowflake -DartifactId=snowflake-ingest-sdk -Dversion=0.10.3 -Durl=file:./local-maven-repo/ -DrepositoryId=local-maven-repo -DupdateReleaseInfo=true -Dfile=local-maven-repo/snowflake-ingest-sdk.jar

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -359,6 +359,7 @@ public class SnowflakeSinkConnectorConfig {
 
     // This API is called by framework to ensure the validity when connector is started or when a
     // validate REST API is called
+    @Override
     public void ensureValid(String name, Object value) {
       assert value instanceof String;
       final String strValue = (String) value;
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -327,7 +327,7 @@ public class SnowflakeSinkConnectorConfig {
         .define(
             PROVIDER_CONFIG,
             Type.STRING,
-            KafkaProvider.NONE.name(),
+            KafkaProvider.UNKNOWN.name(),
             KAFKA_PROVIDER_VALIDATOR,
             Importance.LOW,
             "Whether kafka is running on Confluent code, self hosted or other managed service");
@@ -357,6 +357,8 @@ public class SnowflakeSinkConnectorConfig {
   public static class KafkaProviderValidator implements ConfigDef.Validator {
     public KafkaProviderValidator() {}
 
+    // This API is called by framework to ensure the validity when connector is started or when a
+    // validate REST API is called
     public void ensureValid(String name, Object value) {
       assert value instanceof String;
       final String strValue = (String) value;
@@ -379,7 +381,7 @@ public class SnowflakeSinkConnectorConfig {
   /* Enum which represents allowed values of kafka provider. (Hosted Platform) */
   public enum KafkaProvider {
     // Default value, when nothing is provided. (More like Not Applicable)
-    NONE,
+    UNKNOWN,
 
     // Kafka/KC is on self hosted node
     SELF_HOSTED,
@@ -388,6 +390,7 @@ public class SnowflakeSinkConnectorConfig {
     CONFLUENT,
     ;
 
+    // All valid enum values
     public static final List<String> PROVIDER_NAMES =
         Arrays.stream(KafkaProvider.values())
             .map(kafkaProvider -> kafkaProvider.name().toLowerCase())
@@ -398,7 +401,7 @@ public class SnowflakeSinkConnectorConfig {
     public static KafkaProvider of(final String kafkaProviderStr) {
 
       if (Strings.isNullOrEmpty(kafkaProviderStr)) {
-        return KafkaProvider.NONE;
+        return KafkaProvider.UNKNOWN;
       }
 
       for (final KafkaProvider b : KafkaProvider.values()) {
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java [hunk 1]
```diff
@@ -336,4 +352,64 @@ public String toString() {
           + " <topic-1>:<table-1>,<topic-2>:<table-2>,... ";
     }
   }
+
+  /* Validator to validate Kafka Provider values which says where kafka is hosted */
+  public static class KafkaProviderValidator implements ConfigDef.Validator {
+    public KafkaProviderValidator() {}
+
+    public void ensureValid(String name, Object value) {
```
sfc-gh-rramachandran, line 360: Instead of throwing can it return a boolean? Also where is this method used?

sfc-gh-rramachandran (commented):
Minor comments, can you add javadoc for public methods?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-japatel: This validator is used while defining the provider config and it is being invoked from the framework.

sfc-gh-japatel: Ahh, I should add override annotation as well..

sfc-gh-rramachandran (approved):
LGTM!

sfc-gh-tzhang:
What's the reason we decide to put this in userAgentSuffix instead of through the client telemetry framework?

sfc-gh-japatel:
> LGTM!

Thanks @sfc-gh-rramachandran (FYI, I added pom.xml and the tests have passed.)

sfc-gh-japatel:
> What's the reason we decide to put this in userAgentSuffix instead of through the client telemetry framework?

We had a chat with confluent folks and this was the standard way to do it. Here are some examples
. MongoDB: https://github.com/mongodb/mongo-kafka/pull/60
. CosmosDB: https://github.com/microsoft/kafka-connect-cosmosdb/blob/05ba65d4fbf68c88a53d4cb77008295a0d70ea64/src/main/java/com/azure/cosmos/kafka/connect/sink/CosmosDBSinkTask.java#L45
. S3 Sink: https://github.com/confluentinc/kafka-connect-storage-cloud/blob/58f04361193e186c0c983a635294b6dddac886c3/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java#L69)

Also, for us to add anything in client telemetry, we had to go through some sort of security review. Technically user agent is also available in our snowhouse so in terms of implementation and ease of use there was hardly any difference.

sfc-gh-tzhang:
> > What's the reason we decide to put this in userAgentSuffix instead of through the client telemetry framework?
> 
> We had a chat with confluent folks and this was the standard way to do it. Here are some examples
> . MongoDB: [mongodb/mongo-kafka#60](https://github.com/mongodb/mongo-kafka/pull/60)
> . CosmosDB: https://github.com/microsoft/kafka-connect-cosmosdb/blob/05ba65d4fbf68c88a53d4cb77008295a0d70ea64/src/main/java/com/azure/cosmos/kafka/connect/sink/CosmosDBSinkTask.java#L45
> . S3 Sink: https://github.com/confluentinc/kafka-connect-storage-cloud/blob/58f04361193e186c0c983a635294b6dddac886c3/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java#L69)
> 
> Also, for us to add anything in client telemetry, we had to go through some sort of security review. Technically user agent is also available in our snowhouse so in terms of implementation and ease of use there was hardly any difference.

Thanks! I believe we were saying that we also want to collect the values in the configuration file (like buffer flush time, buffer size, etc.) so we would have a much better idea on why small files are generated and whether the default needs to be updated, and I assume that will be through the client telemetry, right?

sfc-gh-japatel:
> > > What's the reason we decide to put this in userAgentSuffix instead of through the client telemetry framework?
> > 
> > 
> > We had a chat with confluent folks and this was the standard way to do it. Here are some examples
> > . MongoDB: [mongodb/mongo-kafka#60](https://github.com/mongodb/mongo-kafka/pull/60)
> > . CosmosDB: https://github.com/microsoft/kafka-connect-cosmosdb/blob/05ba65d4fbf68c88a53d4cb77008295a0d70ea64/src/main/java/com/azure/cosmos/kafka/connect/sink/CosmosDBSinkTask.java#L45
> > . S3 Sink: https://github.com/confluentinc/kafka-connect-storage-cloud/blob/58f04361193e186c0c983a635294b6dddac886c3/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java#L69)
> > Also, for us to add anything in client telemetry, we had to go through some sort of security review. Technically user agent is also available in our snowhouse so in terms of implementation and ease of use there was hardly any difference.
> 
> Thanks! I believe we were saying that we also want to collect the values in the configuration file (like buffer flush time, buffer size, etc.) so we would have a much better idea on why small files are generated and whether the default needs to be updated, and I assume that will be through the client telemetry, right?

Correct, those things are more granular and usually these kind of information should go in client telemetry.

PR #318, opened 2021-06-09
Add retry mechanism if failure in end to end test because of rebalance

sfc-gh-japatel:


.github/workflows/End2EndTestApacheAws.yml
```diff
@@ -60,8 +60,10 @@ jobs:
         ./build_apache.sh ../../snowflake-kafka-connector package
 
     - name: End to End Test of Apache Plarform 2.5
+      uses: nick-invision/retry@v2
```
sfc-gh-zli, line 63: Is this action github verified on the github action market place? We are only allowed to use those verified actions.

sfc-gh-japatel: Yes this was from marketplace..

sfc-gh-zli: I searched for it on github action marketplace, the developer is not verified by github. So we cannot use this one.
Screen Shot 2021-06-09 at 12 20 36 AM
Screen Shot 2021-06-09 at 12 20 50 AM

sfc-gh-japatel: Ahh I see.. No worries. I will meddle with other settings like max.poll.time.ms because I saw those errors popping up multiple times.

Issue #319, opened 2021-06-15
Need to add support for Plain snowflake UserName and Password


sql-dr-watson:
The current implementation of this connector does not allow for use of standard UserName and Password.  This is not easily updatable for key rotation and stops the adoption of the go-snowflake dynamic credential generation from HashiCorp Vault.

Logic could be:
if PrivateKey != null and PrivateKeyPassphrase != null
 -> prop.privateKey & prop.privateKeyPassphrase = input 
else if password != null
-> prop.password = input
else
-> error!

sql-dr-watson:
https://github.com/snowflakedb/snowflake-kafka-connector/pull/320

sfc-gh-japatel:
Feel free to open. Commented on the #320

PR #320, opened 2021-06-15
updated to add password support

sql-dr-watson:


sfc-gh-japatel:
Hey, why do you need this?

sql-dr-watson:
https://github.com/snowflakedb/snowflake-kafka-connector/issues/319

HashiCorp Vault Snowflake Provider does not support RSA Tokens, but does do User/Pass

sql-dr-watson:
https://www.vaultproject.io/api-docs/secret/databases/snowflake

sfc-gh-japatel:
How is the password you added as a config, used to connect to snowflake? Sorry, I dont think I quite understand this. Can you describe a usecase?

sql-dr-watson:
@sfc-gh-japatel 
The use case is for us to be able to use username & password instead of RSA Key Pair auth in the snowflake kafka connect worker config.  
So instead of 
snowflake.username
snowflake.privateKey
snowflake.privateKeyPassphrase

we would just set
snowflake.username
snowflake.password

sfc-gh-japatel:
I am still confused here and not 100% clear on what the username and password are used for. 

This kafka connector internally uses snowpipe REST API which uses key pair based authentication. 
https://docs.snowflake.com/en/user-guide/data-load-snowpipe-rest-load.html
https://docs.snowflake.com/en/user-guide/data-load-snowpipe-rest-gs.html#step-3-configure-security-per-user

Can you please elaborate on how this username and password will be used? Are these stored in hashicorp? If yes, there is no way for us to fetch that from KC. 

May be this is a feature request for hashicorp to generate a key pair based authentication from username and password?

PR #322, opened 2021-06-23
Increase poll time, reduce no of records in a poll and increase task …

sfc-gh-japatel:
…for flaky test


sfc-gh-rramachandran (approved):
LGTM!

PR #323, opened 2021-07-08
Added JMX metrics which will be exported to Mbean Server

sfc-gh-japatel:
- Here are the screenshots of the jconsole. 
- Here are the settings needed on client side - (Which is already documented here)
  - https://docs.confluent.io/home/connect/monitoring.html#using-jmx-to-monitor-kconnect
  - export JMX_PORT=PORT
- If running remotely:
  - ``` KAFKA_JMX_OPTS="-Dcom.sun.management.jmxremote=true -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false -Djava.rmi.server.hostname=your.kafka.broker.hostname -Djava.net.preferIPv4Stack=true"```

Screen Shot 2021-07-07 at 5 53 04 PM
Screen Shot 2021-07-07 at 5 53 17 PM

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeErrors.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeErrors.java
@@ -230,7 +230,8 @@ public enum SnowflakeErrors {
   ERROR_5017(
       "5017", "Invalid api call to cached put", "Cached put only support AWS, Azure and GCS."),
   ERROR_5018("5018", "Failed to execute cached put", "Error in cached put command"),
-  ERROR_5019("5019", "Failed to get stage storage type", "Error in get storage type");
+  ERROR_5019("5019", "Failed to get stage storage type", "Error in get storage type"),
+  ERROR_5020("5020", "Failed to register MBean in MbeanServer", "Object Name is invalid");
 
   // properties
 
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryBasicInfo.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryBasicInfo.java
@@ -1,6 +1,13 @@
 package com.snowflake.kafka.connector.internal;
 
+import java.lang.management.ManagementFactory;
+import javax.management.JMException;
+import javax.management.MBeanServer;
+import javax.management.MalformedObjectNameException;
+import javax.management.ObjectName;
 import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.node.ObjectNode;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public abstract class SnowflakeTelemetryBasicInfo {
   final String tableName;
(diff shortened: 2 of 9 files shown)


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeErrors.java
```diff
@@ -230,7 +230,8 @@
   ERROR_5017(
       "5017", "Invalid api call to cached put", "Cached put only support AWS, Azure and GCS."),
   ERROR_5018("5018", "Failed to execute cached put", "Error in cached put command"),
-  ERROR_5019("5019", "Failed to get stage storage type", "Error in get storage type");
+  ERROR_5019("5019", "Failed to get stage storage type", "Error in get storage type"),
+  ERROR_5020("5020", "Failed to register MBean in MbeanServer", "Object Name is invalid");
```
sfc-gh-tjones, line 234: Do we log the object name somewhere?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryBasicInfo.java
```diff
@@ -10,12 +17,87 @@
...
+   * @param pipeName the name of the context
+   * @return the JMX metric name
+   * @throws com.snowflake.kafka.connector.internal.SnowflakeKafkaConnectorException if the name is
+   *     invalid
+   */
+  public static ObjectName metricName(String pipeName, String connectorName) {
+    final String metricName = JMX_METRIC_PREFIX + ":type=" + connectorName + ",name=" + pipeName;
+    try {
+      return new ObjectName(metricName);
+    } catch (MalformedObjectNameException e) {
```
sfc-gh-tjones, line 99: Worth logging the error message here?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryBasicInfo.java
```diff
@@ -10,12 +17,87 @@
...
+      throw SnowflakeErrors.ERROR_5020.getException();
```
sfc-gh-japatel, line 100: I can put log line here @sfc-gh-tjones

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryBasicInfo.java
```diff
@@ -10,12 +17,87 @@
...
+          Logging.logMessage("Unable to register the MBean '{}': {}", mBeanName, e.getMessage()));
+    }
+  }
+
+  /**
+   * Unregisters a metrics MBean from the platform MBean server. The method is intentionally
+   * synchronized to prevent preemption between registration and un-registration.
+   */
+  public synchronized void unregisterMBean() {
+    if (this.mBeanName != null && enableJMXMonitoring) {
```
sfc-gh-rramachandran, line 71: Should we have this check in registerMBean too for safety?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryBasicInfo.java
```diff
@@ -10,12 +17,87 @@
...
+  public synchronized void registerMBean() {
+    try {
+      final MBeanServer mBeanServer = ManagementFactory.getPlatformMBeanServer();
+      if (mBeanServer == null) {
+        LOGGER.info(Logging.logMessage("JMX not supported, bean '{}' not registered", mBeanName));
+        return;
+      }
+      mBeanServer.registerMBean(this, mBeanName);
+      LOGGER.info(Logging.logMessage("Registered Mbean:{}", mBeanName));
+    } catch (JMException e) {
```
sfc-gh-rramachandran, line 60: Thinking out loudly here, is it okay to fail the sinkService start if any other exception is thrown? Should we have a catch all exception here for safety?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryBasicInfo.java
```diff
@@ -10,12 +17,87 @@
...
+            Logging.logMessage(
+                "Unable to unregister the MBean '{}': {}", mBeanName, e.getMessage()));
+      }
+    }
+  }
+
+  /**
+   * Create a JMX metric name for the given metric.
+   *
+   * @param pipeName the name of the context
```
sfc-gh-rramachandran, line 90: Nit: Add @param connectorName

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryBasicInfo.java
```diff
@@ -10,12 +17,87 @@
...
-      final String tableName, final String stageName, final String pipeName) {
+      final String tableName,
+      final String stageName,
+      final String pipeName,
+      final String connectorName,
+      final boolean enableJMXMonitoring) {
     this.tableName = tableName;
     this.stageName = stageName;
     this.pipeName = pipeName;
+    this.enableJMXMonitoring = enableJMXMonitoring;
```
sfc-gh-rramachandran, line 39: Nit: Any assert needed for connectorName?

sfc-gh-rramachandran (approved):
Left some minor comments, looks good overall. You should probably present this in a tech sync.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryBasicInfo.java
sfc-gh-japatel: Name seems like a required property. So may be assert is not needed. (And we do throw exception if name is empty or invalid)
https://docs.snowflake.com/en/user-guide/kafka-connector-install.html#required-properties

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryBasicInfo.java
sfc-gh-japatel: Yes, makes sense to catch all. Also just realised `ManagementFactory.getPlatformMBeanServer` throws SecurityException (Extends RTE) so it might be worth to catch all exceptions. I will do that instead of JMException

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryBasicInfo.java
sfc-gh-japatel: Will do.

sfc-gh-japatel:
> Left some minor comments, looks good overall. You should probably present this in a tech sync.

Thanks, I will push this change and in another PR, will add option to enable JMX metrics as a user config.

PR #324, opened 2021-07-20
Add JMX option for custom metrics

sfc-gh-japatel:
- Default is true
- False would not emit custom metrics to JMX


sfc-gh-rramachandran (approved):
LGTM! Wow that is a lot of boilerplate code for adding a config, is this unique to our connector or a common pattern across other connectors?

sfc-gh-japatel:
> LGTM! Wow that is a lot of boilerplate code for adding a config, is this unique to our connector or a common pattern across other connectors?

I think this is common across other connectors. I have seen the Config file generally pretty huge depending on the config options that we provide. It all comes back to this http://neugierig.org/software/blog/2018/07/options.html

PR #325, opened 2021-07-23
Added dropwizard JMX metrics

sfc-gh-japatel:
- Dropwizard: https://metrics.dropwizard.io/4.2.0/getting-started.html

- Refactored a few boilterplate code. (Thanks to Dropwizard)
- Added latency metric (Using Timer Metric)
- Using gauge for most of other metric.
- Used Meter for failedFile metrics.

Diff'ing against #324 
- Will push this after pushing #324 

Reviewers:
- Required @sfc-gh-rramachandran 
- FYI @sfc-gh-tjones @sfc-gh-psaraiya 

Here is how the changes looks like: (For 3 partitioned topic, we will have separate section and subsections for categories)
Screen Shot 2021-07-23 at 11 22 16 AM

- Dashboard metric for latency. (99th percentile)
Screen Shot 2021-07-23 at 11 21 54 AM


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -235,13 +242,15 @@ public void close(Collection<TopicPartition> partitions) {
                 tp.partition());
           }
         });
+    unregisterAllSnowflakeJMXMetrics();
```
sfc-gh-rramachandran, line 245: I am not entirely sure if we should handle any exceptions are thrown in the close method before reaching unregisterMetrics. Would it make sense to put the existing code in try { } and the unregister call in a finally { }?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -235,13 +242,15 @@ public void close(Collection<TopicPartition> partitions) {
...
   }
 
   @Override
   public void closeAll() {
     this.isStopped = true; // release all cleaner and flusher threads
     pipes.forEach((name, context) -> context.close());
     pipes.clear();
+    unregisterAllSnowflakeJMXMetrics();
```
sfc-gh-rramachandran, line 253: Same comment as above, try and finally unregister?

sfc-gh-rramachandran (approved):
LGTM! This looks great!

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-japatel: Good point, yes. Let me do that. I will wrap it around try-catch-finally

sfc-gh-japatel: Actually, now that I think about this, we don't need to do this in try-catch-finally. We are doing unregistering outside of foreach so it should be fine since we catch all exception inside for each. 

And I will lazily unregister all metrics before we try to register during partition assignment. (Coming in next PR)

I will keep this as is.

sfc-gh-tjones (approved):
Awesome stuff!

PR #326, opened 2021-07-29
Add buffer metrics as Histogram

sfc-gh-japatel:
- Added histogram for buffer metric
- Improved unregistration code. 
- Lazily invoke remove metrics before registering them. (This can happen if there are rebalances) So rebalancing would start from all over again. 
- Fixed test
Example metric with 1000 records and 10 as buffer threshold. 
Screen Shot 2021-07-28 at 7 48 51 PM
Screen Shot 2021-07-28 at 7 48 44 PM


sfc-gh-rramachandran (approved):
LGTM!

PR #327, opened 2021-08-02
Rename latency metric for consistency

sfc-gh-japatel:
- Just for consistency since all other metric names have a `-` and not an underscore


sfc-gh-rramachandran (approved):
LGTM

PR #334, opened 2021-09-13
Regionless url regex support

sfc-gh-japatel:
SNOW-445660

**Commit 1:**
- The string before google reformat didnt have + inside the matcher.
Changing this to a single line without a concatenation so that the
matcher string doesnt confuse with +

**Commit 2:**
- Adding a hyphen in account matcher group to support regionless URL along with tests.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeURL.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeURL.java
@@ -34,9 +34,11 @@ public class SnowflakeURL extends Logging {
 
   private int port;
 
+  private static final String SNOWFLAKE_URL_REGEX_PATTERN =
+      "^(https?://)?((([\\w\\d]+)(\\.[\\w\\d-]+){2,})(:(\\d+))?)/?$";
+
   public SnowflakeURL(String urlStr) {
-    Pattern pattern =
-        Pattern.compile("^(https?://)?((([\\w\\d]+)(\\" + ".[\\w\\d-]+){2,})(:(\\d+))?)/?$");
+    Pattern pattern = Pattern.compile(SNOWFLAKE_URL_REGEX_PATTERN);
 
     Matcher matcher = pattern.matcher(urlStr.trim().toLowerCase());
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeURL.java
```diff
@@ -34,9 +34,27 @@
...
+   *
+   * <p>Group 1: If https is present or not. (Not required)
+   *
+   * <p>Group 2: Is the entire URL including the port number
+   *
+   * <p>Group 3: URL until .com
+   *
+   * <p>Group 4: Account name (may include org-account/alias)
+   *
+   * <p>Group 5: (Everything after accountname or org-accountname until .com)
```
sfc-gh-rramachandran, line 49: NIT: Group 6?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeURL.java
```diff
@@ -34,9 +34,27 @@
...
+   *
+   * <p>Group 7: port number
+   */
+  private static final String SNOWFLAKE_URL_REGEX_PATTERN =
```
sfc-gh-rramachandran, line 53: Is this a copy+paste from snowflakedb repo? If so, can you include a link to that here as a comment?

sfc-gh-rramachandran (approved):
LGTM

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeURL.java
sfc-gh-japatel: Wasnt too useful, since it is not used.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeURL.java
sfc-gh-japatel: Nah, I added this comment for regex amateurs like me. It is to validate the url before generating one for jdbc.

sfc-gh-japatel:
I also tested this E2E on an internal account in testqa8 deployment.

PR #338, opened 2021-09-21
Do not update committed offset(In memory) and eventually to Kafka if insertFiles has failed

sfc-gh-japatel:
Consider this scenario:
1. insertFiles has failed three times which results into no offsets being committed to return response of precommit. (Empty [HashMap](https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java#L291)) 
2. There is no restart/rebalance. 
3. InsertFiles API has been successful for newer offsets. 
4. Because we update the committedOffset, even though we would have unsuccessfully called insertFiles, our committedOffset pointer would have moved forward and we would skip those offsets.


sfc-gh-zli (approved):
lgtm

PR #340, opened 2021-09-30
Enable respective protobuf test version on confluent

sfc-gh-japatel:


test/test_verify.py
```diff
@@ -341,11 +342,11 @@ def runTestSet(driver, testSet, nameSalt, pressure):
                      testAvrosrAvrosr, testNativeStringAvrosr, testNativeStringJsonWithoutSchema,
                      testNativeComplexSmt, testNativeStringProtobuf, testConfluentProtobufProtobuf]
 
-    # Adding StringJsonProxy test at the end
+    # clean is set to true for all tests
     testCleanEnableList1 = [True, True, True, True, True, True, True, True, True, True, True]
     testSuitEnableList1 = []
     if testSet == "confluent":
-        testSuitEnableList1 = [True, True, True, True, True, True, True, True, True, True, False]
+        testSuitEnableList1 = [True, True, True, True, True, True, True, True, True, False, True]
```
sfc-gh-japatel, line 349: I might be wrong here but can we install blueapron protobuf on confluent? (Seems like we can, just confirming)

If yes, I should set second to last value to true too. (corresponding to testNativeStringProtobuf)
CC: @sfc-gh-zli

sfc-gh-zli: Seems like tests are failing?

sfc-gh-japatel: Mostly because of a rebalance.

sfc-gh-japatel:
CC: @sfc-gh-rramachandran / @sfc-gh-anwang

PR #350, opened 2021-11-29
SNOW-508731 update snowflake ingestion service

sfc-gh-anwang:
Added new apis in the SnowflakeIngestionService: 
- configureClient
- getClientStatus
- ingestFilesWithClientInfo

The unit test IngestIonServiceIT was tested against QA env


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
```diff
@@ -254,4 +257,104 @@ public void close() {
...
+    }
+  }
+
+  /**
+   * get the Snowpipe client and return the offset token
+   *
+   * @return a String value contains the offset token
+   */
+  public String getClientStatus() {
+    ClientStatusResponse response;
```
sfc-gh-anwang, line 291: should we pass the clientSequencer and make sure it matches with returned clientSequencer?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
```diff
@@ -254,4 +257,107 @@ public void close() {
   public SimpleIngestManager getIngestManager() {
     return this.ingestManager;
   }
+
+  /**
+   * configure the Snowpipe client and return the client sequencer
+   *
+   * @return a Long value contains the client sequencer
+   */
+  public Long configureClient() {
```
sfc-gh-japatel, line 266: should we return `ConfigureClientResponse` directly? let the caller decide what to use!

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
```diff
@@ -254,4 +257,107 @@ public void close() {
...
+      throw SnowflakeErrors.ERROR_4001.getException("the response of configure client is null");
+    }
+  }
+
+  /**
+   * get the Snowpipe client and return the offset token
+   *
+   * @return a String value contains the offset token
+   */
+  public String getClientStatus() {
```
sfc-gh-japatel, line 290: Same here, lets return the response directly which contains token and sequencer.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
sfc-gh-japatel: If we are not passing it in API call, we can avoid doing it in this method. 
This method can be thought of as a wrapper on top of remote API. Lets avoid adding more checks. 

May be that assert can be done by the caller?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
```diff
@@ -254,4 +257,107 @@ public void close() {
...
+              InternalUtils.backoffAndRetry(
+                  telemetry,
+                  SnowflakeInternalOperations.GET_CLIENT_STATUS_SNOWPIPE_API,
+                  () -> ingestManager.getClientStatus(null));
+    } catch (Exception e) {
+      throw SnowflakeErrors.ERROR_3007.getException(e);
+    }
+    if (response != null) {
+      // offsetToken is ok to be null
+      return response.getOffsetToken();
```
sfc-gh-japatel, line 304: this offsetToken can be null for first call right? Just after configure client it will return null if we didnt call insertFiles.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
```diff
@@ -254,4 +257,107 @@ public void close() {
...
+   * Ingest a list of files with the clientInfo (clientSequencer and offsetToken)
+   *
+   * @param fileNames file name List
+   * @param clientSequencer unique identification of the Snowpipe client
+   */
+  public void ingestFilesWithClientInfo(List<String> fileNames, long clientSequencer) {
+    if (fileNames.isEmpty()) {
+      return;
+    }
+    logDebug("ingest files with client info: {}, clientSequencer: {} ", fileNames, clientSequencer);
```
sfc-gh-japatel, line 320: `Arrays.toString(fileNames)`
https://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html#toString%28java.lang.Object%5B%5D%29

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
```diff
@@ -254,4 +257,107 @@ public void close() {
   public SimpleIngestManager getIngestManager() {
     return this.ingestManager;
   }
+
```
sfc-gh-japatel, line 260: nit: @Override annotation on all implemented methods

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
sfc-gh-japatel: can we just mention in our javadocs that it(Offsettoken) can be null

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
```diff
@@ -254,4 +257,107 @@ public void close() {
...
+  private String getLastOffsetTokenFromBatch(List<String> fileNameBatch) {
+    Long lastFileEndOffset =
+        FileNameUtils.fileNameToEndOffset(fileNameBatch.get(fileNameBatch.size() - 1));
+    for (String fileName : fileNameBatch) {
+      if (lastFileEndOffset < FileNameUtils.fileNameToEndOffset(fileName)) {
+        lastFileEndOffset = FileNameUtils.fileNameToEndOffset(fileName);
+        logWarn("The fileName list is not sequential.");
+      }
+    }
+    return lastFileEndOffset.toString();
```
sfc-gh-japatel, line 361: can you add a unit test for this? You can make this protected and add annotation @visiblefortesting

src/test/java/com/snowflake/kafka/connector/internal/IngestionServiceIT.java
```diff
@@ -62,4 +64,60 @@ public void ingestFileTest() throws Exception {
...
+
+    // Register/Configure a snowpipe client
+    Long clientSequencer = ingestService.configureClient();
+    assert clientSequencer.equals(0l);
+    // Reconfigure the snowpipe client, the clientSequencer should increase
+    clientSequencer = ingestService.configureClient();
+    assert clientSequencer.equals(1l);
+    // Get the client offset token
+    String offsetToken = ingestService.getClientStatus();
+    assert offsetToken == null;
```
sfc-gh-japatel, line 84: 👍

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
sfc-gh-japatel: or you can add multiple files in existing test in different order and see if it works.

src/test/java/com/snowflake/kafka/connector/internal/IngestionServiceIT.java
```diff
@@ -62,4 +64,60 @@ public void ingestFileTest() throws Exception {
 
     assert ingestService.getStageName().equals(stage);
   }
+
+  @Ignore
+  @Test
+  public void ingestFileWithClientInfoTestSuccessful() throws Exception {
+    String file = "{\"aa\":123}";
+    String fileName = FileNameUtils.fileName(TestUtils.TEST_CONNECTOR_NAME, table, 0, 0, 1);
```
sfc-gh-japatel, line 72: can you add comment here for filename. 
I think the parameters are table, partition_no, startOffset, endOffset

src/test/java/com/snowflake/kafka/connector/internal/IngestionServiceIT.java
```diff
@@ -62,4 +64,60 @@ public void ingestFileTest() throws Exception {
 
     assert ingestService.getStageName().equals(stage);
   }
+
+  @Ignore
+  @Test
+  public void ingestFileWithClientInfoTestSuccessful() throws Exception {
+    String file = "{\"aa\":123}";
```
sfc-gh-japatel, line 71: can you also add a test for more than one file?

sfc-gh-japatel (commented):
minor changes, overall LGTM. Lets push it after 5.44 server releases have happened.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
sfc-gh-anwang: Are we just using SnowflakeIngestService as a wrapper? All other apis have some extra logic to make them easier to use. For example, check history convert the response to map. I think it makes sense that we unwrap the response here in the ingestionService. The caller of this function does not need to know what structure of response we received from the SDK. Whenever we update the SDK, we only need to update those functions. This will make maintenance easier.

sfc-gh-japatel: Makes sense based on what is done earlier. Here is the case where I thought returning the response as is would be beneficial. Lets say we combine configure and get api to just configure api and then the configure api also returns offsetToken. In this case having returned the response object would also include offsetToken. And the response object type (`ConfigureClientResponse`) wont change, hopefully only more fields will be added.

sfc-gh-anwang: This also makes sense. Sure, we can return the entire response to caller.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionService.java
```diff
@@ -51,14 +53,14 @@
    *
    * @return a Long value contains the client sequencer
```
sfc-gh-japatel, line 54: nit: return response

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionService.java
```diff
@@ -51,14 +53,14 @@
...
    */
-  Long configureClient();
+  ConfigureClientResponse configureClient();
 
   /**
    * get the Snowpipe client and return the offset token
    *
    * @return a String value contains the offset token
```
sfc-gh-japatel, line 61: nit: return response

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
```diff
@@ -261,9 +261,10 @@ public SimpleIngestManager getIngestManager() {
   /**
    * configure the Snowpipe client and return the client sequencer
    *
-   * @return a Long value contains the client sequencer
+   * @return ConfigureClientResponse contains the client sequencer
    */
```
sfc-gh-japatel: May be you can remove this javadocs from here since we already have it in interface?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
```diff
@@ -275,19 +276,20 @@ public Long configureClient() {
...
+    if (response != null) {
+      return response;
     } else {
       throw SnowflakeErrors.ERROR_4001.getException("the response of configure client is null");
     }
   }
 
   /**
-   * get the Snowpipe client and return the offset token
+   * get the Snowpipe client and return the ConfigureClientResponse
```
sfc-gh-japatel, line 287: same as above.

src/test/java/com/snowflake/kafka/connector/internal/IngestionServiceIT.java
```diff
@@ -69,22 +73,24 @@ public void ingestFileTest() throws Exception {
   @Test
   public void ingestFileWithClientInfoTestSuccessful() throws Exception {
     String file = "{\"aa\":123}";
+    // File Name Format: app/table/partition/start_end_timeStamp.fileFormat.gz
```
sfc-gh-japatel, line 76: thank you!

src/test/java/com/snowflake/kafka/connector/internal/IngestionServiceIT.java
```diff
@@ -97,27 +103,53 @@ public void ingestFileWithClientInfoTestSuccessful() throws Exception {
...
+  @Ignore
+  @Test
+  public void ingestMultipleFilesWithClientInfoTest() {
+    String fileContent = "{\"aa\":123}";
+    // start offset = 0, end offset = 1
+    String fileName1 = FileNameUtils.fileName(TestUtils.TEST_CONNECTOR_NAME, table, 0, 0, 1);
+    // start offset = 2, end offset = 3
+    String fileName2 = FileNameUtils.fileName(TestUtils.TEST_CONNECTOR_NAME, table, 0, 2, 3);
+    // start offset = 1, end offset = 2
+    String fileName3 = FileNameUtils.fileName(TestUtils.TEST_CONNECTOR_NAME, table, 0, 1, 2);
```
sfc-gh-japatel, line 141: So this is fine for the test but this is not how it happens in code. 
0-1 -> includes two offsets 0 and 1
2-3 -> includes two offsets 2 and 3
3-4 -> includes two offsets 3 and 4

the test looks good though!

sfc-gh-anwang: I see, I'll update it.

sfc-gh-japatel (approved):
lgtm!

PR #353, opened 2021-11-30
Change name to delivery.guarantee instead of processing.gurantee

sfc-gh-japatel:
I got this idea from https://docs.confluent.io/5.5.1/connect/managing/confluent-hub/kafka-connect-maven-plugin/kafka-connect-mojo.html

Which seems a better option.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -93,7 +93,7 @@ public class SnowflakeSinkConnectorConfig {
   // By default it will be None since this is not enforced and only used for monitoring
   public static final String PROVIDER_CONFIG = "provider";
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -532,7 +532,7 @@ public String toString() {
    * Enum which represents the type of processing guarantees that the customer want (either
```
sfc-gh-anwang, line 532: Update javadoc

sfc-gh-anwang (approved):
Minor comment, LGTM!

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-japatel: Will do!

PR #355, opened 2021-12-01
SNOW-508732 Update Sink Service EOS

sfc-gh-anwang:
When EOS is enabled: 
- Call configure client and get status APIs when initializing ServiceContext
- Discard the record if the record offset is smaller or equal to offset persisted in snowflake
- Insert file api also contains client info (clientSequencer and offsetToken)

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -127,8 +127,8 @@ public class SnowflakeSinkConnectorConfig {
    * Get a property from the config map
    *
    * @param config connector configuration
-   * @param key name of the key to be retrived
-   * @return proprity value or null
+   * @param key name of the key to be retrieved
+   * @return property value or null
    */
   static String getProperty(final Map<String, String> config, final String key) {
     if (config.containsKey(key) && !config.get(key).isEmpty()) {
@@ -253,7 +253,7 @@ public class SnowflakeSinkConnectorConfig {
             "",
             topicToTableValidator,
             Importance.LOW,
-            "Map of topics to tables (optional). Format : comma-seperated tuples, e.g."
+            "Map of topics to tables (optional). Format : comma-separated tuples, e.g."
                 + " <topic-1>:<table-1>,<topic-2>:<table-2>,... ",
             CONNECTOR_CONFIG,
             0,
@@ -386,7 +386,7 @@ public class SnowflakeSinkConnectorConfig {
     }
 
     public String toString() {
-      return "Topic to table map format : comma-seperated tuples, e.g."
+      return "Topic to table map format : comma-separated tuples, e.g."
           + " <topic-1>:<table-1>,<topic-2>:<table-2>,... ";
     }
   }
@@ -404,7 +404,6 @@ public class SnowflakeSinkConnectorConfig {
       // The value can be null or empty.
       try {
         KafkaProvider kafkaProvider = KafkaProvider.of(strValue);
-        LOGGER.debug("KafkaProvider value is:{}", kafkaProvider.name());
       } catch (final IllegalArgumentException e) {
         throw new ConfigException(PROVIDER_CONFIG, value, e.getMessage());
       }
@@ -429,7 +428,6 @@ public class SnowflakeSinkConnectorConfig {
       try {
         IngestionDeliveryGuarantee ingestionDeliveryGuarantee =
             IngestionDeliveryGuarantee.of(strValue);
-        LOGGER.debug("DeliveryGuarantee type is:{}", ingestionDeliveryGuarantee.name());
       } catch (final IllegalArgumentException e) {
         throw new ConfigException(PROVIDER_CONFIG, value, e.getMessage());
       }
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
@@ -504,7 +504,7 @@ class SnowflakeSinkServiceV1 extends Logging implements SnowflakeSinkService {
       // when exactly_once is enabled,fetch clientSequencer and offsetPersistedInSnowflake
       if (ingestionDeliveryGuarantee
           == SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.EXACTLY_ONCE) {
-        initClientInfo();
+        initClientInfoForExactlyOnceDelivery();
       }
 
       try {
@@ -515,7 +515,11 @@ class SnowflakeSinkServiceV1 extends Logging implements SnowflakeSinkService {
       }
     }
 
-    private void initClientInfo() {
+    /**
+     * Initialize the client info (clientSequencer and offsetPersistedInSnowflake) by calling
+     * ingestion service API configureClient and getClientStatus
+     */
+    private void initClientInfoForExactlyOnceDelivery() {
       ConfigureClientResponse configureClientResponse = ingestionService.configureClient();
       this.clientSequencer.set(configureClientResponse.getClientSequencer());
       ClientStatusResponse clientStatusResponse = ingestionService.getClientStatus();
@@ -527,7 +531,8 @@ class SnowflakeSinkServiceV1 extends Logging implements SnowflakeSinkService {
           this.offsetPersistedInSnowflake.set(Long.parseLong(offsetToken));
         }
         logInfo(
-            "Initialized client info, clientSequencer: {}, offsetToken: {}.",
+            "Initialized client info for pipe:{}, clientSequencer:{}, offsetToken:{}.",
+            this.pipeName,
             this.clientSequencer.get(),
             this.offsetPersistedInSnowflake.get());
       } catch (NumberFormatException e) {
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
@@ -843,6 +843,13 @@ class SnowflakeSinkServiceV1 extends Logging implements SnowflakeSinkService {
       if (ingestionDeliveryGuarantee
           == SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.EXACTLY_ONCE) {
         ingestionService.ingestFilesWithClientInfo(fileNamesCopy, this.clientSequencer.get());
+        String offsetToken = ingestionService.getClientStatus().getOffsetToken();
+        // Update server side offset
+        if (offsetToken == null) {
+          this.serverSideOffset.set(-1);
+        } else {
+          this.serverSideOffset.set(Long.parseLong(offsetToken));
+        }
       } else {
         ingestionService.ingestFiles(fileNamesCopy);
       }
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
@@ -436,7 +436,9 @@ class SnowflakeSinkServiceV1 extends Logging implements SnowflakeSinkService {
 
     // exactly once semantics
     private final AtomicLong clientSequencer = new AtomicLong(-1);
-    private final AtomicLong serverSideOffset = new AtomicLong(-1);
+    // This offset is updated when Snowflake has received offset from ingest file and has queued it
+    // For verifying the ingestion status, we use the insertReport api
+    private final AtomicLong offsetPersistedInSnowflake = new AtomicLong(-1);
 
     private ServiceContext(
         String tableName,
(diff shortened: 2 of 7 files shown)


sfc-gh-japatel:
Can we please add description in this PR?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -432,6 +434,10 @@ protected static String getNameIndex(String topic, int partition) {
     private boolean hasInitialized = false;
     private boolean forceCleanerFileReset = false;
 
+    // exactly once semantics
+    private final AtomicLong clientSequencer = new AtomicLong(-1);
+    private final AtomicLong serverSideOffset = new AtomicLong(-1);
```
sfc-gh-japatel, line 439: may be something like `offsetPersistedInSnowflake`

Can you also add a comment saying `this is source of truth` Also this value doesnt necessarily mean it is available to query but just a notification that snowflake has received this offset and has queued it?

sfc-gh-japatel: For verifying whether that offset is ingested, we still have to rely on insertReport api

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -493,6 +501,23 @@ private void init(long recordOffset) {
       // recover will only check pipe status and create pipe if it does not exist.
       recover(pipeCreation);
 
+      // when exactly_once is enabled,fetch clientSequencer and offsetPersistedInSnowflake
+      if (ingestionDeliveryGuarantee
```
sfc-gh-japatel, line 505: lets extract this out to a new method so that this init method is not noisy. 
Also since it calls only during start, we should log the offsetToken and clientSequencer.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -813,7 +842,19 @@ private long getOffset() {
...
       // This api should throw exception if backoff failed.
       // fileNamesCopy after this call is emptied (clears the input list)
-      ingestionService.ingestFiles(fileNamesCopy);
+      if (ingestionDeliveryGuarantee
+          == SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.EXACTLY_ONCE) {
+        ingestionService.ingestFilesWithClientInfo(fileNamesCopy, this.clientSequencer.get());
+        String offsetToken = ingestionService.getClientStatus().getOffsetToken();
+        // Update server side offset
+        if (offsetToken == null) {
+          this.offsetPersistedInSnowflake.set(-1);
```
sfc-gh-japatel, line 851: Is this even possible? Should we throw an exception if offsetToken we receive from SF is null?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -813,7 +842,19 @@ private long getOffset() {
...
+        } else {
+          this.offsetPersistedInSnowflake.set(Long.parseLong(offsetToken));
```
sfc-gh-japatel, line 853: Just thinking out loud here. Why do we want to set this value after returning from another rest call for getClientStatus? 

Cant we rely on `insertFilesWithClientInfo`'s successful call? Basically if it is a success, the `this.offsetPersistedInSnowflake.set(lastOffsetFromFrom_fileNamesCopy)`

I am okay with one more though for completeness.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-anwang: when we create new pipe, the offset token is null when we call get pipe status

sfc-gh-anwang: we can not put null offset token in insert call though

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -651,10 +676,14 @@ private void insert(final SinkRecord record) {
...
-
+      // only get offset token once when service context is initialized
       // ignore ingested files
-      if (record.kafkaOffset() > processedOffset.get()) {
+      // if ingestionDeliveryGuarantee is AT_LEAST_ONCE, ignore offsetPersistedInSnowflake
+      // else discard the record if the record offset is smaller or equal to server side offset
+      if ((ingestionDeliveryGuarantee
+                  == SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.AT_LEAST_ONCE
+              || record.kafkaOffset() > this.offsetPersistedInSnowflake.get())
+          && record.kafkaOffset() > processedOffset.get()) {
```
sfc-gh-japatel, line 686: in case of exactly_once, can we set `processedOffset` = `offsetPersistedInSnowflake`?
Then we can avoid this complicated if logic!

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-anwang: Sounds good

src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
```diff
@@ -1147,4 +1148,229 @@ public void testCleanerRecoverListCount() throws Exception {
...
+    offset = 0;
+    // Create sink record with same offset
+    SinkRecord record2 =
+        new SinkRecord(
+            topic, partition, Schema.STRING_SCHEMA, "test", input.schema(), input.value(), offset);
+    service2.insert(record2);
+    // The record should not be inserted in the buffer
+    Assert.assertTrue(
+        ((SnowflakeSinkServiceV1) service2)
+            .isPartitionBufferEmpty(SnowflakeSinkServiceV1.getNameIndex(topic, partition)));
```
sfc-gh-japatel, line 1309: nice!

src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
```diff
@@ -1147,4 +1148,229 @@ public void testCleanerRecoverListCount() throws Exception {
...
+    SinkRecord record2 =
+        new SinkRecord(
+            topic, partition, Schema.STRING_SCHEMA, "test", input.schema(), input.value(), offset);
+    // Should update the client sequencer
+    service2.insert(record2);
+    // call snowpipe ingest api
+    service.callAllGetOffset();
+
+    service.closeAll();
+    service2.closeAll();
```
sfc-gh-japatel, line 1374: but we are changing the offset here to 1? So it should ingest it?

sfc-gh-anwang: Sorry, the comment wasn't very clear, the old service calls get offset, so the api call should throw exception due to outdated client sequencer.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-anwang: proceedOffset can sometime be more update to date than the offsetPersistedInSnowflake, so this might not actually work

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-anwang: if we only need the offsetPersistedInSnowflake during rebalancing, this call is kind of redundant already. i am also ok with directly setting the offset to the lastOffsetFromFilename.

src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
```diff
@@ -1147,4 +1148,229 @@ public void testCleanerRecoverListCount() throws Exception {
...
+            .build();
+    offset = 1;
+    // Create sink record with same offset
+    SinkRecord record2 =
+        new SinkRecord(
+            topic, partition, Schema.STRING_SCHEMA, "test", input.schema(), input.value(), offset);
+    // Should update the client sequencer
+    service2.insert(record2);
+    // call snowpipe ingest api
+    service.callAllGetOffset();
```
sfc-gh-japatel, line 1371: so this will throw an exception? Can you catch the exception and assert the exception type? I think we discussed that it will throw IngestException so may be we can just catch it? 

And then may be we can call  `service2.callAllGetOffset();` to see if offset 1 is getting ingested? 

In an ideal world we would expect the offset 0 to be ingested by `service2` (The newly woken up instance after rebalance) since precommit failed in `service0` and offset 0 was never ingested in SF.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-japatel: > when we create new pipe, the offset token is null when we call get pipe status

This will not happen in this call method right? because we are calling `ingestFilesWithClientInfo`

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-japatel: We can eventually get away with it if we have enough data about its completeness. Can you put a log here?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-japatel: Thank you, the idea here is to make the new code as little as possible to invoke when default/exactly once is set.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-japatel: Also, lets verify we have a logger in `insertFilesWithClientInfo's` on what we are passing in as offsetToken.. that way we can decide if this is a redundant call

sfc-gh-japatel (commented):
no major changes but will wait for one more pass. This is great stuff Andy. Good to see all tests being added.

src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
sfc-gh-japatel: I ran this test. we should catch exception and assert. 
I see all retries with 
```Message: Pipe has older version of clientSequencer. Please try again with latest clientSequencer.,
Data: null
}```
Which seems corrcet and behavior also seemed to be correct.

i.e on this exception technically there is another node with higher sequencer trying to ingest the data. so this instance of connector should stop working. i.e stop abruptly.

**Comment by sfc-gh-japatel** [comment]
Also, we might have to add google guava dependency 31.0jre since we removed that dependency from snowflake-java sdk . 

We can check that too in pom.xml and confluent_pom.xml

```
<dependency>
            <groupId>com.google.guava</groupId>
            <artifactId>guava</artifactId>
            <version>31.0.1-jre</version>
        </dependency>
```

**Comment by sfc-gh-anwang** [comment]
> Also, we might have to add google guava dependency 31.0jre since we removed that dependency from snowflake-java sdk .
> 
> We can check that too in pom.xml and confluent_pom.xml
> 
> ```
> <dependency>
>             <groupId>com.google.guava</groupId>
>             <artifactId>guava</artifactId>
>             <version>31.0.1-jre</version>
>         </dependency>
> ```

I added this dependency on my local branch, I'll check that in then.

File: src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java [hunk 9]
```diff
@@ -526,6 +515,26 @@ private void init(long recordOffset) {
       }
     }
 
+    private void initClientInfo() {
+      ConfigureClientResponse configureClientResponse = ingestionService.configureClient();
```
sfc-gh-japatel, line 519: A short description here would be very helpful. I know we usually dont put Javadocs on private method but it's just easier to follow the code. 

Also lets rename this to `initClientInfoForExactlyOnceDelivery`

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -526,6 +515,26 @@ private void init(long recordOffset) {
...
+      try {
+        if (offsetToken == null) {
+          this.offsetPersistedInSnowflake.set(-1);
+        } else {
+          this.offsetPersistedInSnowflake.set(Long.parseLong(offsetToken));
+        }
+        logInfo(
+            "Initialized client info, clientSequencer: {}, offsetToken: {}.",
+            this.clientSequencer.get(),
+            this.offsetPersistedInSnowflake.get());
```
sfc-gh-japatel, line 532: Awesome. can we also print pipeName from servicecontext?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -127,8 +127,8 @@ static void setFieldToDefaultValues(Map<String, String> config, String field, Lo
    * Get a property from the config map
    *
    * @param config connector configuration
-   * @param key name of the key to be retrived
-   * @return proprity value or null
+   * @param key name of the key to be retrieved
+   * @return property value or null
```
sfc-gh-japatel, line 131: thank you!

sfc-gh-japatel (approved):
lgtm! Way to go! 

@sfc-gh-anwang and I have performed End to End test locally and we will push this once we release 5.44 server side and we release Snowflake-ingest-java next version. 

@sfc-gh-anwang can you file a JIRA to add a new template to `/test/rest_request_template/` directory with `exactly_once`?

PR #356, opened 2021-12-08
Bugfix/break-on-reader-schema-error

anekdoti:
The Avro converter for the Snowflake Kafka connector has two options in case of a record value or key can not be consumed. By default, the record is dropped with a log message. To the contrary, if break.on.schema.registry.error is enabled, an exception is raised and the task should fail.

Previously, the second option was only implemented for the case that the writer schema can not be found in the schema registry. With the introduction of a reader schema that is used to read all keys / values, another failure condition got introduced: it is possible, that a record key / value can not be read with the given reader schema, e.g., because the reader schema contains a mandatory field that is not in the record key / value.

Until know, such an error lead to a dropped record and a log message. With this PR, the error leads to an exception and a failing task when break.on.schema.registry.error is enabled.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java
+++ b/src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java
@@ -168,7 +168,8 @@ public class SnowflakeAvroConverter extends SnowflakeConverter {
               id));
     } catch (Exception e) {
       if (breakOnSchemaRegistryError) {
-        throw SnowflakeErrors.ERROR_0010.getException(e);
+        throw SnowflakeErrors.ERROR_0010.getException(
+                "Failed to parse AVRO " + "record\n" + e.toString());
       } else {
         return logErrorAndReturnBrokenRecord(e, bytes);
       }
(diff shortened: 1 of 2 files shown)


src/main/java/com/snowflake/kafka/connector/records/SnowflakeAvroConverter.java
```diff
@@ -168,8 +168,7 @@ public SchemaAndValue toConnectData(final String s, final byte[] bytes) {
               id));
     } catch (Exception e) {
       if (breakOnSchemaRegistryError) {
-        throw SnowflakeErrors.ERROR_0010.getException(
-                "Failed to parse AVRO " + "record\n" + e.toString());
+        throw SnowflakeErrors.ERROR_0010.getException(e);
```
sfc-gh-japatel, line 171: should we keep the error message?

anekdoti: The original implementation dropped further meta data about the exception e, e.g., the stack strace. As I thought the exceptions from the Avro implementation are informative enough I decided to use the getException(...) implementation that just wraps the original exception. I see that this is a trade-off, and I am completely fine with reverting here.

anekdoti: I have reverted to the original error message now.

sfc-gh-japatel:
Apologies for delay, I was OOO. This looks good to me. Will try to release it in next version. (1.7.0)

sfc-gh-japatel (approved):
lgtm (Also, CLA has been previously signed)

PR #360, opened 2021-12-10
release KC 1.6.2 - Patch for log4j vulnerability

sfc-gh-japatel:
- Although the dependency is test only.

We will push this to master.


sfc-gh-rramachandran (approved):
LGTM!

PR #372, opened 2022-01-10
Bootstrap test containers for easier IT tests - Just basic

sfc-gh-japatel:
Kicking off Integration testing with TestContainers. https://www.testcontainers.org/

This is the defacto integration testing toolkit. It is much easier to set this up compared to what we have right now. 

It is also much simpler to set it up for a onboarding. 

I created a repository for setting up docker containers for Kafka, kafka connect and other related services. 
https://github.com/sfc-gh-japatel/snowflake-kafka-docker


sfc-gh-japatel:
This is not building right now since the new dependency is built with https://github.com/christophschubert/cp-testcontainers

sfc-gh-japatel:
> This is not building right now since the new dependency is built with https://github.com/christophschubert/cp-testcontainers

We either need to update our Java compiler plugin to 11.

PR #379, opened 2022-01-14
Refactor SinkServiceV1, PartitionBuffer and RecordService

sfc-gh-japatel:
- Just a Refactoring PR benefitted for StreamingIngest Based implementations.
- Also added some javadocs for more readability.
- No functional change.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/PartitionBuffer.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/PartitionBuffer.java
@@ -1,0 +1,64 @@
+package com.snowflake.kafka.connector.internal;
+
+import org.apache.kafka.connect.sink.SinkRecord;
+
+public abstract class PartitionBuffer<T> {
+  private int numOfRecord;
+  private int bufferSize;
+  private long firstOffset;
+  private long lastOffset;
+
+  public int getNumOfRecord() {
+    return numOfRecord;
+  }
+
+  public int getBufferSize() {
+    return bufferSize;
+  }
+
+  public long getFirstOffset() {
+    return firstOffset;
+  }
+
+  public long getLastOffset() {
+    return lastOffset;
+  }
+
+  public void setNumOfRecord(int numOfRecord) {
+    this.numOfRecord = numOfRecord;
+  }
+
+  public void setBufferSize(int bufferSize) {
+    this.bufferSize = bufferSize;
+  }
+
+  public void setFirstOffset(long firstOffset) {
+    this.firstOffset = firstOffset;
+  }
+
+  public void setLastOffset(long lastOffset) {
+    this.lastOffset = lastOffset;
+  }
+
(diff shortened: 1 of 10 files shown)


src/main/java/com/snowflake/kafka/connector/internal/PartitionBuffer.java
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * <p>Long lived would mean the data in partition would stay across two put APIs since the buffer
+ * thresholds were not met.
+ *
+ * <p>Please check respective implementation class for more details.
+ *
+ * @param <T> Return type of {@link #getData()}
+ */
+public abstract class PartitionBuffer<T> {
+  private int numOfRecord;
```
sfc-gh-tzhang, line 26: is numOfRecord = lastOffset-firstOffset? If yes, could we use it instead?

sfc-gh-tzhang (approved):
LGTM!

src/main/java/com/snowflake/kafka/connector/internal/PartitionBuffer.java
sfc-gh-mnaides: If I recall correctly there can be gaps in offsets so we probably want to stick with numOfRecords

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -1147,64 +1093,49 @@ public MetricRegistry getMetricRegistry() {
...
+     *
+     * <p>Please note {@link #insert(SinkRecord)} API is called from {@link
+     * com.snowflake.kafka.connector.SnowflakeSinkTask#put(Collection)} API and it is possible the
+     * buffered data is present across multiple PUT apis.
+     *
+     * <p>Check the usage of {@link #getData()} to understand when we would empty this buffer and
+     * when we would generate files in internal stage for snowpipe to ingest later using Snowpipe's
+     * REST APIs
+     */
+    private class SnowpipeBuffer extends PartitionBuffer<String> {
```
sfc-gh-mnaides, line 1107: We should come up with how we want to differentiate between file based snowpipe and snowpipe streaming

sfc-gh-mnaides (approved):
LGTM, we might want to take another pass at how we're naming things to differentiate between Snowpipe as it is today and Snowpipe Streaming

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-japatel: Yes. We will defined a new PartitionBuffer Type. (More on that in this PR: https://github.com/snowflakedb/snowflake-kafka-connector/pull/382/files)

PR #381, opened 2022-01-14
Two new Configs for Streaming Snowpipe

sfc-gh-japatel:
Introduce two new configs:

- snowflake.role.name
- snowflake.ingestion.method = SNOWPIPE/SNOWPIPE_STREAMING

PR #382, opened 2022-01-14
Implement SinkService for Streaming Ingest

sfc-gh-japatel:
Get the basics going for Streaming Ingest

- Implementing SinkServiceV2 for Streaming Ingest
- TopicPartitionChannel is a wrapper around StreamingIngestChannel

Doesn't include tests but will add it once I get some feedback.


src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -0,0 +1,426 @@
...
+ * <p>This implementation of SinkService uses Streaming Snowpipe (Streaming Ingestion)
+ *
+ * <p>Hence this initializes the channel, opens, closes. The StreamingIngestChannel resides inside
+ * {@link TopicPartitionChannel} which is per partition.
+ */
+public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(SnowflakeSinkServiceV2.class);
+
+  private static String STREAMING_CLIENT_PREFIX_NAME = "KC_CLIENT_";
```
sfc-gh-tzhang, line 49: Do we need a different name here for V2?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -0,0 +1,426 @@
...
+  private SnowflakeSinkConnectorConfig.BehaviorOnNullValues behaviorOnNullValues;
+
+  // default is true unless the configuration provided is false;
+  // If this is true, we will enable Mbean for required classes and emit JMX metrics for monitoring
+  private boolean enableCustomJMXMonitoring = SnowflakeSinkConnectorConfig.JMX_OPT_DEFAULT;
+
+  // default is at_least_once semantic (To begin with)
+  // TODO: SNOW-526435
+  private SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee ingestionDeliveryGuarantee =
+      SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.AT_LEAST_ONCE;
```
sfc-gh-tzhang, line 78: this won't apply here, right?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -0,0 +1,426 @@
...
+  public void startTask(String tableName, String topic, int partition) {
+    String partitionChannelKey = partitionChannelKey(topic, partition);
+    LOGGER.info(
+        "Opening a channel with name:{} for table name:{}, topic:{}, partition:{}",
+        partitionChannelKey,
+        tableName,
+        topic,
+        partition);
+    // the table should be present before opening a channel so lets do a table existence check here
+    createTableIfNotExists(tableName);
```
sfc-gh-tzhang, line 141: This is interesting, how would you create the table since you don't have the schema, do you?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -0,0 +1,426 @@
...
+  @Override
+  public void insert(Collection<SinkRecord> records) {
+    // note that records can be empty
+    for (SinkRecord record : records) {
+      // check if need to handle null value records
+      if (recordService.shouldSkipNullValue(record, behaviorOnNullValues)) {
+        continue;
+      }
+      // Might happen a count of record based flushing
+      insert(record);
```
sfc-gh-tzhang, line 166: What will happen if the insertion failed? I guess we have roughly two cases, if the error is channel related, we would want to reopen the channel and start with the last offset token. If the error is record related, we would want to insert it to the DLQ?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -0,0 +1,426 @@
...
+    return topic + "_" + partition;
+  }
+
+  // ------ Streaming Ingest Related Functions ------ //
+  private OpenChannelRequest getOpenChannelRequest(
+      final String channelName, final String tableName) {
+    return OpenChannelRequest.builder(channelName)
+        .setDBName(this.connectorConfig.get("snowflake.database.name"))
+        .setSchemaName(this.connectorConfig.get("snowflake.schema.name"))
+        .setTableName(tableName)
```
sfc-gh-tzhang, line 395: is the table name is fully qualified table name or without db and schema?

sfc-gh-tzhang: And we also add a required onErrorOption field, you probably want to update the jar file you use

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -0,0 +1,426 @@
...
+    LOGGER.info(
+        "Opening a channel with name:{} for table name:{}, topic:{}, partition:{}",
+        partitionChannelKey,
+        tableName,
+        topic,
+        partition);
+    // the table should be present before opening a channel so lets do a table existence check here
+    createTableIfNotExists(tableName);
+    SnowflakeStreamingIngestChannel partitionChannel =
+        streamingIngestClient.openChannel(getOpenChannelRequest(partitionChannelKey, tableName));
```
sfc-gh-tzhang, line 143: What if the channel has been used before? Do we need to start from the last offset token?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -0,0 +1,426 @@
...
+    Properties streamingClientProps = new Properties();
+    streamingClientProps.putAll(connectorConfig);
+    this.taskId = connectorConfig.getOrDefault(Utils.TASK_ID, "-1");
+    this.streamingIngestClientName =
+        STREAMING_CLIENT_PREFIX_NAME + conn.getConnectorName() + "_" + taskId;
+    this.streamingIngestClient =
+        SnowflakeStreamingIngestClientFactory.builder(this.streamingIngestClientName)
+            .setProperties(streamingClientProps)
+            .build();
+    this.partitionsToChannel = new HashMap<>();
```
sfc-gh-tzhang, line 128: Doesn't this need to be thread safe?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -0,0 +1,426 @@
...
+          >= (getFlushTime() * 1000)) {
+        LOGGER.info("Time based flush for channel:{}", partitionChannel.getChannelName());
+        LOGGER.info(
+            "Current:{}, previousFlushTime:{}, threshold:{}",
+            System.currentTimeMillis(),
+            partitionChannel.getPreviousFlushTimeStampMs(),
+            (getFlushTime() * 1000));
+        partitionChannel.insertBufferedRows();
+      }
+    }
```
sfc-gh-tzhang, line 181: This is interesting, ideally we want a background thread to do this, I assume this is probably the old behavior?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -0,0 +1,426 @@
...
+      final String channelName, final String tableName) {
+    return OpenChannelRequest.builder(channelName)
+        .setDBName(this.connectorConfig.get("snowflake.database.name"))
+        .setSchemaName(this.connectorConfig.get("snowflake.schema.name"))
+        .setTableName(tableName)
+        .build();
+  }
+
+  private void closeStreamingClient() {
+    // do we need to close the client? If I close, I will have to re init the client upon rebalance
```
sfc-gh-tzhang, line 400: good question, if you look at `SnowflakeStreamingIngestClient.close`, it does two things:
1. it will flush all the rows in all the channels, and mark the client as closed so no new inserts are allowed
2. It will check whether any of the channels have uncommitted rows, after flushing

I'm not sure how rebalance works, so I can't comment

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -0,0 +1,350 @@
...
+    bufferLock.lock();
+    try {
+      intermediateBuffer = streamingBuffer;
+      streamingBuffer = new StreamingBuffer();
+    } finally {
+      // release lock
+      bufferLock.unlock();
+    }
+    // intermediate buffer can be empty here if time interval reached but kafka produced no records.
+    if (intermediateBuffer.isEmpty()) {
```
sfc-gh-tzhang, line 211: a small optimization is that we could check the streamingBuffer before taking the lock, and skip if it's empty. I know there could be insert rows after the check, but I assume those rows could be flushed later.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -0,0 +1,350 @@
...
+    InsertValidationResponse response =
+        channel.insertRows(
+            (Iterable<Map<String, Object>>) intermediateBuffer.getData(),
+            Long.toString(intermediateBuffer.getLastOffset()));
+
+    // need to update the flush time(last I called insertrows)
+    this.previousFlushTimeStampMs = System.currentTimeMillis();
+
+    // should we retry if response.hasErrors() == true?
+    // how can we retry if buffer was cleared?
```
sfc-gh-tzhang, line 225: If the error is row related, usually I think retry won't help, you just need to insert it into the DLQ. If the error is channel related, we want to retry, but we will need to get the latest offset first.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -0,0 +1,350 @@
...
+        "Fetching last committed offset for partition channel:{}",
+        this.channel.getFullyQualifiedName());
+    String lastOffsetCommitted = channel.getLatestCommittedOffsetToken();
+    LOGGER.info(
+        "Last committed offset for partition channel:{} is :{}",
+        this.channel.getFullyQualifiedName(),
+        lastOffsetCommitted);
+    if (Strings.isNullOrEmpty(lastOffsetCommitted)
+        || lastOffsetCommitted.equalsIgnoreCase("null")) {
+      return committedOffset.get();
```
sfc-gh-tzhang, line 261: could we have a null or empty offset for kafka?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -0,0 +1,350 @@
...
+    } else {
+      return committedOffset.updateAndGet(operand -> Long.parseLong(lastOffsetCommitted));
+    }
+  }
+
+  public void closeChannel() {
+    try {
+      this.channel.close().get(1000, TimeUnit.MILLISECONDS);
```
sfc-gh-tzhang, line 269: could we start with no timeout? I'm not sure if 1 second is enough :)

sfc-gh-tzhang (commented):
Good stuff! Most of my comments are error handling related and let's see if the channel could be restarted itself when there is failure, but this PR is a good starting point!

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: From the initial thoughts, I don't think SnowflakeSinkServiceV2 is needed in this name. `KC_CLIENT ` seems a shorter version and we will also have connector name and taskID in streaming client name. 

Also, there isn't a similar name in V1. (Since there is no client name in ingest SDK for snowpipe)

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: I am hoping it doesnt apply here, but I havent much thought about this config for Streaming Ingestion. 

I know internally we do have rowSequencer which will help us get to EXACTLY_ONCE.

Also, this member (ingestionDeliveryGuarantee) isnt used anywhere in this file. So we are safe for now and I have logged the JIRA to tackle this separately. 

Potentially we will not support this config if streaming ingestion is used.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Not needed. Reason being:
1. Each of this instances are only assigned a set of partitions. 
2. Even if there are multiple tasks (SnowflakeSinkTask) assigned to a node (In case of multiple partitions) each of this tasks are only assigned its own set of partitions and no two partitions will overlap across nodes and tasks. 

So we are kind of okay to not make this thread safe. 
Reference:
- https://docs.confluent.io/platform/current/connect/devguide.html#sink-tasks
- https://kafka.apache.org/24/javadoc/org/apache/kafka/connect/sink/SinkTask.html
`Partition Assignment: After initialization, Connect will assign the task a set of partitions using open(Collection). These partitions are owned exclusively by this task until they have been closed with close(Collection).`

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: For now we decided we will go with just the two column schema right? Same as what we have right now.. 
(metadata and content both being variant columns)

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Hmm good question. So only during first init (open) https://kafka.apache.org/24/javadoc/org/apache/kafka/connect/sink/SinkTask.html we wont need the offset token. 

But lets say there is a rebalance, do we need to pass in offsetToken in openChannel? Also, the partitionChannelKey is still the same since the key is same before and after rebalance. 

I thought we would fetch offsetToken only when we get records from kafka and when we need to reset the offsets based on what we get from snowflake (getOffsetToken API)

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: so lines 169-181
- It is time based flush.
- Every 5second or so we have been invoked with put API which checks if time based flush has been hit. 
- It is kind of unimportant right now since we will not have any flush. (i.e we will not have any buffer)

Also, yes you are right, this is the V1 behavior.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: > is the table name is fully qualified table name or without db and schema?

No, only table name since DB and schema are added explicitly in config. And, table name is not provided, `tablename = f(topicName)` and there are some rules which we will stick with. 

https://docs.snowflake.com/en/user-guide/kafka-connector-install.html#required-privileges

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Also looking at code, offsetToken is fetched from an internal call. So I am assuming from the client side it is automatically handled if the same name is used?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: > And we also add a required onErrorOption field, you probably want to update the jar file you use

Will do. Thanks.

sfc-gh-mnaides:
FYI I'm having a hard time with some of the dependencies across PRs.  For example this PR merges into japatel-SNOW-447424-new-configs and https://github.com/snowflakedb/snowflake-kafka-connector/pull/381 merges japatel-SNOW-447424-new-configs into japatel-279791-master-streaming-snowpipe-prpr.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Thanks Toby. 

So rebalancing is pretty straight forward. It's like a server restart but new partitions can be assigned to same node and current partitions can be given to other node. 

So when a rebalance happens:
1. close() and 
2. open() 
APIS are called in that order. 


Follow up question. 
1. Should I use client close or channel close?
2. Do I need to init channel in open ()? I only do openChannel which is fine for first time because start() will be called but start() wont be called after rebalance. 
  a. So, will openChannel init the client or do I need to invoke isClosed()?

pom.xml
```diff
@@ -68,6 +68,12 @@
                 <enabled>true</enabled>
             </snapshots>
         </repository>
+
+<!--        Placeholder for local import of an unpublished library to maven repo. (i.e this is a local repository)-->
+        <repository>
+            <id>local-maven-repo</id>
+            <url>file:///${project.basedir}/local-maven-repo</url>
```
sfc-gh-mnaides, line 75: What about pushing a version of the sdk to nexus so we don't have to commit the binary to the repo?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Makes sense about the distinction.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: yes.. for first time. default is 0 to begin with when we return in precommit..

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Ahh I see. will do.

sfc-gh-japatel:
> FYI I'm having a hard time with some of the dependencies across PRs. For example this PR merges into japatel-SNOW-447424-new-configs and #381 merges japatel-SNOW-447424-new-configs into japatel-279791-master-streaming-snowpipe-prpr.

Sorry for confusion. 
Thats right, thats how I have kept my base branch. Ultimately all commits will go onto `japatel-279791-master-streaming-snowpipe-prpr`

And all PRs are on top of each other. 

So once I push `japatel-SNOW-447424-new-configs` into `japatel-279791-master-streaming-snowpipe-prpr`, I will rebase 
`japatel-SNOW-447418-implement-sinkservicev2` with `japatel-279791-master-streaming-snowpipe-prpr`

pom.xml
sfc-gh-japatel: Yeah, thats how I would like it too. 
Are we talking about a nexus repo internally(snowflake.int..) or onto public nexus?

I think there would be problems with accessing internal repo from github actions. 

Fetching it from public maven would be great. For that we would have to do a beta release of our sdk if we are ready. 

CC: @sfc-gh-tzhang

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -0,0 +1,427 @@
...
+ */
+public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(SnowflakeSinkServiceV2.class);
+
+  private static String STREAMING_CLIENT_PREFIX_NAME = "KC_CLIENT_";
+
+  // Assume next three values are a threshold after which we will call insertRows API
+  // Set in config (Time based flush) in seconds
+  private long flushTime;
```
sfc-gh-tjones, line 53: Let's add a time unit.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -0,0 +1,427 @@
...
+  // Set in config (buffer size based flush) in bytes
+  private long fileSize;
```
sfc-gh-tjones, line 55: Let's add a size unit.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-tjones, line 53: Can these all be `final`?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -0,0 +1,427 @@
...
+  // Set in config (Threshold before we call insertRows API) corresponds to # of
+  // records in kafka
+  private long recordNum;
+
+  // Used to connect to Snowflake
+  private final SnowflakeConnectionService conn;
+
+  private final RecordService recordService;
+  private final SnowflakeTelemetryService telemetryService;
+  private Map<String, String> topic2TableMap;
```
sfc-gh-tjones, line 66: nit, `topicToTableMap`

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-tjones: I'd say let's do `EXACTLY_ONCE` out the gate. It'll be weird otherwise to go to at least once using snowpipe (current), exactly once using snowpipe (being rolled out), at least once using streaming ingest (this pr), exactly once using streaming ingest. Actually, that would be a regression in behavior and we should definitely avoid that.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Will do. (Although we will remove all this in later versions since we dont need buffer logic). But let me change it.

sfc-gh-japatel: Yes.. one time setup during ctor. let me make them final.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: will do.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: I will change only variable names since changing method names would mix with feature and refactor.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Will do. Copy paste issue.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: That is true.. Will keep this in mind and tackle the JIRA SNOW-526435 in separate PR.

sfc-gh-japatel: We do have logic in place based on what we implemented for snowpipe and also we have similar APIs. so not too tough to do `exactly_once` from initial look.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-tzhang: Oh, i see, I think we just want to error out if the table doesn't exist, otherwise it feels like a regression from the customer's point after we have schematization

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-tzhang: yes, you're correct, we would fetch the offset token as part of the openChannel, so I assume you need to somehow reset the offset in Kafka after reopening each channel?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-tzhang: hmmm, I thought the conclusion was that the first version will still honor the flush time/buffer size, meaning we will have two buffers?

pom.xml
sfc-gh-tzhang: i think we're ready, let's do it! Is there a way to prevent random customers from using the beta release?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-tzhang: 1. You could just use client.close if all channels need to be closed, but one question I have is what would you do if there is a failure?
2. You need to reinit the channel if you have close it. Actually, will the `partitionsToChannel` be gone during rebalance? 

I guess it will be easier to schedule a quick call, i'm not sure about your timezone, do you want to schedule a chat on my calendar?

pom.xml
sfc-gh-mnaides: We can definitely do it from our internal nexus repo, I'll defer to @sfc-gh-tzhang and @sfc-gh-tjones on whether we're ready to push a beta to public maven

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: I think if table doesnt exist, it's a cleaner approach. Sorry, I dont understand how it will be a regression.

sfc-gh-japatel:
@sfc-gh-tzhang added tests which are running fine against QA1 but has been ignored for now.. 

Please take a look once you have time. Also modified few code path. 
1. Reinit channel. 
2. precommit offset returned == offset number found in SF.

pom.xml
sfc-gh-japatel: @sfc-gh-mnaides I am fine keeping it as is. And once it is released, I will remove jar and fetch it from maven. 

@sfc-gh-tzhang - Umm, we will have beta name in our release which itself will prevent customers from using it IMO.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: I added 2. in my recent commits. 

for 1, I will handle in other PR.(
failure handling)

sfc-gh-japatel: (I realized this while writing tests)

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Created a JIRA to discuss this further. SNOW-536661

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Here is what happens during rebalance. 
1. open(Partitions) followed by close(Partitions)
2. open(Partitions) is also called immediately after connect starts. 

We might not have to call close() channel from close(partition) because we will have to reopen them. More on that in further PRs. I will go with this current implementation which is much cleaner although involves few more remote calls.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: precommit = API kafka calls us immediately after put. We send offset no safe to commit back to kafka. 

This committedOffset = offset Ingested in SF + 1

Idea behind sending +1 is that kafka will send us offset starting from that number because we already ingested previous one.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -0,0 +1,371 @@
...
+                    insertError.getException().getMessage());
+              });
+    }
+  }
+
+  // should we rely on precommits frequency to give us the last committed offset token?
+  // or should we have a BG thread polling on every channel to update the Atomic long and simply
+  // return the atomic long's current value
+  // For now we will rely on preCommit API to fetch the inserted offsets into Snowflake
+  // TODO: SNOW-529755 POLL committed offsets in backgraound thread
```
sfc-gh-tzhang, line 251: Is there any correctness issue if we reply on precommit frequency?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -0,0 +1,371 @@
...
+    LOGGER.debug(
+        "Fetching last committed offset for partition channel:{}",
+        this.channel.getFullyQualifiedName());
+    String lastOffsetCommitted = channel.getLatestCommittedOffsetToken();
+    LOGGER.info(
+        "Last committed offset for partition channel:{} is :{}",
+        this.channel.getFullyQualifiedName(),
+        lastOffsetCommitted);
+    if (Strings.isNullOrEmpty(lastOffsetCommitted)
+        || lastOffsetCommitted.equalsIgnoreCase("null")) {
```
sfc-gh-tzhang, line 277: I don't think the lastOffsetCommitted could be "null" string, it should be either valid and NULL

sfc-gh-tzhang (approved):
The change overall looks good now! I left some comments and questions, PTAL, will give a final look when it's merged into master. Thanks!

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceFactory.java
```diff
@@ -43,9 +44,7 @@ private SnowflakeSinkServiceBuilder(
       if (ingestionType == IngestionMethodConfig.SNOWPIPE) {
         this.service = new SnowflakeSinkServiceV1(conn);
       } else {
-        // Use SinkServiceV2
-        // TODO:SNOW-447418
-        service = null;
+        service = new SnowflakeSinkServiceV2(conn, connectorConfig);
```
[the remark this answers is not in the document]
sfc-gh-japatel: Will do!

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Update: We will keep the buffer params, essentially saying we will have double buffer. 
As far as BG thread is concerned, we dont necessarily need to do for PrPr, will log a JIRA and see if this can be done.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -0,0 +1,441 @@
...
+    return partitionsToChannel.size();
+  }
+
+  @Override
+  public void callAllGetOffset() {
+    // undefined
+  }
+
+  @Override
+  public void closeAll() {
```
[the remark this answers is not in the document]
sfc-gh-japatel: `CloseAll()` method is something which is not invoked from Kafka framework. Only `close()` is called. `CloseAll()` is infact called if there is a dangling reference to this service instance when a connector starts. (Check usage of `closeAll()`)

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -0,0 +1,441 @@
...
+    closeStreamingClient();
+  }
+
+  @Override
+  public void setIsStoppedToTrue() {}
+
+  /* Undefined */
+  @Override
+  public boolean isClosed() {
+    return false;
```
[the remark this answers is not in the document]
sfc-gh-japatel: This is not for client closed. This is undefined since it is overridden from SnowflakeSinkService interface.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Okay makes sense, I confirmed from gs side, we only set a valid one and if it is not set, we return null.

sfc-gh-japatel: I will remove the second condition. Thanks

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Answering specific questions as this can easily be forgotten. 

> You could just use client.close if all channels need to be closed, but one question I have is what would you do if there is a failure?
I will handle this in failure handling PR. If closing has failed, it should be fine since there is no potential impact. May be reopening wouldn't work since it was open. 

> You need to reinit the channel if you have close it. Actually, will the partitionsToChannel be gone during rebalance?
partitionsToChannel will be gone if close() completed successfully.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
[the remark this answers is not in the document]
sfc-gh-tzhang: then can we just get the list of partitions from `partitionsToChannel` and then call `close` on it? It looks more clear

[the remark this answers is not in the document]
sfc-gh-japatel: In close, partitions are passed in so we use that. In closeAll we use partitionsToChannel

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Not really, precommit is called immediately after put (in the same poll thread). Also even if there are not messages, we still received a call to put with 0 messages which means precommit will be called.  

offset.flush.interval.ms is defaulted to 60 seconds. And this can be set at connect level. 

Also, it is not necessary we need to commit data in precommit. It is very much possible that 
- put(1 record) -> puts in buffer
- precommit(1 record) we still return 0 since insertrows was not called. 

But in subsequent put APIs calls which will again call precommit will fetch offsetToken 1 since insertRows would have been called ultimately. 


There is a potential problem that flush set by customers is < offset.flush.interval.ms, in this case latency will be high since flush time would have passed but we could only know that after `offset.flush.interval.ms`

sfc-gh-japatel: For context here is kafka code: https://github.com/apache/kafka/blob/830d83e2cde4f33735128b533cacc8b8ea23a2c0/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSinkTask.java#L210

PR #389, opened 2022-02-01
Enable V2IT tests to run against qa1

sfc-gh-japatel:


sfc-gh-japatel:
Just realised this will fail since dev vpn is required for qa1! Will close this.

PR #390, opened 2022-02-02
SNOW-529753 Handle rebalance in SnowflakeSinkTask

sfc-gh-japatel:
Rebalance [documentation](https://kafka.apache.org/11/javadoc/org/apache/kafka/connect/sink/SinkTask.html): 

- Close() will not invoke close on channel
  - `close()` in sinktask is invoked before rebalance and `open()` is called. 
  - It will also not clear the partitionsToChannel map. 
    - One can argue that this map can grow but most of the times, same partitions are assigned and very rarely partitions are added. Even then this map cant grow more than no of partitions. 
- Made channel object in TopicPartitionChannel mutable. This is because TopicPartitionChannel object doesnt need to be wiped off when rebalance happens, but client is closed (Which results into channel being closed) So we have to open channel after rebalance. 
- Moved code for more readability. 
- Added test which replicates the rebalance scenario and also kind of tests exactly once. 

Note: EOS case will be handled in separate PR but this one handles the case when there was a rebalance and not restart.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
@@ -125,19 +125,27 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
   @Override
   public void startTask(String tableName, String topic, int partition) {
     String partitionChannelKey = partitionChannelKey(topic, partition);
-    LOGGER.info(
-        "Opening a channel with name:{} for table name:{}, topic:{}, partition:{}",
-        partitionChannelKey,
-        tableName,
-        topic,
-        partition);
+
     // the table should be present before opening a channel so lets do a table existence check here
     createTableIfNotExists(tableName);
+
+    createOrUpdatePartitionsToChannelCache(partitionChannelKey, tableName);
+  }
+
+  /* We will always open a new channel. This will either update the TopicPartitionChannel's streamingChannel or will create a new instance of TopicPartitionChannel. */
+  private void createOrUpdatePartitionsToChannelCache(
+      final String partitionChannelKey, final String tableName) {
+    // Always open a new channel.
     SnowflakeStreamingIngestChannel partitionChannel =
         openChannelForTable(partitionChannelKey, tableName);
-    TopicPartitionChannel topicPartitionChannel =
-        new TopicPartitionChannel(partitionChannel, this.conn, tableName);
-    partitionsToChannel.putIfAbsent(partitionChannelKey, topicPartitionChannel);
+    if (partitionsToChannel.containsKey(partitionChannelKey)) {
+      // we just update the channel instance for this partition
+      partitionsToChannel.get(partitionChannelKey).updateStreamingIngestChannel(partitionChannel);
+    } else {
+      // Add channel and related metadata to partitionsToChannel map
+      TopicPartitionChannel topicPartitionChannel = new TopicPartitionChannel(partitionChannel);
+      partitionsToChannel.put(partitionChannelKey, topicPartitionChannel);
+    }
   }
 
   /**
@@ -184,8 +192,9 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
   @Override
   public void insert(SinkRecord record) {
     String partitionChannelKey = partitionChannelKey(record.topic(), record.kafkaPartition());
-    // init a new topic partition
-    if (!partitionsToChannel.containsKey(partitionChannelKey)) {
+    // init a new topic partition if it not present in cache or if channel is closed
+    if (!partitionsToChannel.containsKey(partitionChannelKey)
+        || partitionsToChannel.get(partitionChannelKey).isChannelClosed()) {
       LOGGER.warn(
           "Topic: {} Partition: {} hasn't been initialized by OPEN " + "function",
           record.topic(),
@@ -246,17 +255,24 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
     closeStreamingClient();
   }
 
+  /**
+   * This function is called during rebalance. Please ensure we dont wipe of state when we are
+   * invoking this function.
+   *
+   * <p>The channel is not closed since we would need to reiniate the channel.
+   *
+   * <p>Also, wiping off partitionsToChannel map would mean we would will need to reinitialise and
+   * we will lose processedOffset information.
+   *
+   * @param partitions a list of topic partition
+   */
   @Override
   public void close(Collection<TopicPartition> partitions) {
     partitions.forEach(
         topicPartition -> {
-          String partitionChannelKey =
-              partitionChannelKey(topicPartition.topic(), topicPartition.partition());
-          LOGGER.info("Closing partition channel:{}", partitionChannelKey);
-          partitionsToChannel.get(partitionChannelKey).closeChannel();
-          partitionsToChannel.remove(partitionChannelKey);
+          LOGGER.info(
+              "Closing partition {}, topic:{}", topicPartition.topic(), topicPartition.partition());
         });
-
     closeStreamingClient();
   }
 
@@ -396,6 +412,7 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
     if (streamingIngestClient.isClosed()) {
       initStreamingClient();
     }
+    LOGGER.info("Opening a channel with name:{} for table name:{}", channelName, tableName);
     return streamingIngestClient.openChannel(channelRequest);
   }
 
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
@@ -256,15 +256,12 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
   }
 
   /**
-   * This function is called during rebalance. Please ensure we dont wipe of state when we are
-   * invoking this function.
+   * This function is called during rebalance. Please ensure we dont wipe off state when we are
+   * invoking this function. i.e after rebalance, we would still want to preserve metadata like
+   * streamingBuffer, processedOffset
    *
-   * <p>The channel is not closed since we would need to re-initiate the channel.
-   *
-   * <p>Also, wiping off partitionsToChannel map would mean we would need to reinitialise and we
-   * will lose processedOffset information.
-   *
-   * <p>We will only close the client.
+   * <p>All the channels are closed. The client is still active. Upon rebalance, (inside {@link
+   * com.snowflake.kafka.connector.SnowflakeSinkTask#open(Collection)} we will reopen the channel.
    *
    * @param partitions a list of topic partition
    */
@@ -272,10 +269,17 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
   public void close(Collection<TopicPartition> partitions) {
     partitions.forEach(
         topicPartition -> {
+          final String partitionChannelKey =
+              partitionChannelKey(topicPartition.topic(), topicPartition.partition());
+          TopicPartitionChannel topicPartitionChannel =
+              partitionsToChannel.get(partitionChannelKey);
+          topicPartitionChannel.closeChannel();
           LOGGER.info(
-              "Closing partition {}, topic:{}", topicPartition.topic(), topicPartition.partition());
+              "Closing partitionChannel:{}, partition:{}, topic:{}",
+              topicPartitionChannel.getChannelName(),
+              topicPartition.topic(),
+              topicPartition.partition());
         });
-    closeStreamingClient();
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -123,19 +123,27 @@ public SnowflakeSinkServiceV2(
...
     // the table should be present before opening a channel so lets do a table existence check here
     createTableIfNotExists(tableName);
+
+    createOrUpdatePartitionsToChannelCache(partitionChannelKey, tableName);
+  }
+
+  /* We will always open a new channel. This will either update the TopicPartitionChannel's streamingChannel or will create a new instance of TopicPartitionChannel. */
+  private void createOrUpdatePartitionsToChannelCache(
+      final String partitionChannelKey, final String tableName) {
+    // Always open a new channel.
```
sfc-gh-tzhang, line 136: Have you consider the suggestion about don't close any channel during `close` and only open the channel if `partitionsToChannel.containsKey = false`?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -244,17 +253,26 @@ public void closeAll() {
...
+  /**
+   * This function is called during rebalance. Please ensure we dont wipe of state when we are
+   * invoking this function.
+   *
+   * <p>The channel is not closed since we would need to re-initiate the channel.
+   *
+   * <p>Also, wiping off partitionsToChannel map would mean we would need to reinitialise and we
+   * will lose processedOffset information.
+   *
+   * <p>We will only close the client.
```
sfc-gh-tzhang, line 265: I don't think this will work, closing the client will close all channels

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -308,7 +317,14 @@ public long getPreviousFlushTimeStampMs() {
   }
 
   public String getChannelName() {
-    return this.channel.getFullyQualifiedTableName();
+    return this.channel.getFullyQualifiedName();
```
sfc-gh-tzhang, line 320: ah, good catch!

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -308,7 +317,14 @@ public long getPreviousFlushTimeStampMs() {
...
+  }
+
+  public String toString() {
+    return "StreamingChannelName:"
+        + getChannelName()
+        + ", FullyQualifiedTableName:{}"
+        + this.channel.getFullyQualifiedTableName();
```
sfc-gh-tzhang, line 327: why not just use this.channel.getFullyQualifiedName? it will contains both channel name and table name

sfc-gh-japatel: Sounds like an idea.. let me do that..

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: The only reason why I am doing a closing of client is because I want to close all underlying channels. 

consider this scenario: there is a rebalance but the task is still running, in that case, if I am trying to reopen a channel which was already opened, it will throw up because the channel name can also be deterministic right? 

What would be side effect of closing and reopening of channels?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: same as above.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -123,19 +123,27 @@ public SnowflakeSinkServiceV2(
...
+      final String partitionChannelKey, final String tableName) {
+    // Always open a new channel.
     SnowflakeStreamingIngestChannel partitionChannel =
         openChannelForTable(partitionChannelKey, tableName);
-    TopicPartitionChannel topicPartitionChannel =
-        new TopicPartitionChannel(partitionChannel, this.conn, tableName);
-    partitionsToChannel.putIfAbsent(partitionChannelKey, topicPartitionChannel);
+    if (partitionsToChannel.containsKey(partitionChannelKey)) {
+      // we just update the channel instance for this partition
+      partitionsToChannel.get(partitionChannelKey).updateStreamingIngestChannel(partitionChannel);
```
sfc-gh-mnaides, line 141: Double checking we're ok with not updating offsets and the StreamingBuffer in this case where we just set a new StreamingIngestChannel

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -38,11 +36,14 @@
...
 public class TopicPartitionChannel {
   private static final Logger LOGGER = LoggerFactory.getLogger(TopicPartitionChannel.class);
-  private static final ObjectMapper MAPPER = new ObjectMapper();
 
   // more of previousInsertRowTsMs
   private long previousFlushTimeStampMs;
 
+  /**
+   * States whether this channel has had any data before. If this is false, the
+   * topicPartitionChannel has recently been initialised and didnt receive any records before
```
sfc-gh-mnaides, line 45: nit:  The comment is a little confusing since the variable `hasChannelInitialized ` is false if the channel has recently been initialized

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: that is correct! We will only update channel since we closed it when close() on sinktask was called. (partition rebalance)

I think it is okay to not reset buffer and processed offset. 

One of the primary reason I didnt want to recreate topicpartitonchannel object was to preserve processed offset after rebalance. so that we could start off from where we left..

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-tzhang: no, what I mean here is that all the channels will be closed as well when you close the client, so what you said here and at line 260 is not correct

sfc-gh-tzhang: Also, can we update `closeAll` to call `closeStreamingClient` directly instead of closing channels one by one?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-tzhang: Actually, I just checked the close channel logic, we won't call flush if the channel is empty, so for the rebalance case, we should only flush once even we have 100 channels, what we should do is instead of calling `closeStreamingClient`, just close the channels one by one.

sfc-gh-japatel: Well, the closing of client internally calls markClosed() which is what each channel close does. So this is much simpler IMO..

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: https://github.com/snowflakedb/snowflake-kafka-connector/pull/391/files#

Check topic partition channel logic. https://github.com/snowflakedb/snowflake-kafka-connector/blob/japatel-SNOW-536429-exactly-once/src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java#L117

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Hmm okay, I think what you are saying is we dont need to close client but we can close all channels. so that we dont need to reopen the client?

Agreed that line 260 is misleading..

sfc-gh-japatel: Let me clean this up..

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: hmm, may be I can rename it to something more descriptive.. 
`hasChannelReceivedAnyRecordsBefore`

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: But I see the benefit of not closing the client. I will remove that from close() and keep it in closeAll() only.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Now that I am thinking, there is also no harm in calling getCommittedOffsetToken in `startTask()` too to fetch the latest committed offset.

sfc-gh-japatel: I will handle it in subsequent PRs.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -244,18 +253,31 @@ public void closeAll() {
...
         topicPartition -> {
-          String partitionChannelKey =
+          final String partitionChannelKey =
               partitionChannelKey(topicPartition.topic(), topicPartition.partition());
-          LOGGER.info("Closing partition channel:{}", partitionChannelKey);
-          partitionsToChannel.get(partitionChannelKey).closeChannel();
-          partitionsToChannel.remove(partitionChannelKey);
+          TopicPartitionChannel topicPartitionChannel =
+              partitionsToChannel.get(partitionChannelKey);
+          topicPartitionChannel.closeChannel();
```
sfc-gh-tzhang, line 274: We would need to handle the potential exceptions thrown during close channel, we could just catch it and ignore it because we will start with the latest committed offset

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskForStreamingIT.java
```diff
@@ -0,0 +1,201 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Ignore;
+import org.junit.Test;
+
+/**
+ * Sink Task IT test which uses {@link
+ * com.snowflake.kafka.connector.internal.streaming.SnowflakeSinkServiceV2}
+ */
+public class SnowflakeSinkTaskForStreamingIT {
```
sfc-gh-tzhang, line 37: would be good to add some test cases to test failure cases, for example, what if closeChannel throws exception?

sfc-gh-tzhang (approved):
LGTM!

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Sounds good, let me do that..thanks

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskForStreamingIT.java
sfc-gh-japatel: Makes sense. I will add it along with the code change where I will catch all exceptions in close()

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: [SNOW-542734](https://snowflakecomputing.atlassian.net/browse/SNOW-542734)

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Actually, I already catch that in `closeChannel()`

sfc-gh-mnaides (approved):

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskForStreamingIT.java
sfc-gh-japatel: Will add it in TopicPartitionChannel UT which is not part of this change..although the close/reopen case is taken care of in this IT test.

PR #391, opened 2022-02-02
SNOW-536429 ☝️ exactly once for Streaming Ingest

sfc-gh-japatel:
TLDR:

- No need to pass in `delivery.guarantee` into config.
- If they pass in, it should be EXACTLY_ONCE
- `AT_LEAST_ONCE` will error out. 

Logic:
- Cases where we need to be cautious. 
  - The insertRows was called but precommit didnt send that because of a rebalance or shutdown. 
  - In this case the offset can be resent back to KC and then we will discard it because we will also fetch it from Snowflake. 

- On a connector start, just fetch offsetToken for this channel. 
  - The channel would have been already opened before. (from startTask)
  - This value will be the source of truth and we will compare it with the incoming offset from Kafka.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
@@ -171,11 +171,16 @@ public class SnowflakeSinkTask extends SinkTask {
         Boolean.parseBoolean(parsedConfig.get(SnowflakeSinkConnectorConfig.REBALANCING));
 
     // default to snowpipe
+    // If it is snowpipe_streaming, set delivery guarantee to exactly once.
     IngestionMethodConfig ingestionType = IngestionMethodConfig.SNOWPIPE;
     if (parsedConfig.containsKey(SnowflakeSinkConnectorConfig.INGESTION_METHOD_OPT)) {
       ingestionType =
           IngestionMethodConfig.valueOf(
               parsedConfig.get(SnowflakeSinkConnectorConfig.INGESTION_METHOD_OPT).toUpperCase());
+      if (ingestionType.equals(IngestionMethodConfig.SNOWPIPE_STREAMING)) {
+        ingestionDeliveryGuarantee =
+            SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.EXACTLY_ONCE;
+      }
     }
 
     conn =
--- a/src/main/java/com/snowflake/kafka/connector/Utils.java
+++ b/src/main/java/com/snowflake/kafka/connector/Utils.java
@@ -548,6 +548,14 @@ public class Utils {
                     config.get(INGESTION_METHOD_OPT)));
             configIsValid = false;
           }
+          // setting delivery guarantee to EOS.
+          // It is fine for customer to not set this value if Streaming SNOWPIPE is used.
+          deliveryGuarantee =
+              SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.of(
+                  config.getOrDefault(
+                      DELIVERY_GUARANTEE,
+                      SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.EXACTLY_ONCE.name()));
+
           if (deliveryGuarantee.equals(
               SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.AT_LEAST_ONCE)) {
             LOGGER.error(
--- a/src/main/java/com/snowflake/kafka/connector/Utils.java
+++ b/src/main/java/com/snowflake/kafka/connector/Utils.java
@@ -552,10 +552,11 @@ public class Utils {
               SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.AT_LEAST_ONCE)) {
             LOGGER.error(
                 Logging.logMessage(
-                    "Config:{} should be:{}",
+                    "Config:{} should be:{} if ingestion method is:{}",
                     DELIVERY_GUARANTEE,
-                    SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.EXACTLY_ONCE
-                        .toString()));
+                    SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.EXACTLY_ONCE.toString(),
+                    IngestionMethodConfig.SNOWPIPE_STREAMING.toString()));
+            configIsValid = false;
           }
         }
       } catch (ConfigException exception) {
--- a/src/main/java/com/snowflake/kafka/connector/Utils.java
+++ b/src/main/java/com/snowflake/kafka/connector/Utils.java
@@ -512,11 +512,15 @@ public class Utils {
       }
     }
 
+    // This is default for SNOWPIPE based KC because that is GA.
+    SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee deliveryGuarantee =
+        SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.AT_LEAST_ONCE;
     try {
-      SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.of(
-          config.getOrDefault(
-              DELIVERY_GUARANTEE,
-              SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.AT_LEAST_ONCE.name()));
+      deliveryGuarantee =
+          SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.of(
+              config.getOrDefault(
+                  DELIVERY_GUARANTEE,
+                  SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.AT_LEAST_ONCE.name()));
     } catch (IllegalArgumentException exception) {
       LOGGER.error(
           Logging.logMessage(
@@ -524,6 +528,8 @@ public class Utils {
       configIsValid = false;
     }
 
+    // For snowpipe_streaming, role should be non empty and delivery guarantee should be exactly
+    // once. (Which is default)
     if (config.containsKey(INGESTION_METHOD_OPT)) {
       try {
         // This throws an exception if config value is invalid.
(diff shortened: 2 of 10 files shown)


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -171,11 +171,16 @@ public void start(final Map<String, String> parsedConfig) {
         Boolean.parseBoolean(parsedConfig.get(SnowflakeSinkConnectorConfig.REBALANCING));
 
     // default to snowpipe
+    // If it is snowpipe_streaming, set delivery guarantee to exactly once.
     IngestionMethodConfig ingestionType = IngestionMethodConfig.SNOWPIPE;
     if (parsedConfig.containsKey(SnowflakeSinkConnectorConfig.INGESTION_METHOD_OPT)) {
```
sfc-gh-tjones, line 176: Can the user set this? If they can and they set the delivery guarantee to anything other than exactly once we should error out.

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -542,6 +548,24 @@ static String validateConfig(Map<String, String> config) {
...
+
+          if (deliveryGuarantee.equals(
+              SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.AT_LEAST_ONCE)) {
+            LOGGER.error(
+                Logging.logMessage(
+                    "Config:{} should be:{} if ingestion method is:{}",
+                    DELIVERY_GUARANTEE,
+                    SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.EXACTLY_ONCE.toString(),
+                    IngestionMethodConfig.SNOWPIPE_STREAMING.toString()));
+            configIsValid = false;
```
sfc-gh-tjones, line 567: Presumably something checks this and throws an exception if false?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -129,19 +135,28 @@ public void insertRecordToBuffer(SinkRecord record) {
...
+      if (offsetToken == null) {
+        this.offsetPersistedInSnowflake.set(-1);
+      } else {
+        this.offsetPersistedInSnowflake.set(Long.parseLong(offsetToken));
+      }
+      LOGGER.info(
+          "Fetched offsetToken:{} for channelName:{} during channel initialization.",
+          this.offsetPersistedInSnowflake.get(),
+          this.getChannelName());
+    } catch (NumberFormatException e) {
```
sfc-gh-tjones, line 151: Should we throw an exception here? Silently continuing is a bit scary IMO

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -129,19 +135,28 @@ public void insertRecordToBuffer(SinkRecord record) {
...
+      LOGGER.error(
+          "The offsetToken string does not contain a parsable long:{},"
+              + "offsetTokenLastRetrieved:{}. ",
```
sfc-gh-tjones, line 154: Should we log the value of `offsetToken` here?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -91,14 +95,16 @@ protected void updateStreamingIngestChannel(SnowflakeStreamingIngestChannel upda
...
+
       this.hasChannelInitialized = true;
     }
 
-    // ignore ingested files
-    if (record.kafkaOffset() > processedOffset.get()) {
+    // discard the record if the record offset is smaller or equal to server side offset, or if
+    // record is smaller than any other record offset we received before
+    if (record.kafkaOffset() > this.offsetPersistedInSnowflake.get()
+        && record.kafkaOffset() > processedOffset.get()) {
```
sfc-gh-tjones, line 107: Not sure if this is logged already but I think it would be good to capture what KC decides to use when it starts for a given topic partition in the event that we discover a bug in which we pick the wrong offset token.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -129,19 +135,28 @@ public void insertRecordToBuffer(SinkRecord record) {
...
-  private void init(long recordOffset) {
+  /* For EOS, fetch the offset from snowflake during only during first time after connector starts and record is received in Kafka Connector */
+  private void fetchLastCommittedOffsetToken() {
+    String offsetToken = this.channel.getLatestCommittedOffsetToken();
     try {
-      startCleaner(recordOffset);
-    } catch (Exception e) {
-      LOGGER.error("Cleaner and Flusher threads shut down before initialization");
+      if (offsetToken == null) {
+        this.offsetPersistedInSnowflake.set(-1);
```
sfc-gh-tjones, line 143: Should we pull `-1L` as a constant somewhere and treat that as "NO_OFFSET_TOKEN" here and elsewhere?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -71,6 +71,10 @@
   // added to buffer before calling insertRows
   private final AtomicLong processedOffset; // processed offset
 
+  // This offset is updated when Snowflake has received offset from insertRows API
+  // We will update this value after calling offsetToken API for this channel
+  private final AtomicLong offsetPersistedInSnowflake = new AtomicLong(-1);
```
sfc-gh-tjones, line 76: Why is this an `AtomicLong`?

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-japatel: Correct!

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-japatel: Correct, we will err out if they mistakenly set this as atleast once

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: This shouldn't happen, and is just a cautious check. But yes, we should fail starting the connector if this happens.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: yes, my bad. thanks

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Yeah, good idea, let me log it in when hasChannelInitialized= false;

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Will do. "no magic constants"

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Umm good point. It doesnt need to be AtomicLong at the moment.

sfc-gh-japatel: Agreed, there is no atomic operation needed and it is single threaded at the moment.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -171,11 +171,16 @@ public void start(final Map<String, String> parsedConfig) {
...
       ingestionType =
           IngestionMethodConfig.valueOf(
               parsedConfig.get(SnowflakeSinkConnectorConfig.INGESTION_METHOD_OPT).toUpperCase());
+      if (ingestionType.equals(IngestionMethodConfig.SNOWPIPE_STREAMING)) {
+        ingestionDeliveryGuarantee =
+            SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.EXACTLY_ONCE;
+      }
```
sfc-gh-tjones, line 183: Should this be in the `validateConfig` file below? Might be easier to consolidate the logic that determines which ingestion delivery guarantee to use.

sfc-gh-tjones (approved):

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-japatel: Validation is already done in Utils.validateConfig.. here we just fetch the values from config so that we can decide which SnowflakeSinkService to initialize.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -131,19 +146,31 @@ public void insertRecordToBuffer(SinkRecord record) {
...
+    String offsetToken = this.channel.getLatestCommittedOffsetToken();
     try {
-      startCleaner(recordOffset);
-    } catch (Exception e) {
-      LOGGER.error("Cleaner and Flusher threads shut down before initialization");
+      if (offsetToken == null) {
+        this.offsetPersistedInSnowflake = NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE;
+      } else {
+        this.offsetPersistedInSnowflake = Long.parseLong(offsetToken);
+      }
```
sfc-gh-tjones, line 158: Nit, this can be consolidated to `this.offsetPersistedInSnowflake = offsetToken == null ? NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE : Long.parseLong(offsetToken);`

sfc-gh-tjones (approved):

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Will do, however, I will handle that as part of https://github.com/snowflakedb/snowflake-kafka-connector/pull/395/commits (Which is still in draft) That PR refactors some of the common code paths.

PR #392, opened 2022-02-03
Implement ErrantRecordReporter - kafka records from Sink Connector to DLQ

sfc-gh-japatel:
- This is only available after 2.6.0
- Tested this with confluent 6.2.0

Background
- For snowpipe based implementation of KC, we used to send broken records coming from Native converters into tableStage. 
  - Native converters are custom converters created for snowflake so that we can put records into tablestage. https://www.confluent.io/blog/kafka-connect-deep-dive-converters-serialization-explained/
  - Take a look at `SnowflakeRecordContent` class to know more about what extra fields we add in converter logic before it reaches Kafka connector. 
  - Also take a look at `SnowflakeConverter` abstract class and its implementation which essentially creates JSON bytes from kafka record and then send it to KC. 

Segway to why we dont need TableStage
- We can send records to DLQ. 
- DLQ is like any other topic but just has special name. 
- Messages can be fed into snowflake again from DLQ. (Using that at source topic and using SF KC with a simpler converter like String)
- It can be used for any failures.   
  - Network failures. 
  - Connection issues with Snowflake. 
  - Failures after multiple retries. 
  - Broken records. (When we cannot convert data to schema based json)
    - These broken records are special records which are sent to KC when custom converters are used. 
    - They can directly be sent to DLQ if community converters are used. 
    - So only benefit of using custom converters is to be able to send it to tablestage. 

This PR:
This PR is just setting up the framework and is backwards compatible with Custom Snowflake converter. It is just that it wont work for Apache Kafka version < 2.6.0

For version < 2.6.0 we can only support logging it as error and stopping connector.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
@@ -18,6 +18,8 @@ package com.snowflake.kafka.connector;
 
 import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.DELIVERY_GUARANTEE;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.snowflake.kafka.connector.dlq.KafkaRecordErrorReporter;
 import com.snowflake.kafka.connector.internal.Logging;
 import com.snowflake.kafka.connector.internal.SnowflakeConnectionService;
 import com.snowflake.kafka.connector.internal.SnowflakeConnectionServiceFactory;
@@ -30,11 +32,14 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Supplier;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.errors.RetriableException;
+import org.apache.kafka.connect.sink.ErrantRecordReporter;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.apache.kafka.connect.sink.SinkTask;
 import org.slf4j.Logger;
@@ -170,6 +175,8 @@ public class SnowflakeSinkTask extends SinkTask {
     enableRebalancing =
         Boolean.parseBoolean(parsedConfig.get(SnowflakeSinkConnectorConfig.REBALANCING));
 
+    KafkaRecordErrorReporter kafkaRecordErrorReporter = noOpKafkaRecordErrorReporter();
+
     // default to snowpipe
     // If it is snowpipe_streaming, set delivery guarantee to exactly once.
     IngestionMethodConfig ingestionType = IngestionMethodConfig.SNOWPIPE;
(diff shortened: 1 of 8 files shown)


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -415,4 +424,46 @@ void processRebalancingTest() {
       }
     }
   }
+
+  /* Useful to report a record back to DLQ if error tolerance is specified */
```
sfc-gh-tjones, line 428: Nit `Used` instead of `Useful`?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -415,4 +424,46 @@ void processRebalancingTest() {
...
+      try {
+        ErrantRecordReporter errantRecordReporter = context.errantRecordReporter();
+        if (errantRecordReporter != null) {
+          result =
+              (record, error) -> {
+                try {
+                  // Blocking this until record is delivered to DLQ
+                  errantRecordReporter.report(record, error).get();
+                } catch (InterruptedException | ExecutionException e) {
+                  final String errMsg = "ERROR reporting records to ErrantRecordReporter";
```
sfc-gh-tjones, line 441: Nit, inline `errMsg` below?

sfc-gh-tjones (approved):

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-japatel: Wanted to use it for logging as well as in connect exception.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-japatel: Will do!

sfc-gh-mnaides (approved):

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -128,7 +135,10 @@ public void insertRecordToBuffer(SinkRecord record) {
 
       // broken record
       if (isRecordBroken(snowflakeRecord)) {
-        // write it to DLQ SNOW-451197
+        // check for error tolerance and log tolerance values
+        // errors.log.enable and errors.tolerance
+        LOGGER.debug("Broken record offset:{}, topic:{}", record.kafkaOffset(), record.topic());
+        this.kafkaRecordErrorReporter.reportError(record, new DataException("Broken Record"));
```
sfc-gh-tzhang, line 141: I assume there will follow-up PRs to send records to DLQ when insertRows API failed?

sfc-gh-tzhang (approved):

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkService.java
```diff
@@ -137,6 +138,9 @@
   void setDeliveryGuarantee(
       SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee ingestionDeliveryGuarantee);
 
+  /* Set Error reporter which can be used to send records to DLQ (Dead Letter Queue) */
+  default void setErrorReporter(KafkaRecordErrorReporter kafkaRecordErrorReporter) {}
```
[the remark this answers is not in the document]
sfc-gh-japatel: yes this is in v2 only. I added a default interface method so as to not add anything in v1.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: yes..

PR #397, opened 2022-02-11
Refactor fetching offsetToken method to be reusable for multiple calls

sfc-gh-japatel:
- Renamed few fields and added more javadocs.
- Add test for close channel exception

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
@@ -135,24 +135,23 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
     // the table should be present before opening a channel so lets do a table existence check here
     createTableIfNotExists(tableName);
 
-    createOrUpdatePartitionsToChannelCache(partitionChannelKey, tableName);
+    createStreamingChannelForTopicPartition(partitionChannelKey, tableName);
   }
 
-  /* We will always open a new channel. This will either update the TopicPartitionChannel's streamingChannel or will create a new instance of TopicPartitionChannel. */
-  private void createOrUpdatePartitionsToChannelCache(
+  /**
+   * Always opens a new channel and creates a new instance of TopicPartitionChannel.
+   *
+   * <p>This is essentially a blind write to partitionsToChannel. i.e we dont check if it is present
+   * or not.
+   */
+  private void createStreamingChannelForTopicPartition(
       final String partitionChannelKey, final String tableName) {
     // Always open a new channel.
     SnowflakeStreamingIngestChannel partitionChannel =
         openChannelForTable(partitionChannelKey, tableName);
-    if (partitionsToChannel.containsKey(partitionChannelKey)) {
-      // we just update the channel instance for this partition
-      partitionsToChannel.get(partitionChannelKey).updateStreamingIngestChannel(partitionChannel);
-    } else {
-      // Add channel and related metadata to partitionsToChannel map
-      TopicPartitionChannel topicPartitionChannel =
-          new TopicPartitionChannel(partitionChannel, this.kafkaRecordErrorReporter);
-      partitionsToChannel.put(partitionChannelKey, topicPartitionChannel);
-    }
+    TopicPartitionChannel topicPartitionChannel =
+        new TopicPartitionChannel(partitionChannel, this.kafkaRecordErrorReporter);
+    partitionsToChannel.put(partitionChannelKey, topicPartitionChannel);
   }
 
   /**
@@ -203,7 +202,7 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
     if (!partitionsToChannel.containsKey(partitionChannelKey)
         || partitionsToChannel.get(partitionChannelKey).isChannelClosed()) {
       LOGGER.warn(
-          "Topic: {} Partition: {} hasn't been initialized by OPEN " + "function",
+          "Topic: {} Partition: {} hasn't been initialized by OPEN function",
           record.topic(),
           record.kafkaPartition());
       startTask(
@@ -263,12 +262,14 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
   }
 
   /**
-   * This function is called during rebalance. Please ensure we dont wipe off state when we are
-   * invoking this function. i.e after rebalance, we would still want to preserve metadata like
-   * streamingBuffer, processedOffset
+   * This function is called during rebalance.
    *
    * <p>All the channels are closed. The client is still active. Upon rebalance, (inside {@link
    * com.snowflake.kafka.connector.SnowflakeSinkTask#open(Collection)} we will reopen the channel.
+   *
+   * <p>We will wipe the cache partitionsToChannel so that in {@link
+   * com.snowflake.kafka.connector.SnowflakeSinkTask#open(Collection)} we reinstantiate and fetch
+   * offsetToken
    *
    * @param partitions a list of topic partition
    */
@@ -287,6 +288,7 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
               topicPartition.topic(),
               topicPartition.partition());
         });
+    partitionsToChannel.clear();
   }
 
   @Override
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
@@ -231,7 +231,7 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
     String partitionChannelKey =
         partitionChannelKey(topicPartition.topic(), topicPartition.partition());
     if (partitionsToChannel.containsKey(partitionChannelKey)) {
-      return partitionsToChannel.get(partitionChannelKey).getCommittedOffset();
+      return partitionsToChannel.get(partitionChannelKey).getOffsetSafeToCommitToKafka();
     } else {
       LOGGER.warn(
           "Topic: {} Partition: {} hasn't been initialized to get offset",
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
@@ -35,32 +35,35 @@ import org.slf4j.LoggerFactory;
  *
  * <p>The number of TopicPartitionChannel objects can scale in proportion to the number of
  * partitions of a topic.
+ *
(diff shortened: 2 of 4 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -97,29 +95,27 @@ public TopicPartitionChannel(
 
     this.streamingBuffer = new StreamingBuffer();
     this.processedOffset = new AtomicLong(-1);
-    this.committedOffset = new AtomicLong(0);
+    this.offsetSafeToCommitToKafka = new AtomicLong(0);
```
sfc-gh-tzhang, line 104: should this also start with an invalid value (like -1)?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -97,29 +95,27 @@ public TopicPartitionChannel(
...
-          record.kafkaPartition());
-      // This will only be called once at the beginning when an offset arrives for first time
-      // after connector starts
-      fetchLastCommittedOffsetToken();
-
-      this.hasChannelReceivedAnyRecordsBefore = true;
-    }
+    initChannel(record);
 
     // discard the record if the record offset is smaller or equal to server side offset, or if
```
sfc-gh-tzhang, line 121: > discard the record if the record offset is smaller or equal to server side offset

Is this possible to happen? If no, could you add a logwarn?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -158,28 +154,34 @@ public void insertRecordToBuffer(SinkRecord record) {
...
+   *
+   * <p>Note: This code is execute only for the first time connector starts or restarts after task
+   * was killed previously.
+   *
+   * <p>This doesnt get invoked on rebalance.
+   *
+   * @param record record this partition received.
+   */
+  private void initChannel(final SinkRecord record) {
+    if (!hasChannelReceivedAnyRecordsBefore) {
```
sfc-gh-tzhang, line 174: don't we always recreate the TopicPartitionChannel object during start/rebalance? If yes, this should always be false?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -289,56 +291,79 @@ public void insertBufferedRows() {
...
-    LOGGER.info(
-        "Last committed offset for partition channel:{} is :{}",
-        this.channel.getFullyQualifiedName(),
-        lastOffsetCommitted);
-    if (Strings.isNullOrEmpty(lastOffsetCommitted)) {
-      return committedOffset.get();
+  public long getOffsetSafeToCommitToKafka() {
+    final long committedOffsetInSnowflake = fetchLatestCommittedOffsetFromSnowflake();
+    if (committedOffsetInSnowflake == NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE) {
+      return offsetSafeToCommitToKafka.get();
```
sfc-gh-tzhang, line 317: could we just return -1? It's easier to understand

sfc-gh-tzhang (approved):

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: No, actually since offsets start from 0 in kafka, by default is 0. if we send -1, it will throw up!

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Yess. if we had not committed that offset, upon restart it might send same offset again.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Ahh yes!

sfc-gh-japatel: No, we cant return -1; (As above, it it init to 0)

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Yes, good idea to add logwarn. might be helpful while debugging. I will also print processoffset and committedoffset from snowflaake.

sfc-gh-tzhang: if this is a valid case, then no need to add logwarn

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: ok, then could we return 0 instead? (or at least add a comment saying this is always 0, otherwise the logic is confusing by only looking at the code)

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Correct but we call this from insertRecord. So from insertRecord, `hasChannelReceivedAnyRecordsBefore` will be false only once. (First record)

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: The only reason I didnt want to return 0 was it offsetSafeToCommitToKafka could be non zero. 

i.e it is possible  `fetchLatestCommittedOffsetFromSnowflake` gave us invalid token, in that case we would like to return the offset which was last sent to kafka. (which can be any number)

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: FYI, I changed this name to `precomputeOffsetTokenForChannel`

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -158,28 +155,37 @@ public void insertRecordToBuffer(SinkRecord record) {
...
-        this.offsetPersistedInSnowflake = NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE;
-      } else {
-        this.offsetPersistedInSnowflake = Long.parseLong(offsetToken);
-      }
+  /**
+   * Pre-computes the offset token which might have been persisted into Snowflake for this channel.
+   *
+   * <p>It is possible the offset token is null. In this case, -1 is set
+   *
+   * <p>Note: This code is execute only for the first time when
```
sfc-gh-mnaides, line 163: nit: incomplete comment?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -289,56 +295,79 @@ public void insertBufferedRows() {
...
+  }
+
+  /* Returns the offset Token persisted into snowflake */
+  @VisibleForTesting
+  protected long fetchLatestCommittedOffsetFromSnowflake() {
+    long latestCommittedOffsetInSnowflake;
+    LOGGER.debug(
+        "Fetching last committed offset for partition channel:{}",
+        this.channel.getFullyQualifiedName());
+    String offsetToken = this.channel.getLatestCommittedOffsetToken();
```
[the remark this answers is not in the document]
sfc-gh-japatel: Yes, I will be doing that in another PR.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -135,24 +135,23 @@ public void startTask(String tableName, String topic, int partition) {
...
+    createStreamingChannelForTopicPartition(partitionChannelKey, tableName);
   }
 
-  /* We will always open a new channel. This will either update the TopicPartitionChannel's streamingChannel or will create a new instance of TopicPartitionChannel. */
-  private void createOrUpdatePartitionsToChannelCache(
+  /**
+   * Always opens a new channel and creates a new instance of TopicPartitionChannel.
+   *
+   * <p>This is essentially a blind write to partitionsToChannel. i.e we dont check if it is present
+   * or not.
```
sfc-gh-tjones, line 145: Can you elaborate why this is a blind write?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -135,24 +135,23 @@ public void startTask(String tableName, String topic, int partition) {
...
-      partitionsToChannel.get(partitionChannelKey).updateStreamingIngestChannel(partitionChannel);
-    } else {
-      // Add channel and related metadata to partitionsToChannel map
-      TopicPartitionChannel topicPartitionChannel =
-          new TopicPartitionChannel(partitionChannel, this.kafkaRecordErrorReporter);
-      partitionsToChannel.put(partitionChannelKey, topicPartitionChannel);
-    }
+    TopicPartitionChannel topicPartitionChannel =
+        new TopicPartitionChannel(partitionChannel, this.kafkaRecordErrorReporter);
+    partitionsToChannel.put(partitionChannelKey, topicPartitionChannel);
```
sfc-gh-tjones, line 154: Nit, just inline  the above `TopicPartitionChannel` creation here.

sfc-gh-tjones (approved):

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Blind write because partitionsToChannel map is not checked before puting the key. 
May be it is not 100% appropriate since it is mostly related to database?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Taken care in next PR.

PR #401, opened 2022-02-17
SNOW-530369 insert rows failure handling and delivery to DLQ

sfc-gh-japatel:
Error handling in insertRows + Logging/delivery to DLQ

- Add few configs for error handling. 
- [Here](https://www.confluent.io/blog/kafka-connect-deep-dive-error-handling-dead-letter-queues/) is how it will work. 
  - error.tolerance = NONE 
    - fail fast
  - error.tolerance = ALL
    - similar to on error continue. 
    - can provide DLQ to send those records to DLQ (`errors.deadletterqueue.topic.name`)
  - errors.log.enable = false by default
    - can be used to log errors in log file. 

- Fallback for SFException
 - No retry for insertRows. 
 - Reopen the channel and reset kafka with offset found in snowflake. 
 - Send retriableexception back from the put API. https://kafka.apache.org/31/javadoc/org/apache/kafka/connect/sink/SinkTask.html#put(java.util.Collection)

- Added test
- Will fix SNOW-545729 in later commits.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -125,7 +125,7 @@ public class SnowflakeSinkConnectorConfig {
   public static final String ERROR_GROUP = "ERRORS";
   public static final String ERRORS_TOLERANCE_CONFIG = "errors.tolerance";
   public static final String ERRORS_TOLERANCE_DISPLAY = "Error Tolerance";
-  public static final ErrorTolerance ERRORS_TOLERANCE_DEFAULT = ErrorTolerance.NONE;
+  public static final String ERRORS_TOLERANCE_DEFAULT = ErrorTolerance.NONE.toString();
   public static final String ERRORS_TOLERANCE_DOC =
       "Behavior for tolerating errors during Sink connector's operation. 'NONE' is set as default"
           + " and denotes that it will be fail fast. i.e any error will result in an immediate task"
@@ -142,11 +142,11 @@ public class SnowflakeSinkConnectorConfig {
   public static final String ERRORS_DEAD_LETTER_QUEUE_TOPIC_NAME_CONFIG =
       "errors.deadletterqueue.topic.name";
   public static final String ERRORS_DEAD_LETTER_QUEUE_TOPIC_NAME_DISPLAY =
-      "Send error records to the dead letter queue(DLQ)";
+      "Send error records to the Dead Letter Queue (DLQ)";
   public static final String ERRORS_DEAD_LETTER_QUEUE_TOPIC_NAME_DEFAULT = "";
   public static final String ERRORS_DEAD_LETTER_QUEUE_TOPIC_NAME_DOC =
       "Whether to output conversion errors to the dead letter queue "
-          + "By default messages are not outputted to the dead letter queue. "
+          + "By default messages are not sent to the dead letter queue. "
           + "Requires property `errors.tolerance=all`.";
 
   public static void setDefaultValues(Map<String, String> config) {
@@ -442,9 +442,9 @@ public class SnowflakeSinkConnectorConfig {
         .define(
             ERRORS_TOLERANCE_CONFIG,
             Type.STRING,
-            ERRORS_TOLERANCE_DEFAULT.value(),
+            ERRORS_TOLERANCE_DEFAULT,
             ErrorTolerance.VALIDATOR,
-            Importance.MEDIUM,
+            Importance.LOW,
             ERRORS_TOLERANCE_DOC,
             ERROR_GROUP,
             0,
@@ -454,7 +454,7 @@ public class SnowflakeSinkConnectorConfig {
             ERRORS_LOG_ENABLE_CONFIG,
             Type.BOOLEAN,
             ERRORS_LOG_ENABLE_DEFAULT,
-            Importance.MEDIUM,
+            Importance.LOW,
             ERRORS_LOG_ENABLE_DOC,
             ERROR_GROUP,
             1,
@@ -464,7 +464,7 @@ public class SnowflakeSinkConnectorConfig {
             ERRORS_DEAD_LETTER_QUEUE_TOPIC_NAME_CONFIG,
             Type.STRING,
             ERRORS_DEAD_LETTER_QUEUE_TOPIC_NAME_DEFAULT,
-            Importance.MEDIUM,
+            Importance.LOW,
             ERRORS_DEAD_LETTER_QUEUE_TOPIC_NAME_DOC,
             ERROR_GROUP,
             2,
@@ -680,14 +680,11 @@ public class SnowflakeSinkConnectorConfig {
     /** Tolerate all errors. */
     ALL;
 
-    public String value() {
-      return name().toLowerCase(Locale.ROOT);
-    }
-
     /* Validator to validate behavior.on.null.values which says whether kafka should keep null value records or ignore them while ingesting into snowflake table. */
     public static final ConfigDef.Validator VALIDATOR =
         new ConfigDef.Validator() {
-          private final ConfigDef.ValidString validator = ConfigDef.ValidString.in(names());
+          private final ConfigDef.ValidString validator =
+              ConfigDef.ValidString.in(ErrorTolerance.names());
 
           @Override
           public void ensureValid(String name, Object value) {
@@ -697,7 +694,6 @@ public class SnowflakeSinkConnectorConfig {
             validator.ensureValid(name, value);
           }
 
-          // Overridden here so that ConfigDef.toEnrichedRst shows possible values correctly
           @Override
           public String toString() {
             return validator.toString();
@@ -715,5 +711,30 @@ public class SnowflakeSinkConnectorConfig {
 
       return result;
     }
+
+    @Override
+    public String toString() {
+      return name().toLowerCase(Locale.ROOT);
+    }
   }
+
+  /**
+   * Boolean Validator of passed booleans in configurations (TRUE or FALSE). This validator is case
+   * insensitive
+   */
+  public static final ConfigDef.Validator BOOLEAN_VALIDATOR =
+      new ConfigDef.Validator() {
+        private final ConfigDef.ValidString validator =
+            ConfigDef.ValidString.in(
+                Boolean.TRUE.toString().toLowerCase(Locale.ROOT),
+                Boolean.FALSE.toString().toLowerCase(Locale.ROOT));
+
+        @Override
+        public void ensureValid(String name, Object value) {
+          if (value instanceof String) {
+            value = ((String) value).toLowerCase(Locale.ROOT);
+          }
+          this.validator.ensureValid(name, value);
+        }
+      };
 }
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -40,13 +40,13 @@ import org.slf4j.LoggerFactory;
 public class SnowflakeSinkConnectorConfig {
 
   static final String NAME = Utils.NAME;
-  static final String TOPICS = "topics";
+  public static final String TOPICS = "topics";
 
   // Connector config
   private static final String CONNECTOR_CONFIG = "Connector Config";
-  static final String BUFFER_COUNT_RECORDS = "buffer.count.records";
+  public static final String BUFFER_COUNT_RECORDS = "buffer.count.records";
   public static final long BUFFER_COUNT_RECORDS_DEFAULT = 10000;
-  static final String BUFFER_SIZE_BYTES = "buffer.size.bytes";
+  public static final String BUFFER_SIZE_BYTES = "buffer.size.bytes";
   public static final long BUFFER_SIZE_BYTES_DEFAULT = 5000000;
   public static final long BUFFER_SIZE_BYTES_MIN = 1;
   static final String TOPICS_TABLES_MAP = "snowflake.topic2table.map";
@@ -57,7 +57,7 @@ public class SnowflakeSinkConnectorConfig {
   // Time in seconds
   public static final long BUFFER_FLUSH_TIME_SEC_MIN = 10;
   public static final long BUFFER_FLUSH_TIME_SEC_DEFAULT = 120;
-  static final String BUFFER_FLUSH_TIME_SEC = "buffer.flush.time";
+  public static final String BUFFER_FLUSH_TIME_SEC = "buffer.flush.time";
 
   // Snowflake connection and database config
   private static final String SNOWFLAKE_LOGIN_INFO = "Snowflake Login Info";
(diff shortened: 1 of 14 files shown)


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -121,6 +121,34 @@
...
+  public static final boolean ERRORS_LOG_ENABLE_DEFAULT = false;
+  public static final String ERRORS_LOG_ENABLE_DOC =
+      "If true, write/log each error along with details of the failed operation and record"
+          + " properties to the Connect log. Default is 'false', so that only errors that are not"
+          + " tolerated are reported.";
+
+  public static final String ERRORS_DEAD_LETTER_QUEUE_TOPIC_NAME_CONFIG =
+      "errors.deadletterqueue.topic.name";
+  public static final String ERRORS_DEAD_LETTER_QUEUE_TOPIC_NAME_DISPLAY =
+      "Send error records to the dead letter queue(DLQ)";
```
sfc-gh-tjones, line 145: nit, capitalize each letter and add a space between "queue" and ("DLQ)"

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -121,6 +121,34 @@
...
+  public static final String ERRORS_DEAD_LETTER_QUEUE_TOPIC_NAME_DEFAULT = "";
+  public static final String ERRORS_DEAD_LETTER_QUEUE_TOPIC_NAME_DOC =
+      "Whether to output conversion errors to the dead letter queue "
+          + "By default messages are not outputted to the dead letter queue. "
```
sfc-gh-tjones, line 149: "outputted" sounds weird, maybe just change to "sent"?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -611,4 +670,50 @@ public String toString() {
...
+    }
+
+    /* Validator to validate behavior.on.null.values which says whether kafka should keep null value records or ignore them while ingesting into snowflake table. */
+    public static final ConfigDef.Validator VALIDATOR =
+        new ConfigDef.Validator() {
+          private final ConfigDef.ValidString validator = ConfigDef.ValidString.in(names());
+
+          @Override
+          public void ensureValid(String name, Object value) {
+            if (value instanceof String) {
```
sfc-gh-tjones, line 694: What happens if it's not a string?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-japatel: Will do! thanks

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-japatel: This validator usually runs when we pass in a config value which is already in form of json and each values are strings. There is technically no need to check this but I followed how other `ensureValids` were written.

sfc-gh-tzhang:
btw, nice label! I like it :)

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -611,4 +670,71 @@ public String toString() {
...
+  /* https://www.confluent.io/blog/kafka-connect-deep-dive-error-handling-dead-letter-queues/ */
+  public enum ErrorTolerance {
+
+    /** Tolerate no errors. */
+    NONE,
+
+    /** Tolerate all errors. */
+    ALL;
+
+    /* Validator to validate behavior.on.null.values which says whether kafka should keep null value records or ignore them while ingesting into snowflake table. */
```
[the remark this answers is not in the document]
sfc-gh-japatel: Will change. Somehow I cant put /***/

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -611,4 +670,71 @@ public String toString() {
...
+            validator.ensureValid(name, value);
+          }
+
+          @Override
+          public String toString() {
+            return validator.toString();
+          }
+        };
+
+    // All valid enum values
```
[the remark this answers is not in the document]
sfc-gh-japatel: will do

sfc-gh-japatel:
> Discussed in the channel about duplicate rows in DLQ (please reset my status after we have a conclusion and the changes are done accordingly)

Thanks for raising this concern but here is what I thought first and then how I researched it:
1. DLQ topics are just like any other kafka topic. 
2. We are sending SinkRecord into kafka through the API https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/errors/WorkerErrantRecordReporter.java#L74
3. Each SinkRecord has a topic and partition/offset. 
4. I digged deeper into how this is implemented on kafka side. 
  a. Error handling is essentially another SinkConnector with different properties just like SnowflakeSInkConnector. https://github.com/apache/kafka/blob/trunk/connect/runtime/src/test/java/org/apache/kafka/connect/integration/ErrantRecordSinkConnector.java#L29
  b. Once the API delivers it to above SinkConnector, it caches it in memory with a unique key derived from and offset and partition for dlq topic. Check [here](https://github.com/apache/kafka/blob/trunk/connect/runtime/src/test/java/org/apache/kafka/connect/integration/ErrantRecordSinkConnector.java#L50-L57)  
  c. And the sinkrecords in kafka dlq are committed accordingly with its own set of offset no(starting from 0)

So overall, I think we are safe to not worry about duplicates in DLQ.

sfc-gh-japatel:
@sfc-gh-tzhang can you re-review after resolving the concern you mentioned earlier?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -316,40 +332,136 @@ public void insertBufferedRows() {
...
+          this.channel.getFullyQualifiedName(),
+          streamingBuffer.getNumOfRecord(),
+          streamingBuffer.getFirstOffset(),
+          streamingBuffer.getLastOffset());
+      return this.channel.insertRows(
+          streamingBuffer.getData(), Long.toString(streamingBuffer.getLastOffset()));
+    }
+  }
+
+  /** We will reopen the channel on {@link SFException} and call insert rows API again. */
```
sfc-gh-tzhang, line 416: We should reopen the channel and restart from the latest committed offset token, right?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -547,8 +659,8 @@ private long fetchLatestCommittedOffsetFromSnowflake() {
   private SnowflakeStreamingIngestChannel openChannelForTable() {
     OpenChannelRequest channelRequest =
         OpenChannelRequest.builder(this.channelName)
-            .setDBName(this.databaseName)
-            .setSchemaName(this.schemaName)
+            .setDBName(this.sfConnectorConfig.getOrDefault(Utils.SF_DATABASE, null))
+            .setSchemaName(this.sfConnectorConfig.getOrDefault(Utils.SF_SCHEMA, null))
```
sfc-gh-tzhang, line 663: this doesn't seems right, we have to have database and schema, right?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -640,9 +752,12 @@ protected SnowflakeStreamingIngestChannel getChannel() {
     // Every List element corresponds to one record(offset) of a partition
     private final List<Map<String, Object>> tableRows;
 
+    private final List<SinkRecord> sinkRecords;
```
sfc-gh-tzhang, line 755: This doesn't look good as we're doubling the memory usage, could we reuse the original buffer?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Correct but I am relying on the open channel API to send an exception if it passes null.

sfc-gh-japatel: Infact this should never happen and would fail earlier.

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -566,6 +569,19 @@ static String validateConfig(Map<String, String> config) {
                     IngestionMethodConfig.SNOWPIPE_STREAMING.toString()));
             configIsValid = false;
           }
+
+          /**
+           * Only checking in streaming since we are utilizing the values before we send it to
+           * DLQ/output to log file
+           */
+          if (config.containsKey(ERRORS_TOLERANCE_CONFIG)) {
```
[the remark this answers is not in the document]
sfc-gh-japatel: sorry, I am missing something here. ensurevalid is only used if config has that key because they are not mandatory.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -316,40 +332,136 @@ public void insertBufferedRows() {
...
+   * net.snowflake.ingest.streaming.InsertValidationResponse.InsertError} in form of response inside
+   * {@link InsertValidationResponse}
+   *
+   * @param buffer buffer to insert into snowflake
+   * @return InsertValidationResponse object sent from insertRows API.
+   */
+  private InsertValidationResponse insertRowsWithFallback(StreamingBuffer buffer) {
+    Fallback<InsertValidationResponse> reopenChannelFallbackExecutorForInsertRows =
+        Fallback.builder(() -> insertRowsFallbackSupplier(buffer))
+            .handle(SFException.class)
```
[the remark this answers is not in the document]
sfc-gh-japatel: I do want to refactor common logic for retry into another class but I will do that after PrPr. 

Based on chat with @sfc-gh-mnaides and @sfc-gh-tjones only SFExceptions are worth retrying. I generally dont like the idea of catch all exceptions because that essentially will hide away other problems.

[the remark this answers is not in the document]
sfc-gh-japatel: We will have logs on which of two operations (open channel or insertRows) failed. This logger is kind of useful to know that even fallback has failed.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: I suppose that is only for get offset token? 
Idea here is to retry sending the same set of rows and worry about resetting offset token only in getOffsetToken. resetting it in getoffsettoken makes sense because the rows will be sent again and hence insertRows can try inserting lost rows. 

Adding get offsetToken here just complicate things here since we would be sending rows for current offsets which might be higher than what we might reset later.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Agreed, unfortunately we will have to deal with that since we are buffering. There is no way to send records to DLQ without even buffering them. 

Infact I just realised we format the record into JSON and create new sink records from kafka records. I will fix that in another PR. 

Happy to discuss this more since I am out of ideas on how to regenerate sinkrecords back for buffered records. 

If we dont buffer at all, this issue will solved.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: If you follow [this](https://snowflake.slack.com/archives/G01F1B8PLDU/p1645582692556109) thread, it's possible that the queued chunks got skipped when we see a bigger client/channel sequencer, so you will need to fetch the latest offset token first

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: Why can't we keep the original SinkRecord and only convert during flushing?

src/main/java/com/snowflake/kafka/connector/Utils.java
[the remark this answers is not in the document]
sfc-gh-japatel: Unfortunately if checks cant be moved in ensureValid, ensureValid is usually called by connector framework and has one overridden method which is called during validate API of connector.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Ahh okay. I kind of understand.. So calling insertRows here wouldnt make sense right?
Reopen and then reset.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: It's not harmful to keep this but let me remove it. I will just replace it with `map.get()`

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Hmm good idea but that might require quite a bit of refactoring.. Can I deal with that in separate PR?

sfc-gh-japatel:
@sfc-gh-tzhang Here is what the recent commits does:
1. Reopen channel and call offsettoken. reset offset in kafka. 
2. Reuse same code as for offset token fallback. 
3. Throws `RetriableException`.

sfc-gh-tjones (approved):

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -316,40 +333,141 @@ public void insertBufferedRows() {
...
+    }
+
+    @Override
+    public InsertValidationResponse get() throws Throwable {
+      LOGGER.debug(
+          "Invoking insertRows API for channel:{}, noOfRecords:{}, startOffset:{}, endOffset:{}",
+          this.channel.getFullyQualifiedName(),
+          streamingBuffer.getNumOfRecords(),
+          streamingBuffer.getFirstOffset(),
+          streamingBuffer.getLastOffset());
```
sfc-gh-tzhang, line 413: nit: use the same pattern
```suggestion
          this.channel.getFullyQualifiedName(),
          this.streamingBuffer.getNumOfRecords(),
          this.streamingBuffer.getFirstOffset(),
          this.streamingBuffer.getLastOffset());
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: sounds good, please add a TODO with JIRA link

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -316,40 +333,141 @@ public void insertBufferedRows() {
...
+        LOGGER.warn(
+            "Although config:{} is set, Dead Letter Queue topic:{} is not set.",
+            ERRORS_TOLERANCE_CONFIG,
+            ERRORS_DEAD_LETTER_QUEUE_TOPIC_NAME_CONFIG);
+      } else {
+        for (int i = 0; i < insertErrors.size(); i++) {
+          // Write now I am assuming each error maps to its original row which is incorrect.
+          // TODO: SNOW-545729 would handle mapping errors with original SinkRecord.
+          this.kafkaRecordErrorReporter.reportError(
+              insertedRecordsToBuffer.get(i), insertErrors.get(i).getException());
```
sfc-gh-tzhang, line 465: I think you can update this now since you're using the new beta release

sfc-gh-tzhang (approved):
Ship it!

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Oops, I think the rebase was mysterious.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Updated.

sfc-gh-japatel:
@sfc-gh-tzhang @sfc-gh-tjones fixed the bug in last commit on this PR!

PR #407, opened 2022-02-23
SNOW-550868 Change buffer size from integer to long

sfc-gh-japatel:
- Max int value - 2147483647 = ~1.999 GB
- We are bounded by Heap size but we dont want to be limited by max int value.


sfc-gh-tjones (approved):

sfc-gh-tzhang:
I'm curious about whether we hit any problem with `int`? Or this is just something you believe we should do?

sfc-gh-japatel:
> I'm curious about whether we hit any problem with `int`? Or this is just something you believe we should do?

We didnt but I think we should do.

PR #412, opened 2022-03-15
NO-SNOW: BUGFIX - Remove unnecessary check for buffer count and add a test

sfc-gh-japatel:
Bug:
No need to check for count of records in buffer == (lastOffset number - firstOffset number + 1)

This will not be true when there are bad records sent to Kafka but we would not buffer those bad records resulting into incorrect exception being thrown rather than inserting remaining rows. 

Test included


sfc-gh-tjones (approved):
Nice!

sfc-gh-tzhang (approved):

PR #413, opened 2022-03-16
Add a debug statement for insert rows failure

sfc-gh-japatel:
Found this could be helpful. I wasnt able to figure out why the insertRows API was failing.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
@@ -433,6 +433,12 @@ public class TopicPartitionChannel {
     Fallback<Object> reopenChannelFallbackExecutorForInsertRows =
         Fallback.builder(() -> insertRowsFallbackSupplier(buffer))
             .handle(SFException.class)
+            .onFailedAttempt(
+                event ->
+                    LOGGER.warn(
+                        String.format(
+                            "Failed Attempt to invoke the insertRows API for buffer:%s", buffer),
+                        event.getLastException()))
             .onFailure(
                 event ->
                     LOGGER.error(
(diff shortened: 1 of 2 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -433,6 +433,12 @@ private InsertValidationResponse insertRowsWithFallback(StreamingBuffer buffer)
     Fallback<Object> reopenChannelFallbackExecutorForInsertRows =
         Fallback.builder(() -> insertRowsFallbackSupplier(buffer))
             .handle(SFException.class)
+            .onFailedAttempt(
+                event ->
+                    LOGGER.warn(
+                        String.format(
+                            "Failed Attempt to invoke the insertRows API for buffer:%s", buffer),
```
sfc-gh-tzhang, line 440: InsertRows shouldn't fail other than invalid or closed channel, what's the reason that you saw?

sfc-gh-tzhang (approved):

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: I did see 429 from one of the APIs. (I think it was blob upload)

sfc-gh-tzhang: what do you mean by 429?

sfc-gh-japatel: Sorry I meant 429 response code: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429

sfc-gh-tzhang: hmmm, that's weird. We're not calling any REST APIs in the InsertRows API, i assume we're talking about `SnowflakeStreamingIngestChannelInternal.insertRows`?

PR #415, opened 2022-03-23
NO-SNOW: fix security vulnerability

sfc-gh-tzhang:
This should fix https://github.com/snowflakedb/snowflake-kafka-connector/issues/411

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -349,12 +349,12 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-core</artifactId>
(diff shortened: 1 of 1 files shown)


sfc-gh-tjones (approved):

sfc-gh-japatel:
I somehow missed this but we also need to update confluent pom

sfc-gh-tzhang:
> I somehow missed this but we also need to update confluent pom

No worries, I believe we need to update it again (even though the original one mentioned that anything higher than 2.13 should be sufficient)

PR #420, opened 2022-04-04
Add more connector properties during kafka connect start

sfc-gh-japatel:
- No behavior change to existing data format. (Anyways it was not visible to
  customers) But the format is backward compatible for existing SQL
  queries
- Adding new fields to the data node of Json sent during kafka connecter
  start.
- Added unit test for checking format.
- Verified results internally. (on Snowhouse view)

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryServiceV1.java
@@ -1,8 +1,20 @@
 package com.snowflake.kafka.connector.internal.telemetry;
 
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.BUFFER_COUNT_RECORDS;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.BUFFER_FLUSH_TIME_SEC;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.BUFFER_SIZE_BYTES;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.DELIVERY_GUARANTEE;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.INGESTION_METHOD_DEFAULT_SNOWPIPE;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.INGESTION_METHOD_OPT;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.KEY_CONVERTER_CONFIG_FIELD;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.VALUE_CONVERTER_CONFIG_FIELD;
+
+import com.google.common.annotations.VisibleForTesting;
 import com.snowflake.kafka.connector.Utils;
 import com.snowflake.kafka.connector.internal.Logging;
 import java.sql.Connection;
+import java.util.Map;
 import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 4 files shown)


sfc-gh-tjones (approved):

src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryServiceV1.java
```diff
@@ -109,6 +139,25 @@ public void reportKafkaPartitionStart(final SnowflakeTelemetryBasicInfo pipeCrea
     send(TelemetryType.KAFKA_PIPE_START, msg);
   }
 
+  /**
+   * JsonNode data is wrapped into another ObjectNode which looks like this:
```
sfc-gh-tzhang, line 143: What's the advantage of having nested json instead of flat schema?

src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryServiceV1.java
```diff
@@ -140,6 +189,49 @@ private String getTaskID() {
...
+  /**
+   * Adds specific user provided connector properties to ObjectNode
+   *
+   * @param userProvidedConfig user provided key value pairs in a Map
+   * @param dataObjectNode Object node in which specific properties to add
+   */
+  protected void addUserConnectorPropertiesToDataNode(
+      final Map<String, String> userProvidedConfig, ObjectNode dataObjectNode) {
+    // maxTasks value isn't visible if the user leaves it at default. So 0
+    // means unknown.
```
sfc-gh-tzhang, line 201: but do you prevent user from setting maxTasks to 0?

src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryServiceV1.java
```diff
@@ -140,6 +189,49 @@ private String getTaskID() {
...
+    int maxTasks =
+        (userProvidedConfig.containsKey("tasks.max"))
+            ? Integer.parseInt(userProvidedConfig.get("tasks.max"))
```
sfc-gh-tzhang, line 204: nit: do you really need to parse it?

sfc-gh-tzhang (approved):
lgtm, left a few minor comments

src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryServiceV1.java
sfc-gh-tzhang: Actually why not just use string? it will NULL if user doesn't set it

src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryServiceV1.java
sfc-gh-japatel: I don't see any advantage per se apart from having a distinct hierarchy but I wanted to keep the same structure since otherwise we would have to run two different queries if we change and put it into outer node.

src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryServiceV1.java
sfc-gh-japatel: Yeah, not needed. will remove.

src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryServiceV1.java
sfc-gh-japatel: there is actually no checks for tasks.max.

sfc-gh-japatel: It will change the values moving forward. i.e if nothing is given, it will parse it to null instead of 0.

PR #421, opened 2022-04-04
Add huge avro schema test for buffer bytes threshold

sfc-gh-japatel:
Just more test for Avro, util function will be useful later too!


sfc-gh-tjones (approved):
Woohoo more tests! Thank you!

PR #423, opened 2022-04-12
no-snow BUG_FIX Error handling put batch

sfc-gh-japatel:
Bug:
- In case of `insertRows` failures, subsequent records from this batch (Batch sent in put API of SinkTask) were successfully calling insertRows since we were reopening the channel. 

Lets say 101-500 for partition 0 is sent as part of a batch and threshold is 100 records. 
- `101-200` fails to get inserted into Snowflake and it reopens the channel
- `201-300` should not be sent before until we receive `101-200` or may be any non overlapping sets which contains `101-200`
- We cannot send a RetriableException to Kafka since we dont want to retry this batch in case `offsetCommittedInKafka` is less than 100. (Lets say 25 is offsetToken present in Snowflake)
- `26` should be the first offset this partition should receive and we would not know if records 101-200 would be part of which buffer. (It is possible that after resetting offsets in kafka, we would receive 26-600)

Hence, this change prevents records to wrongly being added into buffer and only wait for right offset number before we can start buffering rest of them. 

Introducing a boolean `isOffsetResetInKafka` which will be set to true in case of failures and fallback mechanism. 

Added test cases, + manual end to end tests. 

TLDR: 

- Block the PUT API (or block inserting into the KC buffer)
- Cleanup the KC buffer
- Reopen the channel
- Reset KC offset
- Unblock PUT API


src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -76,7 +77,8 @@ public class TopicPartitionChannel {
    * States whether this channel has received any records before.
    *
    * <p>If this is false, the topicPartitionChannel has recently been initialised and didnt receive
-   * any records before or TopicPartitionChannel was recently created.
+   * any records before or TopicPartitionChannel was recently created. (Or may be there was some
+   * error in some API and hence we had to fetch it again?)
```
sfc-gh-tjones, line 81: Why the `?`

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -86,12 +88,31 @@ public class TopicPartitionChannel {
...
 
   // used to communicate to the streaming ingest's insertRows API
   // This is non final because we might decide to get the new instance of Channel
   private SnowflakeStreamingIngestChannel channel;
 
+  /**
+   * Offsets are reset in kafka when one of following cases arises in which we rely on source of
+   * truth (Which is Snowflake's committed offsetToken)
+   *
+   * <p>1. If channel fails to fetch offsetToken in kafka, we reopen the channel and try to fetch
```
sfc-gh-tjones, line 102: Nit, it might be more clear to use `<ol>` and `<li` here instead of paragraphs.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -86,12 +88,31 @@ public class TopicPartitionChannel {
...
+   * offset from Snowflake again
+   *
+   * <p>2. If channel fails to ingest a buffer(Buffer containing rows/offsets), we reopen the
```
sfc-gh-tjones, line 105: Nit, space between `buffer (`

sfc-gh-tjones, line 105: When we say "ingest" here do we mean in the Client SDK side of things or in the Snowflake side of things? If the Client SDK rejects a set of rows we should back off rather than immediately re-opening the channel.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -86,12 +88,31 @@ public class TopicPartitionChannel {
...
+   * channel and try to fetch offset from Snowflake again
+   *
+   * <p>In both above cases, we ask Kafka to send back offsets after offset already present in
```
sfc-gh-tjones, line 108: Nit, I think you're missing some words after "send back offsets"

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -128,6 +149,11 @@ public class TopicPartitionChannel {
   private final AtomicLong offsetSafeToCommitToKafka;
 
   // added to buffer before calling insertRows
+  // Reset to earlier value if we were not successfully able to call insertRows.
+  // Do we even need this processedOffset
+  // Why do we even need to rely on a value which is cached in KC?
+  // Source of truth is only Snowflake.
+  // May be remove this and just rely on this boolean (hasChannelReceivedAnyRecordsBefore)
   private final AtomicLong processedOffset; // processed offset
```
sfc-gh-tjones, line 157: ? Should we just remove this then? This is a confusing set of comments, it's unclear if this is a `TODO` or `I'm not sure what's going on here` type of statement

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -237,7 +267,55 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
       if (copiedStreamingBuffer != null) {
         insertBufferedRecords(copiedStreamingBuffer);
       }
+    } else {
+      LOGGER.debug(
+          "Skip adding offset:{} to buffer for channel:{} because"
+              + " offsetPersistedInSnowflake:{}, processedOffset:{}",
+          kafkaSinkRecord.kafkaOffset(),
+          this.getChannelName(),
+          this.offsetPersistedInSnowflake.get(),
```
sfc-gh-tjones, line 276: We might want to pull this out above the `if` statement and use it in the `if` conditional and in the else block as the value could change by the time we log this which could make debugging tricky.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -237,7 +267,55 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
...
+          this.processedOffset.get());
+    }
+  }
+
+  /**
+   * If kafka offset was recently reset, we will skip adding any more records to buffer until we see
+   * a desired offset from kafka.
+   *
+   * <p>Desired Offset from Kafka = (offset present in snowflake + 1)
```
sfc-gh-tjones, line 285: `persisted` instead of `present`?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -237,7 +267,55 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
...
+   *
+   * <p>In other words
+   *
+   * <p>Desired Offset from Kafka = Offset which was reset into kafka. (Check {link {@link
+   * TopicPartitionChannel#resetChannelMetadataAfterRecovery\}}
```
sfc-gh-tjones, line 290: `\` at the end should be removed?

sfc-gh-tjones, line 290: I'd drop this, I think the above comment covers it.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -237,7 +267,55 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
...
+   *
+   * @param kafkaSinkRecord Record to check for above condition only in case of failures
+   *     (isOffsetResetInKafka = true)
+   * @return true if this record can be skipped to add into buffer, false otherwise.
+   */
+  private boolean shouldIgnoreAddingRecordToBuffer(SinkRecord kafkaSinkRecord) {
+    // we should not buffer any record whose value is more than what we reset in kafka in case of
+    // failures
+    if (this.isOffsetResetInKafka.get()
```
sfc-gh-tjones, line 299: Should we pull out `offsetPersistedInKafa.get()` above the `if`? Can it change midway through?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -237,7 +267,55 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
...
+          kafkaSinkRecord.kafkaOffset(),
+          this.getChannelName(),
+          this.offsetPersistedInSnowflake.get());
+      return true;
+    } else if (this.isOffsetResetInKafka.get()
+        && (kafkaSinkRecord.kafkaOffset() - offsetPersistedInSnowflake.get()) == 1L) {
+      LOGGER.debug(
+          "Got the desired offset:{} from Kafka, we can add this offset to buffer for channel:{}",
+          kafkaSinkRecord.kafkaOffset(),
+          this.getChannelName());
```
sfc-gh-tjones, line 314: Will this be a noisy log line?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -377,26 +455,33 @@ InsertValidationResponse insertBufferedRecords(StreamingBuffer streamingBufferTo
...
+          this.getChannelName(),
+          streamingBufferToInsert,
+          response.hasErrors());
+      if (response.hasErrors()) {
+        handleInsertRowsFailures(
+            response.getInsertErrors(), streamingBufferToInsert.getSinkRecords());
+      }
+      return response;
+    } catch (TopicPartitionChannelInsertionException ex) {
+      // Suppressing the exception because other channels might still continue to ingest
```
sfc-gh-tjones, line 479: Eh? I don't think we should be suppressing errors here? Can you explain why we'd want to swallow exceptions?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelInsertionException.java
```diff
@@ -0,0 +1,19 @@
+package com.snowflake.kafka.connector.internal.streaming;
+
+/**
+ * Class for exceptions that occur while interacting with Snowflake through Streaming Snowpipe.
```
sfc-gh-tjones, line 4: Nit, `Snowpipe Streaming`

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelInsertionException.java
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * <p>Use this exception when a particular channel (Topic Partition) fails to insert Rows into
+ * Snowflake Table.
```
sfc-gh-tjones, line 7: Client or Server side failure with respect to inserting rows?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Not needed, will remove.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Will do!

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: I don't think it differentiates at the moment. It will just reopen the channel in case of SFException. SFException is overloaded in the sense it is sent as exception when error occurs from both sides. 

Ideally we should backoff only if it sends 429?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: No, I think it is meant to be like that. Let me make it more clear. 
`In both above cases, we ask Kafka to send back offsets, strictly from offset number after the offsetToken present in Snowflake`

I will replace it.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Sorry, I didn't clean this. 🥇 

`processedOffset` is actually needed in case when offsetToken is not updated and it still gives us null but we know what we have processed up until that point. 

will remove the comment/TODO

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Actually the `offsetPersistedInSnowflake` would not change elsewhere. It will only change after it fetches the offsetToken when this `put` operation is completed and it goes into `preCommit`. It happens in the same thread and `offsetPersistedInSnowflake` is not set in separate thread.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: yes!

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Ahh thanks for the catch!

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: It's a debug level log and I think we wont run into this issue many times I am assuming. It will help us debug more to see if we got desired offset.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: No, it cant change midway.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelInsertionException.java
sfc-gh-japatel: I dont differentiate it at the moment. So may be both.

sfc-gh-japatel:
@sfc-gh-tjones this is ready for another look.

sfc-gh-tzhang:
> * In case of `insertRows` failures, subsequent records from this batch (Batch sent in put API of SinkTask) were successfully calling insertRows since we were reopening the channel.

I believe I understand the issue based on our chat, but I don't think I understand your solution. At a high level, if we get an invalid channel exception, we need to 
1. Block the PUT API (or block inserting into the KC buffer)
2. Cleanup the KC buffer
3. Reopen the channel
4. Reset KC offset
5. Unblock PUT API
Does above make sense or you're doing a different approach?

sfc-gh-japatel:
> > * In case of `insertRows` failures, subsequent records from this batch (Batch sent in put API of SinkTask) were successfully calling insertRows since we were reopening the channel.
> 
> I believe I understand the issue based on our chat, but I don't think I understand your solution. At a high level, if we get an invalid channel exception, we need to
> 
> 1. Block the PUT API (or block inserting into the KC buffer)
> 2. Cleanup the KC buffer
> 3. Reopen the channel
> 4. Reset KC offset
> 5. Unblock PUT API
>    Does above make sense or you're doing a different approach?

100% this is the right approach.. And this is how it is implemented right now..

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelInsertionException.java
sfc-gh-tjones: Ack, we may want to clarify that here.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -86,12 +88,33 @@ public class TopicPartitionChannel {
...
+   * </ol>
+   *
+   * <p>In both cases above, we ask Kafka to send back offsets, strictly from offset number after
+   * the offset present in Snowflake. i.e of Snowflake has offsetToken = x, we are asking Kafka to
+   * start sending offsets again from x + 1
+   *
+   * <p>This boolean is used to indicate that we reset offset in kafka and we will only buffer once
+   * we see the offset which is one more than an offset present in Snowflake.
+   */
+  private AtomicBoolean isOffsetResetInKafka = new AtomicBoolean(false);
```
sfc-gh-tjones, line 116: When we chatted offline about this we discussed whether this should be `AtomicBoolean` and if that's representative of a code smell because of potential threading issues that we may not be aware of. Do we need to use `AtomicBoolean` here or can we instead use `boolean`?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tjones: I see makes sense. Maybe pull it out anyway so we can reduce the number of calls to `get()`?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -237,7 +264,52 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
...
+          this.offsetPersistedInSnowflake.get());
+      return true;
+    } else if (this.isOffsetResetInKafka.get()
+        && (kafkaSinkRecord.kafkaOffset() - offsetPersistedInSnowflake.get()) == 1L) {
+      LOGGER.debug(
+          "Got the desired offset:{} from Kafka, we can add this offset to buffer for channel:{}",
+          kafkaSinkRecord.kafkaOffset(),
+          this.getChannelName());
+      this.isOffsetResetInKafka.set(false);
+      return false;
```
sfc-gh-tjones, line 310: Can we fall through below?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tjones: Oh I see, so this is partially related to how KC handles failures?

sfc-gh-tjones (approved):
This LGTM. I have some additional questions/comments but I don't think any are blockers ATM. Let's have Toby or Matt review as well and then let's ship it!

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: Also, per row logging seems too much?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -237,7 +264,52 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
...
+   * <p>Check {link {@link TopicPartitionChannel#resetChannelMetadataAfterRecovery}} for reset logic
+   *
+   * @param kafkaSinkRecord Record to check for above condition only in case of failures
+   *     (isOffsetResetInKafka = true)
+   * @return true if this record can be skipped to add into buffer, false otherwise.
+   */
+  private boolean shouldIgnoreAddingRecordToBuffer(SinkRecord kafkaSinkRecord) {
+    // we should not buffer any record whose value is more than what we reset in kafka in case of
+    // failures
+    if (this.isOffsetResetInKafka.get()
```
sfc-gh-tzhang, line 293: Reformat suggestion:
if (!this.isOffsetResetInKafka.get())
{
        return false; (or true?)
}

if kafkaSinkRecord.kafkaOffset() - offsetPersistedInSnowflake.get()) == 1L
{
        do something
}
else
{
        ignore
}

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -237,7 +264,52 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
...
+        && (kafkaSinkRecord.kafkaOffset() - offsetPersistedInSnowflake.get()) != 1L) {
+      // ignore
+      LOGGER.debug(
+          "Skip/Ignore adding offset:{} to buffer for channel:{} because we recently encountered"
+              + " error and reset offset in Kafka. offsetPersistedInSnowflake:{}",
+          kafkaSinkRecord.kafkaOffset(),
+          this.getChannelName(),
+          this.offsetPersistedInSnowflake.get());
```
sfc-gh-tzhang, line 301: Per row logging seems too much, why not log it at the PUT batch level?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -237,7 +264,52 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
...
+      return true;
+    } else if (this.isOffsetResetInKafka.get()
+        && (kafkaSinkRecord.kafkaOffset() - offsetPersistedInSnowflake.get()) == 1L) {
+      LOGGER.debug(
+          "Got the desired offset:{} from Kafka, we can add this offset to buffer for channel:{}",
+          kafkaSinkRecord.kafkaOffset(),
+          this.getChannelName());
+      this.isOffsetResetInKafka.set(false);
```
sfc-gh-tzhang, line 309: Is it possible that the offset is set to true after the if check above, then set to false again here?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -86,12 +88,33 @@ public class TopicPartitionChannel {
...
   // used to communicate to the streaming ingest's insertRows API
   // This is non final because we might decide to get the new instance of Channel
   private SnowflakeStreamingIngestChannel channel;
 
+  /**
+   * Offsets are reset in kafka when one of following cases arises in which we rely on source of
+   * truth (Which is Snowflake's committed offsetToken)
+   *
+   * <ol>
+   *   <li>If channel fails to fetch offsetToken in kafka, we reopen the channel and try to fetch
```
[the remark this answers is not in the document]
sfc-gh-japatel: Yes,  `from snowflake` will change.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -86,12 +88,33 @@ public class TopicPartitionChannel {
...
+   *   <li>If channel fails to ingest a buffer(Buffer containing rows/offsets), we reopen the
+   *       channel and try to fetch offset from Snowflake again
+   * </ol>
+   *
+   * <p>In both cases above, we ask Kafka to send back offsets, strictly from offset number after
+   * the offset present in Snowflake. i.e of Snowflake has offsetToken = x, we are asking Kafka to
+   * start sending offsets again from x + 1
+   *
+   * <p>This boolean is used to indicate that we reset offset in kafka and we will only buffer once
+   * we see the offset which is one more than an offset present in Snowflake.
```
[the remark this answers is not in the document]
sfc-gh-japatel: Will do!

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Yes, per partition logic need not be thread safe but for sake of not mixing two things, I went ahead with keeping everything threadsafe. 

I will open up another PR for making all related changes.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Will put it out at top.. 
Re: per row logging-> only in case of skipped offsets + debug level log, so should be fine IMO.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Sure, can do this!

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Yess.. removed return false from `else if`

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Actually not possible because of synchronous nature of one partition.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Correct, more of how this doesnt qualify for `ConnectException` which can cause the `SinkTask` to fail.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -445,43 +527,47 @@ private InsertValidationResponse insertRowsWithFallback(StreamingBuffer buffer)
...
+      throws TopicPartitionChannelInsertionException {
     final long offsetRecoveredFromSnowflake =
         streamingApiFallbackSupplier(StreamingApiFallbackInvoker.INSERT_ROWS_FALLBACK);
-    LOGGER.warn(
-        "{} Fetched offsetToken:{} for channel:{}",
-        StreamingApiFallbackInvoker.INSERT_ROWS_FALLBACK,
-        offsetRecoveredFromSnowflake,
-        this.getChannelName());
-    throw new RetriableException(
+    throw new TopicPartitionChannelInsertionException(
```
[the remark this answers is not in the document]
sfc-gh-japatel: If I dont throw an exception, the insertRows call which happened earlier would be thought of as a successful call and I couldnt return an empty `InsertValidationResponse`

More or so here is the answer: https://github.com/snowflakedb/snowflake-kafka-connector/pull/423/files#r848941980

Check `insertBufferedRecords`. If I dont throw the exception, the insertRows API would come out as a successful call, which I didnt wanted. I specifically wanted to log the original error. (Which I pass into `insertRowsFallbackSupplier`)

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelInsertionException.java
```diff
@@ -0,0 +1,15 @@
+package com.snowflake.kafka.connector.internal.streaming;
+
+/**
+ * Class for exceptions that occur while interacting with Snowflake through Snowpipe Streaming.
```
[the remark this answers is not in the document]
sfc-gh-japatel: `SFException` is not what I wanted to throw because we reopen the channel on that. I couldnt through ConnectException either since it would fail the task, hence I decided to create another RTE instance which could be more suitable for insertRows failures which can help us reset offset + get desired offset in Kafka. 

I used `RetriableException` before but that was incorrect. https://github.com/snowflakedb/snowflake-kafka-connector/pull/423/files#r848941980

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelInsertionException.java
sfc-gh-japatel: True that, may be I misunderstood where the exception was coming from. Most likely it was coming from Client SDK since the clientseq was bumped up and it never reached server. Let me clarify.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Printing batch level wont be possible right now because we are inserting every record one by one! Would be possible when I acquire lock at batch rather than at record level.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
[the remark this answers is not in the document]
sfc-gh-tzhang: ok, then we need a custom exception in this case? Can we just throw a RuntimeException? Or maybe create an exception with a more general name so all other places could use it as well if needed?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: Could you add a TODO with the same JIRA? I believe this is a perf critical piece

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -237,7 +268,51 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
...
+      if ((kafkaSinkRecord.kafkaOffset() - offsetPersistedInSnowflake.get()) != 1L) {
+        // ignore
+        LOGGER.debug(
+            "Ignore adding offset:{} to buffer for channel:{} because we recently encountered"
+                + " error and reset offset in Kafka. offsetPersistedInSnowflake:{}",
+            kafkaSinkRecord.kafkaOffset(),
+            this.getChannelName(),
+            this.offsetPersistedInSnowflake.get());
+        return true;
+      } else if ((kafkaSinkRecord.kafkaOffset() - offsetPersistedInSnowflake.get()) == 1L) {
```
sfc-gh-tzhang, line 307:
```suggestion
      } else
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -237,7 +268,51 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
...
+        LOGGER.debug(
+            "Got the desired offset:{} from Kafka, we can add this offset to buffer for channel:{}",
+            kafkaSinkRecord.kafkaOffset(),
+            this.getChannelName());
+        this.isOffsetResetInKafka.set(false);
+      }
     }
+    return false;
```
sfc-gh-tzhang, line 315: nit: you only need return false in the end, so something like
if (this.isOffsetResetInKafka.get())
{

}
return false;

sfc-gh-tzhang (approved):
lgtm, left a few comments

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: then we don't need this to be atomic, right?

sfc-gh-japatel: True, I will change everything to use non atomic variables in upcoming PR.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: Yeah, I have that only return false in the end? What am I missing?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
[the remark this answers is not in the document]
sfc-gh-japatel: I specifically wanted it to represent an exception happened while doing `insert`. 
I can change the name in future if we need more use cases?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: you have another return false at line 296

sfc-gh-japatel: Ahh oops! Yes! Thanks Toby.

PR #429, opened 2022-04-20
SNOW-521833 Release branch v1.6.6

sfc-gh-japatel:
Update new version of KC with updated ingest sdk to 0.10.6

- Wont push this to master.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,7 +12,7 @@
 
     <groupId>com.snowflake</groupId>
     <artifactId>snowflake-kafka-connector</artifactId>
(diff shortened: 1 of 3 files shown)


sfc-gh-japatel:
@sfc-gh-tzhang I am releasing 1.6.6 with this changes only. 

The GFMT is failing because it uses older version and we explicitly set it in later commits. 

So, I think we are good to go with this release. 

Let me know if you feel otherwise.

sfc-gh-tzhang (approved):

PR #427, opened 2022-04-19
SNOW-447423 Enable streaming tests 

sfc-gh-japatel:
- Enabling more tests. 
- Once Integration tests pass for prod1, I will enable params for gcs and azure too. 

Adding snowflake role to the profile file and encrypting again. 
Steps:
- gpg --symmetric --cipher-algo AES256 profile_gcs.json

PR #431, opened 2022-04-23
Adding end to end tests for Snowpipe Streaming

sfc-gh-japatel:
Adding two End to End tests which will run against all PRs
Here is what the output looks like

TLDR: 
- Checks uniqueness in all three partitions
- Checks count of all three partitions
- Both Json and Avro tests
```
Sending in Partition:0
Sending in Partition:1
Sending in Partition:2
20:09:42  === Done TestSnowpipeStreamingStringAvro ===

20:09:42  === Sleep 60 secs before verify result in Snowflake DB ===

20:10:42  === Verify TestSnowpipeStreamingStringAvro ===
Count records in table travis_correct_snowpipe_streaming_string_avro_xyz=3000
Topic:travis_correct_snowpipe_streaming_string_avro_xyz count is exactly 3000
Duplicates:None
[(1000, 0), (1000, 1), (1000, 2)]
20:10:44  === Passed TestSnowpipeStreamingStringAvro ===

20:10:44  === All test passed ===

20:10:45  === Round 1 ===

20:10:45  === Drop table travis_correct_snowpipe_streaming_string_avro_xyz ===
20:10:46  === Drop stage SNOWFLAKE_KAFKA_CONNECTOR_travis_correct_snowpipe_streaming_string_avro_xyz_STAGE_travis_correct_snowpipe_streaming_string_avro_xyz ===
20:10:46  === Drop pipe SNOWFLAKE_KAFKA_CONNECTOR_travis_correct_snowpipe_streaming_string_avro_xyz_PIPE_travis_correct_snowpipe_streaming_string_avro_xyz_0 ===
20:10:46  === Done ===

20:10:46  === All clean done ===
```


sfc-gh-japatel:
E2E tests will fail I need to base this on top of https://github.com/snowflakedb/snowflake-kafka-connector/pull/427

test/test_suit/test_snowpipe_streaming_string_json.py
```diff
@@ -0,0 +1,76 @@
...
+
+        for p in range(self.partitionNum):
+            print("Sending in Partition:" + str(p))
+            key = []
+            value = []
+            for e in range(self.recordNum):
+                value.append(json.dumps(
+                    {'numbernumbernumbernumbernumbernumbernumbernumbernumbernumbernumbernumber': str(e)}
+                ).encode('utf-8'))
+            self.driver.sendBytesData(self.topic, value, key, partition=p)
```
sfc-gh-tzhang, line 38: seems like `key` is always empty?

sfc-gh-tzhang (approved):

test/test_suit/test_snowpipe_streaming_string_avro_sr.py
```diff
@@ -0,0 +1,91 @@
+from test_suit.test_utils import RetryableError, NonRetryableError
+from time import sleep
+from confluent_kafka import avro
+
+# SR -> Schema Registry
+# Runs only in confluent test suite environment
+class TestSnowpipeStreamingStringAvroSR:
```
sfc-gh-tzhang, line 7: Might be worth to run with and without schema registry?

sfc-gh-tzhang: > But since we got away with SnowflakeConverterWithoutSchemaRegistry, we are implicitly saying we wont support avro without SR.

hmmm, this sounds like an issue, why we plan to stop supporting this? Do we know how many customers are using AVRO without SR before making this decision?

sfc-gh-japatel: Because of error handling. This was already decided when we went ahead with a plan of deprecating this custom snowflake converters because it will not support error handling to DLQ. 

There are other open source converters like https://github.com/farmdawgnation/registryless-avro-converter

Which can be used. 

> Do we know how many customers are using AVRO without SR before making this decision?
Nope. May be we can ask this to confluent. 

Also, we will not support for Snowpipe Streaming only - which is only PrPr. For snowpipe it is still supported and hopefully we will get more metrics from the changes I did - but requires customers to update it to latest version.

PR #435, opened 2022-05-02
Snowflake validation timeout

SajanaW:
## Problem

We're running into a number of issues in production where Snowflake connection calls hang, which leads to synchronous validation calls hanging on the UI. These synchronous call are then timed out by the gateway - which lead to an ominous error message on the UI: `UNKNOWN ERROR`.

## Solution

Add a network time out to validation calls


sfc-gh-japatel:
Do you know what API is timing out? Is it the snowpipe API? 

If it's snowpipe, it doesnt go through JDBC, they are rest APIs (jdbc session less)

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
```diff
@@ -59,6 +59,8 @@ public class SnowflakeSinkConnector extends SinkConnector {
   // Using setupComplete to synchronize
   private boolean setupComplete;
 
+  private int VALIDATION_NETWORK_TIMEOUT = 45000;
```
sfc-gh-japatel, line 62: nit: `static final`

sfc-gh-japatel: Can you use TimeUnit.SECONDS.toMillis(45) to indicate this is seconds? 
Also `VALIDATION_NETWORK_TIMEOUT_SECONDS`

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
```diff
@@ -113,7 +114,7 @@ static String timestampToDate(long time) {
    * @param sslEnabled if ssl is enabled
    * @return a Properties instance
    */
-  static Properties createProperties(Map<String, String> conf, boolean sslEnabled) {
+  static Properties createProperties(Map<String, String> conf, boolean sslEnabled, int networkTimeout) {
```
sfc-gh-japatel, line 117: javadoc for @networktimeout and `networkTimeoutSeconds`

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceFactory.java
```diff
@@ -18,6 +18,9 @@ public static class SnowflakeConnectionServiceBuilder extends Logging {
     private SnowflakeURL url;
     private String connectorName;
     private String taskID = "-1";
+    // 0 specifies no network timeout is set
+    // https://docs.snowflake.com/en/user-guide/jdbc-parameters.html#networktimeout
+    private int networkTimeOut = 0;
```
sfc-gh-japatel, line 23: `networkTimeOutSeconds`

sfc-gh-japatel:
lgtm overall but we will need to go through legal agreement, where you need to sign CLA. I will send details to you over slack. Thanks for contribution!

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
SajanaW: Since the API takes in MS. I converted all to MS instead

SajanaW:
> Do you know what API is timing out? Is it the snowpipe API?

This is a fix for the Jdbc network calls that happen to verify user credentials/ DB information

sfc-gh-japatel:
lgtm! I will hold off pushing as of yet..I will have to run the merge by cloning this PR.

PR #436, opened 2022-05-02
 Snowflake Cleaner thread should kill the task if it fails to verify files have landed in Snowpipe after retries

SajanaW:
## Problem
Every Snowflake task spins up a cleaner thread. By design, every record is written to an internal stage and then it signals to a corresponding Snow pipe that this file should be ingested. The cleaner thread then tracks the written file on the stage and the corresponding ingestion job- and after Snow pipe has successfully ingested the file, the cleaner will delete the file.

In other words, a sink operation to snowflake is not-atomic and could fail after it is written to the stage but BEFORE it's ingested by Snow pipe. When this happens this connector does not kill the task: it continues to write to the stage and the cleaner continues to retry.

In the best case, this confuses the customer because the connector continues to work without failing. 
In the worst case, this could build up a huge backlog on the stage so that when we read the filenames from the stage - it fills up a list to a huge number of nodes (21 M for example) which kills heap space. 

## Solution
In order to bubble up the right message to the customer we should just kill the task when we encounter an issue with moving data from the internal stage to SNOW PIPE.

The main concern here is that we've already committed offsets for the records which have failed to move to SNOW PIPE. However since we load files from the stage and clean up based on their ingestion report - we are effectively still not losing data (it's persisted in the stage).


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceFactory.java
```diff
@@ -146,6 +146,14 @@ public SnowflakeSinkServiceBuilder setSinkTaskContext(SinkTaskContext sinkTaskCo
       return this;
     }
 
+    public SnowflakeSinkServiceBuilder setMaxCleanerRetries(
```
sfc-gh-japatel, line 149: nit: java docs on public methods.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -130,6 +135,16 @@ public void startTask(final String tableName, final TopicPartition topicPartitio
...
+    Exception potentialEx = failedCleanerException.get();
+    if (potentialEx != null) {
+      for (ServiceContext pipe : pipes.values()) {
+        // flush current buffers
+        pipe.flushBuffer();
+      }
+
+      throw new ConnectException("Cleaner encountered an exception", potentialEx);
+    }
+
```
sfc-gh-japatel, line 147: can you extract this into a private method? may be call it something like `cleanerThreadExceptionHandling()`?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -575,6 +596,10 @@ private void startCleaner(long recordOffset, SnowflakeTelemetryPipeCreation pipe
                     e.getMessage(),
                     e.getStackTrace());
                 telemetryService.reportKafkaConnectFatalError(e.getMessage());
+                if (maxCleanerRetries >= 0 &&  pipeStatus.cleanerRestartCount.intValue() == maxCleanerRetries) {
+                  failedCleanerException.set(e);
+                  break;
+                }
                 forceCleanerFileReset = true;
```
sfc-gh-japatel, line 603: This will be unreachable if above `if` case is true and we will never call `resetCleanerFiles` right?

src/test/java/com/snowflake/kafka/connector/internal/CleanerTest.java
```diff
@@ -0,0 +1,60 @@
...
+        service.startTask(table, topic, partition);
+
+        SnowflakeConverter converter = new SnowflakeJsonConverter();
+        SchemaAndValue input =
+                converter.toConnectData(topic, "{\"name\":\"test\"}".getBytes(StandardCharsets.UTF_8));
+        long offset = 0;
+        SinkRecord record1 =
+                new SinkRecord(
+                        topic, partition, Schema.STRING_SCHEMA, "test", input.schema(), input.value(), offset);
+        // trigger cleaner failure
```
sfc-gh-japatel, line 52: can you add a comment here saying what is the cleaner failure reason?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -130,6 +135,16 @@ public void startTask(final String tableName, final TopicPartition topicPartitio
 
   @Override
   public void insert(final Collection<SinkRecord> records) {
+    Exception potentialEx = failedCleanerException.get();
+    if (potentialEx != null) {
+      for (ServiceContext pipe : pipes.values()) {
+        // flush current buffers
+        pipe.flushBuffer();
```
sfc-gh-japatel, line 142: Should we flush? Flushing here will only increase number of files present on internal stage. the `init()` method also does a list operation on fresh start of the task. 

The committing of offsets happen in `getOffset` when we call ingestFiles API in snowpipe. 

Also, if we are sending connect exception, all the `ServiceContext` objects will be reinitialized because we will have to restart the task.. Restart task-> `start()` method of SinkTask?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
SajanaW: Yup. If we've retried the max amount of times, we shouldn't do a forced cleaner file reset (which happens for each retry).

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
SajanaW: > Also, if we are sending connect exception, all the ServiceContext objects will be reinitialized because we will have to restart the task.. Restart task-> start() method of SinkTask? (Correct me if this is not the right API called on task restart) 

Yes this is right. Once the task fails - and if it's restarted by the user- we would reinit the `ServiceContext` objects. 

So the initial reasoning behind this was I thought we were committing offsets for successful write to the stage. But after another pass, I think we're committing offsets only based on successful ingestion into snowpipe. Therefore we can rely on the Kafka Connect framework to provide us with the right records after a restart so that we can avoid data loss. 

In fact, flushing the buffer is incorrect since it might end up writing duplicates to the stage - which might betray EOS semantics. (though I guess with a consistent naming scheme for the files in the stage - this write could be idempotent).  

I'll get rid of the flush!

sfc-gh-japatel: Sounds good!

sfc-gh-japatel: It's actually not idempotent since the file name has timestamp of flush. So it is probably better to not flush (Thanks for changing it)

sfc-gh-japatel:
lgtm. Thanks
I will hold off pushing this until you get to sign CLA and will open a new PR which should pass merge gates too.

sfc-gh-japatel (approved):
lgtm

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
chrisVia: > So the initial reasoning behind this was I thought we were committing offsets for successful write to the stage. But after another pass, I think we're committing offsets only based on successful ingestion into snowpipe. 

Are you sure this is correct? We observed that commiting only after confirmation of the ingestion-report that snowpipe has processed the files and their contents have made it to the snowflake table was _pre v1.4 behaviour_. Since this PR (https://github.com/snowflakedb/snowflake-kafka-connector/pull/152), it seemed to us that the connector intentionally changed to a more aggressive (and potentially premature) offset strategy where it commits immediately after records are flushed (and an insert-request is made to snowpipe).

I think this implies your initial observation is correct - if snowpipe fails to ingest records on the stage and it kills the connector tasks, when the user restarts the task, Kafka will not resend those failed records. They will be left in  orphaned files on the stage. They likely won't be purged automatically by `filterFileReprocess`, but they also won't be reingested by snowpipe since there is nothing to make an ingest-request on its behalf. This seems pretty alarming to me as it suggests that there is _effectively_ data-loss here. The connectors job is to get data from Kafka to snowflake tables, not merely to snowflake _stages_. It seems dangerous then to consider killing the connector task without moving the current offset-commit strategy back to the more conservative approach pre v1.4, no?

PR #437, opened 2022-05-06
SNOW-521833 Add more logs for debugging missing offsets

sfc-gh-japatel:
Add more logs for debugging

Follow up:
- We might want to call ingest Files API for files leaked on internal stage but were either not deleted or were not being put into ingest files API. There are few advantages
  - If before restart/rebalance (Assignment to new node) we had not called ingest Files for this files, we will reprocess those offsets. It is kind of redundant to do it also very expensive to just list the stage and then search for all those files in loadHistory scan from last 1 hour. 
  - If we had called insertFiles before rebalance, the new insertFiles call would be redundant because of snowpipe's deduplication logic on file name. It would be a no-op. 
- We will reset the offset in kafka to the last offset number from the list operation.


sfc-gh-japatel:
> Add more logs for debugging

sfc-gh-tjones (approved):

sfc-gh-tzhang (approved):

PR #438, opened 2022-05-13
Fidelity branch v1.6.4

sfc-gh-japatel:


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -97,6 +97,10 @@ public class SnowflakeSinkConnectorConfig {
   public static final String JMX_OPT = "jmx";
   public static final boolean JMX_OPT_DEFAULT = true;
 
+  // TESTING
+  public static final String REBALANCING = "snowflake.test.rebalancing";
+  public static final boolean REBALANCING_DEFAULT = false;
+
```
sfc-gh-tzhang, line 103: Please remove all the code related to rebalance testing

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -501,8 +501,14 @@ public void moveToTableStage(
       // get
       InputStream file;
       try {
+        logInfo(
+            "Downloading file: {} from stage: {} to move to table stage: {}",
+            name,
+            stageName,
+            tableName);
```
sfc-gh-tzhang, line 508: Let's try to keep the change small, all the logs related to file moving/deletion could be removed. If we want to look at this particular issue, Pfizer has everything as well

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -466,6 +466,7 @@ private ServiceContext(
     }
 
     private void init(long recordOffset) {
+      final long startTime = System.currentTimeMillis();
```
sfc-gh-tzhang, line 469: Could you explain why we're interested in the time spent in `init`?

sfc-gh-tzhang (commented):
Looks like most of changes here are not helping with the hang issue, could we focus on that? One change would be useful is to add a heartbeat log somewhere

sfc-gh-japatel:
Using this branch as testing for CP 5.3.5 and AK 2.6.1 (which is LCM of fixes, unfortunately 2.3.2 was never released)

PR #439, opened 2022-06-01
SNOW-514303

sfc-gh-zefan:
Modified code to allow n:1 topics:table mapping.


sfc-gh-japatel:
Is this ready for review or can be deleted? @sfc-gh-zefan

PR #450, opened 2022-06-24
Snowflake Cleaner thread should kill the task if it fails to verify files have landed in Snowpipe after retries

sagarrao12:
Duplicate of https://github.com/snowflakedb/snowflake-kafka-connector/pull/436


sfc-gh-japatel:
lgtm! I will be coauthor and run tests.

sfc-gh-japatel:
CLA has been signed by @sagarrao12

chrisVia:
Hello, in the linked PR, I've raised some concerns that this change may exacerbate existing potential data-loss issues caused by this connector having adopted a premature offset-commit strategy.

While _technically_ this might not be complete data-loss (the data which did not make it through snowpipe is still recoverable from the stage as you mention, unless externaly purged), I'm of the opinion that it would be _effectively_ data-loss. The purpose of the connector is to move data from Kafka into snowflake _tables_, not snowflake stages.

Imagine the scenario where:
1. The first record of topic `T`, partition `0` is flushed to Snowflake stage and an insert-file request is sent to the corresponding snowpipe
2. Snowflake connector immediately updates its committed offset, and instructs Kafka-Connect to advance its consumer-offsets
3. A while later, the cleaner receives the ingestReport from snowflake and realizes snowpipe has _failed_ to proccess that record. Cleaner restarts, retries, but the issue persists, we exhaust the max-retries and intentionally fail the connector task.
4. Lets say whatever underlying issue was causing snowpipe to fail to process the flushed files is solved, then we restart the connector task. But now, the connector now starts consuming data from offset 1 of topic `T`, partition `0` because the connector had already committed the offset (even though the record did not make it all the way to the table).
5. So instead of reprocessing the first record, we start with the second record of topic `T`, partition `0` being flushed to Snowflake stage. This time, the snowpipe works and the second record is sent to the snowflake table. But the poor first record is still just stuck in an orphaned file on the stage!

This sounds like data-loss to me. People querying the snowflake table will have no idea that they are missing records that are left orphaned on the connector's snowflake stage. I definitely think bubbling up the cleaner failures to connect task failures is super valuable, but if we do this, shouldn't the committable offset logic for the connector be reverted to the more conservative pre v1.4 behaviour where we don't commit offsets on record flush, but only once the ingest-report verifies the flushed records have made it through snowpipe)?

sfc-gh-japatel:
> 

Thanks @chrisVia for the explanation, your example scenario makes sense to me but let me explain how we have defined data loss and offset commit behavior. 

We rely on Snowpipe to ingest the file but if the file under ingestion is not completed, it is more of a data issue and should not be thought as data loss. I understand eventually the data needs to be in snowflake table and not snowflake table stage but putting data into snowflake table stage was our way to handle the exceptions. (Instead of DLQ which is the traditional way)

Also, on top of that snowpipe guarantees that if file was accepted by it, it will either be ingested or failed to ingest due to for instance data formatting issue. So we commit back to kafka once we get a positive response from Snowpipe that the file was successfully received by it. We use the insertReport API more as a way to delete files from internal stage.

chrisVia:
Thanks for the reply . I've thought about this some more and looked at the implementation and I think my example above wasn't quite appropriate for this specific PR. Even if snowpipe reports a failed ingested file, that is not an _exception_ and thus won't kill the connector task under this PR. I guess this PR is more for general unanticipated exceptions, not snowpipe ingest failures. It's not entirely clear to me now what sorts of exceptions this PR is actually meant to catch and translate into connector-task failures (@sagarrao12 I'm curious what specific cleaner behaviour/incident you were looking to handle here?).

I would understand the current offset commit behaviour assuming it is this project's opinion that Snowpipe ingest failures are purely data issues and should be treated in a dead-letter-queue pattern (ie. non-blocking and _non-fatal_) . But then I'm a bit confused about what this PR is actually trying to solve for 😕.

sagarrao12:
hey @chrisVia , thanks for your comments. TBH, I have picked up this change very recently so I dont have the full context. To answer your question, the test in this PR is throwing a RuntimeException which is very general purpose so I can't say for sure what specific errors this PR aims to solve. From what I have understood, it's mainly to put a cap on the number of retries for the cleaner thread which can lead to confusion or a huge backlog on the stage. Maybe @sfc-gh-japatel , might know of anything that was brought up in his earlier discussions/PR review ..

sfc-gh-japatel:
Picking this from the original PR's description: 

> In the best case, this confuses the customer because the connector continues to work without failing.
> In the worst case, this could build up a huge backlog on the stage so that when we read the filenames from the stage - it fills up a list to a huge number of nodes (21 M for example) which kills heap space.

This is a good enough reason to stop the connector instead of continuing it and creating huge backlog in internal stage. 

Anyways this is parameter protected so we can always disable this feature.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -617,6 +630,10 @@ private void startCleaner(long recordOffset, SnowflakeTelemetryPipeCreation pipe
                     e.getMessage(),
                     e.getStackTrace());
                 telemetryService.reportKafkaFatalError(e.getMessage());
+                if (maxCleanerRetries >= 0 &&  pipeStatus.cleanerRestartCount.get() == maxCleanerRetries) {
+                  failedCleanerException.set(e);
+                  break;
+                }
```
sfc-gh-japatel, line 636: Is this true?

sfc-gh-japatel: Can you fix the tests which are failing? https://github.com/snowflakedb/snowflake-kafka-connector/pull/453

sagarrao12: > Can you fix the tests which are failing? #453

@sfc-gh-japatel , I looked at the failing tests and noticed this one =>

```
19:26:11  === Sleep 60 secs before verify result in Snowflake DB ===
  File "test_verify.py", line 550, in execution
    driver.verifyWithRetry(test.verify, r)
  File "test_verify.py", line 112, in verifyWithRetry
    func(round)
  File "/home/runner/work/snowflake-kafka-connector/snowflake-kafka-connector/test/test_suit/test_pressure_restart.py", line 50, in verify
    self.driver.createConnector(self.getConfigFileName(), self.nameSalt)
  File "test_verify.py", line 337, in createConnector
    print(datetime.now().strftime("%H:%M:%S "), json.loads(r.content.decode("utf-8"))["name"], r.status_code)
19:27:11  === generate sink connector rest reqeuest from ./rest_request_template ===
```

Seems unrelated to the changes in this PR.. Or am I looking at the wrong test failures?

sagarrao12: > Is this true?

hmm I think it should be `if (maxCleanerRetries >= 0 &&  pipeStatus.cleanerRestartCount.get() >= maxCleanerRetries) {`

Because, resetCleanerFiles() always increments the cleanerRestartCount and if `maxCleanerRetries` is set to 0, then we will never reach the if condition. WDYT?

sagarrao12: @sfc-gh-japatel did you get a chance to look at my comments? Thanks!

sfc-gh-japatel: I was talking about the Integration test. 
`testCleanerRecover` is the failing test.

sfc-gh-japatel: I tried fixing the test but didnt spend too much time on it, can you run the test and fix the PR? I can help you with the profile.json file which you might not have.

sagarrao12: Sure @sfc-gh-japatel  I would fix it. Can you plz provide with the profile.json ?

PR #454, opened 2022-06-29
Pr 449 sagarrao

sfc-gh-japatel:
Open PR request: https://github.com/snowflakedb/snowflake-kafka-connector/pull/449

This is to push to release version 1.7.3

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
@@ -22,7 +22,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
-
 import org.apache.kafka.common.config.Config;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.connect.connector.Task;
@@ -236,9 +235,9 @@ public class SnowflakeSinkConnector extends SinkConnector {
     try {
       testConnection =
           SnowflakeConnectionServiceFactory.builder()
-                  .setProperties(connectorConfigs)
-                  .setNetworkTimeout(VALIDATION_NETWORK_TIMEOUT_MS)
-                  .build();
+              .setProperties(connectorConfigs)
+              .setNetworkTimeout(VALIDATION_NETWORK_TIMEOUT_MS)
+              .build();
 
     } catch (SnowflakeKafkaConnectorException e) {
       LOGGER.error(
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
@@ -59,6 +59,8 @@ public class SnowflakeSinkConnector extends SinkConnector {
   // Using setupComplete to synchronize
   private boolean setupComplete;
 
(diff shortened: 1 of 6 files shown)


sfc-gh-tjones (approved):

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
```diff
@@ -59,6 +60,8 @@ public class SnowflakeSinkConnector extends SinkConnector {
   // Using setupComplete to synchronize
   private boolean setupComplete;
 
+  private static final long VALIDATION_NETWORK_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(45);
```
sfc-gh-tzhang, line 63: I thought you have added a few timeout values in another PR, what's the difference between this and the others?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
```diff
@@ -231,7 +234,11 @@ public Config validate(Map<String, String> connectorConfigs) {
     SnowflakeConnectionService testConnection;
     try {
       testConnection =
-          SnowflakeConnectionServiceFactory.builder().setProperties(connectorConfigs).build();
+          SnowflakeConnectionServiceFactory.builder()
+              .setProperties(connectorConfigs)
+              .setNetworkTimeout(VALIDATION_NETWORK_TIMEOUT_MS)
```
sfc-gh-tzhang, line 239: Should we added it to `connectorConfigs` or move this to the same place whether you added the other timeout constants? What's so special about this one?

src/test/java/com/snowflake/kafka/connector/internal/ConnectionServiceIT.java
```diff
@@ -99,7 +99,7 @@ public void createConnectionService() {
         });
 
     SnowflakeURL url = TestUtils.getUrl();
-    Properties prop = InternalUtils.createProperties(TestUtils.getConf(), url.sslEnabled());
+    Properties prop = InternalUtils.createProperties(TestUtils.getConf(), url.sslEnabled(), 0);
```
sfc-gh-tzhang, line 102: could we add a test to verify non-zero timeout?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
sfc-gh-japatel: This timeout is for JDBC connection. https://docs.snowflake.com/en/user-guide/jdbc-parameters.html#networktimeout
Others were for httpclient which is used to call Snowpipe REST APIs.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
sfc-gh-japatel: Mentioned above. 
These timeouts is for checking if db exists, schema exists etc. (Connection through JDBC)

src/test/java/com/snowflake/kafka/connector/internal/ConnectionServiceIT.java
sfc-gh-japatel: [sagarrao12](https://github.com/sagarrao12)

PR #459, opened 2022-07-08
SNOW-514303 Support n:1 topics:table

sfc-gh-zefan:
Support multiple topics mapping to one table by deleting the check in the configuration.


test/rest_request_template/travis_correct_multiple_topic_to_one_table_snowpipe.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "SNOWFLAKE_CONNECTOR_NAME",
+  "config": {
+    "connector.class": "com.snowflake.kafka.connector.SnowflakeSinkConnector",
+    "snowflake.role.name": "TESTROLE_KAFKA",
+    "topics": "SNOWFLAKE_TEST_TOPIC",
+    "snowflake.topic2table.map": "NUM_TOPICS<3>NUM_TOPICS",
```
sfc-gh-tzhang, line 7: could you explain what is `NUM_TOPICS<3>NUM_TOPICS`? My understanding is that we need to provide a map here, so something like `topic1:table,topic2:table`

sfc-gh-zefan: Enclosed in <> is the number of topics to be generated. I added this to make the number of topics both configurable and compatible to the old behavior of the test script.

sfc-gh-zefan: It is basically equivalent to `topic1:table,topic2:table,topic3:table`, except that the name of the table and topics are not hard-coded.

sfc-gh-tzhang (approved):
LGTM!

src/test/java/com/snowflake/kafka/connector/ConnectorConfigTest.java
```diff
@@ -127,7 +127,7 @@ public void testDuplicatedTopic() {
     Utils.validateConfig(config);
   }
 
-  @Test(expected = SnowflakeKafkaConnectorException.class)
+  @Test()
```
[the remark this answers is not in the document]
sfc-gh-zefan: done

PR #460, opened 2022-07-12
SNOW-625820 Schema mapping

sfc-gh-zefan:
Added logic to add the 'RECORD_CONTENT' column if not present.
Added tests.


src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -177,7 +190,20 @@ public Map<String, Object> getProcessedRecordForStreamingIngest(SinkRecord recor
     final Map<String, Object> streamingIngestRow = new HashMap<>();
     for (JsonNode node : row.content.getData()) {
       try {
-        streamingIngestRow.put(TABLE_COLUMN_CONTENT, MAPPER.writeValueAsString(node));
+        if (schematizationEnable) {
```
sfc-gh-tzhang, line 193: Why we still need to convert the SinkRecord to a SnowflakeTableRow first and then to a MAP object, could we convert the SinkRecord to MAP directly?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -326,6 +349,43 @@ public boolean isTableCompatible(final String tableName) {
     return compatible;
   }
 
+  public void appendMetaColIfNotExist(final String tableName) {
```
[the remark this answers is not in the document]
sfc-gh-zefan: We might check the privilege beforehand to see if we have such permission, and error out if not.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
```diff
@@ -211,6 +211,7 @@ public interface SnowflakeConnectionService {
    * @param content file content
    */
   void putToTableStage(String tableName, String fileName, byte[] content);
+
```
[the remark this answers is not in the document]
sfc-gh-zefan: It was added by the formatter.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -127,6 +129,16 @@ public SnowflakeSinkServiceV2(
     this.behaviorOnNullValues = SnowflakeSinkConnectorConfig.BehaviorOnNullValues.DEFAULT;
 
     this.connectorConfig = connectorConfig;
+
+    if (connectorConfig.containsKey(SnowflakeSinkConnectorConfig.SCHEMATIZATION_ENABLE_CONFIG)) {
```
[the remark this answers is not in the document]
sfc-gh-zefan: If I get this correctly, this modification should be made in the SnowflakeSinkServiceV1.java, right?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
```diff
@@ -234,4 +235,6 @@ public interface SnowflakeConnectionService {
 
   /** @return the raw jdbc connection */
   Connection getConnection();
+
+  void appendMetaColIfNotExist(String tableName);
```
sfc-gh-japatel, line 239: nit: javadocs please

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -326,6 +326,47 @@ public boolean isTableCompatible(final String tableName) {
...
```
sfc-gh-japatel, line 329: nit: @ override annotation for all overridden methods from interface

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -326,6 +326,48 @@ public boolean isTableCompatible(final String tableName) {
...
+        if (result.getString(1).equals("RECORD_METADATA")) {
+          hasMeta = true;
+          if (result.getString(2).equals("VARIANT")) {
+            isVariant = true;
+          }
+          break;
+        }
+      }
+    } catch (SQLException e) {
+      logError("table {} doesn't exist", tableName);
```
sfc-gh-tzhang, line 352: Why we only log here instead of thrown the exception and stop the connector?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -326,6 +326,48 @@ public boolean isTableCompatible(final String tableName) {
...
+    }
+    try {
+      if (!isVariant) {
+        String MetaQuery;
```
sfc-gh-tzhang, line 356:
```suggestion
        String metaQuery;
```

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -326,6 +326,48 @@ public boolean isTableCompatible(final String tableName) {
...
+          MetaQuery = "alter table identifier(?) add RECORD_METADATA VARIANT";
+        } else {
+          throw SnowflakeErrors.ERROR_2012.getException("table name: " + tableName);
+        }
+        stmt = conn.prepareStatement(MetaQuery);
+        stmt.setString(1, tableName);
+        stmt.executeQuery();
+      }
+    } catch (SQLException e) {
+      throw SnowflakeErrors.ERROR_2013.getException(e);
```
sfc-gh-tzhang, line 367: could you add the table name here as well?

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-tzhang: I see, would it be hard to extract the error detection logic out as a helper function and use it in both cases?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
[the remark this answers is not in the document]
sfc-gh-tzhang: sounds good, did you add the logic somewhere?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-zefan: Because this is unlikely to happen. The table should pass the existence checking before a meta column can be appended.

sfc-gh-zefan: I could add an error about this, but it might only happen when the table is deleted immediately after the existence checking.

Issue #456, opened 2022-07-06
Still getting "Connector name is empty or invalid" with invalid Snowflake identifiers 


kguelseven:
Hi folks

I just run into the issue described in #62 with v1.8.0 and checked #115. I think these lines should be swapped for the validation to work with the converted connector name. 

https://github.com/snowflakedb/snowflake-kafka-connector/blob/d4816a598af7e8151406432b4983564757d0b2dc/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java#L87-L90

sfc-gh-japatel:
What is the application name you are using? Can you paste the error message?

kguelseven:
The connector is named korhan.snowflake-connector. 

> 2022-07-06 20:18:06,703 ERROR
[SF_KAFKA_CONNECTOR] name is empty or invalid. It should match Snowflake object identifier syntax. Please see the documentation. (com.snowflake.kafka.connector.Utils) [connector-thread-korhan.snowflake-connector]
2022-07-06 20:18:06,703 ERROR WorkerConnector{id=korhan.snowflake-connector} Error while starting connector (org.apache.kafka.connect.runtime.WorkerConnector) [connector-thread-korhan.snowflake-connector]
com.snowflake.kafka.connector.internal.SnowflakeKafkaConnectorException:
[SF_KAFKA_CONNECTOR] Exception: Invalid input connector configuration
[SF_KAFKA_CONNECTOR] Error Code: 0001
[SF_KAFKA_CONNECTOR] Detail: input kafka connector configuration is null, missing required values, or wrong input value

kguelseven:
Hi @sfc-gh-japatel could you look at my suggestion? The fix seems to be trivial in this case.

PR #461, opened 2022-07-14
Fix for invalid connector name - Issue 456

sfc-gh-japatel:
Added test for #456 

Just a swap of two lines.


sfc-gh-tzhang (approved):
lgtm

sfc-gh-tzhang:
Looks like the the GCS merge gate keeps failing, do we have any plan to fix it?

sfc-gh-japatel:
> Looks like the the GCS merge gate keeps failing, do we have any plan to fix it?

No plans. 
- We can either ignore it right now(which we are)
- Increase overall timeout because GCS is a bit slower than other two. 
  - Will need to spend more time on debugging it though. Will file a jira.

Issue #443, opened 2022-06-13
SnowflakeSinkConnectorConfig: key/value converters misplaced in teh const definitions


dlg99:
Should be switched, like `KEY_CONVERTER_CONFIG_FIELD = "key.converter"` otherwise will lead to unexpected problems downstream

https://github.com/snowflakedb/snowflake-kafka-connector/blob/4b90cdeb5695543a3dec6299aa0018b7ed92b423/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java#L163-L165

sfc-gh-tzhang:
Fix should be part of 1.8.1

Issue #87, opened 2020-01-03
Snowflake Connector log statements are in a different format than the configured Kafka Connect log format


yarmiganosca:
Example of Snowflake Connector log statement:
```
[SecurityManager-1(2981)] INFO net.snowflake.ingest.connection.SecurityManager - Creating Token with subject OPTORO.STAGING_KAFKA_CONNECT
```

Example of Kafka Connect log statement:
```
{"timestamp":"2020-01-02 11:12:33,806","level":"INFO","logger":"org.apache.kafka.connect.runtime.rest.RestServer","thread":"qtp1871312485-27317","message":"127.0.0.1 - - [02/Jan/2020:11:12:33 +0000] \"GET /connectors HTTP/1.1\" 200 255  1"}
```

This prevents us from ingesting the Connector's log statements into our logging pipeline.

yarmiganosca:
It also prevents us from even knowing when events that triggered log line actually happened, which is a bit of a problem.

sfc-gh-zli:
That component belongs to Snowflake Ingest SDK https://github.com/snowflakedb/snowflake-ingest-java. Please file an issue in their repo if this hasn't been fixed.

seruman:
@sfc-gh-zli besides `SecurityManager`, if multiple connectors are running on the same Kafka Connect cluster, there's no way to distinguish their logs. 

```
2020-06-16 17:49:51,416 INFO   ||  
[SF_KAFKA_CONNECTOR] SnowflakeSinkTask[ID:0]:put 220 records   [com.snowflake.kafka.connector.SnowflakeSinkTask]
```

```
"2020-06-16 17:52:48,184 INFO   ||  
[SF_KAFKA_CONNECTOR] searched 4 files in ingest report, found 3   [com.snowflake.kafka.connector.internal.SnowflakeIngestionServiceV1]
```

Also log aggregator usually uses new line to separate new logs. Using new lines inside sometimes causes our log aggregator to merge them to following lines.

**eg**, following line ingested as a single log;
```
2020-06-16 17:52:34,940 INFO   ||  
[SF_KAFKA_CONNECTOR] SnowflakeSinkTask[ID:0]:preCommit   [com.snowflake.kafka.connector.SnowflakeSinkTask]
[task-thread-$connector_name-0] INFO net.snowflake.ingest.SimpleIngestManager - Sending Request UUID - 
```

sfc-gh-zli:
[...]

PR #463, opened 2022-07-15
NO-SNOW: Fix two issues in Kafka Connector

sfc-gh-tzhang:
Remove new line for every KC log - https://github.com/snowflakedb/snowflake-kafka-connector/issues/87

Fix the mismatch of the config field name - https://github.com/snowflakedb/snowflake-kafka-connector/issues/443

### DIFF
--- a/.github/workflows/GoogleJavaFormat.yml
+++ b/.github/workflows/GoogleJavaFormat.yml
@@ -16,5 +16,5 @@ jobs:
         uses: axel-op/googlejavaformat-action@v3
         with:
           skipCommit: true
-          version: 1.10.0
+          version: 1.15.0
           args: "-n --set-exit-if-changed"
--- a/.github/workflows/GoogleJavaFormat.yml
+++ b/.github/workflows/GoogleJavaFormat.yml
@@ -16,5 +16,5 @@ jobs:
         uses: axel-op/googlejavaformat-action@v3
         with:
           skipCommit: true
-          version: 1.15.0
+          version: 1.10.0
           args: "-n --set-exit-if-changed"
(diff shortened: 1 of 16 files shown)


.github/workflows/GoogleJavaFormat.yml
```diff
@@ -16,5 +16,5 @@ jobs:
         uses: axel-op/googlejavaformat-action@v3
         with:
           skipCommit: true
-          version: 1.10.0
+          version: 1.15.0
```
sfc-gh-japatel, line 19: We deliberately added a version here 1.10.0 
Can you change wiki: https://snowflakecomputing.atlassian.net/wiki/spaces/EN/pages/1673103743/Kafka+Connector+Setup+Debugging if we need 1.15.0? 

Or you could revert it to 1.10.0 so that zekai and I dont have to update. Either Version is fine

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -160,9 +161,9 @@ public class SnowflakeSinkConnectorConfig {
    *     href="https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSinkTask.java#L332">Kafka
    *     Code for Converter</a>
    */
-  public static final String KEY_CONVERTER_CONFIG_FIELD = "value.converter";
+  public static final String KEY_CONVERTER_CONFIG_FIELD = "key.converter";
```
sfc-gh-japatel, line 164: Oops, thanks!

.github/workflows/GoogleJavaFormat.yml
sfc-gh-japatel: This is failing the gfmt too, might want to do it in separate PR though if you decide 1.15.0

sfc-gh-japatel (approved):
Couple of comments, but lgtm, thanks!

.github/workflows/GoogleJavaFormat.yml
sfc-gh-tzhang: revert back to 1.10.0

PR #464, opened 2022-07-20
NO-SNOW Rotate private keys for all three sfctest0 accounts

sfc-gh-japatel:
Rotate keys

Steps:
1. Delete the .gpg keys from your local mac. (Updating gpg wont work with decryption)
2. After rotating public key and updating each of your profile.json for aws, azure and gcs, 
  - Run `gpg --symmetric --cipher-algo AES256 profile.json`
3. Check decryption through ` gpg --quiet --batch --yes --decrypt --passphrase="$SNOWFLAKE_TEST_PROFILE_SECRET" \
--output profile_2.json profile.json.gpg`


sfc-gh-tzhang (approved):
thanks Jay!

PR #466, opened 2022-07-21
Schema mapping

sfc-gh-zefan:
Added logic to add the 'RECORD_CONTENT' column if not present.
Added tests.


sfc-gh-japatel:
This PR has 330 commits, can you rebase master and force push?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -152,6 +152,8 @@ public class SnowflakeSinkConnectorConfig {
           + "By default messages are not sent to the dead letter queue. "
           + "Requires property `errors.tolerance=all`.";
 
+  public static final String SCHEMATIZATION_ENABLE_CONFIG = "schematization.enable";
```
sfc-gh-japatel, line 155: I would keep this as `enable.schematization`

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -364,6 +364,16 @@ static String validateConfig(Map<String, String> config) {
       if (!BufferThreshold.validateBufferThreshold(config, IngestionMethodConfig.SNOWPIPE)) {
         configIsValid = false;
       }
+      if (config.containsKey(SnowflakeSinkConnectorConfig.SCHEMATIZATION_ENABLE_CONFIG)) {
```
sfc-gh-japatel, line 367: did you mean to check `snowpipe_streaming` here?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -364,6 +364,16 @@ static String validateConfig(Map<String, String> config) {
...
+        boolean enableSchematization =
+            Boolean.parseBoolean(
+                config.get(SnowflakeSinkConnectorConfig.SCHEMATIZATION_ENABLE_CONFIG));
+        if (enableSchematization) {
+          configIsValid = false;
+          LOGGER.error(
+              Logging.logMessage("Schematization is only available with SNOWPIPE_STREAMING."));
```
sfc-gh-japatel, line 374: Lets use the enum here for printing.. `IngestionMethodConfig`

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
```diff
@@ -234,4 +235,12 @@ public interface SnowflakeConnectionService {
 
   /** @return the raw jdbc connection */
   Connection getConnection();
+
+  /**
+   * append a VARIANT type column "RECORD_METADATA" to the table if not present only called when
```
sfc-gh-japatel, line 240:
```
<p> Append a VARIANT type column "RECORD_METADATA" to the table if it is not present.

<p> This method is only called when schematization is enabled
```

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
```diff
@@ -234,4 +235,12 @@ public interface SnowflakeConnectionService {
...
+   * schematization enabled
+   *
+   * @param tableName table name
+   */
```
sfc-gh-japatel, line 244:
```suggestion

  /**
   * <p> Append a VARIANT type column "RECORD_METADATA" to the table if it is not present.
   * <p> This method is only called when schematization is enabled
   *
   * @param tableName table name
   */
```

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-zefan: This check is to make sure snowpipe & enable.schematization are not set at the same time.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-zefan: Done.

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-japatel: nvm, didnt see the above if case. Thanks

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -326,6 +326,49 @@ public boolean isTableCompatible(final String tableName) {
...
+    boolean isVariant = false;
+    try {
+      stmt = conn.prepareStatement(query);
+      stmt.setString(1, tableName);
+      result = stmt.executeQuery();
+      while (result.next()) {
+        if (result.getString(1).equals("RECORD_METADATA")) {
+          hasMeta = true;
+          if (result.getString(2).equals("VARIANT")) {
+            isVariant = true;
```
sfc-gh-japatel, line 346: nit: These are all magic numbers here. (1) and (2). 

Might be good to explain what 1 and 2 means here. column name and data type but would be a good call out..

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -326,6 +326,49 @@ public boolean isTableCompatible(final String tableName) {
...
+    boolean hasMeta = false;
+    boolean isVariant = false;
+    try {
+      stmt = conn.prepareStatement(query);
+      stmt.setString(1, tableName);
+      result = stmt.executeQuery();
+      while (result.next()) {
+        if (result.getString(1).equals("RECORD_METADATA")) {
+          hasMeta = true;
+          if (result.getString(2).equals("VARIANT")) {
```
sfc-gh-japatel, line 345: will this always be variant or VARIANT? needs `equals ignorecase`?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -326,6 +326,49 @@ public boolean isTableCompatible(final String tableName) {
...
+    PreparedStatement stmt = null;
+    ResultSet result = null;
+    boolean hasMeta = false;
+    boolean isVariant = false;
+    try {
+      stmt = conn.prepareStatement(query);
+      stmt.setString(1, tableName);
+      result = stmt.executeQuery();
+      while (result.next()) {
+        if (result.getString(1).equals("RECORD_METADATA")) {
```
sfc-gh-japatel, line 343: consider extracting "RECORD_METADATA" to a static final string, something like
```
private static final String METADATA_COLUMN = "RECORD_METADATA"
```

and then using it here. I know table creation has not used it but may be we can start.

sfc-gh-japatel:
Why closed? @sfc-gh-zefan

sfc-gh-japatel:
Can you also add a JIRA?

sfc-gh-zefan:
I tried to do the rebase and force push. This will automatically close the related pr.

sfc-gh-zefan:
During rebasing there might be something wrong and a pr couldn't be created. I might create a brand new branch and copy all the changes so that it is clean. I will update the suggested changes on the way.

PR #467, opened 2022-07-21
SNOW-625820 Schema mapping

sfc-gh-zefan:
Added logic to add the 'RECORD_METADATA' column if not present.
Added tests.


src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -57,6 +64,8 @@ public class RecordService extends Logging {
   private static final String KEY_SCHEMA_ID = "key_schema_id";
   static final String HEADERS = "headers";
 
+  private boolean schematizationEnable = false;
```
sfc-gh-japatel, line 67: `enableSchematization`

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -87,6 +96,17 @@ public void setMetadataConfig(SnowflakeMetadataConfig metadataConfigIn) {
     metadataConfig = metadataConfigIn;
   }
 
+  public boolean setSchematizationEnable(final Map<String, String> connectorConfig) {
```
sfc-gh-japatel, line 99: Java docs for public methods.

test/test_verify.py
```diff
@@ -437,7 +448,7 @@ def runTestSet(driver, testSet, nameSalt, pressure):
 
     testStringJsonProxy = TestStringJsonProxy(driver, nameSalt)
 
-    # Run this test on both confluent and apache kafka
+    # # Run this test on both confluent and apache kafka
```
sfc-gh-japatel, line 451: do you need this change?

test/test_suit/test_schema_mapping.py
```diff
@@ -0,0 +1,72 @@
+from test_suit.test_utils import RetryableError, NonRetryableError
+import json
+
+
+class TestSchemaMapping:
```
sfc-gh-japatel, line 5: A small description here would be great.

sfc-gh-japatel:
In the description you have `Added logic to add the 'RECORD_CONTENT' column if not present.`
Do you mean record_metadata?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -326,6 +330,50 @@ public boolean isTableCompatible(final String tableName) {
...
+      }
+    } catch (SQLException e) {
+      logError("table {} doesn't exist", tableName);
+      throw SnowflakeErrors.ERROR_2014.getException("table name: " + tableName);
+    }
+    try {
+      if (!isVariant) {
+        String metaQuery;
+        if (!hasMeta) {
+          metaQuery = "alter table identifier(?) add RECORD_METADATA VARIANT";
```
sfc-gh-japatel, line 364: Do we need to make any changes in the docs when we publish about schema mapping, about some specific privilege/permissions?

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -177,7 +207,20 @@ public Map<String, Object> getProcessedRecordForStreamingIngest(SinkRecord recor
...
+        if (enableSchematization) {
+          Iterator<String> fieldNames = node.fieldNames();
+          while (fieldNames.hasNext()) {
+            String fieldName = fieldNames.next();
+            JsonNode fieldNode = node.get(fieldName);
+            String filedContent =
+                fieldNode.isTextual()
+                    ? fieldNode.textValue()
+                    : MAPPER.writeValueAsString(fieldNode);
+            streamingIngestRow.put(fieldName, filedContent);
```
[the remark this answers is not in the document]
sfc-gh-japatel: Also, I am assuming we only care about one level deep jsonNode? all rest of them are going into a variant column?

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -177,7 +207,20 @@ public Map<String, Object> getProcessedRecordForStreamingIngest(SinkRecord recor
     final Map<String, Object> streamingIngestRow = new HashMap<>();
     for (JsonNode node : row.content.getData()) {
       try {
-        streamingIngestRow.put(TABLE_COLUMN_CONTENT, MAPPER.writeValueAsString(node));
+        if (enableSchematization) {
+          Iterator<String> fieldNames = node.fieldNames();
+          while (fieldNames.hasNext()) {
+            String fieldName = fieldNames.next();
```
sfc-gh-japatel, line 213: nit: more like a columnName?

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -177,7 +207,20 @@ public Map<String, Object> getProcessedRecordForStreamingIngest(SinkRecord recor
...
+            JsonNode fieldNode = node.get(fieldName);
+            String filedContent =
```
sfc-gh-japatel, line 215: nit: more like a `columnValue`

sfc-gh-japatel (changes requested):
I think overall looks great, mostly cosmetic changes. Will wait for one more pass. 

Also, tests looks great, thanks for adding them!

test/test_suit/test_schema_mapping.py
```diff
@@ -0,0 +1,72 @@
...
+    def __init__(self, driver, nameSalt):
+        self.driver = driver
+        self.fileName = "travis_correct_schema_mapping"
+        self.topic = self.fileName + nameSalt
+
+        self.driver.snowflake_conn.cursor().execute(
+            "Create or replace table {} (PERFORMANCE_STRING STRING, PERFORMANCE_CHAR CHAR, PERFORMANCE_HEX BINARY, RATING_INT NUMBER, RATING_DOUBLE DOUBLE, APPROVAL BOOLEAN, APPROVAL_DATE DATE, APPROVAL_TIME TIME)".format(self.topic))
```
sfc-gh-japatel, line 12: can you squeeze in a variant here too??

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -177,7 +207,20 @@ public Map<String, Object> getProcessedRecordForStreamingIngest(SinkRecord recor
...
+                fieldNode.isTextual()
+                    ? fieldNode.textValue()
+                    : MAPPER.writeValueAsString(fieldNode);
```
sfc-gh-japatel, line 218: can you add a UT which will test this line? (I assume this is when a field itself is a json string)

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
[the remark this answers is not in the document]
sfc-gh-zefan: will do; Yep

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-zefan: The assumption is that the user themself needs to create the table with the correct schema. They will need the privilege to alter the table if the metadata column is added by us. In that case they need one of 'INSERT', 'ALL' or 'OWNERSHIP'.

test/test_suit/test_schema_mapping.py
```diff
@@ -0,0 +1,76 @@
+from test_suit.test_utils import RetryableError, NonRetryableError
+import json
+
+# test if each type of data fit into the right column with the right type
+# also test if the metadata column is automatically added
+class TestSchemaMapping:
```
sfc-gh-tzhang, line 6: if the kafka message has two columns c1 and C1, will we be able to mapped to the current column in the table?

sfc-gh-zefan: I believe they will be mapped into the same column C1 with the current behavior. I think we could add an option in the sf config to allow uppercase letter, and we can add the quotation marks in the connector.

sfc-gh-tzhang: Are you saying the second value will override the first value?

sfc-gh-zefan: I suppose so. I think this happens in the SDK.

sfc-gh-tzhang: Could we add a verification step in KC to check whether there're two columns with the same name if ignoring cases, and fail saying we can't supported?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -368,6 +370,10 @@ private boolean isContentBroken(final Object content) {
     return content != null && ((SnowflakeRecordContent) content).isBroken();
   }
 
+  /**
+   * TODO: SNOW-630885 - When schematization is enabled, we should create the map directly from the
+   * SinkRecord instead of first turning it into json
+   */
```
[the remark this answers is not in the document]
sfc-gh-zefan: Indeed there's no change here, but I believe this is where the change should be made. A function similar to the current handleNativeRecord should be added to produce a map return value.

test/test_suit/test_schema_mapping.py
sfc-gh-zefan: Just added the check and tested it. I could add a UT if you think it's necessary

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -186,6 +253,11 @@ public Map<String, Object> getProcessedRecordForStreamingIngest(SinkRecord recor
         e.printStackTrace();
       }
     }
+
+    if (hasCaseInsensitiveDuplicates(streamingIngestRow)) {
+      throw SnowflakeErrors.ERROR_0025.getException();
+    }
+
```
sfc-gh-tzhang, line 260: discussed offline, we don't need to do this anymore, we could have a function called `formatColumnName` which convert everything to uppercase if no double quotes, please take a look at how [ingest SDK](https://github.com/snowflakedb/snowflake-ingest-java/blob/83008cd9259fb18518f3167261aacdc664bdb992/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java#L617) is handling it

sfc-gh-tzhang: And could you add a test on double quotes column name?

sfc-gh-tzhang (approved):
LGTM other than my comments, thanks!

src/test/java/com/snowflake/kafka/connector/records/RecordContentTest.java
```diff
@@ -230,4 +231,44 @@ public void testConvertToJsonReadOnlyByteBuffer() {
     Schema schema = SchemaBuilder.bytes().build();
     assert RecordService.convertToJson(schema, buffer).toString().equals(expected);
   }
+
+  @Test
+  public void testSchematizationStringField() {
+    RecordService service = new RecordService();
+    SnowflakeJsonConverter jsonConverter = new SnowflakeJsonConverter();
```
sfc-gh-japatel, line 238: can you instead use JsonConverter in both the tests below? 

We are not using JsonConverter for `SNOWPIPE_STREAMING`

test/test_suit/test_schema_mapping.py
sfc-gh-japatel: Thanks @sfc-gh-zefan

sfc-gh-japatel (approved):
apart from one change in test, lgtm! Approving since your test has jsonconverter so I am assuming its working! 

Great stuff @sfc-gh-zefan

sfc-gh-japatel:
My only concern is if it is okay to not push it to master? 

You can push it to some `schema-mapping-master` branch.. 

Only reason to do this is we will be releasing new versions and we will have to pick our commits. Once we are ready for PrPr, we can bring the commits from `schema-mapping-master` to `master`

Thoughts @sfc-gh-tzhang @sfc-gh-zefan ?

PR #470, opened 2022-08-08
SNOW-628677 Auto Table Creation

sfc-gh-zefan:
Create the table automatically with the schema stored in schema-registry.

Only the Confluent Avro convertor is supported in this PR.

There are some known limitation to the Kafka connector schematization up to now as linked in the JIRA, which needs proper documentation.
[SNOW-644961](https://snowflakecomputing.atlassian.net/browse/SNOW-644961)


sfc-gh-zefan:
So far I use value.converter==io.confluent.connect.avro.AvroConverter to see if schema registry could be used. It will be updated to include other valid converters.

pom.xml
```diff
@@ -283,14 +283,14 @@
             <groupId>org.bouncycastle</groupId>
             <artifactId>bcpkix-fips</artifactId>
             <version>1.0.3</version>
-            <scope>provided</scope>
+            <!--            <scope>provided</scope>-->
```
sfc-gh-japatel, line 286: nit: remove

pom.xml
```diff
@@ -379,10 +379,12 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/io.confluent/kafka-connect-avro-data -->
```
sfc-gh-japatel, line 382: is this required?

pom.xml
```diff
@@ -394,8 +396,8 @@
         <dependency>
             <groupId>io.confluent</groupId>
             <artifactId>kafka-connect-avro-converter</artifactId>
-            <version>5.5.1</version>
-            <scope>test</scope>
+            <version>7.1.1</version>
+            <!--            <scope>test</scope>-->
```
sfc-gh-japatel, line 400: remove comments

pom.xml
```diff
@@ -477,5 +479,17 @@
             <version>0.9.0.0</version>
             <scope>test</scope>
         </dependency>
+        <!--        <dependency>-->
```
sfc-gh-japatel, line 482: remove commented code

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -152,6 +152,8 @@ public class SnowflakeSinkConnectorConfig {
           + "By default messages are not sent to the dead letter queue. "
           + "Requires property `errors.tolerance=all`.";
 
+  public static final String SCHEMATIZATION_ENABLE_CONFIG = "enable.schematization";
```
sfc-gh-japatel, line 155:
```suggestion
  public static final String ENABLE_SCHEMATIZATION_CONFIG = "enable.schematization";
```

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -622,4 +642,68 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
       }
     }
   }
+
+  public static Map<String, String> getSchemaFromSchemaRegistry(
+      final String topicName, final String schemaRegistryURL) {
+    return getSchemaFromSchemaRegistry(topicName, schemaRegistryURL, "value");
+  }
+
+  public static Map<String, String> getSchemaFromSchemaRegistry(
```
sfc-gh-japatel, line 651: consider making this method private

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -622,4 +642,68 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
       }
     }
   }
+
+  public static Map<String, String> getSchemaFromSchemaRegistry(
```
sfc-gh-japatel, line 646: nit: Javadocs for public methods 
(We follow some of effective java guidelines.
https://snowflakecomputing.atlassian.net/wiki/spaces/EN/pages/1960706106/Effective+Java+Guidelines+at+Snowflake)

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -622,4 +642,68 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
...
+      final String topicName, final String schemaRegistryURL) {
+    return getSchemaFromSchemaRegistry(topicName, schemaRegistryURL, "value");
+  }
+
+  public static Map<String, String> getSchemaFromSchemaRegistry(
+      final String topicName, final String schemaRegistryURL, final String type) {
```
sfc-gh-japatel, line 652: Can you also put a java doc here? 
Can you also explain what is String type? Why is it hardcoded to "value"

Better name?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -622,4 +642,68 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
...
+    Map<String, String> srConfig = new HashMap<>();
+    srConfig.put("schema.registry.url", schemaRegistryURL);
+    AvroConverterConfig avroConverterConfig = new AvroConverterConfig(srConfig);
```
sfc-gh-japatel, line 655: for my understanding, does providing schema registry mean the converter will always be avro?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -491,14 +498,44 @@ private void closeStreamingClient() {
...
-        throw SnowflakeErrors.ERROR_5003.getException("table name: " + tableName);
+        this.conn.appendMetaColIfNotExist(tableName);
       }
     } else {
       LOGGER.info("Creating new table {}.", tableName);
-      this.conn.createTable(tableName);
+      if (this.enableSchematization) {
+        if (connectorConfig
+            .get("value.converter")
+            .equals("io.confluent.connect.avro.AvroConverter")) {
```
sfc-gh-japatel, line 515: can you extract this out into a new method? Usually the idea is to keep only one job in a method. (Also called single responsibility principle, mostly applied to classes but can also be applied to this method) https://www.baeldung.com/java-single-responsibility-principle

From the readers perspective, this are just too many if else cases.. 

I would may be also keep this in a separate class may be called SchematizationUtils. 

In this class you can keep a connection object for creating classes
schema registry for fetching schema 
and all other future schema evolution utils can be kept in this class. 

Benefits:
1. SRP
2. Unit Testing 
3. Readability

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -991,13 +998,22 @@ protected long getApproxSizeOfRecordInBytes(SinkRecord kafkaSinkRecord) {
...
-      sinkRecordBufferSizeInBytes += ((String) entry.getValue()).length() * 2L; // 1 char = 2 bytes
+      Object value = entry.getValue();
+      if (value instanceof String) {
+        sinkRecordBufferSizeInBytes += ((String) value).length() * 2L; // 1 char = 2 bytes
+      } else {
+        // for now it could only be a list of string
+        for (String s : (List<String>) value) {
+          sinkRecordBufferSizeInBytes += s.length() * 2L;
+        }
+      }
```
sfc-gh-japatel, line 1009: do you need this now? We can add it once we support list? Add a TODO JIRA here if needed.

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -87,6 +98,27 @@ public void setMetadataConfig(SnowflakeMetadataConfig metadataConfigIn) {
...
+   * extract enableSchematization from the connector config and set the value for the recordService
+   *
+   * <p>The extracted boolean is returned for external usage.
+   *
+   * @param connectorConfig the connector config map
+   * @return a boolean indicating whether schematization is enabled
+   */
+  public boolean setEnableSchematizationFromConfig(final Map<String, String> connectorConfig) {
+    if (connectorConfig.containsKey(SnowflakeSinkConnectorConfig.SCHEMATIZATION_ENABLE_CONFIG)) {
+      enableSchematization =
```
sfc-gh-japatel, line 111: nit: this.schematization

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -87,6 +98,27 @@ public void setMetadataConfig(SnowflakeMetadataConfig metadataConfigIn) {
...
 
+  /**
+   * extract enableSchematization from the connector config and set the value for the recordService
+   *
+   * <p>The extracted boolean is returned for external usage.
+   *
+   * @param connectorConfig the connector config map
+   * @return a boolean indicating whether schematization is enabled
+   */
+  public boolean setEnableSchematizationFromConfig(final Map<String, String> connectorConfig) {
```
sfc-gh-japatel, line 109: can you name this setAndGetEnable
```suggestion
  public boolean setAndGetEnableSchematizationFromConfig(final Map<String, String> connectorConfig) {
```

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -87,6 +98,27 @@ public void setMetadataConfig(SnowflakeMetadataConfig metadataConfigIn) {
...
+    if (connectorConfig.containsKey(SnowflakeSinkConnectorConfig.SCHEMATIZATION_ENABLE_CONFIG)) {
+      enableSchematization =
+          Boolean.parseBoolean(
+              connectorConfig.get(SnowflakeSinkConnectorConfig.SCHEMATIZATION_ENABLE_CONFIG));
+    }
+    return enableSchematization;
+  }
+
+  public void setEnableSchematization(final boolean enableSchematizationIn) {
```
sfc-gh-japatel, line 118: annotate this with @ VisibleForTesting

sfc-gh-japatel: and javadoc which would entail this is only for testing. (Annotation does mean that so you can javadoc, but javadoc for @param would be great)

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -177,18 +211,56 @@ public Map<String, Object> getProcessedRecordForStreamingIngest(SinkRecord recor
     final Map<String, Object> streamingIngestRow = new HashMap<>();
     for (JsonNode node : row.content.getData()) {
       try {
-        streamingIngestRow.put(TABLE_COLUMN_CONTENT, MAPPER.writeValueAsString(node));
+        if (enableSchematization) {
```
sfc-gh-japatel, line 214: See my earlier comment for extracting this into a separate method in record service. Lot of things going on this method. IMO, deserves a new method.

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-zefan: Not really. It could also be in JsonSchema or in Protobuf. I didn't find any usage case in our code though. Maybe it isn't a safe assumption.

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -177,18 +211,56 @@ public Map<String, Object> getProcessedRecordForStreamingIngest(SinkRecord recor
...
     return streamingIngestRow;
   }
 
+  private String formatColumnName(String columnName) {
+    // the columnName has been checked and guaranteed not to be null or empty
+    return (columnName.charAt(0) == '"' && columnName.charAt(columnName.length() - 1) == '"')
+        ? columnName.substring(1, columnName.length() - 1)
+        : columnName.toUpperCase();
+  }
+
```
sfc-gh-japatel, line 263: can you write a blurb here mentioning what you do? may be in form of javadocs

test/rest_request_template/travis_correct_auto_table_creation.json
```diff
@@ -0,0 +1,29 @@
+{
```
sfc-gh-japatel, line 1: Thanks for adding Unit and End to end test. 
Can you also add IT test specifically testing the newly added code in Record Service? 

You can take a look at testNativeAvroInputIngestion test in SinkServiceV2IT test which will test your newly added code. 
Two reasons to add them:
1. Integration test is fastest way to check your expected behavior without hoping your kafka and kafka connector works or not. Although your End to end test will cover the behavior but it wont give you fast feedback. 
2. Code coverage.

sfc-gh-japatel: You will only have to add a new schematization config in your new test.

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-zefan: This is actually copied from the Ingest SDK to mimic the behavior. I will add the blurb.

pom.xml
sfc-gh-japatel: I am not against a version bump here but just want to know the reason. It is entirely fine to just keep it the latest and greatest.

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-japatel: I am okay with the limitation that schematization only works with avro right now, but can we mention it on top of the method?
There are two ways to tackle this. 
1. The config which customer passes has the value.converter 
   a. You can check before fetching schema from SR. 
2. If they dont have value.converter, it is likely they are fetching it from connect distributed properties value and you will get null from value.converter. 
   a. In this case, you will have to rely on calling the SR from the URL and hope it returns else you can say Schematization is not supported for non avro converters.

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-zefan: The schema registry could store schema for either the key and the value (for Kafka). Since most of the time we want the schema for the value, it is set as the default value.

sfc-gh-zefan: Any suggestion for the method name?

sfc-gh-japatel: I see. If we are only doing it for "value", we should explicitly mention it in javadocs. 

Can you start a JIRA for documentation purpose - I think this will be very helpful to announce this feature without looking at the code and finding out what the limitation is. Few I can think of are
1. non - array data type
2. only avro converter
3. Only value converter. 

May be "getValueSchemaFromSchemaRegistry"

pom.xml
sfc-gh-zefan: With the older version I couldn't use some functions to setup the schema registry client. I don't know if there's better way tho

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-zefan: Is it ok if I leave the behavior correct on the connector side?

pom.xml
```diff
@@ -359,7 +358,7 @@
         <dependency>
             <groupId>io.confluent</groupId>
             <artifactId>kafka-schema-registry-client</artifactId>
-            <version>6.2.0</version>
+            <version>7.1.1</version>
```
sfc-gh-japatel, line 361: Latest is https://mvnrepository.com/artifact/io.confluent/kafka-schema-registry-client?repo=confluent-packages 7.2.1

pom.xml
```diff
@@ -379,10 +378,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/io.confluent/kafka-connect-avro-data -->
         <dependency>
             <groupId>io.confluent</groupId>
             <artifactId>kafka-avro-serializer</artifactId>
-            <version>5.5.1</version>
+            <version>7.1.1</version>
```
sfc-gh-japatel, line 385: https://mvnrepository.com/artifact/io.confluent/kafka-avro-serializer?repo=confluent-packages
Lets do 7.2.1

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +643,130 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
       }
     }
   }
+
+  /**
+   * get schema with its subject being [topicName]-value
+   *
+   * @param topicName
+   * @param schemaRegistryURL
+   * @return
```
sfc-gh-japatel, line 652: fill in return too

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +643,130 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
...
+    return new CachedSchemaRegistryClient(
+        avroConverterConfig.getSchemaRegistryUrls(),
+        avroConverterConfig.getMaxSchemasPerSubject(),
+        Collections.singletonList(new AvroSchemaProvider()),
+        srConfig,
+        avroConverterConfig.requestHeaders());
+  }
+
+  @VisibleForTesting
+  public static Map<String, String> getSchemaFromSchemaRegistryClient(
```
sfc-gh-japatel, line 674: nit: Javadocs please add a comment for "value" too

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +643,130 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
...
+  /**
+   * From the connector config extract whether the avro value converter is used
+   *
+   * @param connectorConfig
+   * @return whether the avro value converter is used
+   */
+  public static boolean usesAvroValueConverter(final Map<String, String> connectorConfig) {
+    List<String> validAvroConverter = new ArrayList<>();
+    validAvroConverter.add("io.confluent.connect.avro.AvroConverter");
+    validAvroConverter.add("com.snowflake.kafka.connector.records.SnowflakeAvroConverter");
```
sfc-gh-japatel, line 732: We will not support this. please remove this. 
For Snowpipe streaming we only support confluent AvroConverter.

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +643,130 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
...
+    validAvroConverter.add("io.confluent.encryption.connect.SecuredAvroConverter");
+    // the last one in doubt
```
sfc-gh-japatel, line 734: ?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +643,130 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
...
+   * From the connector config extract whether the avro value converter is used
+   *
+   * @param connectorConfig
+   * @return whether the avro value converter is used
+   */
+  public static boolean usesAvroValueConverter(final Map<String, String> connectorConfig) {
+    List<String> validAvroConverter = new ArrayList<>();
+    validAvroConverter.add("io.confluent.connect.avro.AvroConverter");
+    validAvroConverter.add("com.snowflake.kafka.connector.records.SnowflakeAvroConverter");
+    validAvroConverter.add("io.confluent.encryption.connect.SecuredAvroConverter");
```
sfc-gh-japatel, line 733: do we need this?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -131,6 +132,52 @@ public void createTable(final String tableName) {
...
   }
 
+  public void createTableWithSchema(final String tableName, final Map<String, String> schema) {
+    if (schema.isEmpty()) {
+      createTableWithOnlyMetadataColumn(tableName);
+      return;
+    }
+    checkConnection();
+    InternalUtils.assertNotEmpty("tableName", tableName);
+    String query = "create table if not exists identifier(?)";
```
sfc-gh-japatel, line 142: nit: createTableQuery

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -131,6 +132,52 @@ public void createTable(final String tableName) {
...
+    String logColumn = "[RECORD_METADATA (VARIANT)";
+    for (Map.Entry<String, String> field : schema.entrySet()) {
+      query += ", " + field.getKey() + " " + field.getValue();
+      logColumn += ", " + field.getKey() + " (" + field.getValue() + ")";
+    }
+    query += ")";
+    logColumn += "]";
+
+    try {
+      PreparedStatement stmt = conn.prepareStatement(query);
```
sfc-gh-japatel, line 153: lets do a `logInfo` here too. In case of failures, we would want to know what was the query getting executed.

sfc-gh-japatel: first line of try catch

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +643,130 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
...
+    // the last one in doubt
+    if (connectorConfig.containsKey(SnowflakeSinkConnectorConfig.VALUE_CONVERTER_CONFIG_FIELD)) {
+      String valueConverter =
+          connectorConfig.get(SnowflakeSinkConnectorConfig.VALUE_CONVERTER_CONFIG_FIELD);
+      return validAvroConverter.contains(valueConverter);
+    } else {
+      return false;
+    }
```
sfc-gh-japatel, line 741: Umm, I am not entirely convinced this is true. 
In case the value converter is not used, it is still possible they have avro converter since it can be defined at connector JVM level and not at specific config level. 

So returning false is kind of harmful here. 

In this case (i.e instead of returning false) we can check if we can fetch schema from schema registry. (Which is not possible until we get records)
Or fetch schema registry URL- again can be set at worker level. 

The documentation says it can be defined at worker level and can be overridden by connector. 
https://docs.confluent.io/home/connect/self-managed/userguide.html#configuring-key-and-value-converters

I dont see a way to fetch this properties at worker level. 

So we can either enforce this at connector level or we can do hacky way which is to wait till we see the record. 

My preference is we should ask customers to provide this at connector level too since they will anyways need to give us schema registry URL anyways. 
so on top of schematization, we will need value and key converters with url.

sfc-gh-zefan: If it returns false we will create a table with only the RECORD_METADATA column, and the schema evolution should handle the things since then, which is not bad, right?

I also prefer enforcing it at the connector level. May I add this to the documentation JIRA that the user need to set these fields on the connector config if they want to use this feature?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +643,130 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
...
+  }
+
+  /**
+   * From get the schemaMap for the table from topics. Topics will be collected from topicToTableMap
+   * or connectorConfig
```
sfc-gh-japatel, line 746:
```suggestion
   *<p>Get the schema for the table from topics. 
   *<p>Topics will be collected from topicToTableMap or connectorConfig
```

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +643,130 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
...
+  public static Map<String, String> getSchemaMap(
+      final Map<String, String> topicToTableMap, final Map<String, String> connectorConfig) {
+    Map<String, String> schemaMap = new HashMap<>();
+    if (!topicToTableMap.isEmpty()) {
+      for (String topic : topicToTableMap.keySet()) {
+        Map<String, String> tempMap =
+            Utils.getValueSchemaFromSchemaRegistryURL(
+                topic, connectorConfig.get("value.converter.schema.registry.url"));
+        schemaMap.putAll(tempMap);
+      }
```
sfc-gh-japatel, line 766: I dont understand what will be done here and how this schema for all topics will work? Can you explain? 
Can you also add a test case for this?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +643,130 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
...
+    } else {
+      // if topic is not present in topic2table map, the table name must be the same with the
+      // topic
+      schemaMap =
+          Utils.getValueSchemaFromSchemaRegistryURL(
+              connectorConfig.get("topics"),
+              connectorConfig.get("value.converter.schema.registry.url"));
```
sfc-gh-japatel, line 773: can you extract `value.converter.schema.registry.url` into a public static final string in constants.java file
I see this being used at multiple places.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -131,6 +132,52 @@ public void createTable(final String tableName) {
...
+      stmt.setString(1, tableName);
+      stmt.execute();
+      stmt.close();
+    } catch (SQLException e) {
+      throw SnowflakeErrors.ERROR_2007.getException(e);
+    }
+
+    logInfo("create table {} with only metadata column", tableName);
```
sfc-gh-japatel: Created*

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -131,6 +132,52 @@ public void createTable(final String tableName) {
...
+    try {
+      PreparedStatement stmt = conn.prepareStatement(query);
+      stmt.setString(1, tableName);
+      stmt.execute();
+      stmt.close();
+    } catch (SQLException e) {
+      throw SnowflakeErrors.ERROR_2007.getException(e);
+    }
+
+    logInfo("create table {} with schema {}", tableName, logColumn);
```
sfc-gh-japatel: Created*

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +642,125 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
...
+      schemaMetadata = schemaRegistry.getLatestSchemaMetadata(subjectName);
+    } catch (Exception e) {
+      throw SnowflakeErrors.ERROR_0012.getException();
+    }
+    Map<String, String> schemaMap = new HashMap<>();
+    if (schemaMetadata != null) {
+      AvroSchema schema = new AvroSchema(schemaMetadata.getSchema());
+      for (Schema.Field field : schema.rawSchema().getFields()) {
+        Schema fieldSchema = field.schema();
+        if (!schemaMap.containsKey(field.name())) {
```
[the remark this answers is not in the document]
sfc-gh-zefan: This won't happen actually. Will delete this if

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-japatel: > If it returns false we will create a table with only the RECORD_METADATA column, and the schema evolution should handle the things since then, which is not bad, right?
Sounds good to me! Please mention this in the else case. 

Yes please add this to documentation. Thank you!

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -579,6 +581,89 @@ public void testNativeAvroInputIngestion() throws Exception {
     service.closeAll();
   }
 
+  @Test
+  public void testTableCreationAndNativeAvroInputIngestionWithSchematization() throws Exception {
```
sfc-gh-japatel, line 585: Awesome! Great to see this tests @sfc-gh-zefan

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -579,6 +581,89 @@ public void testNativeAvroInputIngestion() throws Exception {
...
+    Map<String, String> config = TestUtils.getConfForStreaming();
+    config.put(SnowflakeSinkConnectorConfig.ENABLE_SCHEMATIZATION_CONFIG, "true");
+    config.put(
+        SnowflakeSinkConnectorConfig.VALUE_CONVERTER_CONFIG_FIELD,
+        "io.confluent.connect.avro.AvroConverter");
```
sfc-gh-japatel, line 590: you could use the static final field I mentioned earlier in the tests too.

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -579,6 +581,89 @@ public void testNativeAvroInputIngestion() throws Exception {
...
+    config.put("value.converter.schema.registry.url", "http://fake-url");
```
sfc-gh-japatel, line 591: same for this

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -579,6 +581,89 @@ public void testNativeAvroInputIngestion() throws Exception {
...
+            + "            \"name\":\"value_schema\",\n"
+            + "            \"fields\":[\n"
+            + "                {\"name\":\"id\",\"type\":\"int\"},\n"
+            + "                {\"name\":\"first_name\",\"type\":\"string\"},\n"
+            + "                {\"name\":\"rating\",\"type\":\"float\"},\n"
+            + "                {\"name\":\"approval\",\"type\":\"boolean\"},\n"
+            + "               "
+            + " {\"name\":\"info_map\",\"type\":{\"type\":\"map\",\"values\":\"string\"}}\n"
+            + "            ]\n"
+            + "        }";
```
sfc-gh-japatel, line 622: If you could extra this to some field at class level or in constants. 

Also kudos for this schema, I know how boring and convoluted this struct and schema field is.

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -579,6 +581,89 @@ public void testNativeAvroInputIngestion() throws Exception {
...
+    SnowflakeSinkService service =
+        SnowflakeSinkServiceFactory.builder(conn, IngestionMethodConfig.SNOWPIPE_STREAMING, config)
+            .setRecordNumber(1)
+            .setErrorReporter(new InMemoryKafkaRecordErrorReporter())
+            .setSinkTaskContext(new InMemorySinkTaskContext(Collections.singleton(topicPartition)))
+            .addTask(table, new TopicPartition(topic, partition))
+            .build();
+    // Table should be created at the start of the task
+
+    service.insert(avroRecordValue);
```
sfc-gh-japatel, line 659: I would even go one more step and assert the columns returned by desc table/topic_name and make this a complete IT test. 

I think it is very important to also do a select * from the table and assert values from each column with what you expect it to be. 

Please add it in TestUtil.java class

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +642,125 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
       }
     }
   }
+
+  /**
+   * get schema with its subject being [topicName]-value
```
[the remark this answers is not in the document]
sfc-gh-zefan: schema is stored in the schema registry in terms of subject, in our case the subject name could either be "<topicName>-value" or "<topicName>-key". These names are set by the producer (or schema registry client) automatically. 

We don't need to do anything with "<topicName>-key" because the content of the key will only be present in the RECORD_METADATA, so this method will only fetch the schema from "<topicName>-value"

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-zefan: Done

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-tzhang: Not sure if I understand this thread, why can't we try to get it from the worker level if we don't see it in the configuration file?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +645,141 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
...
+   */
+  public static Map<String, String> getSchemaMapForTable(
+      final String tableName,
+      final Map<String, String> topicToTableMap,
+      final String schemaRegistryURL) {
+    return getSchemaMapForTableWithSchemaRegistryClient(
+        tableName, topicToTableMap, getAvroSchemaRegistryClientFromURL(schemaRegistryURL));
+  }
+
+  public static Map<String, String> getSchemaMapForTableWithSchemaRegistryClient(
```
sfc-gh-tzhang, line 765: comment please

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
```diff
@@ -244,4 +245,24 @@ public interface SnowflakeConnectionService {
...
+   */
+  void createTableWithSchema(String tableName, Map<String, String> schemaMap);
+
+  /**
+   * Create a table with only the RECORD_METADATA column. The rest of the columns might be added
+   * through schema evolution
+   *
+   * @param tableName table name
+   */
+  void createTableWithOnlyMetadataColumn(String tableName);
```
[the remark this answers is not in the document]
sfc-gh-tzhang: this looks unresolved, are you planning to take it out of this PR?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -509,7 +509,22 @@ private void createTableIfNotExists(final String tableName) {
...
+          this.conn.createTableWithSchema(
+              tableName,
+              Utils.getSchemaMapForTable(
+                  tableName,
+                  this.topicToTableMap,
+                  this.connectorConfig.get(
+                      SnowflakeSinkConnectorConfig.VALUE_SCHEMA_REGISTRY_CONFIG_FIELD)));
+        } else {
+          throw SnowflakeErrors.ERROR_5021.getException();
+        }
```
sfc-gh-tzhang, line 524: It looks weird that we're throwing in else branch, should we throw it inside `usesAvroValueConverter`?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +645,141 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
...
+    } catch (IOException | RestClientException e) {
+      throw SnowflakeErrors.ERROR_0012.getException(e);
+    }
+    Map<String, String> schemaMap = new HashMap<>();
+    if (schemaMetadata != null) {
+      AvroSchema schema = new AvroSchema(schemaMetadata.getSchema());
+      for (Schema.Field field : schema.rawSchema().getFields()) {
+        Schema fieldSchema = field.schema();
+        String columnName = field.name().toUpperCase();
+        // avro does not support double quotes so the columnName will be in uppercase anyway
```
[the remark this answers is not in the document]
sfc-gh-zefan: we should error out

sfc-gh-tzhang (approved):
LGTM, let's go!

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +645,156 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
...
+    return new CachedSchemaRegistryClient(
+        avroConverterConfig.getSchemaRegistryUrls(),
+        avroConverterConfig.getMaxSchemasPerSubject(),
+        Collections.singletonList(new AvroSchemaProvider()),
+        srConfig,
+        avroConverterConfig.requestHeaders());
+  }
+
+  /**
+   * get schema with its subject being [topicName]-[type]
```
sfc-gh-japatel, line 663:
```suggestion
   * Get schema with its subject being [topicName]-[type]
```

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-japatel: bump..

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +645,156 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
       }
     }
   }
+
+  private static SchemaRegistryClient getAvroSchemaRegistryClientFromURL(
+      final String schemaRegistryURL) {
+    Map<String, String> srConfig = new HashMap<>();
+    srConfig.put("schema.registry.url", schemaRegistryURL);
```
sfc-gh-japatel, line 652: you can use SCHEMA_REGISTRY_URL_CONFIG from AbstractKafkaSchemaSerDeConfig.java

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +642,125 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
...
+        new CachedSchemaRegistryClient(
+            avroConverterConfig.getSchemaRegistryUrls(),
+            avroConverterConfig.getMaxSchemasPerSubject(),
+            Collections.singletonList(new AvroSchemaProvider()),
+            srConfig,
+            avroConverterConfig.requestHeaders());
+    String subjectName = topicName + "-" + type;
+    SchemaMetadata schemaMetadata;
+    try {
+      schemaMetadata = schemaRegistry.getLatestSchemaMetadata(subjectName);
```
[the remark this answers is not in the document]
sfc-gh-japatel: Can you check if it is possible to use TopicNameStrategy class?

src/test/java/com/snowflake/kafka/connector/UtilsTest.java
```diff
@@ -133,4 +136,23 @@ public void testSetJDBCLoggingDir() {
...
+  @Test
+  public void testCollectSchemaFromTopics() throws Exception {
+    SchemaRegistryClient schemaRegistry = new MockSchemaRegistryClient();
+    schemaRegistry.register(
+        "topic0-value", new AvroSchema(TestUtils.AVRO_SCHEMA_FOR_SCHEMA_COLLECTION_0));
+    schemaRegistry.register(
+        "topic1-value", new AvroSchema(TestUtils.AVRO_SCHEMA_FOR_SCHEMA_COLLECTION_1));
+    Map<String, String> topicToTableMap = new HashMap<>();
+    topicToTableMap.put("topic0", "table");
+    topicToTableMap.put("topic1", "table");
```
sfc-gh-japatel, line 149: if topic0 table1 and topic1 and table2 are mapped, we will not create the schema map right?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +645,156 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
       }
     }
   }
+
+  private static SchemaRegistryClient getAvroSchemaRegistryClientFromURL(
```
sfc-gh-japatel, line 650: Consider moving all of these functions to a new Util file called SchemaUtils

src/test/java/com/snowflake/kafka/connector/UtilsTest.java
sfc-gh-zefan: if we query for table1 or table2 we will still get the schema map

src/test/java/com/snowflake/kafka/connector/internal/TestUtils.java
```diff
@@ -129,6 +129,64 @@ public class TestUtils {
           + "}";
   public static final String JSON_WITHOUT_SCHEMA = "{\"userid\": \"User_1\"}";
 
+  public static final String AVRO_SCHEMA_FOR_TABLE_CREATION =
```
sfc-gh-japatel, line 132: All of these can go into SchemaTestUtils

src/test/java/com/snowflake/kafka/connector/internal/TestUtils.java
```diff
@@ -664,4 +722,56 @@ public static long getOffsetTokenForChannelAndTable(String tableName, final Stri
...
+
+  /**
+   * Check if one row retrieved from the table matches the provided content
+   *
+   * <p>The assumption is that the rows in the table are the same.
+   *
+   * @param tableName the name of the table
+   * @param contentMap the provided content map from columnName to their value
+   */
+  public static void checkTableContentOneRow(String tableName, Map<String, Object> contentMap)
```
sfc-gh-japatel, line 756: Thanks for this function!

test/rest_request_template/travis_correct_auto_table_creation.json
```diff
@@ -0,0 +1,29 @@
...
+  "name": "SNOWFLAKE_CONNECTOR_NAME",
+  "config": {
+    "connector.class": "com.snowflake.kafka.connector.SnowflakeSinkConnector",
+    "snowflake.role.name": "TESTROLE_KAFKA",
+    "topics": "SNOWFLAKE_TEST_TOPIC",
+    "tasks.max": "1",
+    "buffer.flush.time": "10",
+    "buffer.count.records": "100",
```
sfc-gh-japatel, line 9: Can you also add one more config for topic to table map? Should be fairly easy.

sfc-gh-zefan: You mean add another test with the topic to table map config set?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -631,4 +645,156 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
...
+  public static boolean usesAvroValueConverter(final Map<String, String> connectorConfig) {
+    List<String> validAvroConverter = new ArrayList<>();
+    validAvroConverter.add(SnowflakeSinkConnectorConfig.CONFLUENT_AVRO_CONVERTER);
+    if (connectorConfig.containsKey(SnowflakeSinkConnectorConfig.VALUE_CONVERTER_CONFIG_FIELD)) {
+      String valueConverter =
+          connectorConfig.get(SnowflakeSinkConnectorConfig.VALUE_CONVERTER_CONFIG_FIELD);
+      return validAvroConverter.contains(valueConverter);
+    }
+    throw SnowflakeErrors.ERROR_5021.getException();
+  }
```
sfc-gh-japatel, line 749: consider adding a UT for the else case

test/rest_request_template/travis_correct_auto_table_creation.json
sfc-gh-japatel: Yes, correct

sfc-gh-japatel:
There is a code coverage report both possible to look at through Intellij and through https://github.com/snowflakedb/snowflake-kafka-connector/pull/470#issuecomment-1207565562 can you try and add few more UT cases for missed lines? We want to be more than 90% for newly added code.

sfc-gh-japatel:
Please also link the documentation JIRA in the description.

sfc-gh-japatel (approved):
overall lgtm! Very minor suggestions, hoping you could add few more tests to improve the code coverage.

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
```diff
@@ -0,0 +1,176 @@
...
+import io.confluent.kafka.serializers.AbstractKafkaSchemaSerDeConfig;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.apache.avro.Schema;
+
+public class SchematizationUtils {
```
sfc-gh-japatel, line 21: If you get time, please add Javadocs to the public class.

test/rest_request_template/travis_correct_auto_table_creation.json
```diff
@@ -2,7 +2,6 @@
   "name": "SNOWFLAKE_CONNECTOR_NAME",
   "config": {
     "connector.class": "com.snowflake.kafka.connector.SnowflakeSinkConnector",
-    "snowflake.role.name": "TESTROLE_KAFKA",
```
sfc-gh-japatel: you dont need this?

sfc-gh-japatel: Similarly for other tests?

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
sfc-gh-japatel: Thanks for refactoring

test/rest_request_template/travis_correct_auto_table_creation_topic2table.json
```diff
@@ -0,0 +1,29 @@
+{
```
sfc-gh-japatel, line 1: Looks good! great stuff

test/rest_request_template/travis_correct_auto_table_creation.json
sfc-gh-zefan: There's a field with the same name below. This was mistakenly introduced by me at some time.

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
sfc-gh-zefan: I would add the Javadocs in the schema evolution PR

PR #472, opened 2022-08-10
SNOW-630885 Schema Evolution

sfc-gh-zefan:
Schema evolution through alter table request sent from Kafka connector.

Add logic to alter table when we see "extra column" or "missing nonnullable value" error in the response from the Ingest SDK. We will the reopen the channel and retry insertion until the maximum number of attempts is reached.

Add tests related to schema evolution.
Add tests related to the new behavior of automatic table creation with schema evolution enabled.


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -154,6 +154,12 @@ public class SnowflakeSinkConnectorConfig {
 
   public static final String ENABLE_SCHEMATIZATION_CONFIG = "enable.schematization";
 
+  // the amount of waiting before next insertion of buffered records is retried for schema evolution
```
sfc-gh-tzhang, line 157: please add TODO along with JIRA number so that we won't forget to remove it

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
```diff
@@ -90,6 +91,10 @@ public interface SnowflakeConnectionService {
    */
   boolean isTableCompatible(String tableName);
 
+  boolean hasSchemaEvolutionPermission(String tableName, String role);
```
sfc-gh-tzhang, line 94: coment please

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -330,6 +380,25 @@ public boolean isTableCompatible(final String tableName) {
...
+      stmt.setString(1, tableName);
+      result = stmt.executeQuery();
+      while (result.next()) {
+        System.out.println(result.getString(1) + " (" + result.getString(2) + ")");
+      }
+    } catch (SQLException e) {
+      logDebug("table {} doesn't exist", tableName);
+    }
+  }
+
```
sfc-gh-tzhang, line 401: I think this is only for testing?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -372,6 +441,99 @@ public void appendMetaColIfNotExist(final String tableName) {
...
+  @Override
+  public boolean hasSchemaEvolutionPermission(String tableName, String role) {
+    checkConnection();
+    InternalUtils.assertNotEmpty("tableName", tableName);
+    String query = "show grants on table identifier(?)";
+    ResultSet result = null;
+    boolean hasRolePrivilege = false;
+    boolean hasTableOptionEnabled = true;
+    String myRole = formatRoleName(role);
+    // the schema evolution permission is still in PrPr, so it is left as true here
```
sfc-gh-tzhang, line 461: any idea about the timeline? And could you create a JIRA dependency and assign to the correct person?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -372,6 +441,99 @@ public void appendMetaColIfNotExist(final String tableName) {
...
+            || result.getString(2).equals("ALL")
+            || result.getString(2).equals("OWNERSHIP")) {
+          hasRolePrivilege = true;
+        }
+      }
+      stmt.close();
+    } catch (SQLException e) {
+      hasRolePrivilege = true;
+      // if the table doesn't exist, we will create it, and thus we will have the privilege
+      // throw SnowflakeErrors.ERROR_2014.getException(e);
```
sfc-gh-tzhang, line 481: uncomment this?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -372,6 +441,99 @@ public void appendMetaColIfNotExist(final String tableName) {
...
+  }
+
+  /**
+   * Transform the columnName to uppercase unless it is enclosed in double quotes
+   *
+   * <p>In that case, drop the quotes and leave it as it is.
+   *
+   * @param roleName
+   * @return Transformed roleName
+   */
```
sfc-gh-tzhang, line 493: the comment doesn't look correct

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -269,7 +315,11 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
       // If we found reaching buffer size threshold or count based threshold, we will immediately
       // flush (Insert them)
       if (copiedStreamingBuffer != null) {
-        insertBufferedRecords(copiedStreamingBuffer);
+        if (this.enableSchematization && this.enableSchemaEvolution) {
+          insertBufferedRecordsWithRetry(copiedStreamingBuffer);
```
sfc-gh-tzhang, line 319: do you think these two functions could be combined? If the only difference is the number of reties, we could pass a retry count of 0 to `insertBufferedRecords` and combine these two?

sfc-gh-tzhang (commented):
Thanks for the change! Did a really quick walkthrough and left some comments, will take a look again after you merge and remove the auto table creation related change

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-zefan: There are a few differences that come to my mind:
1. The `insertBufferedRecords` will add rows with any types of error to DLQ while `insertBufferedRecordsWithRetry` will add rows that cannot or failed to be reinserted to DLQ.
2. The `insertBufferedRecords` will only reopen the channel when an SFException is thrown while the `insertBufferedRecordsWithRetry` will reopen the channel with each retry, which also leads to a difference that the the channel offset needs to be reset for the latter.

I see that there are now many functions look like twins, but I think combining them will make the code quite messy.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -402,6 +429,145 @@ public void appendMetaColIfNotExist(final String tableName) {
...
+
+  /**
+   * Check if the role used by the connector has the permission to do schema evolution on the table
+   * through a cached map
+   *
+   * @param tableName the name of the table
+   * @return whether we have the permission to do schema evolution on the table
+   */
+  @Override
+  public boolean getSchemaEvolutionPermission(String tableName) {
```
sfc-gh-tzhang, line 485:
```suggestion
  public boolean hasSchemaEvolutionPermission(String tableName) {
```

sfc-gh-tzhang: And is this cached? What if the permission is updated?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -402,6 +429,145 @@ public void appendMetaColIfNotExist(final String tableName) {
...
+   *
+   * @param tableName the name of the table
+   * @param columnToType the mapping from the columnNames to their types
+   */
+  @Override
+  public void appendColumns(String tableName, Map<String, String> columnToType) {
+    checkConnection();
+    InternalUtils.assertNotEmpty("tableName", tableName);
+    StringBuilder query = new StringBuilder("alter table identifier(?) add column ");
+    boolean first = true;
```
sfc-gh-tzhang, line 500: please add comment or improve the naming

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -402,6 +429,145 @@ public void appendMetaColIfNotExist(final String tableName) {
...
+        first = false;
+      } else {
+        query.append(", add column ");
+        logColumn.append(",");
+      }
+      query
+          .append(columnName)
+          .append(" ")
+          .append(columnToType.get(columnName))
+          .append(" comment 'column created by schema evolution'");
```
sfc-gh-tzhang, line 513: let's include kc somewhere in the comment

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -402,6 +429,145 @@ public void appendMetaColIfNotExist(final String tableName) {
...
+        first = false;
+      } else {
+        query.append(", ");
+        logColumn.append(", ");
+      }
+      query
+          .append(columnName)
+          .append(" drop not null, ")
+          .append(columnName)
+          .append(" comment 'column altered to be nullable by schema evolution'");
```
sfc-gh-tzhang, line 553: let's include kc somewhere in the comment

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeErrors.java
```diff
@@ -166,6 +166,8 @@ public enum SnowflakeErrors {
       "Failed to append RECORD_METADATA column, please check that you have permission to do so."),
   ERROR_2014(
       "2014", "Table not exists", "Table not exists. It might have been deleted externally."),
+  ERROR_2015("2015", "Failed to append columns", "Failed to append columns"),
+  ERROR_2016("2016", "Failed to drop nullability", "Failed to drop nullability"),
```
sfc-gh-tzhang, line 170: drop NOT NULL

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -161,6 +168,26 @@ public void createTableWithSchema(final String tableName, final Map<String, Stri
     logInfo("Created table {} with schema {}", tableName, logColumn.toString());
   }
 
+  public void createTableWithOnlyMetadataColumn(final String tableName) {
+    checkConnection();
+    if (!getSchemaEvolutionPermission(tableName)) {
+      throw SnowflakeErrors.ERROR_5021.getException();
+    }
+    InternalUtils.assertNotEmpty("tableName", tableName);
+    String createTableQuery = "create table if not exists identifier(?) (record_metadata variant)";
```
sfc-gh-tzhang, line 177: could we add a comment for CREATE TABLE as well?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -170,6 +176,34 @@ public class TopicPartitionChannel {
...
   private final BufferThreshold streamingBufferThreshold;
 
+  // Whether schematization has been enabled.
+  private boolean enableSchematization;
+
+  private boolean enableSchemaEvolution = false;
+
+  private SnowflakeConnectionService conn;
+
+  public TopicPartitionChannel(
```
sfc-gh-tzhang, line 186: comment please

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -189,7 +223,8 @@ public TopicPartitionChannel(
       final BufferThreshold streamingBufferThreshold,
       final Map<String, String> sfConnectorConfig,
       KafkaRecordErrorReporter kafkaRecordErrorReporter,
-      SinkTaskContext sinkTaskContext) {
+      SinkTaskContext sinkTaskContext,
+      SnowflakeConnectionService conn) {
```
sfc-gh-tzhang, line 227: update comment please

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -200,10 +235,16 @@ public TopicPartitionChannel(
     this.channel = Preconditions.checkNotNull(openChannelForTable());
     this.kafkaRecordErrorReporter = Preconditions.checkNotNull(kafkaRecordErrorReporter);
     this.sinkTaskContext = Preconditions.checkNotNull(sinkTaskContext);
+    this.conn = conn;
+    if (conn != null) {
```
sfc-gh-tzhang, line 239: when the conn could be null?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -246,6 +287,7 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
       bufferLock.lock();
       try {
         this.streamingBuffer.insert(kafkaSinkRecord);
+        LOGGER.debug("Record with offset [{}] inserted into buffer", kafkaSinkRecord.kafkaOffset());
```
sfc-gh-tzhang, line 290: do we need to say which partition is this?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -270,7 +312,13 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
...
       // flush (Insert them)
       if (copiedStreamingBuffer != null) {
-        insertBufferedRecords(copiedStreamingBuffer);
+        if (this.enableSchematization && this.enableSchemaEvolution) {
+          insertBufferedRecordsWithRetry(
+              copiedStreamingBuffer,
+              SnowflakeSinkConnectorConfig.RETRY_INSERTION_ATTEMPT_MAX_COUNT_DEFAULT);
+        } else {
+          insertBufferedRecords(copiedStreamingBuffer);
+        }
```
sfc-gh-tzhang, line 321: why we can't compute the retry count first and only call `insertBufferedRecordsWithRetry`?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -451,30 +505,59 @@ protected void insertBufferedRecordsIfFlushTimeThresholdReached() {
...
+              + " insertResponseHasErrors:{}",
           this.getChannelName(),
           streamingBufferToInsert,
           response.hasErrors());
       if (response.hasErrors()) {
-        handleInsertRowsFailures(
-            response.getInsertErrors(), streamingBufferToInsert.getSinkRecords());
+        handleInsertRowsWithRetryFailures(response.insertErrors, response.recordWithError);
+      }
+      // with the attempts the channel was reopened (several times), and the channel offset was
```
sfc-gh-tzhang, line 549: could you clarify why the channel needs to be reopened multiple times?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -451,30 +505,59 @@ protected void insertBufferedRecordsIfFlushTimeThresholdReached() {
...
+      // reset to be before the current buffer. The offset needs to be reset to be the end of the
+      // buffer to continue ingestion properly
+      // even if the retry does not succeed the reset is necessary to make the behavior consistent
+      // with that without schema evolution
+      if (response.getChannelReopened()) {
+        resetChannelMetadataAfterRecovery(
+            StreamingApiFallbackInvoker.INSERT_ROWS_FALLBACK, lastKafkaOffsetInFailedBuffer);
+        // TODO: last record offset equivalent to snowflake offset?
+        // TODO: How multiple tasks work?
```
sfc-gh-tzhang, line 558: these needs to be investigated before merging?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -486,23 +569,148 @@ InsertValidationResponse insertBufferedRecords(StreamingBuffer streamingBufferTo
...
-   *
-   * <p>It can also send errors {@link
-   * net.snowflake.ingest.streaming.InsertValidationResponse.InsertError} in form of response inside
-   * {@link InsertValidationResponse}
-   *
-   * @param buffer buffer to insert into snowflake
-   * @return InsertValidationResponse object sent from insertRows API.
-   */
-  private InsertValidationResponse insertRowsWithFallback(StreamingBuffer buffer) {
+  public class InsertRowsWithRetryResponse {
```
sfc-gh-tzhang, line 572: could you add comment on these functions?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -486,23 +569,148 @@ InsertValidationResponse insertBufferedRecords(StreamingBuffer streamingBufferTo
...
+      Map<String, String> schemaMap = new HashMap<>();
+      if (sfConnectorConfig
+          .get(SnowflakeSinkConnectorConfig.VALUE_CONVERTER_CONFIG_FIELD)
+          .equals(SnowflakeSinkConnectorConfig.CONFLUENT_AVRO_CONVERTER)) {
+        schemaMap =
+            SchematizationUtils.getSchemaMapForTopic(
+                topicPartition.topic(),
+                sfConnectorConfig.get(
+                    SnowflakeSinkConnectorConfig.VALUE_SCHEMA_REGISTRY_CONFIG_FIELD));
+      }
```
sfc-gh-tzhang, line 634: do we need to get this everytime we insert to failed buffer?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -486,23 +569,148 @@ InsertValidationResponse insertBufferedRecords(StreamingBuffer streamingBufferTo
...
+     * From the response collect rows with error and append those with unretryable errors to
+     * recordWithError. Insert the rest into a StreamingBuffer waiting for retry insertion, and
+     * collect the column to be added at the same time. After that the table is altered.
+     *
+     * @param conn
+     * @param tableName
+     * @param response
+     * @return
+     */
+    public StreamingBuffer CollectRowsWithErrorsAndAlterTable(
```
sfc-gh-tzhang, line 680:
```suggestion
    public StreamingBuffer collectRowsWithErrorsAndAlterTable(
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -600,12 +867,11 @@ private void handleInsertRowsFailures(
...
             ERRORS_DEAD_LETTER_QUEUE_TOPIC_NAME_CONFIG);
       } else {
+        int index = 0;
         for (InsertValidationResponse.InsertError insertError : insertErrors) {
           // Map error row number to index in sinkRecords list.
-          int rowIndexToOriginalSinkRecord = (int) insertError.getRowIndex();
           this.kafkaRecordErrorReporter.reportError(
-              insertedRecordsToBuffer.get(rowIndexToOriginalSinkRecord),
-              insertError.getException());
+              insertedRecordsToBuffer.get(index++), insertError.getException());
```
sfc-gh-tzhang, line 874: i'm not sure if this is correct, the rowIndex may not be +1 every time?

test/rest_request_template/travis_correct_schema_evolution_avro_sr.json
```diff
@@ -0,0 +1,29 @@
+{
```
sfc-gh-tzhang, line 1: these are some really good tests, thanks!

sfc-gh-tzhang (changes requested):
This is a good start, left some comments, thanks!

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
```diff
@@ -17,8 +17,23 @@
...
 import org.apache.avro.Schema;
+import org.apache.kafka.connect.data.ConnectSchema;
 
+/** This is a class containing the helper functions related to schematization */
 public class SchematizationUtils {
+  static final String EXTRA_COLUMNS_PREFIX = "Extra columns: ";
+
+  static final String DEPRECATED_EXTRA_COLUMNS_PREFIX = "Extra column: ";
+  static final String EXTRA_COLUMNS_SUFFIX =
+      ". Columns not present in the table shouldn't be specified.";
```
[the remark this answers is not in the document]
sfc-gh-zefan: I was worrying about a dot could be inside the columnName...

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
```diff
@@ -173,4 +202,222 @@ public static Map<String, String> getSchemaMapForTableWithSchemaRegistryClient(
...
+      //  The recordMap arg should also be removed since it's only used here
+      int startIndex =
+          message.indexOf(DEPRECATED_NONNULLABLE_COLUMNS_PREFIX)
+              + DEPRECATED_NONNULLABLE_COLUMNS_PREFIX.length();
+      int endIndex = message.indexOf(NONNULLABLE_COLUMNS_SUFFIX);
+      String columnName = message.substring(startIndex, endIndex);
+      // need to find the columnName before formatColumnName
+      // this is wrong when the columnName was enclosed in double quotes
+      nonNullableColumns.add(columnName);
+    }
```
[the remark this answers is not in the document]
sfc-gh-zefan: Not really? Because the old behavior of the SDK is that it will return the formatted columnName

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
```diff
@@ -173,4 +202,222 @@ public static Map<String, String> getSchemaMapForTableWithSchemaRegistryClient(
...
+
+  private static String getTypeFromJsonObject(Object value) {
+    if (value == null) {
+      return "VARIANT";
+    }
+    org.apache.kafka.connect.data.Schema.Type schemaType =
+        ConnectSchema.schemaType(value.getClass());
+    if (schemaType == null) {
+      return "VARIANT";
+    }
```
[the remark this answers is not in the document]
sfc-gh-zefan: When it is not a type java recognizes it will return null. I believe this could only happen when some data corruption occurs.

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
```diff
@@ -173,4 +202,222 @@ public static Map<String, String> getSchemaMapForTableWithSchemaRegistryClient(
...
+   */
+  public static String formatColumnName(String columnName) {
+    // the columnName has been checked and guaranteed not to be null or empty
+    return (columnName.charAt(0) == '"' && columnName.charAt(columnName.length() - 1) == '"')
+        ? columnName.substring(1, columnName.length() - 1)
+        : columnName.toUpperCase();
+  }
+
+  /**
+   * Transform the roleName to uppercase unless it is enclosed in double quotes
```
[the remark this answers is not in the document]
sfc-gh-zefan: Yes, a lower case roleName doesn't work. What should I do when two functions share the same logic? I find calling one function inside another quite messy.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-zefan: I guess the costumer will have to restart the connector. It makes no sense to query the permission every time we failed to send the records.

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
```diff
@@ -173,4 +202,222 @@ public static Map<String, String> getSchemaMapForTableWithSchemaRegistryClient(
...
+  /**
+   * extra a list of columnNames from their string representation
+   *
+   * <p>input: "["a", B]", output: [""a"", "B"]
+   *
+   * @param message part of the error message that contains a list of columnNames
+   * @return the list of columnNames
+   */
+  @VisibleForTesting
+  public static List<String> getColumnNamesFromMessage(String message) {
```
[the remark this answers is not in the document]
sfc-gh-zefan: Makes sense. That will be safer and easier for processing.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-zefan: When the topicPartitionChannel is used in tests

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-zefan: The error and the buffer have been reorganized so that they have a matching index, to wit, record[i] has error[i]

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
```diff
@@ -55,7 +70,8 @@ public static Map<String, String> getAvroSchemaFromSchemaRegistryClient(
     try {
       schemaMetadata = schemaRegistry.getLatestSchemaMetadata(subjectName);
     } catch (IOException | RestClientException e) {
-      throw SnowflakeErrors.ERROR_0012.getException(e);
+      // suppress the excpetion and return empty map to indicate a failure in fetching schemas
```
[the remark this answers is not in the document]
sfc-gh-zefan: But we don't want to deal with the exception, we just want to know that the schema was not retrieved, so maybe it's ok if we just catch the exception here

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-zefan: You mean to add the explanation to the comment?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-zefan: Actually it's only used for debugging. I will delete this log line.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-zefan: Ideally this won't happen many times, unless all the records have errors that cannot be resolved and need to be inserted to DLQ (otherwise we will error out pretty fast).

We could also cache the schema and call this function when we see new schema ID.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
```diff
@@ -91,6 +91,41 @@ public interface SnowflakeConnectionService {
    */
   boolean isTableCompatible(String tableName);
 
+  /**
+   * Check whether the user has the role privilge to do schema evolution and whether the schema
```
sfc-gh-japatel, line 95: nit: *privilege
For my education, which role gives schema evolution?

sfc-gh-japatel, line 95: nit: *privilege
For my education, which role gives schema evolution?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
```diff
@@ -91,6 +91,41 @@ public interface SnowflakeConnectionService {
...
+   * evolution option is enabled on the table
```
sfc-gh-japatel, line 96: Is this available on mixed tables?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
```diff
@@ -91,6 +91,41 @@ public interface SnowflakeConnectionService {
...
+   *
+   * @param tableName the name of the table
+   * @param role the role of the user
+   * @return whether schema evolution has the required permission to be performed
```
sfc-gh-japatel, line 100: nit: `whether table has the required permission to be perform schema evolution`

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
```diff
@@ -91,6 +91,41 @@ public interface SnowflakeConnectionService {
...
+   */
+  boolean hasSchemaEvolutionPermission(String tableName);
+
+  /**
+   * Alter table to add columns according to a map from columnNames to their types
+   *
+   * @param tableName the name of the table
+   * @param columnToType the mapping from the columnNames to their types
+   */
+  void appendColumns(String tableName, Map<String, String> columnToType);
```
sfc-gh-japatel, line 119: AppendColumn is a very generalized term, Can we be more specific? May be `appendColumnsToTable`?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -402,6 +438,147 @@ public void appendMetaColIfNotExist(final String tableName) {
...
+      stmt.setString(1, tableName);
+      stmt.execute();
+      stmt.close();
+    } catch (SQLException e) {
+      throw SnowflakeErrors.ERROR_2016.getException(e);
+    }
+
+    logColumn
+        .insert(0, "Following columns' non-nullabilty was dropped for table {}:\n")
+        .append("]");
```
sfc-gh-japatel, line 578: ?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -402,6 +438,147 @@ public void appendMetaColIfNotExist(final String tableName) {
...
+   *
+   * @param tableName the name of the table
+   * @param columnNames the list of columnNames
+   */
+  @Override
+  public void alterNonNullableColumns(String tableName, List<String> columnNames) {
+    checkConnection();
+    InternalUtils.assertNotEmpty("tableName", tableName);
+    StringBuilder query = new StringBuilder("alter table identifier(?) alter ");
+    boolean first = true;
```
sfc-gh-japatel, line 549: nit: `isFirstColumn`

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -152,6 +153,9 @@ public SnowflakeSinkServiceV2(
    */
   @Override
   public void startTask(String tableName, TopicPartition topicPartition) {
+    // check if we have the schema evolution permission and register it
+    this.conn.hasSchemaEvolutionPermissionForRole(
+        tableName, this.connectorConfig.get(SNOWFLAKE_ROLE));
```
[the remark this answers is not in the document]
sfc-gh-japatel: What will happen to the clients using this new version without the appropriate permissions? 
We shouldn't fail the task for them.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -402,6 +438,147 @@ public void appendMetaColIfNotExist(final String tableName) {
...
+   * @return whether schema evolution has the required permission to be performed
+   */
+  @Override
+  public boolean hasSchemaEvolutionPermissionForRole(String tableName, String role) {
+    checkConnection();
+    InternalUtils.assertNotEmpty("tableName", tableName);
+    String query = "show grants on table identifier(?)";
+    ResultSet result = null;
+    boolean hasRolePrivilege = false;
+    boolean hasTableOptionEnabled = true;
```
sfc-gh-japatel, line 458: can you describe what this field mean?

test/test_verify.py
```diff
@@ -464,7 +480,10 @@ def runTestSet(driver, testSet, nameSalt, pressure):
         testSnowpipeStreamingStringJson, testSnowpipeStreamingStringAvro,
         testMultipleTopicToOneTableSnowpipeStreaming, testMultipleTopicToOneTableSnowpipe,
         testSchemaMapping,
-        testAutoTableCreation, testAutoTableCreationTopic2Table
+        testAutoTableCreation, testAutoTableCreationTopic2Table,
+        testSchemaEvolutionJson, testSchemaEvolutionAvroSR,
+        testSchemaEvolutionWithAutoTableCreationJson, testSchemaEvolutionWithAutoTableCreationAvroSR,
+        testSchemaEvolutionNonNullableJson
```
sfc-gh-japatel, line 486: this is not scalable in future. 
Can you may be create a jira to refactor this into a map of test name and whether it is enabled in confluent or kafka? (Someone will pick it up)

src/test/java/com/snowflake/kafka/connector/UtilsTest.java
```diff
@@ -173,4 +175,88 @@ public void testValidAvroValueConverter() {
         "com.snowflake.kafka.connector.records.SnowflakeAvroConverter");
     assert !SchematizationUtils.usesAvroValueConverter(config);
   }
+
+  @Test
+  public void testColumnNamesParsingFromMessage() {
```
sfc-gh-japatel, line 180: move this tests to `SchematizationUtilsTest`

sfc-gh-japatel:
Please add description in PR what the change is supposed to do. This PR is huge and some background would go long way.  
Also, best practice is to split this PR into smaller ones so that you can get feedback. (It is fine since this is time bounded)

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
sfc-gh-zefan: I'm not sure if I understand the question.

The role should have one of "EVOLVE SCHEMA", "ALL" or "OWNERSHIP" privilege to do schema evolution.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
sfc-gh-zefan: This is new for me. 

How can I create a mixed table to have some experiments?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-zefan: I don't know what is wrong here...

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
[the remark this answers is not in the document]
sfc-gh-zefan: We could add a new config specifically for schema evolution, but I think it will be more like an agreement, letting the costumer be aware of the option.

If they don't have the appropriate permission they will still be able to do schema mapping and automatic table creation.

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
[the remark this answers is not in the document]
sfc-gh-tzhang: The change has been merged so you should be able to remove all these and use the lasted SDK

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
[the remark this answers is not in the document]
sfc-gh-zefan: Will do.

sfc-gh-zefan:
I made some change related to the updates in the SDK. Since it won't pass the merge gate, I just put it in a draft PR. https://github.com/snowflakedb/snowflake-kafka-connector/pull/479

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -486,23 +559,154 @@ InsertValidationResponse insertBufferedRecords(StreamingBuffer streamingBufferTo
...
+        InsertValidationResponse response,
+        List<SinkRecord> records) {
+      StreamingBuffer newBufferToInsert =
+          insertRecordsToFailedBuffer(response.getInsertErrors(), records);
+      if (!this.extraColumnToType.isEmpty()) {
+        try {
+          conn.appendColumnsToTable(tableName, this.extraColumnToType);
+        } catch (SnowflakeKafkaConnectorException e) {
+          LOGGER.warn(
+              String.format("[INSERT_BUFFERED_RECORDS] Failure altering table :%s", tableName), e);
```
sfc-gh-tzhang, line 688: what will we do after we log it? Will we retry?

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -246,7 +246,7 @@ private Map<String, Object> getMapFromJsonNodeForStreamingIngest(JsonNode node)
     while (columnNames.hasNext()) {
       String columnName = columnNames.next();
       JsonNode columnNode = node.get(columnName);
-      columnName = formatColumnName(columnName);
+      // column should be passed without processing
```
sfc-gh-tzhang, line 249: nit: please remove, people who read the code won't know the history that column needs processing :)

sfc-gh-tzhang (approved):
Overall LGTM if you could 
- Address the existing comments. 
- I see there're a bunch of TODO without JIRA number, are the investigations done?
- Please make sure @sfc-gh-japatel has looked at the `insertBufferedRecordsWithRetry` and `InsertRowsWithRetryResponse` since I don't think I fully understand the logic :) 

Thanks, almost there!

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
```diff
@@ -173,4 +196,84 @@ public static Map<String, String> getSchemaMapForTableWithSchemaRegistryClient(
...
+        extraColumnToType.put(columnName, type);
+      }
+    }
+    return extraColumnToType;
+  }
+
+  private static String getTypeFromJsonObject(Object value) {
+    if (value == null) {
+      return "VARIANT";
+    }
```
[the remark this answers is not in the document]
sfc-gh-zefan: When the field is there but the value is just null. This is rare but might happen, I suppose.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-zefan: Since this function is called in the `onRetry` clause, yes we will retry.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -402,6 +438,153 @@ public void appendMetaColIfNotExist(final String tableName) {
...
+  }
+
+  /**
+   * Alter table to add columns according to a map from columnNames to their types
+   *
+   * @param tableName the name of the table
+   * @param columnToType the mapping from the columnNames to their types
+   */
+  @Override
+  public void appendColumnsToTable(String tableName, Map<String, String> columnToType) {
```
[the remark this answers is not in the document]
sfc-gh-zefan: Valid point. Just added.

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
```diff
@@ -55,7 +63,9 @@ public static Map<String, String> getAvroSchemaFromSchemaRegistryClient(
     try {
       schemaMetadata = schemaRegistry.getLatestSchemaMetadata(subjectName);
     } catch (IOException | RestClientException e) {
-      throw SnowflakeErrors.ERROR_0012.getException(e);
+      // suppress the excpetion and return empty map to indicate a failure in fetching schemas
+      LOGGER.info("Schema with subject '{}' not found in schema registry", subjectName);
+      return new HashMap<>();
```
sfc-gh-japatel, line 68: return Collections.emptyMap() to indicate immutability
https://stackoverflow.com/questions/14846920/collections-emptymap-vs-new-hashmap

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
```diff
@@ -55,7 +63,9 @@ public static Map<String, String> getAvroSchemaFromSchemaRegistryClient(
     try {
       schemaMetadata = schemaRegistry.getLatestSchemaMetadata(subjectName);
     } catch (IOException | RestClientException e) {
-      throw SnowflakeErrors.ERROR_0012.getException(e);
+      // suppress the excpetion and return empty map to indicate a failure in fetching schemas
+      LOGGER.info("Schema with subject '{}' not found in schema registry", subjectName);
```
sfc-gh-japatel, line 67: How often are we printing this?

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
```diff
@@ -139,6 +149,19 @@ public static Map<String, String> getSchemaMapForTable(
         tableName, topicToTableMap, getAvroSchemaRegistryClientFromURL(schemaRegistryURL));
   }
 
+  /**
+   * Get the schema for a specific topics.
```
sfc-gh-japatel, line 153: nit: topic

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
sfc-gh-japatel: sounds good! 
I just wanted to make sure if this code would work if we disabled schema evolution on this streaming table. But we can add that as an action item to verify before we announce.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
```diff
@@ -91,6 +91,41 @@ public interface SnowflakeConnectionService {
...
+  boolean hasSchemaEvolutionPermissionForRole(String tableName, String role);
+
+  /**
+   * Check if the role used by the connector has the permission to do schema evolution on the table
+   * through a cached map
+   *
+   * @param tableName the name of the table
+   * @return whether we have the permission to do schema evolution on the table
+   */
+  boolean hasSchemaEvolutionPermission(String tableName);
```
sfc-gh-japatel, line 111: Can we combine this two APIs together?

sfc-gh-japatel: We can get rid of this function and can just use above one but pass in the default role?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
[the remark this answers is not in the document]
sfc-gh-japatel: I saw the try catch in this method and wasnt sure what would happen if the exception is not caught in that try catch. The catch only caught SQLException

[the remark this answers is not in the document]
sfc-gh-japatel: My idea was to not invoke any new code unless we have the right permissions or configs are set.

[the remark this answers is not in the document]
sfc-gh-japatel: Here is how big query does this: https://docs.confluent.io/cloud/current/connectors/cc-gcp-bigquery-sink.html#features

`autoCreateTables` and `autoUpdateSchemas`

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
sfc-gh-zefan: So far the `hasSchemaEvolutionPermission(String tableName)` reads the cache. If we let it call the other with the default role, we will be continuously querying the permission with each buffer to be inserted.

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
sfc-gh-zefan: I believe every time we try to get the schema from schema registry but found nothing there.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -450,31 +502,52 @@ protected void insertBufferedRecordsIfFlushTimeThresholdReached() {
    * Invokes insertRows API using the provided offsets which were initially buffered for this
    * partition. This buffer is decided based on the flush time threshold, buffered bytes or number
    * of records
+   *
+   * <p>When the response contains any extra column error, we will attempt to alter the table
+   * accordinly, reopen the channel and reinsert the records
```
sfc-gh-japatel, line 507: nit: accordingly
re-insert

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelTest.java
```diff
@@ -435,8 +435,7 @@ public void testInsertRows_GetOffsetTokenFailureAfterReopenChannel() throws Exce
       TopicPartitionChannel.StreamingBuffer streamingBuffer =
           topicPartitionChannel.new StreamingBuffer();
       streamingBuffer.insert(records.get(0));
-      InsertValidationResponse response =
-          topicPartitionChannel.insertBufferedRecords(streamingBuffer);
+      topicPartitionChannel.insertBufferedRecordsWithRetry(streamingBuffer, 1);
```
sfc-gh-japatel, line 438: not major: To avoid so many changes in all files you can instead do this:
Recreate private insertBufferedRecordsWithRetry(streamingBuffer, 1) instead of adding more parameters into the function and call this newly created method from the existing method. 

so insertBufferedRecords(buffer) calls insertBufferedRecirds(buffer, 1)
and tests can continue to call insertBufferedRecords(buffer)

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -170,6 +176,36 @@ public class TopicPartitionChannel {
   // Used to identify when to flush (Time, bytes or number of records)
   private final BufferThreshold streamingBufferThreshold;
 
+  // Whether schematization has been enabled.
+  private boolean enableSchematization;
+
+  // Whether schema evolution could be done on this channel
+  private boolean enableSchemaEvolution = false;
```
sfc-gh-japatel, line 183: I feel this property should be passed in from SnowflakeSinkServiceV2 and shouldnt be at channel level, instead should be at a table/connector level no? Hence making this private final just makes more sense.

src/main/java/com/snowflake/kafka/connector/SchematizationUtils.java
sfc-gh-japatel: This be too much if someone doesn't opt into schematization

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelTest.java
sfc-gh-zefan: lol, actually that was exactly what I've done several commits ago, but Toby thinks this creates extra redundancy, so it is changed to what it is now.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-japatel: was confused with insert(0, "") API. nvm 
can you comment we are adding at beginning

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -170,6 +176,36 @@ public class TopicPartitionChannel {
   // Used to identify when to flush (Time, bytes or number of records)
   private final BufferThreshold streamingBufferThreshold;
 
+  // Whether schematization has been enabled.
+  private boolean enableSchematization;
```
sfc-gh-japatel, line 180: make this private final. and move this to section where all private final instances are kept.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -170,6 +176,36 @@ public class TopicPartitionChannel {
...
+    this(
+        streamingIngestClient,
+        topicPartition,
+        channelName,
+        tableName,
+        streamingBufferThreshold,
+        sfConnectorConfig,
+        kafkaRecordErrorReporter,
+        sinkTaskContext,
+        null);
```
sfc-gh-japatel, line 206: This is not required for this PR but:
I do feel this can be refactored and we dont need to pass in connectionService. 

Consider creating a class called SchemaManager which can be created from SnowflakeSinkServiceV2 and it will have enableSchemaMapping, enableSchemaEvolution and connectionService instance. 

All the helper methods related to schematization can reside in that and we can avoid creating a huge monolith like TopicPartitionChannel. 

Please file a JIRA for this.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -450,31 +502,52 @@ protected void insertBufferedRecordsIfFlushTimeThresholdReached() {
    * Invokes insertRows API using the provided offsets which were initially buffered for this
    * partition. This buffer is decided based on the flush time threshold, buffered bytes or number
    * of records
+   *
+   * <p>When the response contains any extra column error, we will attempt to alter the table
+   * accordingly, reopen the channel and re-insert the records
+   *
+   * <p>Rows with error other than that will be added to DLQ if enabled
```
sfc-gh-japatel, line 509: can you add a jira to test this when we will send to DLQ?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -450,31 +502,52 @@ protected void insertBufferedRecordsIfFlushTimeThresholdReached() {
...
-            response.getInsertErrors(), streamingBufferToInsert.getSinkRecords());
+        handleInsertRowsWithRetryFailures(response.insertErrors, response.recordWithError);
       }
-      return response;
+      // the channel was reopened (several times, due to a racing condition in altering table or a
+      // random failure in reinsert the rows), and the channel offset was
+      // reset to be before the current buffer. The offset needs to be reset to be the end of the
+      // buffer to continue ingestion properly
+      // even if the retry does not succeed the reset is necessary to make the behavior consistent
+      // with that without schema evolution
```
sfc-gh-japatel, line 543: can you complete the sentence

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -486,23 +559,154 @@ InsertValidationResponse insertBufferedRecords(StreamingBuffer streamingBufferTo
...
-   *
-   * <p>It can also send errors {@link
-   * net.snowflake.ingest.streaming.InsertValidationResponse.InsertError} in form of response inside
-   * {@link InsertValidationResponse}
-   *
-   * @param buffer buffer to insert into snowflake
-   * @return InsertValidationResponse object sent from insertRows API.
-   */
-  private InsertValidationResponse insertRowsWithFallback(StreamingBuffer buffer) {
+  public class InsertRowsWithRetryResponse {
```
sfc-gh-japatel, line 562: java docs for this class.

sfc-gh-japatel: Can you rename this to InsertRowsReponse and instead put retry count inside this class. Class name should be generic and more details should be part of the instance variables.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-zefan: This is actually a complete sentence but I got confused when I look back at it 😆 . Changed

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -486,23 +559,154 @@ InsertValidationResponse insertBufferedRecords(StreamingBuffer streamingBufferTo
...
+    public List<InsertValidationResponse.InsertError> insertErrors;
+    public List<SinkRecord> recordWithError;
```
sfc-gh-japatel, line 564: recordsWithError

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -525,17 +729,67 @@ private InsertValidationResponse insertRowsWithFallback(StreamingBuffer buffer)
...
+        RetryPolicy.builder()
+            .withMaxAttempts(maxInsertAttempt)
+            .handle(SFException.class)
+            .handleResultIf(
+                result ->
+                    result instanceof InsertValidationResponse
+                        && ((InsertValidationResponse) result).hasErrors())
+            .onRetry(
+                // when schematization is not enabled, maxInsertAttempt will be 0 and this block
+                // will never be executed. Instead it will go to onFailure immediately
```
sfc-gh-japatel, line 742: should this be 1 or 0?

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelTest.java
```diff
@@ -591,7 +598,7 @@ public void testInsertRows_ValidationResponseHasErrors_ErrorTolerance_ALL_LogEna
         topicPartitionChannel.new StreamingBuffer();
     streamingBuffer.insert(records.get(0));
 
-    assert topicPartitionChannel.insertBufferedRecords(streamingBuffer).hasErrors();
+    assert topicPartitionChannel.insertBufferedRecordsWithRetry(streamingBuffer, 1).hasErrors();
```
sfc-gh-japatel, line 601: consider adding tests with maxRetry count >1
i think that is where the new code will kick in right?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel, line 571: Javadocs on this class

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -486,23 +559,154 @@ InsertValidationResponse insertBufferedRecords(StreamingBuffer streamingBufferTo
...
+      return this.isChannelReopened;
+    }
+
+    /**
+     * Add rows with their errors to the response. The index of the record matches that of the
+     * error.
+     *
+     * <p>Used when all retries failed
+     */
+    public void addRecordsWithErrorsToResponse(
```
sfc-gh-japatel, line 605: can you comment here saying this is also reached when there are no retries (or in non schematization case too?)

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: and on all private fields and public methods

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: +1

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: remove this if not needed

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: It is possible not all rows are erroneous. we should revert back to original code here.

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelTest.java
sfc-gh-japatel: I see, okay

test/rest_request_template/travis_correct_schema_evolution_w_auto_table_creation_avro_sr.json
```diff
@@ -0,0 +1,29 @@
+{
```
sfc-gh-japatel, line 1: wonderful, all these tests are amazing to see!

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -134,15 +140,23 @@ public void createTable(final String tableName) {
 
   public void createTableWithSchema(final String tableName, final Map<String, String> schema) {
     if (schema.isEmpty()) {
-      throw SnowflakeErrors.ERROR_5021.getException();
+      createTableWithOnlyMetadataColumn(tableName);
+      return;
     }
+    String comment = "comment 'created by automatic table creation from Snowflake Kafka Connector'";
```
sfc-gh-japatel, line 146: all of these can be public static final values

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -161,6 +175,28 @@ public void createTableWithSchema(final String tableName, final Map<String, Stri
...
 
+  public void createTableWithOnlyMetadataColumn(final String tableName) {
+    checkConnection();
+    if (!hasSchemaEvolutionPermission(tableName)) {
+      throw SnowflakeErrors.ERROR_5021.getException();
+    }
+    InternalUtils.assertNotEmpty("tableName", tableName);
+    String createTableQuery =
+        "create table if not exists identifier(?) (record_metadata variant comment 'created by"
+            + " automatic table creation from Snowflake Kafka Connector')";
```
sfc-gh-japatel, line 186: use comment here?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-japatel: rename it to `snowflakeCreatedComment`

sfc-gh-japatel (approved):
Thanks Zekai, overall lgtm! But I have few concerns:
1. Refactoring 
2. Code coverage numbers are reduced 
3. These are a lot of changes in one PR. Given the time we had, you have done phenomenal job to make it able to even release it to our customers with some conditions/limitations. 
Great job on adding tests too, but I feel we can add even more.

I would rather consider pushing it to a different branch and then fetch it once we are ready so as to not have tech debt

sfc-gh-tzhang:
Will try to see if we could use the schema in the record instead of getting it from schema registry, which limits us the converts that could be supported

PR #473, opened 2022-08-12
Disable whitesource checks as we migrated to synk

sfc-gh-kdama:
Disable whitesource checks as we migrated to synk

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -60,6 +60,6 @@ mvn --settings $OSSRH_DEPLOY_SETTINGS_XML -DskipTests clean deploy
 #confluent release
 mvn -f pom_confluent.xml --settings $OSSRH_DEPLOY_SETTINGS_XML -DskipTests clean package
 #white source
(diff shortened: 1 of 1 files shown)


deploy.sh
```diff
@@ -60,6 +60,6 @@ mvn --settings $OSSRH_DEPLOY_SETTINGS_XML -DskipTests clean deploy
 #confluent release
 mvn -f pom_confluent.xml --settings $OSSRH_DEPLOY_SETTINGS_XML -DskipTests clean package
 #white source
-whitesource/run_whitesource.sh
+# whitesource/run_whitesource.sh
```
sfc-gh-japatel, line 63: Instead of comment, just remove it? unless we are planning to use in future? also delete the whitesource.sh script?

sfc-gh-japatel (approved):

deploy.sh
sfc-gh-kdama: Lets Remove this with next release.

PR #476, opened 2022-08-18
SNOW-645776 Run different versions of confluent and AK using matrix strategy + Refactor

sfc-gh-japatel:
Changes:
- Introducing Matrix for cloud type and version type in AK and Confluent versions
  - Confluent 5.5.11, 6.2.6 and 7.2.1
  - Apache Kafka 2.5.1, 2.8.1, 3.2.0 (These are more or less what Confluent versions belong to) 
  - Example https://docs.confluent.io/platform/current/release-notes/index.html
- Upgrade Python to 3.9
- ubuntu-20.04

Benefits:
- Less code to maintain.
- Takes only 1 hour to run the entire suite compared to 2 hours.
- Moved pressure test to StressTest.yml - this was the flaky test. 
- Future versions compatible by adding only to version matrix
- Runs in parallel
- Keeps up with latests versions. 

Next:
StressTest will also be upgraded to newer versions.


sfc-gh-japatel:
This is ready for Review!

sfc-gh-tzhang:
> Since I introduced continue-on-error: True
> 
> We have to go to summary section of each test to see if we have any failed tests.
> There have been lots of complaints on this, hopefully github fixes this.

Could you explain what does this mean? The merge gate will pass even if there is a test failure?

sfc-gh-japatel:
> > Since I introduced continue-on-error: True
> > We have to go to summary section of each test to see if we have any failed tests.
> > There have been lots of complaints on this, hopefully github fixes this.
> 
> Could you explain what does this mean? The merge gate will pass even if there is a test failure?

Yes, that is correct understanding. But I removed it so removing it from the description..

test/apache_properties/schema-registry.properties
```diff
@@ -25,7 +25,7 @@ listeners=http://0.0.0.0:8081
 # (see zookeeper docs for details).
 # This is a comma separated host:port pairs, each corresponding to a zk
 # server. e.g. "127.0.0.1:3000,127.0.0.1:3001,127.0.0.1:3002".
-kafkastore.connection.url=localhost:2181
+#kafkastore.connection.url=localhost:2181
```
sfc-gh-tzhang, line 28: could you explain the change in this file?

sfc-gh-tzhang (approved):
Thanks Jay! Could you explain what you did to reduce the entire to 1 hour?

test/apache_properties/schema-registry.properties
sfc-gh-japatel: I ran into schema registry issues while running version 7.2.1 in confluent. the fix was to not use kafkastore.

sfc-gh-japatel:
> Thanks Jay! Could you explain what you did to reduce the suite runtime to 1 hour?

I removed the stress tests from normal pull request and merge to master, instead run it only in a cron'ed schedule.

PR #478, opened 2022-08-23
SNOW-645776 Part 2 Stress Tests run on Confluent's Different Versions Against Different Cloud

sfc-gh-japatel:
This Change:
- Required to identify if we can support future versions of Confluent Platform. 
- End to End tests are already passing, these two tests were the most flakiest. 
- Refactor stress tests to different functions so as to not run along
  with End to End Tests
- Stress Tests wont run on master push or PR, will only run once every
  week on saturday (Check Schedule in StressTest.yml file)

Future Improvements: (Filed JIRAs)
- Run a lot of Job Pre Steps inside a docker container and just pull that image. 
- Refactor test suits to a map instead of a list of Tests and List of True False, makes it easier for reading and enabling/disabling tests.

### DIFF
--- a/.github/workflows/StressTest.yml
+++ b/.github/workflows/StressTest.yml
@@ -1,9 +1,6 @@
 name: Kafka Connector Stress Test
 
 on:
-  schedule:
-    - cron:  '0 8 * * 6' # at 8.00 on Saturday
-  # This is for Testing against this PR, remove it later before pushing
   push:
     branches: [ master ]
   pull_request:
--- a/.github/workflows/StressTest.yml
+++ b/.github/workflows/StressTest.yml
@@ -3,10 +3,19 @@ name: Kafka Connector Stress Test
 on:
   schedule:
     - cron:  '0 8 * * 6' # at 8.00 on Saturday
+  # This is for Testing against this PR, remove it later before pushing
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: '**'
 
 jobs:
   build:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        confluent_version: [ '5.5.11', '6.2.6', '7.2.1' ]
+        snowflake_cloud: [ 'AWS', 'AZURE', 'GCS' ]
     steps:
     - name: Checkout Code
       uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


sfc-gh-tzhang:
Does this mean that we won't have Azure and GCS coverage for merge date?

sfc-gh-japatel:
> Does this mean that we won't have Azure and GCS coverage for merge date?

No, it runs stress tests in all three clouds.

sfc-gh-tzhang:
> > Does this mean that we won't have Azure and GCS coverage for merge date?
> 
> No, it runs stress tests in all three clouds. The other PR also runs in all three clouds. So no change, the flakiness was because of too many tests running on 1 server and because of duplicates that gets inserted into the table. The tests we have verify exact table count as what we inserted, but this stress test has chaos test too which pauses, stops, restarts the connector.

Assuming I understand correctly, we move the stress tests out of the merge gate and the stress tests will now be run in a certain schedule, my concern is that no one will really look at it anymore and we're loosing coverage in the merge gate. Does this sound like a valid concern? Should we simplify the stress tests and add part of them to the merge gate to make sure at least all the behaviors are tested?

sfc-gh-japatel:
> > > Does this mean that we won't have Azure and GCS coverage for merge date?
> > 
> > 
> > No, it runs stress tests in all three clouds. The other PR also runs in all three clouds. So no change, the flakiness was because of too many tests running on 1 server and because of duplicates that gets inserted into the table. The tests we have verify exact table count as what we inserted, but this stress test has chaos test too which pauses, stops, restarts the connector.
> 
> Assuming I understand correctly, we move the stress tests out of the merge gate and the stress tests will now be run in a certain schedule, my concern is that no one will really look at it anymore and we're loosing coverage in the merge gate. Does this sound like a valid concern? Should we simplify the stress tests and add part of them to the merge gate to make sure at least all the behaviors are tested?

Kind of a valid concern but I dont think we are relying on these tests anyways. Anyways I have reduced the no of rounds, so I can keep it in merge gates and not rely on schedule crons which we dont check. 

Until we move away from github actions to Jenkins, I agree we should keep this in merge gate. 

(So the code is actually keeping it in merge gate but I was planning to remove it)

So I will keep your recommendation to keep in merge gates. 

Thanks for bring this up Toby!

.github/workflows/StressTest.yml
```diff
@@ -3,10 +3,19 @@ name: Kafka Connector Stress Test
 on:
   schedule:
     - cron:  '0 8 * * 6' # at 8.00 on Saturday
+  # This is for Testing against this PR, remove it later before pushing
```
sfc-gh-japatel, line 6: I will remove this comment and keep code as is. remove schedule cron.

sfc-gh-tzhang (approved):
LGTM

PR #484, opened 2022-09-08
[SNOW-654492] Enable nonproxyhosts parameter 

sfc-gh-rcheng:
Status: waiting on a testing environment

JDBC allows the customer to create nonProxyHosts. Multiple hosts can be set with a '|' and a '*' wildcard. 

https://docs.snowflake.com/en/user-guide/jdbc-configure.html#specifying-a-proxy-server-by-setting-java-system-properties

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/Utils.java
+++ b/src/main/java/com/snowflake/kafka/connector/Utils.java
@@ -16,18 +16,18 @@
  */
 package com.snowflake.kafka.connector;
 
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG;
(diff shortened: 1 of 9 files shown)


src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -220,6 +213,7 @@ public static void setJDBCLoggingDirectory() {
 
   /**
    * validate whether proxy settings in the config is valid
+   * note: nonproxyhosts have no requirements, so they are not validated here
```
sfc-gh-hachouraria, line 216: Just a single comment (nit): Given the env-vars use `,` as delimiters while JVMs use `|` perhaps it's worth validating that `,` characters are not passed in this value? It could help avoid the troubleshooting required when a user copy-pastes things and does not carefully follow a document.

sfc-gh-rcheng:
There are merge conflicts with master, I'll fix these once the customer confirms that this works for them

PR #486, opened 2022-09-09
[SNOW-658355] Add invalid configuration parameters to the exception

sfc-gh-rcheng:
Status: currently deprioritized. revi would like to make some more extensive changes to parameters, which would nullify a lot of changes in this pr. Waiting for more time when she can make these changes

https://github.com/snowflakedb/snowflake-kafka-connector/pull/489

Issue: Customer would like to know which config parameters are invalid when creating a connector
Changes: changed 'configIsValid' boolean to a hashset to store the parameter names for each invalid parameter


src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -223,13 +222,17 @@ public static void setJDBCLoggingDirectory() {
    *
    * @param config connector configuration
    */
+  // TODO @rcheng: question - can i change this exception to also return the invalid configs?
```
sfc-gh-rcheng, line 225: validateProxySetting is only ever called within a try/catch. It would be more performant and easier to log if we  returned a list of invalid configs. Can I make this change?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -361,66 +364,65 @@ static String validateConfig(Map<String, String> config) {
...
       if (config.containsKey(SnowflakeSinkConnectorConfig.ENABLE_SCHEMATIZATION_CONFIG)
           && Boolean.parseBoolean(
               config.get(SnowflakeSinkConnectorConfig.ENABLE_SCHEMATIZATION_CONFIG))) {
-        configIsValid = false;
         LOGGER.error(
             Logging.logMessage(
                 "Schematization is only available with {}.",
                 IngestionMethodConfig.SNOWPIPE_STREAMING.toString()));
+        invalidConfigParams.add(SnowflakeSinkConnectorConfig.ENABLE_SCHEMATIZATION_CONFIG);
+        // TODO @rcheng: question - can we add logging information in the error? this is confusing for customer
```
sfc-gh-rcheng, line 377: would we want to add more info here? or should we rely on the customer self-debugging with their logs?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -469,14 +471,16 @@ && parseTopicToTableMap(config.get(SnowflakeSinkConnectorConfig.TOPICS_TABLES_MA
...
 
     // Check all config values for ingestion method == IngestionMethodConfig.SNOWPIPE_STREAMING
-    final boolean isStreamingConfigValid = StreamingUtils.isStreamingSnowpipeConfigValid(config);
+    invalidConfigParams.addAll(StreamingUtils.isStreamingSnowpipeConfigValid(config));
 
-    if (!configIsValid || !isStreamingConfigValid) {
-      throw SnowflakeErrors.ERROR_0001.getException();
+    if (!invalidConfigParams.isEmpty()) {
+      String errorMsg = VALIDATION_ERRORMSG_PREFIX + invalidConfigParams;
+      // TODO @rcheng: question - do we want to trust the set.tostring function here?
```
sfc-gh-rcheng, line 482: should i implement a custom toString for the set here?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -569,6 +573,7 @@ public static Map<String, String> parseTopicToTableMap(String input) {
       String topic = tt[0].trim();
       String table = tt[1].trim();
 
+      // TODO @rcheng: question - why the distinction between isInvalid and returning null?
```
sfc-gh-rcheng, line 576: this function behaves differently when given different invalid topic to table maps. I think it should treat invalid table maps consistently - return null for the purposes of this PR. Can i make this change?

src/main/java/com/snowflake/kafka/connector/internal/BufferThreshold.java
```diff
@@ -126,21 +129,26 @@ public long getFlushTimeThresholdSeconds() {
    * @param ingestionMethodConfig ingestion method used. Check {@link IngestionMethodConfig}
    * @return true if all thresholds are valid.
    */
-  public static boolean validateBufferThreshold(
+  public static Set<String> validateBufferThreshold(
```
sfc-gh-rcheng, line 132: the three verify* methods in this class are very similar. can i refactor them into one method?

sfc-gh-rcheng:
overall I saw different design choices in the code that confused me. if they were intentional i'd like to know why, if not i'd like to fix them either in this or a later pr

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -36,9 +36,7 @@
 import java.net.PasswordAuthentication;
 import java.net.URL;
 import java.net.URLConnection;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Random;
+import java.util.*;
```
sfc-gh-japatel, line 39: nit: no * imports
Probably intellij thing. 

Run a google formatter to match style.

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -223,13 +221,17 @@ public static void setJDBCLoggingDirectory() {
    *
    * @param config connector configuration
    */
+  // TODO @rcheng: question - can i change this exception to also return the invalid configs?
+  //        is there a reason we want the functionality of throwing immediately here instead of collecting params?
+  //        this method is only ever called in a try catch
```
sfc-gh-japatel, line 226: yes, that would work! for every invalid config, only until everything is collected we throw an exception. So may be just collecting every config key value pair sounds like a good idea.

src/main/java/com/snowflake/kafka/connector/internal/BufferThreshold.java
sfc-gh-japatel: Sure! Can you rename the methods too since it is not only verifying but also returning? and change javadocs for returning a set instead of a boolean now?

sfc-gh-japatel: The reason I kept them separate was because of those methods had different behaviors to verify.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java
```diff
@@ -161,7 +153,8 @@ public static boolean isStreamingSnowpipeConfigValid(final Map<String, String> i
                     "Config:{} should be present if ingestionMethod is:{}",
                     Utils.SF_ROLE,
                     inputConfig.get(INGESTION_METHOD_OPT)));
-            configIsValid = false;
+            invalidConfigParams.add(Utils.SF_ROLE);
+            // TODO @rcheng: question - can we add logging information in the error? this is confusing for customer
```
sfc-gh-japatel, line 157: sorry what logging information?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -335,7 +337,7 @@ static boolean isValidSnowflakeTableName(String tableName) {
    * @return connector name
    */
   static String validateConfig(Map<String, String> config) {
-    boolean configIsValid = true; // verify all config
+    Set<String> invalidConfigParams = new HashSet<>(); // verify all config and list invalid ones
```
sfc-gh-japatel, line 340: Instead of a set of String, lets create an Immutable class may be called 'InvalidParameters` with few private final strings called configName, configValue and then errorMsg. 

This will reduce a lot of redundant messaging here starting with LOGGER.error(Loggin.logMessage)

That immutable class can have a function called logInvalidProperties which can accept logger

sfc-gh-japatel: WDYT?

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-japatel: sure. I think the general coding guideline is to not have multiple types of changes in one PR. so you can definitely make changes/refactor as required but may be add a TODO to do it in a separate PR?

src/main/java/com/snowflake/kafka/connector/internal/BufferThreshold.java
```diff
@@ -159,27 +167,27 @@ private static boolean verifyBufferFlushTimeThreshold(
...
             Logging.logMessage(
                 "{} should be an integer. Invalid integer was provided:{}",
                 SnowflakeSinkConnectorConfig.BUFFER_FLUSH_TIME_SEC,
                 providedFlushTimeSecondsInStr));
-        return false;
+        return SnowflakeSinkConnectorConfig.BUFFER_FLUSH_TIME_SEC;
       }
     }
-    return true;
+    return "";
```
sfc-gh-japatel, line 181: not a big fan of returning an empty string here, can you return an ImmutablePair with empty things in it which can be a static final constant. (Denoting nothing is wrong with this config)

At the caller side, we can just filter that empty string to just have all invalid Pairs.

sfc-gh-japatel (changes requested):
left some suggestions. 
Can you also add an UT/IT with multiple invalid configs and verify them?

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-rcheng: Sounds good, created a JIRA for it
https://snowflakecomputing.atlassian.net/browse/SNOW-659455

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-rcheng: I decided against keeping the value of the invalid config because I was worried about security (in terms of injection?). But if that is not a concern, I'd be happy to put the invalid config value in the error message along with the invalid key

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java
sfc-gh-rcheng: The ENABLE_SCHEMATIZATION_CONFIG in Utils.java on (roughly) line 368 would be a better example. We add the enable.schematization param to the invalid params when its called on snowpipe, however we don't let the customer know that schematization is only available on streaming snowpipe. From the customer perspective, this could get confusing from just reading the error

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-rcheng: That was my original intention, but I wasn't sure if it was too much for this logging. I'll make the change!

PR #488, opened 2022-09-14
SNOW-625820: multiple fixes for KC schematization schema mapping

sfc-gh-tzhang:
- Update the config key word from `enable.schematization` to `snowflake.enable.schematization`
- Remove some duplicate constants
- Fix a null pointer exception during channel closing
- Update the column name format function to return the quoted column name

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
@@ -1,9 +1,5 @@
 package com.snowflake.kafka.connector.internal.streaming;
 
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.BUFFER_SIZE_BYTES_DEFAULT;
-import static com.snowflake.kafka.connector.internal.streaming.StreamingUtils.STREAMING_BUFFER_COUNT_RECORDS_DEFAULT;
-import static com.snowflake.kafka.connector.internal.streaming.StreamingUtils.STREAMING_BUFFER_FLUSH_TIME_DEFAULT_SEC;
-
 import com.codahale.metrics.MetricRegistry;
 import com.google.common.annotations.VisibleForTesting;
 import com.snowflake.kafka.connector.SchematizationUtils;
@@ -17,12 +13,6 @@ import com.snowflake.kafka.connector.internal.SnowflakeSinkService;
 import com.snowflake.kafka.connector.internal.telemetry.SnowflakeTelemetryService;
 import com.snowflake.kafka.connector.records.RecordService;
 import com.snowflake.kafka.connector.records.SnowflakeMetadataConfig;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Properties;
 import net.snowflake.ingest.streaming.SnowflakeStreamingIngestClient;
 import net.snowflake.ingest.streaming.SnowflakeStreamingIngestClientFactory;
 import net.snowflake.ingest.utils.SFException;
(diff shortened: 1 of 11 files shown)


src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -230,8 +229,9 @@ public Map<String, Object> getProcessedRecordForStreamingIngest(SinkRecord recor
         if (metadataConfig.allFlag) {
           streamingIngestRow.put(TABLE_COLUMN_METADATA, MAPPER.writeValueAsString(row.metadata));
         }
-      } catch (IOException e) {
+      } catch (JsonProcessingException e) {
         // return an exception and propagate upwards
+        // TODO tzhang: do we need to rethrow?
```
sfc-gh-tzhang, line 234: @sfc-gh-japatel I believe we need to rethrow the exception here, don't we?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -293,10 +293,14 @@ public void close(Collection<TopicPartition> partitions) {
               partitionChannelKey(topicPartition.topic(), topicPartition.partition());
           TopicPartitionChannel topicPartitionChannel =
               partitionsToChannel.get(partitionChannelKey);
-          topicPartitionChannel.closeChannel();
+          // Check for null since it's possible that the something goes wrong even before the
```
sfc-gh-tjones, line 296: When can this happen?

sfc-gh-tjones (approved):

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-tzhang: It could happen in many cases, for example, if the table doesn't have a correct schema, or we don't have permission to ALTER COLUMN, see `createTableIfNotExists ` in `SnowflakeSinkServiceV2.startTask` if you're interested

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -263,9 +261,10 @@ private Map<String, Object> getMapFromJsonNodeForStreamingIngest(JsonNode node)
         columnValue = MAPPER.writeValueAsString(columnNode);
       }
       // while the value is always dumped into a string, the Streaming Ingest SDK
-      // will transformed the value according to its type in the table
-      // TODO: SNOW-630885 the record could come directly as a map, and don't have to be
-      //  dumped into a string but the original type could be used.
+      // will be transformed the value according to its type in the table
```
sfc-gh-japatel, line 264:
```suggestion
      // will transform the value according to its type in the table
```

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-japatel: It can only throw exception in case `MAPPER.writeValueAsString` throws exception. 
Is there a need to change this from IoException to JsonProcessExc...?

sfc-gh-japatel: We will need to catch the exception and put it into DLQ? Should this part of separate change?

sfc-gh-tzhang: hmmm, this is your code, not from Zekai, it used to be JsonProcessingException, and he changed it to IoException which i feel like is by accident. And we didn't thrown so I'm not sure what you mean by "put it into DLQ", could you elaborate?

sfc-gh-japatel: Catching JsonExcetion makes sense. Now that I am looking and reading it again,seems like just swallowing it will continue the connector.. 

When I said, put it into DLQ, I meant rather than just continuing with it, does it makes sense to rethrow it and make it deliverable to DLQ. 

Doesnt necessarily need to be in this change as I am not blocking here, just trying to understand how this will work. 

Please file a JIRA.. Now that we will have a schematization go throw same logic, we will need to carefully process this. Earlier the chances of it throwing an exception were less since we were just converting everything into JSON.

sfc-gh-japatel (approved):
lgtm! Thanks

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-tzhang: > When I said, put it into DLQ, I meant rather than just continuing with it, does it makes sense to rethrow it and make it deliverable to DLQ?

Yes, I think we should, and that's the reason I pinged you here. Looks like now we may just skipping rows which is not correct, I will update it in this PR.

sfc-gh-japatel: Thanks Toby!

sfc-gh-tzhang: This is not easy, will do in the next PR

PR #490, opened 2022-09-15
Fix docker-setup yml files

sfc-gh-rcheng:
Problem: Docker doesn't install the kc sink connector correctly when given the '/usr/local/share/kafka/plugins' path. 
Changes:
- copy .jar file to docker-setup/jar folder
- point .yml file to that jar folder
- update .gitignore to ignore .jar and .yml file
- update readme


docker-setup/snowflake-kafka-docker/kafka-connect.yml
```diff
@@ -65,10 +65,7 @@ services:
       - 9876:9876 # JMX port
       - 8085:8085
     volumes:
-      - /usr/local/share/kafka/plugins:/etc/kafka-connect/jars/ # if you want your custom jar, put custom jar in ./jars directory
-    #  - ./prometheus:/opt/jmx-exporter/
-    # volumes:
-    #   - ./config/connect-log4j.properties:/etc/kafka/connect-log4j.properties
```
sfc-gh-tzhang, line 71: should we keep it here? I assume they're needed in some cases

sfc-gh-tzhang (approved):
lgtm, please make sure to get a sign off from @sfc-gh-japatel as well

docker-setup/snowflake-kafka-docker/kafka-connect.yml
sfc-gh-rcheng: They may be needed in non m1 environments, however I haven't needed these. I think its ok to keep it here

sfc-gh-japatel: I havent had a successful run with prometheus and grafanna. 
Can you instead keep it here (Commented) and make a comment here saying we dont need this ATM. Only needed for exporting JMX to prometheus

sfc-gh-japatel (approved):
lgtm!

PR #491, opened 2022-09-16
NO-SNOW: Put rows with JsonProcessingException into the DLQ instead of ignoring them

sfc-gh-tzhang:
Put rows with JsonProcessingException into the DLQ instead of ignoring them


sfc-gh-tzhang:
bump

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -992,23 +993,28 @@ protected long getApproxSizeOfRecordInBytes(SinkRecord kafkaSinkRecord) {
...
+          sinkRecordBufferSizeInBytes += ((String) value).length() * 2L; // 1 char = 2 bytes
+        } else {
+          // for now it could only be a list of string
+          for (String s : (List<String>) value) {
+            sinkRecordBufferSizeInBytes += s.length() * 2L;
+          }
         }
       }
+    } catch (JsonProcessingException e) {
+      // We ignore any errors here because this is just calculating the record size
```
sfc-gh-tjones, line 1015: Should we instead explicitly handle this or log the exception? What happens if we continue?

sfc-gh-tjones (approved):
LGTM barring my question about the `JsonProcessingException`. Looks like we silently swallowed it before anyway but I'm wondering if there's something we should be explicitly doing earlier on.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1082,10 +1088,16 @@ public List<Map<String, Object>> getData() {
...
-          snowflakeTableRowsFromSinkRecords.add(tableRow);
+
+          // Convert this records into Json Schema which has content and metadata, add it to DLQ if
+          // there is an exception
+          try {
+            Map<String, Object> tableRow =
+                recordService.getProcessedRecordForStreamingIngest(snowflakeRecord);
+            snowflakeTableRowsFromSinkRecords.add(tableRow);
+          } catch (JsonProcessingException e) {
+            kafkaRecordErrorReporter.reportError(kafkaSinkRecord, e);
```
sfc-gh-japatel, line 1099: lets add a warn here?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: I feel its okay to skip this and continue since we dont want to calculate/add this record's byte length to memory. A warn should be useful here below in catch exception.

sfc-gh-japatel (approved):
lgtm! Thanks

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: We will call `getProcessedRecordForStreamingIngest` in two places and I added the log in the other place, also I create https://snowflakecomputing.atlassian.net/browse/SNOW-663981 since it's not ideal to process every record twice, ideally we should process every record only once and save that

PR #492, opened 2022-09-20
[SNOW-665920] Add warnings in build script to prevent a privately built jar from going to the customer

sfc-gh-rcheng:
I sent a privately built jar to customer support for testing and didn't realize until after the customer received the jar that it was a mistake. Adding a comment in the build script to prevent future similar mistakes


sfc-gh-rcheng:
New message at end of build looks like this:
image

test/build_apache.sh
```diff
@@ -81,14 +81,17 @@ popd
...
-echo -e "\n=== copied connector to $KAFKA_CONNECT_PLUGIN_PATH ==="
+echo -e "copied connector to $KAFKA_CONNECT_PLUGIN_PATH"
 
 KAFKA_CONNECT_DOCKER_JAR_PATH="$SNOWFLAKE_CONNECTOR_PATH/docker-setup/snowflake-kafka-docker/jars"
 mv $SNOWFLAKE_PLUGIN_PATH/$SNOWFLAKE_PLUGIN_NAME $KAFKA_CONNECT_DOCKER_JAR_PATH || true
-echo -e "\n=== copied connector to $KAFKA_CONNECT_DOCKER_JAR_PATH for docker ==="
\ No newline at end of file
+echo -e "copied connector to $KAFKA_CONNECT_DOCKER_JAR_PATH for docker"
+
+echo -e "=== NOTE: do not share this privately built jar with customers ===\n"
```
sfc-gh-tzhang, line 97: I think we actually do this for some special cases, usually if a feature is in PrPr and we're working with a customer on a fix, then providing a private jar is ok. The case we would want to avoid is that asking customer to run a private jar on their production environment.

sfc-gh-rcheng: Updated the message to reflect special cases and to not let the customer run it in prod

sfc-gh-tzhang (approved):
LGTM!

PR #494, opened 2022-09-21
Modify Codeowners

sfc-gh-japatel:
Adding Revi 
Removed Raghav and Matt (Who are not actively working on KC and they can save some email notifications)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @sfc-gh-japatel @sfc-gh-rramachandran @sfc-gh-mnaides @sfc-gh-tzhang @sfc-gh-tjones
+* @sfc-gh-japatel @sfc-gh-tzhang @sfc-gh-tjones @sfc-gh-rcheng


sfc-gh-tjones (approved):

.github/CODEOWNERS
```diff
@@ -1 +1 @@
-* @sfc-gh-japatel @sfc-gh-rramachandran @sfc-gh-mnaides @sfc-gh-tzhang @sfc-gh-tjones
+* @sfc-gh-japatel @sfc-gh-tzhang @sfc-gh-tjones @sfc-gh-rcheng
```
sfc-gh-tzhang, line 1: I think this won't work long term, we should just add Snowpipe and Snowpipe Streaming team, and whoever wants to be added explicitly could do it themselves, WDYT?

sfc-gh-japatel: Sounds like a plan! I will do it.

sfc-gh-rcheng (approved):

PR #495, opened 2022-09-22
[SNOW-665898] Add log granularity instance id logging for Task

sfc-gh-rcheng:
Changes:
- Refactored LoggerHandler to log a "kcGlobalInstanceId" instead of a "correlationId"
- Added logger specific instanceId to LoggingHandler. This is set during logger creation
- Refactored LoggerHandlerTest with test runners
- Added an instanceId to Task loggers

image

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
@@ -82,7 +82,7 @@ public class SnowflakeSinkConnector extends SinkConnector {
     Utils.checkConnectorVersion();
 
     // initialize logging with correlationId
-    LoggerHandler.setCorrelationUuid(UUID.randomUUID());
+    LoggerHandler.setKcGlobalInstanceId(UUID.randomUUID());
 
     LOGGER.info("SnowflakeSinkConnector:start");
     setupComplete = false;
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
@@ -16,6 +16,8 @@
  */
 package com.snowflake.kafka.connector;
 
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.DELIVERY_GUARANTEE;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.snowflake.kafka.connector.dlq.KafkaRecordErrorReporter;
 import com.snowflake.kafka.connector.internal.LoggerHandler;
@@ -26,14 +28,6 @@ import com.snowflake.kafka.connector.internal.SnowflakeSinkService;
 import com.snowflake.kafka.connector.internal.SnowflakeSinkServiceFactory;
 import com.snowflake.kafka.connector.internal.streaming.IngestionMethodConfig;
 import com.snowflake.kafka.connector.records.SnowflakeMetadataConfig;
-import org.apache.kafka.clients.consumer.OffsetAndMetadata;
-import org.apache.kafka.common.TopicPartition;
-import org.apache.kafka.connect.errors.ConnectException;
-import org.apache.kafka.connect.errors.RetriableException;
-import org.apache.kafka.connect.sink.ErrantRecordReporter;
-import org.apache.kafka.connect.sink.SinkRecord;
-import org.apache.kafka.connect.sink.SinkTask;
-
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
@@ -42,8 +36,13 @@ import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Supplier;
-
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.DELIVERY_GUARANTEE;
+import org.apache.kafka.clients.consumer.OffsetAndMetadata;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.connect.errors.ConnectException;
+import org.apache.kafka.connect.errors.RetriableException;
+import org.apache.kafka.connect.sink.ErrantRecordReporter;
+import org.apache.kafka.connect.sink.SinkRecord;
+import org.apache.kafka.connect.sink.SinkTask;
 
 /**
  * SnowflakeSinkTask implements SinkTask for Kafka Connect framework.
@@ -79,8 +78,8 @@ public class SnowflakeSinkTask extends SinkTask {
   // check connect-distributed.properties file used to start kafka connect
   private final int rebalancingSleepTime = 370000;
 
-  private static final LoggerHandler LOGGER = new LoggerHandler(SnowflakeSinkTask.class.getName(), UUID.randomUUID(),
-    "TASK");
+  private static final LoggerHandler LOGGER =
+      new LoggerHandler(SnowflakeSinkTask.class.getName(), UUID.randomUUID(), "TASK");
 
   /** default constructor, invoked by kafka connect framework */
   public SnowflakeSinkTask() {
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
@@ -16,8 +16,6 @@
  */
 package com.snowflake.kafka.connector;
 
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.DELIVERY_GUARANTEE;
-
 import com.google.common.annotations.VisibleForTesting;
 import com.snowflake.kafka.connector.dlq.KafkaRecordErrorReporter;
 import com.snowflake.kafka.connector.internal.LoggerHandler;
@@ -28,13 +26,6 @@ import com.snowflake.kafka.connector.internal.SnowflakeSinkService;
 import com.snowflake.kafka.connector.internal.SnowflakeSinkServiceFactory;
 import com.snowflake.kafka.connector.internal.streaming.IngestionMethodConfig;
 import com.snowflake.kafka.connector.records.SnowflakeMetadataConfig;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Optional;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeoutException;
-import java.util.function.Supplier;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.connect.errors.ConnectException;
@@ -42,6 +33,17 @@ import org.apache.kafka.connect.errors.RetriableException;
 import org.apache.kafka.connect.sink.ErrantRecordReporter;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.apache.kafka.connect.sink.SinkTask;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Supplier;
+
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.DELIVERY_GUARANTEE;
 
 /**
  * SnowflakeSinkTask implements SinkTask for Kafka Connect framework.
@@ -77,7 +79,8 @@ public class SnowflakeSinkTask extends SinkTask {
   // check connect-distributed.properties file used to start kafka connect
   private final int rebalancingSleepTime = 370000;
 
-  private static final LoggerHandler LOGGER = new LoggerHandler(SnowflakeSinkTask.class.getName());
+  private static final LoggerHandler LOGGER = new LoggerHandler(SnowflakeSinkTask.class.getName(), UUID.randomUUID(),
+    "TASK");
 
   /** default constructor, invoked by kafka connect framework */
   public SnowflakeSinkTask() {
--- a/src/main/java/com/snowflake/kafka/connector/internal/LoggerHandler.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/LoggerHandler.java
@@ -1,36 +1,41 @@
 package com.snowflake.kafka.connector.internal;
 
 import com.snowflake.kafka.connector.Utils;
-import java.util.UUID;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.util.UUID;
 
 /** Attaches additional fields to the logs */
 public class LoggerHandler {
   // static properties and methods
-  private static final UUID CORRELATION_ID_EMPTY = null;
+  private static final String EMPTY_ID_TAG = "";
   private static final Logger META_LOGGER = LoggerFactory.getLogger(LoggerHandler.class.getName());
-  private static UUID loggerCorrelationId = CORRELATION_ID_EMPTY;
+  private static String kcGlobalInstanceIdTag = EMPTY_ID_TAG;
 
   /**
-   * Sets the correlationId for all loggers. This should only be called in start so that the entire
-   * kafka connector instance has the same correlationId logging
+   * Sets the KC global instance id for all loggers.
    *
-   * @param correlationId UUID attached for every log
+   * <p>This should only be called in start so that the entire kafka connector instance has the same
+   * correlationId logging.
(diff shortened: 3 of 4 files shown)


sfc-gh-rcheng:
First commit (refactor to kcglobalinstanceid) is mostly just renaming properties to remove the concept of correlationId. It may be easier to read this PR if you view the commits separate from commit 1

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
```diff
@@ -82,7 +82,7 @@ public void start(final Map<String, String> parsedConfig) {
     Utils.checkConnectorVersion();
 
     // initialize logging with correlationId
-    LoggerHandler.setCorrelationUuid(UUID.randomUUID());
+    LoggerHandler.setKcGlobalInstanceId(UUID.randomUUID());
```
sfc-gh-japatel, line 85: nit: `setKCGlobalInstanceId` or may be `setConnectGlobalInstanceId`

sfc-gh-japatel (commented):
Is the idea here to have one uuid per instance and leverage all tasks in that instance/node to share the same uuid? 
If so, can you add/verify a test with multiple tasks. (For it to work, you will need more than one partition or at least 1 task to 1 partition mapping)

sfc-gh-japatel:
https://cwiki.apache.org/confluence/display/KAFKA/KIP-449%3A+Add+connector+contexts+to+Connect+worker+logs

This is what they have and I wish they had UUID in it.

sfc-gh-rcheng:
> Is the idea here to have one uuid per instance and leverage all tasks in that instance/node to share the same uuid? If so, can you add/verify a test with multiple tasks. (For it to work, you will need more than one partition or at least 1 task to 1 partition mapping)

That was not the goal (but was what the code did at the time of review). The goal is to set a new instance id when a task is started, not for all tasks on that node. From what I understand of the SinkTask, start is a onetime operation run when a task starts ingesting data while open is used post rebalancing. Happy to change where the instanceId is set if there is a better place to put it

https://github.com/a0x8o/kafka/blob/master/connect/api/src/main/java/org/apache/kafka/connect/sink/SinkTask.java

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -77,11 +78,18 @@ public class SnowflakeSinkTask extends SinkTask {
   // check connect-distributed.properties file used to start kafka connect
   private final int rebalancingSleepTime = 370000;
 
-  private static final LoggerHandler LOGGER = new LoggerHandler(SnowflakeSinkTask.class.getName());
+  // the dynamic logger is intended to be attached per task instance. the instance id will be set
+  // during task start,
+  // however if it is not set, it falls back to the static logger
+  private static final LoggerHandler STATIC_LOGGER =
```
sfc-gh-rcheng, line 84: this is only used for the 'getTopicToTableMap' method, which in turn is called in start and in Utils. It maybe better to move the method into Utils so we can remove this static logger

sfc-gh-rcheng:
> https://cwiki.apache.org/confluence/display/KAFKA/KIP-449%3A+Add+connector+contexts+to+Connect+worker+logs
> 
> This is what they have and I wish they had UUID in it.

This is much more readable and concise than using a UUID. It may be better to do something like this where we increment a task counter every time we start a new task instead of setting a new UUID. This would only work if Tasks are only run after a the connector has been start()'ed, which I believe is a safe assumption to make. Alternatively we could use the task counter if a kcGlobalInstanceId has been set, or use a UUID if not

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -77,11 +85,21 @@ public class SnowflakeSinkTask extends SinkTask {
...
+      new LoggerHandler(SnowflakeSinkTask.class.getName() + "_STATIC");
+  private LoggerHandler DYNAMIC_LOGGER;
+
+  private long taskStartTime;
 
   /** default constructor, invoked by kafka connect framework */
   public SnowflakeSinkTask() {
     // nothing
```
sfc-gh-japatel, line 99: nit: remove nothing

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskForStreamingIT.java
```diff
@@ -81,6 +97,116 @@ public void testSinkTask() throws Exception {
     sinkTask.stop();
   }
 
+  @Test
+  public void testMultipleSinkTaskWithLogs() throws Exception {
```
sfc-gh-japatel, line 101: awesome test!

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -77,11 +85,21 @@ public class SnowflakeSinkTask extends SinkTask {
...
 
-  private static final LoggerHandler LOGGER = new LoggerHandler(SnowflakeSinkTask.class.getName());
+  // the dynamic logger is intended to be attached per task instance. the instance id will be set
+  // during task start,
+  // however if it is not set, it falls back to the static logger
+  private static final LoggerHandler STATIC_LOGGER =
+      new LoggerHandler(SnowflakeSinkTask.class.getName() + "_STATIC");
+  private LoggerHandler DYNAMIC_LOGGER;
+
+  private long taskStartTime;
```
sfc-gh-japatel, line 95: can you re-arrange static and non static fields together?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -52,6 +52,8 @@
  * Snowflake via Sink service
  */
 public class SnowflakeSinkTask extends SinkTask {
+  public static final String TASK_INSTANCE_TAG_FORMAT = "SnowflakeSinkTask[ID:{}.{}.{}]";
```
sfc-gh-japatel, line 55: Can this go in LoggerHandler?

sfc-gh-japatel: Also lets add a comment here describing how this format looks like in real life. (May be like an example?)

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -62,7 +64,13 @@ public class SnowflakeSinkTask extends SinkTask {
...
   // account and execute queries
   private SnowflakeConnectionService conn = null;
-  private String id = "-1";
+
+  // tracks number of tasks the config wants to create
+  private String taskConfigId = "-1";
+  // tracks total number of tasks created in this kc instance
+  private static int taskCreationCount = 0;
+  // tracks number of times this task instance has been started
+  private int taskInstanceStartCount;
```
sfc-gh-japatel, line 73: I am a bit confused on the usage of these two. 
Since `taskCreationCount` is static variable, I am assuming it will keep on incrementing until JVM dies. When will this happen? 

taskInstanceStartCount - seems will be incremented every time after rebalance?

sfc-gh-japatel: I would recommend hitting few Connect REST APIs to verify correctness. 
There is a restart task API https://docs.confluent.io/platform/current/connect/references/restapi.html#tasks

And there are two more connector APIs: pause, stop

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -421,21 +463,44 @@ private KafkaRecordErrorReporter createKafkaRecordErrorReporter() {
...
         // Will occur in Connect runtimes earlier than 2.6
-        LOGGER.info("Kafka versions prior to 2.6 do not support the errant record reporter.");
+        this.DYNAMIC_LOGGER.info(
+            "Kafka versions prior to 2.6 do not support the errant record reporter.");
       }
     }
     return result;
   }
 
+  private String getTaskLoggingTag() {
```
sfc-gh-japatel, line 482: move this to logger handler?

sfc-gh-japatel (commented):
So good to see this changes and tests! Great stuff. will wait for your replies to approve. (Mostly testing with REST apis)

Also, looking at the image I feel `SNOWFLAKE_KAFKA_CONNECTOR` is redundant. 
And `SnowflakeSinkTask[ID:{}.{}.{}]` seems a bit convoluted to look at but once you know what is it, seems fine.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-rcheng: I'm confused, remove... the blank line? Not sure what 'remove nothing' means

sfc-gh-rcheng: never mind i see the nothing comment lol

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-rcheng: Added comment. However the instance tag in the LoggerHandler can be used for things other than the tasks, so I think that task instance tag specific strings should be kept in the task class

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-rcheng: taskCreationCount (now totalTaskCreationCount) is to show how many tasks have been created in the KC instance. I originally assumed that it would be cleared when KC is killed, but that's a bad assumption to make. So i've added extra calls in KC start/stop to reset it to 0.

taskInstanceStartCount (now taskOpenCount) is meant to count the number of times the task has been opened, to hopefully show the number of times this task has been rebalanced. I didn't realize that start is only called once, whereas open is called during rebalance (i believe?)

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-rcheng: answered in earlier comment, resolving this one to keep one thread

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-rcheng: Additionally the task instance tag should be kept here so we can emit it with telemetry as well

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
```diff
@@ -79,10 +79,13 @@ public SnowflakeSinkConnector() {
...
   public void start(final Map<String, String> parsedConfig) {
+    // ensure we start counting tasks at 0 for this instance
+    SnowflakeSinkTask.setTotalTaskCreationCount(0);
+
     Utils.checkConnectorVersion();
 
-    // initialize logging with correlationId
-    LoggerHandler.setCorrelationUuid(UUID.randomUUID());
+    // initialize logging with global instance Id
+    LoggerHandler.setConnectGlobalInstanceId(UUID.randomUUID());
```
sfc-gh-tzhang, line 88: Do we really need the full UUID? it seems to long and makes the log even harder to read

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -119,11 +147,22 @@ private SnowflakeSinkService getSink() {
    */
   @Override
   public void start(final Map<String, String> parsedConfig) {
-    long startTime = System.currentTimeMillis();
-    this.id = parsedConfig.getOrDefault(Utils.TASK_ID, "-1");
+    this.taskStartTime = System.currentTimeMillis();
+    long startTime = this.taskStartTime;
```
sfc-gh-tzhang, line 151: why creating a local variable is needed here?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -421,21 +473,35 @@ private KafkaRecordErrorReporter createKafkaRecordErrorReporter() {
...
   }
 
+  private String getTaskLoggingTag() {
+    int countThreshold = 999;
+
+    if (totalTaskCreationCount > countThreshold) {
+      this.DYNAMIC_LOGGER.warn(
+          "More than {} tasks have been created. Resetting to 0", countThreshold);
+      totalTaskCreationCount = 0;
+    }
```
sfc-gh-tzhang, line 499: why this is needed?

src/test/java/com/snowflake/kafka/connector/internal/TestUtils.java
```diff
@@ -16,20 +16,8 @@
...
-import static com.snowflake.kafka.connector.Utils.HTTP_PROXY_PORT;
-import static com.snowflake.kafka.connector.Utils.HTTP_PROXY_USER;
-import static com.snowflake.kafka.connector.Utils.HTTP_USE_PROXY;
-import static com.snowflake.kafka.connector.Utils.JDK_HTTP_AUTH_TUNNELING;
-import static com.snowflake.kafka.connector.Utils.SF_DATABASE;
-import static com.snowflake.kafka.connector.Utils.SF_SCHEMA;
-import static com.snowflake.kafka.connector.Utils.SF_URL;
-import static com.snowflake.kafka.connector.Utils.SF_USER;
+import static com.snowflake.kafka.connector.SnowflakeSinkTask.TASK_INSTANCE_TAG_FORMAT;
+import static com.snowflake.kafka.connector.Utils.*;
```
sfc-gh-tzhang, line 20: please update your intellij settings and avoid using * import

sfc-gh-tzhang (approved):
LGTM, my main comment is that we shouldn't need the entire UUID since the number of instance should be small

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-rcheng: This was created in the interest of reducing log size if the number got too large. I can remove this if we want

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-rcheng: it is unneeded, removed. thanks for the catch

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
sfc-gh-rcheng: @sfc-gh-japatel  and I discussed this before, we'd definitely prefer a way to guarantee uniqueness with a shorter UUID, but we didn't find a great way to achieve that. Open to ideas though!

src/test/java/com/snowflake/kafka/connector/internal/TestUtils.java
sfc-gh-rcheng: ugh I thought I did that already oops. Thanks for the catch! i've turned it off

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -119,11 +137,24 @@ private SnowflakeSinkService getSink() {
...
+
+    // setup logging
+    this.DYNAMIC_LOGGER.debug(
+        "Defining SnowflakeSinkTask instance tag to"
+            + " SnowflakeSinkTask[ID:{taskId}.{taskCreationCount}.{taskStartCount}], where taskId"
+            + " is pulled from the config, taskCreationCount is the total number of tasks created"
+            + " during this run of Snowflake Kafka Connector, and taskStartCount is the number of"
+            + " times this task has been started");
+
+    this.taskInstanceStartCount++;
```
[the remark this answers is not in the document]
sfc-gh-rcheng: start() is meant to be called once to set task configs, open() can be called multiple times, more specifically called during rebalance

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
sfc-gh-tzhang: Given the number of instance being small, i think we will never get a collision if we just take the last section, this is like the chance of a hash collision....Or we could take something from the current timestamp to even lower the chance...

sfc-gh-rcheng: I changed it to hashcode on a UUID and the current time in millis

PR #498, opened 2022-09-28
Fix security vulnerabilities and upgrade KC version to 1.6.9

sfc-gh-rcheng:


sfc-gh-japatel:
lgtm! push only first commit to master..

sfc-gh-japatel (approved):

pom.xml
```diff
@@ -12,7 +12,7 @@
 
     <groupId>com.snowflake</groupId>
     <artifactId>snowflake-kafka-connector</artifactId>
-    <version>1.6.8</version>
+    <version>1.6.9</version>
```
sfc-gh-japatel, line 15: There is some utils.java file as well which uses hardocded 1.6.8, lets fix that too

sfc-gh-rcheng: Ah i was looking in xml files only, thanks for the catch!

sfc-gh-japatel (approved):

PR #500, opened 2022-10-10
New beta version for an already existing 1.8.1

sfc-gh-japatel:
- To release 1.8.1-beta. 

Decided to use 1.8.1-beta instead of 1.8.2-beta since there is no change at all. So better to keep 1.8.1 still but add a beta in the end.


sfc-gh-tzhang (approved):
Maybe it's just me, but I feel like `1.8.1-beta` adds a bit of confusion because people may think `1.8.1` comes after `1.8.1-beta`, so `1.8.2-beta` seems better, but I will the decision to you, thanks!

sfc-gh-rcheng (approved):

sfc-gh-japatel:
> Maybe it's just me, but I feel like `1.8.1-beta` adds a bit of confusion because people may think `1.8.1` comes after `1.8.1-beta`, so `1.8.2-beta` seems better, but I will the decision to you, thanks!

True that! Ahhhhhhh

PR #507, opened 2022-10-28
Include task tag in task telemetry

sfc-gh-rcheng:
Logging has a task tag that will also be added to each task telemetry event

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
@@ -148,6 +148,7 @@ public class SnowflakeSinkTask extends SinkTask {
   @Override
   public void start(final Map<String, String> parsedConfig) {
     this.taskStartTime = System.currentTimeMillis();
+    String taskTag = this.getTaskTag();
 
     // connector configuration
     this.taskConfigId = parsedConfig.getOrDefault(Utils.TASK_ID, "-1");
@@ -159,7 +160,7 @@ public class SnowflakeSinkTask extends SinkTask {
             + " is the number of times this task has been opened and totalTaskCreationCount is the"
             + " total number of tasks created during this run of Snowflake Kafka Connector");
 
-    this.DYNAMIC_LOGGER.setLoggerInstanceTag(this.getTaskLoggingTag());
+    this.DYNAMIC_LOGGER.setLoggerInstanceTag(taskTag);
 
     this.DYNAMIC_LOGGER.info("starting task...");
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryService.java
```diff
@@ -18,6 +18,8 @@ public interface SnowflakeTelemetryService {
    */
   void setTaskID(String taskID);
 
+  void setTelemetryTag(String telemetryTag);
```
sfc-gh-japatel, line 21: nit: java docs

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -488,7 +490,7 @@ private KafkaRecordErrorReporter createKafkaRecordErrorReporter() {
     return result;
   }
 
-  private String getTaskLoggingTag() {
+  private String getTaskTag() {
```
sfc-gh-japatel, line 493: nit: shuold this name be more specific? `getKafkaConnectTaskTag`

src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryServiceFactory.java
```diff
@@ -40,6 +40,15 @@ public SnowflakeTelemetryServiceBuilder setTaskID(String taskID) {
       return this;
     }
 
+    /**
```
sfc-gh-japatel, line 43: nit: one line comment here.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -148,6 +148,7 @@ private SnowflakeSinkService getSink() {
   @Override
   public void start(final Map<String, String> parsedConfig) {
     this.taskStartTime = System.currentTimeMillis();
+    String taskTag = this.getTaskTag();
```
sfc-gh-japatel, line 151: make this `final`

sfc-gh-japatel (approved):
couple of suggestions but otherwise lgtm, thanks! 

Also, google jfmt is failing, lets make sure it is fixed before pushing!

sfc-gh-rcheng:
We don't currently have task level telemetry, only connector level telemetry. I believe task level telemetry should be added, but it is out of scope for this pr, so i'll abandon this and create a corresponding jira. thanks for the review though jay!

PR #508, opened 2022-11-01
Stop Kafka Connector when corresponding table is dropped

sfc-gh-rcheng:
Current behavior: after the target table is dropped, every unused task will automatically recreate the table. This means that dropping a table is not idempotent and that it will be recreated as many times as there are tasks

Goal: prevent the target table from being recreated and stop Kafka Connector if possible

Change: add a boolean that signals if other tasks should be reopened or not and check boolean before opening a task. Unfortunately there is no good way to stop KC since that would be called from the Kafka Connect side, so the best I could figure out was to prevent future tasks from opening

[Current behavior logs](https://github.com/snowflakedb/snowflake-kafka-connector/files/9914322/3.reprowith5sinktask.txt)

[PR change logs](https://github.com/snowflakedb/snowflake-kafka-connector/files/9914321/kafka-connector.txt)


sfc-gh-rcheng:
Currently this change prevents tasks from reopening, however a better solution would be to stop the Snowflake Kafka Connector entirely. Since that logic is handled in Kafka Connector, I wasn't able to figure out a way to stop it entirely. Open to other ideas

sfc-gh-rcheng:
KC logs:
[kafka-connector.txt](https://github.com/snowflakedb/snowflake-kafka-connector/files/9914316/kafka-connector.txt)

sfc-gh-tzhang (changes requested):
This change seems a bit hacky, @sfc-gh-japatel any idea on how we could stop the task from restarting? Also we could have tasks against different tables, this change seems to block all tasks from running?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -369,6 +377,16 @@ public Map<TopicPartition, OffsetAndMetadata> preCommit(
               committedOffsets.put(topicPartition, new OffsetAndMetadata(offSet));
             }
           });
+    } catch (SFException e) {
+      this.DYNAMIC_LOGGER.error("PreCommit SFException: {}", e.getMessage());
+
+      // if open channel failure, throw exception to kill kc
+      if (e.getVendorCode().equals(ErrorCode.OPEN_CHANNEL_FAILURE.getMessageCode())) {
```
sfc-gh-tzhang, line 384: open channel failure could because of something transient so we would want to retry at least a few times?

sfc-gh-rcheng: the retry should be handled lower down in the fallback api, though ill go through and verify that

sfc-gh-japatel:
Take a look at https://kafka.apache.org/20/javadoc/org/apache/kafka/connect/errors/ConnectException.html
We can extend this with TableDeletedException and rethrow this if we see some specific error code.

sfc-gh-japatel (changes requested):
+1 with Toby, seems a hacky way. Let's kill the connector with exception. Also, can we add a test to verify the behavior?

sfc-gh-rcheng:
Did a deep dive into Kafka code. It looks like even throwing a connect exception will not help as tasks catch every exception and simply rebalance. I'll post on a couple forums or look through other connectors for examples on how they stop kafka connector, but it seems like there is no other way to do this

sfc-gh-rcheng:
Discussed in a separate meeting. will solve this a different way

PR #509, opened 2022-11-02
SNOW-686528: Check schema evolution table option 

sfc-gh-tzhang:
- We always enable schema evolution for tables that are created by KC
- Check for schema evolution table option before updating the schema


sfc-gh-tjones (approved):

sfc-gh-japatel (approved):
lgtm

PR #510, opened 2022-11-07
Create release v1.8.2 

sfc-gh-rcheng:
Created v1.8.2 off v1.8.1, adding new changes in


sfc-gh-rcheng:
I'm planning on adding the correlationid in the next version. I added the code however I was unable to easily verify it in snowhouse so I deprioritized the work since it is not necessary to debug Securonix's issue.

PR #511, opened 2022-11-10
Create release 1.8.2

sfc-gh-rcheng:
Bump version 1.8.1 -> 1.8.2

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,7 +12,7 @@
 
     <groupId>com.snowflake</groupId>
     <artifactId>snowflake-kafka-connector</artifactId>
-    <version>1.8.1</version>
+    <version>1.8.2</version>
     <packaging>jar</packaging>
     <name>Snowflake Kafka Connector</name>
     <description>Snowflake Kafka Connect Sink Connector</description>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -478,4 +478,4 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
-</project>
\ No newline at end of file
+</project>
```
sfc-gh-tzhang, line 481: is this expected to be changed?

sfc-gh-tzhang (approved):

pom.xml
sfc-gh-rcheng: that was not, not sure why it got changed. however I don't see any harm in removing the whitespace at the end. leaving as is

pom.xml
```diff
@@ -12,7 +12,7 @@
 
     <groupId>com.snowflake</groupId>
     <artifactId>snowflake-kafka-connector</artifactId>
-    <version>1.8.1</version>
+    <version>1.8.2</version>
```
sfc-gh-japatel, line 15: `Utils.java` `VERSION` also needs to be changed

PR #515, opened 2022-11-22
[SNOW 696506] Add telemetry to show how much data is ingested

sfc-gh-rcheng:
Emits telemetry in buckets showing how much data in bytes is ingested. It is batched to reduce telemetry traffic

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryServiceV1.java
@@ -148,7 +148,11 @@ public class SnowflakeTelemetryServiceV1 extends EnableLogging
   @Override
   public void reportInsertRowsBatchCount(final int recordCount, final long bufferSizeBytes) {
     if (recordCount < 0 || bufferSizeBytes < 0) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryServiceV1.java
```diff
@@ -26,6 +26,7 @@
 public class SnowflakeTelemetryServiceV1 extends EnableLogging
     implements SnowflakeTelemetryService {
   private static final ObjectMapper MAPPER = new ObjectMapper();
+  private static final long INSERTROWS_MIN_BATCH_SIZE = 500;
```
sfc-gh-rcheng, line 29: This was a random number i decided, but its probably too low / sends too many events to the telemetry service. Open to higher numbers

sfc-gh-rcheng: Batching on a timeframe may be better (number of events emitted every 5 min, etc)

sfc-gh-tzhang:
> Emits telemetry in buckets showing how much data in bytes is ingested. It is batched to reduce telemetry traffic. I've asked around but no one has a good estimate of how many events we can send to the telemetry service before we overload it. image

Could you explain what we're trying to monitor and what's the use case here? The reason I'm asking is because the rows are NOT guaranteed to be ingested if you simply add it in `getData`, it's still possible encounter failures later

PR #521, opened 2022-12-22
V1.8.3 release

sfc-gh-tzhang:
- Upgrade to 1.8.3 from 1.8.2 that includes some fixes for schematization
- Upgrade the SDK version to beta.7 for confluent


pom_confluent.xml
```diff
@@ -368,7 +368,7 @@
         <dependency>
             <groupId>net.snowflake</groupId>
             <artifactId>snowflake-ingest-sdk</artifactId>
-            <version>1.0.2-beta.4</version>
+            <version>1.0.2-beta.7</version>
```
sfc-gh-japatel, line 371: was there some perf regression in 1.0.2-beta.5? I saw some jira from @sfc-gh-lsembera

sfc-gh-japatel (approved):
barring a comment, lgtm!

pom_confluent.xml
sfc-gh-tjones: Yes, there was a perf regression relating to timestamps.

sfc-gh-tjones (approved):

pom_confluent.xml
sfc-gh-tzhang: Yes, but beta.5-beta.7 also contain a bunch of corruption/wrong result fixes, so I think it would be good to upgrade since correctness is more important than performance (for example: https://snowflakecomputing.atlassian.net/browse/SNOW-718273 shows we're hitting it in SFCTEST0)

PR #524, opened 2023-01-15
[SNOW-726924] Add jvm.nonProxy.hosts Parameter and update JDBC to 3.13.23

sfc-gh-wfateem:
The feature adds a new parameter `jvm.nonProxy.hosts` that can be configured to bypass the proxy server for connections to specific hosts. This is the equivalent of setting the JVM argument -Dhttp.nonProxyHosts, but in some situations, a user might not have the ability to configure the JVM arguments directly. For example, if they're running the connector in a managed service like AWS MSK Connect.
In order for this feature to work, the JDBC driver version has to be upgraded because there was a defect where it ignored the http.nonProxyHosts argument.

NOTE: If *.snowflakecomputing.com is added to the list of nonProxyHosts then the connector will fail. The parameter is ignored by the ingest-sdk as of v0.10.8, so if you need to bypass the proxy server to get your connection to the snowflakecomputing.com host to work, then this feature won't help. It's safe to add any other hosts to the list.


sfc-gh-wfateem:
Closing this PR. Opening a new one based on main branch.

PR #531, opened 2023-02-01
NO-SNOW: match pom_confluent.xml with pom.xml

sfc-gh-tzhang:


pom.xml
```diff
@@ -242,9 +242,6 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <version>2.9.1</version>
-                <configuration>
-                    <detectJavaApiLink>false</detectJavaApiLink>
-                </configuration>
```
sfc-gh-tzhang, line 247: @sfc-gh-japatel did you add this intentionally in pom.xml but not pom_confluent.xml?

pom_confluent.xml
```diff
@@ -427,6 +427,7 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/io.confluent/kafka-connect-avro-data -->
```
sfc-gh-rcheng, line 430: thanks for adding this in

sfc-gh-rcheng (approved):
Overall, do we want to collapse all our pom.xml files into one? Not sure why we need multiple

pom_confluent.xml
sfc-gh-xhuang: +1

pom.xml
sfc-gh-japatel: It was few years ago but based on this: https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#detectJavaApiLink

I might have seen some warnings while building this repo. May be lets keep it as is or add it to confluent_pom?

sfc-gh-japatel (approved):
lgtm

pom.xml
sfc-gh-tzhang: I don't see any related warning, you probably added as a workaround for some old issue, like https://issues.jenkins.io/browse/JENKINS-55692?src=confmacro. I feel like we should just use the default which is ture.

pom_confluent.xml
```diff
@@ -7,8 +7,7 @@
 
 <project xmlns="http://maven.apache.org/POM/4.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
-                      https://maven.apache.org/xsd/maven-4.0.0.xsd">
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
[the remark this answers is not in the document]
sfc-gh-tzhang: i'm not sure, i just updated the format :(

PR #533, opened 2023-02-02
[SNOW-726924] Add New jvm.nonProxy.hosts Parameter and update JDBC to 3.13.23

sfc-gh-wfateem:
The jvm.nonProxy.hosts parameter can be added to the connector's configuration file and is the equivalent to the JVM argument -Dhttp.nonProxyHosts. If a user provides that parameter and the connector runs in a JVM where -Dhttp.nonProxyHosts was set, then we will merge the two lists of hosts. In the future, we'll want another method of adding proxy configurations without resorting to System.setProperty() calls

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -402,7 +402,7 @@
         <dependency>
             <groupId>net.snowflake</groupId>
             <artifactId>snowflake-jdbc</artifactId>
-            <version>3.13.14</version>
+            <version>3.13.23</version>
         </dependency>
 
         <!-- https://mvnrepository.com/artifact/io.dropwizard.metrics/metrics-core -->
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -78,6 +78,7 @@ public class SnowflakeSinkConnectorConfig {
   private static final String PROXY_INFO = "Proxy Info";
   public static final String JVM_PROXY_HOST = "jvm.proxy.host";
   public static final String JVM_PROXY_PORT = "jvm.proxy.port";
+  public static final String JVM_NON_PROXY_HOSTS = "jvm.nonProxy.hosts";
   public static final String JVM_PROXY_USERNAME = "jvm.proxy.username";
   public static final String JVM_PROXY_PASSWORD = "jvm.proxy.password";
 
@@ -303,13 +304,23 @@ public class SnowflakeSinkConnectorConfig {
             ConfigDef.Width.NONE,
             JVM_PROXY_PORT)
         .define(
+            JVM_NON_PROXY_HOSTS,
+            Type.STRING,
+            "",
+            Importance.LOW,
+            "JVM option: http.nonProxyHosts",
+            PROXY_INFO,
+            2,
+            ConfigDef.Width.NONE,
+            JVM_NON_PROXY_HOSTS)
+        .define(
             JVM_PROXY_USERNAME,
             Type.STRING,
             "",
             Importance.LOW,
             "JVM proxy username",
             PROXY_INFO,
-            2,
+            3,
             ConfigDef.Width.NONE,
             JVM_PROXY_USERNAME)
         .define(
@@ -319,7 +330,7 @@ public class SnowflakeSinkConnectorConfig {
             Importance.LOW,
             "JVM proxy password",
             PROXY_INFO,
-            3,
+            4,
             ConfigDef.Width.NONE,
             JVM_PROXY_PASSWORD)
         // Connector Config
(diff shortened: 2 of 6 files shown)


sfc-gh-japatel (commented):
Can you add test in `ConnectorConfigTest`?

Also, please run java formatter before merging.

pom.xml
```diff
@@ -402,7 +402,7 @@
         <dependency>
             <groupId>net.snowflake</groupId>
             <artifactId>snowflake-jdbc</artifactId>
-            <version>3.13.14</version>
+            <version>3.13.23</version>
```
sfc-gh-japatel, line 405: please add it to pom_confluent.xml as well

sfc-gh-japatel (changes requested):
Couple of comments, will wait for one more revision. Thank you for the contribution

sfc-gh-wfateem:
@sfc-gh-japatel I wasn't too sure what was the right way to reformat the code, so I just used the format.sh script from the snowflake-ingest-java repo.

pom.xml
sfc-gh-tzhang, line 405: Have you try removing this dependency because we already have it in the ingest SDK?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -302,14 +303,24 @@ static ConfigDef newConfigDef() {
             1,
             ConfigDef.Width.NONE,
             JVM_PROXY_PORT)
+        .define(
```
sfc-gh-tzhang, line 306: is there any issue since we update the orderInGroup?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -272,6 +278,17 @@ static boolean enableJVMProxy(Map<String, String> config) {
       System.setProperty(HTTPS_PROXY_HOST, host);
       System.setProperty(HTTPS_PROXY_PORT, port);
 
+      // If the user provided the jvm.nonProxy.hosts configuration then we
+      // will append that to the list provided by the JVM argument
+      // -Dhttp.nonProxyHosts and not override it altogether, if it exists.
+      if (nonProxyHosts != null) {
+        nonProxyHosts =
+            (System.getProperty(HTTP_NON_PROXY_HOSTS) != null)
+                ? System.getProperty(HTTP_NON_PROXY_HOSTS) + "|" + nonProxyHosts
```
sfc-gh-tzhang, line 287: will there be any issue if there're duplicates?

sfc-gh-tzhang:
> @sfc-gh-japatel I wasn't too sure what was the right way to reformat the code, so I just used the format.sh script from the snowflake-ingest-java repo.

running format.sh should be sufficient, the gate already passed with your latest commit

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-wfateem: @sfc-gh-tzhang no, it's safe.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-wfateem: I will need @sfc-gh-japatel to comment on that.

sfc-gh-tzhang: If you're not sure, we can always add it at the end of this group?

sfc-gh-tzhang (approved):
LGTM, thanks!

pom.xml
sfc-gh-wfateem: Yes. I have and it works, but I felt that it would be cleaner to have it as part of another commit that's intentional on removing it and to document why. It also becomes kind of a mystery for someone to figure out what JDBC version we're using in the Kafka connector by simply glancing over the pom.xml file. Not sure, what do you think?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-wfateem: Sorry, I totally misunderstood what you were saying here. 
No, I'm pretty confident this isn't an issue. I initially looked at the Kafka docs to understand what was going on here in the code, and this isn't too important. It's just a way to order our parameters when they're displayed. I'm guessing things like Kafka Conluent Control Center will use this info to display a sink connector's properties in the UI in the same order that we defined them in. It's why I specifically inserted this new parameter in the middle. In my mind it made sense to have it come after proxy host and port, rather than at the end after proxy username and password.

sfc-gh-tjones (approved):

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-japatel: +1, no issues on changing the order, it is used in confluent UI

sfc-gh-japatel (approved):
lgtm! Thank you

PR #540, opened 2023-02-11
Fix client off by one issue

sfc-gh-rcheng:
Client creation logic creates one extra client due to the way the for loop is configured. Made test stricter by exposing the client count.

And fix an IT test that was missed in the last PR.

Thanks @sfc-gh-japatel for catching both of these!

PR #541, opened 2023-02-11
SNOW-740327 Create config for using Arrow BDEC file format + Bump Client SDK 1.0.3-beta

sfc-gh-japatel:
**NOTE**
- Please use this config in extreme scenarios where your workload is failing with default PARQUET as file format. 
- For an end user, it should ideally doesnt matter but this is a backup plan. 
- We don't plan to give this option in PuPr. 

**Other details**
- Also bumps up the latest IngestSDK - 1.0.3-beta
- Added tests
- Defaults to PARQUET - [Set in Ingest SDK]
- In extreme cases, we will ask customers to set this config. 
  - I don't think we need to update documentation for this, please let me know if you feel otherwise.


sfc-gh-japatel:
I will wait to push #540 first.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -112,6 +113,13 @@ public class SnowflakeSinkConnectorConfig {
   public static final String INGESTION_METHOD_DEFAULT_SNOWPIPE =
       IngestionMethodConfig.SNOWPIPE.toString();
 
+  // This is the streaming bdec file type which can be defined in config, default to Parquet
+  public static final String SNOWPIPE_STREAMING_FILE_TYPE = "snowflake.streaming.file.type";
+
+  // default file type to parquet
+  public static final String SNOWPIPE_STREAMING_FILE_TYPE_DEFAULT =
+      SnowpipeStreamingFileType.PARQUET.toString();
```
sfc-gh-tjones, line 121: Should we use write version numbers that we use internally for the Snowpipe Streaming Client SDK?

pom.xml
```diff
@@ -317,7 +317,7 @@
         <dependency>
             <groupId>net.snowflake</groupId>
             <artifactId>snowflake-ingest-sdk</artifactId>
-            <version>1.0.2-beta.7</version>
```
sfc-gh-rcheng, line 320: can we add this version change in the title or something? just thinking in terms of rel notes, we'd wanna call out this upgrade. but afaik rel notes only pulls the titles of each commit?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-rcheng: Can we add a comment about this mapping? Or even a static enum map or something similar. I'm not entirely sure what the version numbers are or how they correspond to the connector

src/main/java/com/snowflake/kafka/connector/internal/ingestsdk/KcStreamingIngestClient.java
```diff
@@ -60,14 +60,17 @@ public KcStreamingIngestClient(SnowflakeStreamingIngestClient client) {
    * <p>Any exceptions will be passed up, a sfexception will be converted to a connectexception
    *
    * @param streamingClientProps the properties for the client
+   * @param parameterOverrides Helps to override any default parameters in streaming client
    * @param clientName the client name to uniquely identify the client
    */
-  protected KcStreamingIngestClient(Properties streamingClientProps, String clientName) {
+  protected KcStreamingIngestClient(
+      Properties streamingClientProps, Map<String, Object> parameterOverrides, String clientName) {
```
sfc-gh-rcheng, line 67: could add the streaming version number to parquet/arrow enum or map here or in the ingestSdkProvider. (I'm assuming its like { (0 -> arrow), (1 -> parquet) }? If not ignore this comment lol)

src/main/java/com/snowflake/kafka/connector/internal/ingestsdk/StreamingClientManager.java
```diff
@@ -52,6 +56,17 @@ protected StreamingClientManager() {
     this.clientCount = -1; // will be incremented when a client is created
   }
 
+  /** Resets existing streaming clients and get a new client manager used in testing. */
+  public static StreamingClientManager resetAndGetEmptyStreamingClientManager() {
```
sfc-gh-rcheng, line 60: code coverage? this file has 100% but I don't see a corresponding test in StreamingClientManagerTest

src/main/java/com/snowflake/kafka/connector/internal/ingestsdk/StreamingClientManager.java
```diff
@@ -72,25 +87,42 @@ public void createAllStreamingClients(
...
     clientProperties.putAll(
         StreamingUtils.convertConfigForStreamingClient(new HashMap<>(connectorConfig)));
 
+    Map<String, Object> parameterOverrides = new HashMap<>();
+    // default file type is PARQUET defined in Ingest SDK dependency
+    String snowpipeStreamingBdecFileType =
+        connectorConfig.getOrDefault(
+            SnowflakeSinkConnectorConfig.SNOWPIPE_STREAMING_FILE_TYPE,
+            SnowpipeStreamingFileType.PARQUET.toString());
+    if (snowpipeStreamingBdecFileType.equalsIgnoreCase(
```
sfc-gh-rcheng, line 101: i think this doesn't have code coverage? didn't run it locally, resolve if it is covered

src/main/java/com/snowflake/kafka/connector/internal/ingestsdk/KcStreamingIngestClient.java
sfc-gh-rcheng: never mind i see the enum at the bottom

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskForStreamingIT.java
```diff
@@ -73,7 +73,8 @@ public void setup() {
   public void after() throws Exception {
     TestUtils.dropTable(topicName);
     IngestSdkProvider.setStreamingClientManager(
-        new StreamingClientManager(new HashMap<>())); // reset to clean initial manager
+        StreamingClientManager
+            .resetAndGetEmptyStreamingClientManager()); // reset to clean initial manager
```
sfc-gh-rcheng, line 77: nice! this cleans it up a lot, lets separate test and prod code by moving it to TestUtils pls

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-rcheng: I agree, i think using the file format name in the connector and adding a conversion enum as jay did below is more readable. Does add a little bit more dependency in testing for the conversion, but worth imo

pom.xml
sfc-gh-japatel: Yes, can add it in title. Thanks for call out. [Have added in description though]

src/main/java/com/snowflake/kafka/connector/internal/ingestsdk/StreamingClientManager.java
sfc-gh-japatel: Ack, will do!

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowpipeStreamingFileType.java
```diff
@@ -0,0 +1,76 @@
...
+   */
+  public Constants.BdecVersion getBdecVersion() {
+    return bdecVersion;
+  }
+
+  /**
+   * Validator to validate {@link
+   * com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig#SNOWPIPE_STREAMING_FILE_TYPE}
+   */
+  public static final ConfigDef.Validator VALIDATOR =
```
sfc-gh-rcheng, line 37: can we add code coverage for this? feels weiiirdd to add validation for a validator, but i'd rather a test fail than the entire test suite fail on this validator if the file type ever changes under us

src/test/java/com/snowflake/kafka/connector/ConnectorConfigTest.java
```diff
@@ -622,4 +623,44 @@ public void testValidSchematizationForStreamingSnowpipe() {
     config.put(Utils.SF_ROLE, "ACCOUNTADMIN");
     Utils.validateConfig(config);
   }
+
+  @Test(expected = SnowflakeKafkaConnectorException.class)
```
sfc-gh-rcheng, line 627: Can we put this in a new test file in /internal/streaming/SnowpipeStreamingFileTypeTest or something similar? since (i think?) this tests the SnowpipeStreamingFileType.VALIDATOR. if not, we should add additional tests for the validator

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskForStreamingIT.java
sfc-gh-japatel: Ack!

src/main/java/com/snowflake/kafka/connector/internal/ingestsdk/StreamingClientManager.java
sfc-gh-japatel: Have a test case, resolving it.

src/test/java/com/snowflake/kafka/connector/ConnectorConfigTest.java
sfc-gh-japatel: IMO, all of these are all tested only through validate logic, so no need to put it in its native file.

sfc-gh-tzhang (commented):
Thanks for making the change! My main concern is the limitation of the current approach:
- If we're adding a new version, additional changes are required in KC

Given that this should be only used with our recommendation, I would suggest to remove some of the validation and make it more flexible. We will rely on the validation in the client SDK if needed.

sfc-gh-rcheng:
@sfc-gh-tzhang 's comment "We will rely on the validation in the client SDK if needed." (still not entirely sure how to reply to that comment?)

I don't think we should rely on the client for validation. Especially when we move to the rowset api, KC will be the only thing running on the customers environment, so I think it should validate and operate independently from the client

src/main/java/com/snowflake/kafka/connector/internal/ingestsdk/StreamingClientManager.java
```diff
@@ -52,6 +56,17 @@ protected StreamingClientManager() {
     this.clientCount = -1; // will be incremented when a client is created
   }
 
+  /** Resets existing streaming clients and get a new client manager used in testing. */
```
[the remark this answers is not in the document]
sfc-gh-japatel: moved it to `TestUtils`

src/main/java/com/snowflake/kafka/connector/internal/ingestsdk/StreamingClientManager.java
```diff
@@ -72,25 +87,42 @@ public void createAllStreamingClients(
...
     Properties clientProperties = new Properties();
     clientProperties.putAll(
         StreamingUtils.convertConfigForStreamingClient(new HashMap<>(connectorConfig)));
 
+    Map<String, Object> parameterOverrides = new HashMap<>();
+    // default file type is PARQUET defined in Ingest SDK dependency
+    String snowpipeStreamingBdecFileType =
+        connectorConfig.getOrDefault(
+            SnowflakeSinkConnectorConfig.SNOWPIPE_STREAMING_FILE_TYPE,
+            SnowpipeStreamingFileType.PARQUET.toString());
```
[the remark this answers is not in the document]
sfc-gh-japatel: Yes, we dont pass anything to the client generator if nothing is set on KC side. (i.e use the default in ingest sdk itself)

the `parameterOverrides` is populated only when arrow is set!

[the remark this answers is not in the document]
sfc-gh-japatel: I will add a comment here

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-japatel: I agree with overall idea behind keeping same versions as of ingest SDK here. 
But here is my overall stance on why I like to decouple the behavior of ingest SDK and KC. 

1. Ingest SDK is just another dependency on SF Kafka Connector. Ingest SDK will evolve over time and we cant be tightly coupled with changes or changes coming into future. 
2. New file type in Ingest SDK would mean we will have a new version of it and KC will depend on it. The existing Unit and IT tests will help us catch any issues pertaining to the supported versions. 
3. Having to depend on ingest sdk would also mean, we will not have our own validators from KC side. 

Although, I also understand that both ingest sdk and KC is being developed by us so no harm is keeping it tightly coupled. 

Happy to make changes if folks like numbered versions here.

src/main/java/com/snowflake/kafka/connector/internal/ingestsdk/KcStreamingIngestClient.java
```diff
@@ -60,14 +60,17 @@ public KcStreamingIngestClient(SnowflakeStreamingIngestClient client) {
...
     try {
       LOGGER.info("Creating Streaming Client: {}", clientName);
       this.client =
           SnowflakeStreamingIngestClientFactory.builder(clientName)
               .setProperties(streamingClientProps)
+              .setParameterOverrides(parameterOverrides)
```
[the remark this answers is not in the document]
sfc-gh-japatel: May be I am missing something, but we can always add more to `parameterOverrides` from the config.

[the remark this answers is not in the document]
sfc-gh-tzhang: Correct, but that requires more changes in KC and a new release, right? I was thinking maybe we can just accept a json and pass to the SDK

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-tzhang: I was not suggesting that we should keep the two tightly coupled, I was just trying to suggest we should make this a more generic change so when a new type is added (or KC needs to override some settings, for example, for Confluent Cloud, it may have certain memory constraint that we need to limit), there is no additional change required other than bumping the release version

sfc-gh-tzhang:
> @sfc-gh-tzhang 's comment "We will rely on the validation in the client SDK if needed." (still not entirely sure how to reply to that comment?)
> 
> I don't think we should rely on the client for validation. Especially when we move to the rowset api, KC will be the only thing running on the customers environment, so I think it should validate and operate independently from the client

For rowset api, i think it will still go through the Ingest SDK, otherwise you need to build the HTTP request again in KC which is a duplication, right?

src/main/java/com/snowflake/kafka/connector/internal/ingestsdk/KcStreamingIngestClient.java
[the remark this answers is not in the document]
sfc-gh-japatel: I see, accepting json makes sense but it just gives more control for configuration, which in past IIRC was not something we encouraged. For example flush time. 

We can add json once we really need more knobs. I will pass this time since we don't anticipate this configuration change. 

I though agree having flexibility to modify config without kc new release. 
So I will make that change to use bdex versions directly.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-japatel: Ack, makes sense. Let me change to BDEC version directly but I am not entirely convinced about accepting json in this PR on in general overall.

sfc-gh-japatel:
@sfc-gh-tzhang @sfc-gh-tjones have made changes to take a simpler route with no enums and validations on KC side apart from allowed enum types. PTAL again!

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-tjones: Strong agree with Toby, let's map to BDEC version directly.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -112,6 +113,16 @@ public class SnowflakeSinkConnectorConfig {
   public static final String INGESTION_METHOD_DEFAULT_SNOWPIPE =
       IngestionMethodConfig.SNOWPIPE.toString();
 
+  // This is the streaming bdec file type which can be defined in config
+  public static final String SNOWPIPE_STREAMING_FILE_TYPE = "snowflake.streaming.file.type";
```
sfc-gh-tzhang, line 117: could we use `file.version` instead of `file.type`?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -474,6 +485,18 @@ static ConfigDef newConfigDef() {
             5,
             ConfigDef.Width.NONE,
             INGESTION_METHOD_OPT)
+        .define(
+            SNOWPIPE_STREAMING_FILE_TYPE,
+            Type.STRING,
+            SNOWPIPE_STREAMING_FILE_TYPE_DEFAULT,
```
sfc-gh-tzhang, line 491: It's hard to keep this in sync with the SDK (think about the case where we give the KC ownership to connector team in the future), I think we should just pass null here

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -474,6 +485,18 @@ static ConfigDef newConfigDef() {
...
+            BDEC_VERSION_VALIDATOR,
```
sfc-gh-tzhang, line 492: I don't think a validator is needed, we already have validation logic in the SDK, it will throw an exception if it's something else

src/main/java/com/snowflake/kafka/connector/internal/ingestsdk/StreamingClientManager.java
```diff
@@ -72,25 +83,46 @@ public void createAllStreamingClients(
...
-        numTasksPerClient);
 
     Properties clientProperties = new Properties();
     clientProperties.putAll(
         StreamingUtils.convertConfigForStreamingClient(new HashMap<>(connectorConfig)));
 
+    // Override only if bdec version is explicitly set in config, default to the version set inside
+    // Ingest SDK
+    Map<String, Object> parameterOverrides = new HashMap<>();
+    // default bdec version is THREE defined in Ingest SDK dependency [THREE corresponds to PARQUET]
```
sfc-gh-tzhang, line 94: I suggest we remove this comment because the default could change later

sfc-gh-tzhang (approved):
Let's remove the default and the validation since the SDK already has both and it will thrown an exception if the value is something else, see `getBlobFormatVersion`, otherwise LGTM!

sfc-gh-japatel:
> Let's remove the default and the validation since the SDK already has both and it will thrown an exception if the value is something else, see `getBlobFormatVersion`, otherwise LGTM!

I wasnt too sure if a null validator/empty validator would have worked, I will may be just use a string validator. Thanks for suggestion.

src/main/java/com/snowflake/kafka/connector/internal/ingestsdk/StreamingClientManager.java
sfc-gh-japatel: Ack .

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-japatel: Sounds good, will keep a string validator though.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-japatel: yes, let me change it

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-japatel: Null is not supported, passed `NO_DEFAULT_VALUE`

sfc-gh-tzhang:
@sfc-gh-japatel do you want to fix https://github.com/snowflakedb/snowflake-kafka-connector/pull/539 in the same PR or it's better to be a separate PR?

sfc-gh-japatel:
> @sfc-gh-japatel do you want to fix #539 in the same PR or it's better to be a separate PR?

I will do it in separate PR.

sfc-gh-japatel:
Unable to push this since the new dependency from ingest SDK (ingest sdk - 1.0.3-beta to jdbc version 3.13.18 gives put command failures in GCS]

I rolledback to older version and ran a test locally, there wasnt any issue. 

I will investigate more on this. 

Eventually, we dont want to rely on jdbc coming from ingest SDK since we have KC's own jdbc dependency.

sfc-gh-japatel:
Passing after pushing #546

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskTestForStreamingIT.java
```diff
@@ -71,8 +70,7 @@ public void after() throws Exception {
     this.sinkTask.stop();
     TestUtils.dropTable(topicName);
     IngestSdkProvider.setStreamingClientManager(
-        new StreamingClientManager(new HashMap<>())); // reset to clean initial manager
-    // TODO @rcheng: use jay's reset method when merged
```
sfc-gh-rcheng, line 75: thanks!

PR #543, opened 2023-02-13
Refactor sink task integration tests to be more readable

sfc-gh-rcheng:


sfc-gh-rcheng:
waiting on https://github.com/snowflakedb/snowflake-kafka-connector/pull/540 to merge first

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskForStreamingIT.java
```diff
@@ -41,109 +41,114 @@
...
-
-  @InjectMocks private SnowflakeSinkTask sinkTask1 = new SnowflakeSinkTask();
-
+  private List<TopicPartition> topicPartitions;
+  private SnowflakeSinkTask sinkTask;
+  private InMemorySinkTaskContext sinkTaskContext;
+  private List<SinkRecord> records;
+  private Map<TopicPartition, OffsetAndMetadata> offsetMap;
+
+  // sets up default objects required for sunny day tests
```
sfc-gh-japatel, line 54: is there some hidden logic behind this phrase `sunny day tests?`

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskForStreamingIT.java
```diff
@@ -41,109 +41,114 @@
...
+    this.sinkTask = new SnowflakeSinkTask();
+    this.sinkTaskContext =
+        new InMemorySinkTaskContext(Collections.singleton(this.topicPartitions.get(0)));
+    this.records = TestUtils.createJsonStringSinkRecords(0, 1, this.topicName, 0);
+    this.offsetMap = new HashMap<>();
+    this.offsetMap.put(this.topicPartitions.get(0), new OffsetAndMetadata(10000));
 
     IngestSdkProvider.getStreamingClientManager()
         .createAllStreamingClients(this.config, "testkcid", 2, 1);
+    assert IngestSdkProvider.getStreamingClientManager().getClientCount() == 2;
```
sfc-gh-japatel, line 72: this is fine for test but since partition count is only 1, should we also keep task count to 1? Just more relatable IMO.

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskForStreamingIT.java
```diff
@@ -160,35 +165,46 @@ public void testTaskToClientMapping() throws Exception {
     sinkTask1.stop();
   }
 
+  @Mock Logger logger = Mockito.mock(Logger.class);
+
+  @InjectMocks @Spy
+  private LoggerHandler loggerHandler = Mockito.spy(new LoggerHandler(this.getClass().getName()));
+
+  @InjectMocks private SnowflakeSinkTask sinkTask1 = new SnowflakeSinkTask();
+
```
sfc-gh-japatel, line 174: can you move this to the top?

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskForStreamingIT.java
sfc-gh-rcheng: oh oops, I've always used 'sunny day tests' as tests that test the ideal state of the code. Ignoring exceptions or anything like that. I'll change the comment, might just be a msft term

sunny day -> ideal test cases
rainy day -> exception or breaking test cases

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskForStreamingIT.java
sfc-gh-rcheng: These arent meant to be used in any other part of the test, so I moved them here for readability. They do have to be declared outside the method for mockito reasons.

Now that they're out here, i do see how its weird to have these variables declared mid test. I'll move them to the top and add a comment about it

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskForStreamingIT.java
sfc-gh-rcheng: yeah entirely right, not sure how i missed this lol

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskForStreamingIT.java
sfc-gh-japatel: Also, just realized we are mocking the behavior in IT, can we instead move this to Unit Tests? [Seems like an anti pattern to me]

And IMO, we dont need to mock any of the behaviors here, I think your test would work if you remove Spy and inject mocks. 

Other way would be to just move out the logs test to its own class - Sorry didnt realize this earlier in your logs changes.

sfc-gh-japatel (approved):
Barring one comment on mocking in IT test, lgtm! [Non blocking] 
It is debatable whether one should use mocks in IT test. There is no general guidelines here but here is one S/O on this discussion. https://stackoverflow.com/questions/52107522/is-it-in-considered-a-good-practice-to-mock-in-integration-test

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskForStreamingIT.java
sfc-gh-japatel: Got it, I thought you are `mocking` seattle weather here!

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskForStreamingIT.java
sfc-gh-rcheng: yeah it prob should be its own class. Wasn't sure where to stick it back then because I didnt see a SnowflakeSinkTaskForStreamingTest. Can't remove the Spy or InjectMocks, I spent way too long figuring it out because the logger is a static final object so this is a not great work around for that. At the same time its testing too many methods for me to comfortably call it a unit test, so maybe we need another 'category' of testing conceptually. regardless for this PR i'll move it to a SnowflakeSinkTaskTest file in the streaming folder

sfc-gh-rcheng: At the same time i dont think this test is strictly necessary, but i would like some form of logging testing i guess? Not that important but makes me feel better about any logging changes

sfc-gh-japatel: Ack, moving to its own class sounds good to me! Thanks

sfc-gh-japatel:
Also can you make sure tests pass in merge gates before pushing? I see failure in `testMultipleSinkTaskWithLogs`

sfc-gh-japatel:
Pushing this to fix the flakiness. I created a SNOW-744735 to reenable it.

PR #546, opened 2023-02-16
Modify pom.xml to use jdbc defined in connector and not from ingest sdk

sfc-gh-japatel:
Background:

- After upgrading ingest sdk to 1.0.3-beta, GCS tests started failing. 
- Above version uses jdbc 3.13.18

- We have our own native jdbc dependency at 3.13.23. 
- Intellij bytecode never showed code corresponding to 3.13.23 and always picked up from 3.13.18 (From Ingest SDK)

Root cause:
- Previously the jdbc dependency was defined after ingest SDK and APPARENTLY ORDERING MATTERS

Solution:
- Exclude jdbc from ingest SDK. 
- Define our own jdbc version in Connector. 

Next steps:
- Fix JDBC version in Ingest SDK to use 3.13.23 instead of 3.13.18

Few more changes:
- Use fail fast strategy to not kill other running tasks in matrix if one fails. We want all of them to run parallely. 
- Fix broken tests after allowed data type formats changed in latest version of ingest SDK. 

**MAVEN  DEPENDENCY  RANT.. SCREEEAMMMM**

### DIFF
--- a/.github/workflows/End2EndTestApache.yml
+++ b/.github/workflows/End2EndTestApache.yml
@@ -10,6 +10,7 @@ jobs:
   build_and_test:
     runs-on: ubuntu-20.04
     strategy:
+      fail-fast: false # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast
       matrix:
         apache_kafka_version: [ '2.5.1', '2.8.1', '3.2.1' ]
         snowflake_cloud: [ 'AWS', 'AZURE', 'GCS' ]
--- a/.github/workflows/End2EndTestConfluent.yml
+++ b/.github/workflows/End2EndTestConfluent.yml
@@ -10,6 +10,7 @@ jobs:
   build_and_test:
     runs-on: ubuntu-20.04
     strategy:
+      fail-fast: false # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast
       matrix:
         confluent_version: [ '5.5.11', '6.2.6', '7.2.1' ]
         snowflake_cloud: [ 'AWS', 'AZURE', 'GCS' ]
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -10,6 +10,7 @@ jobs:
   build_and_test:
     runs-on: ubuntu-20.04
     strategy:
+      fail-fast: false # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast
       matrix:
         snowflake_cloud: [ 'AWS', 'AZURE', 'GCS' ]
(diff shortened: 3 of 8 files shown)


sfc-gh-tjones (approved):

sfc-gh-tzhang:
> Background:
> 
> * After upgrading ingest sdk to 1.0.3-beta, GCS tests started failing.
> * Above version uses jdbc 3.13.18
> * We have our own native jdbc dependency at 3.13.23.
> * Intellij bytecode never showed code corresponding to 3.13.23 and always picked up from 3.13.18 (From Ingest SDK)
> 
> Root cause:
> 
> * Previously the jdbc dependency was defined after ingest SDK and APPARENTLY ORDERING MATTERS
> 
> Solution:
> 
> * Exclude jdbc from ingest SDK.
> * Define our own jdbc version in Connector.
> 
> Next steps:
> 
> * Fix JDBC version in Ingest SDK to use 3.13.23 instead of 3.13.18
> 
> **MAVEN DEPENDENCY RANT.. SCREEEAMMMM**

This is interesting, we have a similar issue reported by the connector folks after upgrading the JDBC to 3.13.18, what's the failure message? Do you know whether it only happens in GCS?

sfc-gh-japatel:
> > Background:
> > 
> > * After upgrading ingest sdk to 1.0.3-beta, GCS tests started failing.
> > * Above version uses jdbc 3.13.18
> > * We have our own native jdbc dependency at 3.13.23.
> > * Intellij bytecode never showed code corresponding to 3.13.23 and always picked up from 3.13.18 (From Ingest SDK)
> > 
> > Root cause:
> > 
> > * Previously the jdbc dependency was defined after ingest SDK and APPARENTLY ORDERING MATTERS
> > 
> > Solution:
> > 
> > * Exclude jdbc from ingest SDK.
> > * Define our own jdbc version in Connector.
> > 
> > Next steps:
> > 
> > * Fix JDBC version in Ingest SDK to use 3.13.23 instead of 3.13.18
> > 
> > **MAVEN DEPENDENCY RANT.. SCREEEAMMMM**
> 
> This is interesting, we have a similar issue reported by the connector folks after upgrading the JDBC to 3.13.18, what's the failure message? Do you know whether it only happens in GCS?

Yes, I only saw this in GCS. The error message is not very descriptive and I tried debugging JDBC with no luck. But it looks like `upload with presigned url failed.` I have asked ecosystem team about this.

sfc-gh-japatel:
All end to end tests are failing in apache kafka and confluent, I will continue to investigate this, ingest SDK was the only thing that was upgraded.

sfc-gh-japatel:
> All end to end tests are failing in apache kafka and confluent, I will continue to investigate this, ingest SDK was the only thing that was upgraded.

The issue is with the date `'APPROVAL_DATE': '15-Jun-2022',` column in `test_schema_mapping.py` test. 

Great test and love how it failed. 

CC: @sfc-gh-tzhang confirming this is expected behavior as I see this in change logs too.

PR #547, opened 2023-02-21
no-snow Add connector name in KC client name for better debugging the client …

sfc-gh-japatel:
…history view

- It becomes increasingly difficult from a user perspective to find out the client name since it contains a kcInstanceId which is a combination of timestamp and UUID. 
- Adding connector name to it helps customer narrow down their search. 

Added verification to existing tests.


sfc-gh-tjones (approved):
Thanks! This will make it a lot easier to track Snowpipe Streaming activity.

sfc-gh-japatel:
Tests failing in azure because of outage in snowpipe! #ESC-819 for enthusiasts

PR #551, opened 2023-02-24
[SNOW-748118] Remove delivery guarantee config param

sfc-gh-rcheng:
Streaming has exactly once delivery hard coded
Snowpipe does not support exactly once delivery anymore

Thus remove the config param and corresponding code/logic


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -452,14 +448,6 @@ static ConfigDef newConfigDef() {
             JMX_OPT_DEFAULT,
             ConfigDef.Importance.HIGH,
             "Whether to enable JMX MBeans for custom SF metrics")
-        .define(
```
sfc-gh-rcheng, line 455: removed from config

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -651,11 +636,8 @@ private void insert(final SinkRecord record) {
       }
       // only get offset token once when service context is initialized
       // ignore ingested files
-      // if ingestionDeliveryGuarantee is AT_LEAST_ONCE, ignore offsetPersistedInSnowflake
-      // else discard the record if the record offset is smaller or equal to server side offset
-      if ((ingestionDeliveryGuarantee
```
sfc-gh-rcheng, line 656: snowpipe doesnt support at least once anymore. remove this check since it is always false and in an OR

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -815,18 +797,13 @@ private long getOffset() {
...
-        } else {
-          this.offsetPersistedInSnowflake.set(Long.parseLong(offsetToken));
-        }
+      ingestionService.ingestFilesWithClientInfo(fileNamesCopy, this.clientSequencer.get());
+      String offsetToken = ingestionService.getClientStatus().getOffsetToken();
+      // Update server side offset
+      if (offsetToken == null) {
+        this.offsetPersistedInSnowflake.set(-1);
       } else {
-        ingestionService.ingestFiles(fileNamesCopy);
```
sfc-gh-rcheng, line 829: removed this else because delivery is always exactly once, so it should not fall into this else statement

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -815,18 +797,13 @@ private long getOffset() {
...
-        ingestionService.ingestFilesWithClientInfo(fileNamesCopy, this.clientSequencer.get());
-        String offsetToken = ingestionService.getClientStatus().getOffsetToken();
-        // Update server side offset
-        if (offsetToken == null) {
-          this.offsetPersistedInSnowflake.set(-1);
-        } else {
-          this.offsetPersistedInSnowflake.set(Long.parseLong(offsetToken));
-        }
+      ingestionService.ingestFilesWithClientInfo(fileNamesCopy, this.clientSequencer.get());
+      String offsetToken = ingestionService.getClientStatus().getOffsetToken();
```
sfc-gh-japatel, line 801: we wont be calling this API anymore, we should remove offsetPersistedInSnowflake field too.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -815,18 +797,13 @@ private long getOffset() {
...
+      // Update server side offset
+      if (offsetToken == null) {
+        this.offsetPersistedInSnowflake.set(-1);
       } else {
-        ingestionService.ingestFiles(fileNamesCopy);
+        this.offsetPersistedInSnowflake.set(Long.parseLong(offsetToken));
```
sfc-gh-japatel, line 806: lets remove both if else.

sfc-gh-japatel (changes requested):
Suggested few changes. 

can you also remove all IT and end to end tests (from test_verify) associated with exactly once? 

Thanks for driving this. 

CC: @sfc-gh-xhuang

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-rcheng: removed

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-rcheng: just curious - why are we not calling this API anymore? and why is the offsetPersistedInSnowflake field not necessary? I'm not entirely sure what the importance of this field is/was

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -455,11 +440,7 @@ private void init(long recordOffset) {
       // recover will only check pipe status and create pipe if it does not exist.
       recover(pipeCreation);
 
-      // when exactly_once is enabled,fetch clientSequencer and offsetPersistedInSnowflake
-      if (ingestionDeliveryGuarantee
-          == SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.EXACTLY_ONCE) {
-        initClientInfoForExactlyOnceDelivery();
-      }
+      initClientInfo();
```
sfc-gh-japatel, line 443: the whole idea behind snowpipe's exactly once api was to use streaming channel like persistence layer behind a snowpipe. 
TLDR: Store offset tokens in pipe's metadata using invisible channel object on server side. (Not putting a lot of information here since this is public repo)

Lets remove all occurrences of clientInfo and calling clientConfigure API. We will not need it.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -473,30 +454,9 @@ private void init(long recordOffset) {
      * Initialize the client info (clientSequencer and offsetPersistedInSnowflake) by calling
      * ingestion service API configureClient and getClientStatus
      */
-    private void initClientInfoForExactlyOnceDelivery() {
+    private void initClientInfo() {
       ConfigureClientResponse configureClientResponse = ingestionService.configureClient();
       this.clientSequencer.set(configureClientResponse.getClientSequencer());
```
sfc-gh-japatel, line 459: we will not need this anymore. We will solely rely on Snowpipe's basic API which doesnt pass in client info.

sfc-gh-japatel (commented):
there are few SF connectors in test/rest_request_template folder pertaining to atleast once and exactly once, can we also remove that? That will reduce end to end running time for few minutes!

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-rcheng: Lets discuss this during our 1:1 tomorrow, I dont have enough context to make these changes currently

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -770,126 +770,6 @@ public void testMissingPropertiesForStreamingClient() {
     }
   }
 
-  /* Service start -> Insert -> Close. service start -> fetch the offsetToken, compare and ingest check data */
-
-  @Test
-  public void testStreamingIngestionWithExactlyOnceSemanticsNoOverlappingOffsets()
```
sfc-gh-japatel, line 776: this is a streaming ingest test!

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -770,126 +770,6 @@ public void testMissingPropertiesForStreamingClient() {
...
-
-    assert service2.getOffset(new TopicPartition(topic, partition)) == offset + 1;
-
-    service2.closeAll();
-  }
-
-  /* Service start -> Insert -> Close. service start -> fetch the offsetToken, compare and ingest check data */
-
-  @Test
-  public void testStreamingIngestionWithExactlyOnceSemanticsOverlappingOffsets() throws Exception {
```
sfc-gh-japatel, line 837: same as above. lets keep it back.

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
sfc-gh-rcheng: should we keep some of the e2e tests as well? or are those situations redundant with streaming's tests?

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
sfc-gh-rcheng: reverted, thanks for the catch

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
sfc-gh-japatel: Snowpipe + Exactly once is totally independent from snowpipe streaming. 
There are exactly once tests for snowpipe streaming because we support that. 

We will no longer support snowpipe with exactly once. 

End to end tests which I was suggesting earlier was all ingestionmethod = snowpipe and delivery guarantee = exactly once

We still want to keep snowpipe streaming tests with exactly once (there is no atleast once in snowpipe streaming)

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -839,6 +764,8 @@ private long getOffset() {
           name ->
               pipeStatus.updateCommitLag(currentTime - FileNameUtils.fileNameToTimeIngested(name)));
 
+      ingestionService.ingestFiles(fileNamesCopy);
```
sfc-gh-japatel, line 767: lets put this above line 758

We want to update committedOffset only after ingestFiles API have succeeded. 

(This file is the only file which could have potential bugs, lets make sure we revert back to the same changes before exactly once change was introduced)

test/test_verify.py
```diff
@@ -588,12 +530,6 @@ def runTestSet(driver, testSet, nameSalt, enable_stress_test):
             errorExit("Unknown testSet option {}, please input confluent, apache or clean".format(testSet))
 
         execution(testSet, testSuitList1, testCleanEnableList1, testSuitEnableList1, driver, nameSalt)
-        ############################ round 1 ############################
-
-        print("Enable Delivery Guarantee tests:" + str(driver.enableDeliveryGuaranteeTests))
```
sfc-gh-japatel, line 593: awesome! less runtime!

sfc-gh-japatel:
Overall lgtm but will wait for one change about committedOffset and insertFile call

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-rcheng: moved the line.

good point, i just did a diff against the 1.6.9 tag, and excluding the log and partitionbuffer changes, the only major change is the 'shouldSkipNullValue' method.

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
sfc-gh-rcheng: sweet, thanks for clearing that up

sfc-gh-rcheng:
> Thanks @sfc-gh-rcheng ! I believe there're changes in the Ingest SDK are also related to exactly once, could you help with removing them as well?

@sfc-gh-japatel and i had some concerns over if customers are relying on that API? Should we do a release with a warning about deprecating those api calls, or can we just remove them?

sfc-gh-japatel:
> > Thanks @sfc-gh-rcheng ! I believe there're changes in the Ingest SDK are also related to exactly once, could you help with removing them as well?
> 
> @sfc-gh-japatel and i had some concerns over if customers are relying on that API? Should we do a release with a warning about deprecating those api calls, or can we just remove them?

We can find usage of this APIs and based on usage we can just remove it since it was just PrPr. But Xin might have better suggestions.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -651,12 +596,8 @@ private void insert(final SinkRecord record) {
       }
       // only get offset token once when service context is initialized
       // ignore ingested files
-      // if ingestionDeliveryGuarantee is AT_LEAST_ONCE, ignore offsetPersistedInSnowflake
-      // else discard the record if the record offset is smaller or equal to server side offset
-      if ((ingestionDeliveryGuarantee
-                  == SnowflakeSinkConnectorConfig.IngestionDeliveryGuarantee.AT_LEAST_ONCE
-              || record.kafkaOffset() > this.offsetPersistedInSnowflake.get())
-          && record.kafkaOffset() > processedOffset.get()) {
+      // discard the record if the record offset is smaller or equal to server side offset
```
sfc-gh-japatel, line 599: nit: this comment in no longer valid

sfc-gh-japatel (approved):
lgtm, please remove the comment. Otherwise LGTM.

sfc-gh-rcheng:
Pretty big changes in master, merged as well as I could. Please take a look to double check I'm not removing anything important!

sfc-gh-tjones (approved):

sfc-gh-japatel (approved):
lgtm

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeErrors.java
```diff
@@ -189,9 +189,6 @@ public enum SnowflakeErrors {
       "Failed to create pipe",
       "User may have insufficient privileges. If this persists, please "
           + "contact Snowflake support. "),
-  ERROR_3006("3006", "Failed to configure client status", "Exception reported by Ingest SDK"),
-  ERROR_3007("3007", "Failed to get client status", "Exception reported by Ingest SDK"),
-  ERROR_3008("3008", "Failed to ingest file with client info", "Exception reported by Ingest SDK"),
```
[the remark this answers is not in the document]
sfc-gh-rcheng: added back in, thanks!

Issue #553, opened 2023-02-28
JWT token is invalid, task failure after upgrading to 1.9.0


acristu:
After upgrading to 1.9.0, after about ~1h or normal operation, "JWT token is invalid" errors started to appear.

Going through these possible causes: https://community.snowflake.com/s/article/Key-Based-Authentication-Failed-with-JWT-token-is-invalid-Error

* credentials are ok, we can connect fine with the same user/key/db/schema/account using the `snowsql` CLI client from all connect nodes in our cluster, so the credentials are fine.
* no relevant time difference between `date +"%T.%N` on the connect node and `SELECT current_timestamp` from snowflake (few ms apart)

Restarting the task and the connector does not help, JWT errors continue and task does not resume.

Rolling back to 1.8.3 fixed the issue. Same connector config for 1.8.3 and 1.9.0, no changes done to the config.  

Maybe the issue is related to this change:
```diff
             <groupId>net.snowflake</groupId>
             <artifactId>snowflake-ingest-sdk</artifactId>
-            <version>1.0.3-beta</version>
+            <version>1.1.0</version>
```

sfc-gh-japatel:
Hi @acristu can you send more logs? stack trace would be really helpful. 
But  on a side note, JWT tokens are rotated every 59 mins if it is coming from Ingest SDK, jdbc will have its own logic for refresh token. Which api is giving this response would be great. thanks for creating the issue.

acristu:
Hi @sfc-gh-japatel , thanks for the quick reply.

Here are some log snippets:
```
[2023-02-28 17:01:44,522] WARN [destination_63ea97792253adec948911fa|task-0] [SF_KAFKA_CONNECTOR] [KC:123402530] [OFFSET_TOKEN_RETRY_POLICY] retry for getLatestCommittedOffsetToken. Retry no:1, message:Get channel status request failed: HTTP Status: 401 ErrorBody: {"
[2023-02-28 17:01:43,522] ERROR [destination_63ea97792253adec948911fa|task-0] Exceptional Status Code from STREAMING_CHANNEL_STATUS: 401, requestId: (net.snowflake.ingest.connection.ServiceResponseHandler:269)"
[...]

PR #554, opened 2023-02-28
no-snow reverting possible commits as reported by #553 - ***Only testing. ***

sfc-gh-japatel:
v1.9.0 has a reported issue for JWT token not valid
#553 

- Customer confirmed that reverting to old version worked for them. 
- There are only few differences in 1.8.3 and 1.9.0
  - JDBC version upgrade to 3.13.23 and ingest sdk to 1.1.0
  - Ingest SDK 1.1.0 had rolledback to jdbc 3.13.15
  - Snowflake KC had 3.13.14 (which was not being used because of ordering of dependencies)
- This change goes back to be reliant on Ingest SDK to use 3.13.15 since that was the last stable one. 

Please note: I will not push this PR before verifying and testing if this is the real root cause, but this is the potential issue.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -313,24 +313,11 @@
             </exclusions>
         </dependency>
 
-        <!--JDBC driver for building connection with Snowflake-->
-        <dependency>
-            <groupId>net.snowflake</groupId>
-            <artifactId>snowflake-jdbc</artifactId>
-            <version>3.13.23</version>
-        </dependency>
-
         <!-- Ingest SDK for copy staged file into snowflake table -->
         <dependency>
             <groupId>net.snowflake</groupId>
             <artifactId>snowflake-ingest-sdk</artifactId>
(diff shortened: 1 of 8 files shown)


sfc-gh-tjones (commented):
What can we do to determine if this is the right fix? Sounds like there are a few possible causes.

sfc-gh-japatel:
> What can we do to determine if this is the right fix? Sounds like there are a few possible causes.

Yes, this changed is still not root caused to the exact issue but is most likely the version upgrade is causing this. 

This was a related change in JDBC when we went with an upgrade from 3.13.15 to 3.13.23
https://github.com/snowflakedb/snowflake-jdbc/pull/715/files

I am not able to repro this so I will abstain from pushing this and creating one more release until I get to root cause.

sfc-gh-japatel:
closing in lieu of #555

PR #555, opened 2023-03-01
NO-SNOW Revert one client change, JDBC upgrade and config for arrow and parquet

### LINKED ISSUE #553: JWT token is invalid, task failure after upgrading to 1.9.0

sfc-gh-japatel:
### This commit is a mega change including all important and relevant reverts from between 1.8.3 and 1.9.0

- It only keeps the 1.1.0 bump of Ingest SDK, connect API bump, few test changes. 

### Summary of what is reverted:
- The fix for one client change is not trivial. I tried couple of changes but it will have other side effects. Behavior will be in contrast to the original philosophy of the change. [i.e create clients from connector and not from tasks] - 3 commits, 1 commit dependent on refactor = 4
- Reverts JDBC changes too. We still haven’t identified if upgrading to 3.13.23 is the cause of JWT token invalid since customer in their other env didn't see this issue. But it is always safe to revert back. - 1 commit
  - This could be a potential bad commit https://github.com/snowflakedb/snowflake-jdbc/pull/715 but we cant be sure. 

### TL;DR:
- Reverts all changes associated with one client change. 
- Reverts changes associated with JDBC. 
  - This change now depends on the version in ingest sdk. (even though we have dependency of 3.13.14, it is depending on version coming from ingest sdk 1.1.0 which is 3.13.15)
- Reverts jvm proxy changes. 
- Reverts the naming change for client. (but client name will come with connector name)
- Reverts config change - will add in next PR. 

Fixes #553


sfc-gh-japatel:
Update: restart, pause and resume APIs are working fine while running the connector. 

I will run this for longer time now

sfc-gh-tzhang:
To confirm, we will
- revert everything after https://github.com/snowflakedb/snowflake-kafka-connector/commit/d2c35614f940f2b20fb345ecadcb3918c610f637 (including this one)
- add back https://github.com/snowflakedb/snowflake-kafka-connector/commit/e58d307732400ba9d7aa43ffbf244736b6baba93, https://github.com/snowflakedb/snowflake-kafka-connector/commit/c5f8aed1a3474fe2cc46ba9fb41faec29341b3d9 and https://github.com/snowflakedb/snowflake-kafka-connector/commit/2786090b8a76b3982ffd85559ab9c180811aef88
- We also need to manually modify some of the failing tests are we fixed

sfc-gh-tzhang (approved):
LGTM, thanks

sfc-gh-japatel:
> To confirm, we will
> 
> * revert everything after [d2c3561](https://github.com/snowflakedb/snowflake-kafka-connector/commit/d2c35614f940f2b20fb345ecadcb3918c610f637) (including this one)
> * add back [e58d307](https://github.com/snowflakedb/snowflake-kafka-connector/commit/e58d307732400ba9d7aa43ffbf244736b6baba93), [c5f8aed](https://github.com/snowflakedb/snowflake-kafka-connector/commit/c5f8aed1a3474fe2cc46ba9fb41faec29341b3d9) and [2786090](https://github.com/snowflakedb/snowflake-kafka-connector/commit/2786090b8a76b3982ffd85559ab9c180811aef88)
> * We also need to manually modify some of the failing tests are we fixed

Yes, that's exactly what I had though of. I will copy this in the description..

PR #560, opened 2023-03-01
NO-SNOW: allow StringConverter when schematization is in the config but set to false

sfc-gh-tzhang:
Bug was introduced in https://github.com/snowflakedb/snowflake-kafka-connector/commit/6b7ed5fea48ff000555948b6c52d3cd1b507dc97, need to include this in 1.9.1 otherwise KC will start to fail as long as the schematization is presented in the config for StringConverter


sfc-gh-tjones (approved):

sfc-gh-tjones (approved):

src/test/java/com/snowflake/kafka/connector/ConnectorConfigTest.java
```diff
@@ -593,4 +593,32 @@ public void testValidSchematizationForStreamingSnowpipe() {
     config.put(Utils.SF_ROLE, "ACCOUNTADMIN");
     Utils.validateConfig(config);
   }
+
+  @Test(expected = SnowflakeKafkaConnectorException.class)
```
sfc-gh-japatel, line 597: 👍

sfc-gh-japatel (approved):

PR #566, opened 2023-03-09
SNOW-759517, SNOW-743629, SNOW-733467: simplify offset management logic in KC

sfc-gh-tzhang:
SNOW-759517, SNOW-743629, SNOW-733467

- Instead of using 5 fields for offset management, we try to simplify the code and use just one, see `processedOffset`. We also keep `offsetPersistedInSnowflake` but that's just for tracking purpose
- We always use the offset token from server side as long as it's valid, when there is no valid offset token at server side, we just trust what kafka sends us
- Add some special handling for schematization schema evolution, and add a workaround for a Ingest SDK issue that causes test flakyness
- Fix tests due to behavior changes and add more validations


src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -77,50 +76,21 @@ public class TopicPartitionChannel {
...
   // This offset is updated when Snowflake has received offset from insertRows API
   // We will update this value after calling offsetToken API for this channel
-  // We will only update it during start of the channel initialization
-  private AtomicLong offsetPersistedInSnowflake =
+  private final AtomicLong offsetPersistedInSnowflake =
+      new AtomicLong(NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE);
+
+  // This offset is updated when KC processed an offset, it's used to make sure that we won't
+  // process records that are already processed. It will be reset to the committed offset every time
+  // we fetch a valid offset token from Snowflake
```
sfc-gh-japatel, line 86: > It will be reset to the committed offset every time we fetch a valid offset token from Snowflake

Can you explain this further? here `committedOffset` is terminology picked from our getLatestComm... API?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -77,50 +76,21 @@ public class TopicPartitionChannel {
...
+  private final AtomicLong processedOffset =
```
sfc-gh-japatel, line 87: nit: can you move these two fields in private final section of the declaration?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -260,6 +217,16 @@ public TopicPartitionChannel(
...
+
+    // Open channel and reset the offset in kafka
+    this.channel = Preconditions.checkNotNull(openChannelForTable());
+    final long lastCommittedOffsetToken = fetchOffsetTokenWithRetry();
+    this.offsetPersistedInSnowflake.set(lastCommittedOffsetToken);
+    this.processedOffset.set(lastCommittedOffsetToken);
+    if (lastCommittedOffsetToken != NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE) {
+      // Reset the offset in kafka, only if we have a valid offset token at server side
+      this.sinkTaskContext.offset(this.topicPartition, lastCommittedOffsetToken);
+    }
```
sfc-gh-japatel, line 229: This sounds like a great idea. Do you mind adding a comment in the constructor definition? 
Technically this is happening even before we receive any offsets from put API. It would be great value addition to our javadocs.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -260,6 +217,16 @@ public TopicPartitionChannel(
...
                 tableName, sfConnectorConfig.get(SNOWFLAKE_ROLE));
+
+    // Open channel and reset the offset in kafka
+    this.channel = Preconditions.checkNotNull(openChannelForTable());
+    final long lastCommittedOffsetToken = fetchOffsetTokenWithRetry();
+    this.offsetPersistedInSnowflake.set(lastCommittedOffsetToken);
+    this.processedOffset.set(lastCommittedOffsetToken);
+    if (lastCommittedOffsetToken != NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE) {
+      // Reset the offset in kafka, only if we have a valid offset token at server side
+      this.sinkTaskContext.offset(this.topicPartition, lastCommittedOffsetToken);
```
sfc-gh-japatel, line 228: I can see a potential exception happening here from kafka side which somwhat says `offset not present` when we try to reset an offset which is not present. Lets say for example a topic is deleted. 

This will force them to delete the table because there is no way to force reset an offset on snowflake side.

sfc-gh-japatel: Also, 
your changes are doing exactly what the docs says
```
When the topic partition is recovered the task would reload offsets from HDFS and use this method to reset the consumer to the offset
```
👍

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -260,6 +217,16 @@ public TopicPartitionChannel(
...
             && this.conn.hasSchemaEvolutionPermission(
                 tableName, sfConnectorConfig.get(SNOWFLAKE_ROLE));
+
+    // Open channel and reset the offset in kafka
+    this.channel = Preconditions.checkNotNull(openChannelForTable());
+    final long lastCommittedOffsetToken = fetchOffsetTokenWithRetry();
+    this.offsetPersistedInSnowflake.set(lastCommittedOffsetToken);
+    this.processedOffset.set(lastCommittedOffsetToken);
+    if (lastCommittedOffsetToken != NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE) {
+      // Reset the offset in kafka, only if we have a valid offset token at server side
```
sfc-gh-japatel, line 227:
```suggestion
      // Reset the consumer offset in kafka, only if we have a valid offset token at server side
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -327,74 +288,6 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
...
-   * @param kafkaSinkRecord Record to check for above condition only in case of failures
-   *     (isOffsetResetInKafka = true)
-   * @return true if this record can be skipped to add into buffer, false otherwise.
-   */
-  private boolean shouldIgnoreAddingRecordToBuffer(SinkRecord kafkaSinkRecord) {
-    if (this.isOffsetResetInKafka.get()) {
-      if ((kafkaSinkRecord.kafkaOffset() - offsetPersistedInSnowflake.get()) != 1L) {
-        // ignore
-        LOGGER.debug(
-            "Ignore adding offset:{} to buffer for channel:{} because we recently encountered"
```
sfc-gh-japatel, line 347: I would actually want to keep this debug statements. Can you put an else in above code and put a debug statement there?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -274,18 +241,12 @@ public TopicPartitionChannel(
...
-
-    // discard the record if the record offset is smaller or equal to server side offset, or if
-    // record is smaller than any other record offset we received before
     final long currentOffsetPersistedInSnowflake = this.offsetPersistedInSnowflake.get();
     final long currentProcessedOffset = this.processedOffset.get();
-    if (kafkaSinkRecord.kafkaOffset() > currentOffsetPersistedInSnowflake
-        && kafkaSinkRecord.kafkaOffset() > currentProcessedOffset) {
+    // Accept the record only if we don't have a valid offset token at server side, or the record
+    // offset is 1 + the processed offset
+    if (currentProcessedOffset == NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE
```
sfc-gh-japatel, line 248: lets add an else to this part and add a debug statement.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -509,19 +402,27 @@ InsertRowsResponse insertBufferedRecords(StreamingBuffer streamingBufferToInsert
...
+      // since it's possible that not all rows are ingested
       if (response.needToResetOffset()) {
-        streamingApiFallbackSupplier(
-            StreamingApiFallbackInvoker.INSERT_ROWS_SCHEMA_EVOLUTION_FALLBACK);
+        final long offsetRecoveredFromSnowflake =
+            streamingApiFallbackSupplier(
+                StreamingApiFallbackInvoker.INSERT_ROWS_SCHEMA_EVOLUTION_FALLBACK);
+        // If there is no valid offset token at server side even after the reset, retry it again
+        if (offsetRecoveredFromSnowflake == NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE) {
+          insertBufferedRecords(streamingBufferToInsert);
```
sfc-gh-japatel, line 424: for my understanding, Is the reason we are retrying is because we have altered the schema in fallback supplier above?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -658,12 +559,15 @@ public InsertRowsResponse get() throws Throwable {
             }
           }
         }
+        // TODO SNOW-758492: for schematization, sleep a few seconds after the insert to avoid
+        // combining channels with different schemas
+        Thread.sleep(2000);
```
sfc-gh-japatel, line 564: are 2 seconds enough?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -958,7 +868,9 @@ private long fetchLatestCommittedOffsetFromSnowflake() {
     try {
       offsetToken = this.channel.getLatestCommittedOffsetToken();
       if (offsetToken == null) {
-        LOGGER.info("OffsetToken not present for channelName:{}", this.getChannelName());
+        LOGGER.warn(
```
sfc-gh-japatel, line 871: Is this a warn? This would create confusions IMO.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -958,7 +868,9 @@ private long fetchLatestCommittedOffsetFromSnowflake() {
...
+            "OffsetToken not present for channelName:{}, will use kafka offset instead",
```
sfc-gh-japatel, line 872:
```suggestion
            "OffsetToken not present for channelName:{}, will rely on Kafka Offsets as source of truth",
```

sfc-gh-japatel (commented):
Overall, I like the refactor and I agree it makes it simpler. Thanks for doing this Toby! 

One major comment is about chaos testing:
Although it has passed all tests on GH Actions, I would be more cautious in pushing this. Can you perform some restart tasks tests in the middle of an ingestion to see how it behaves? I can help you with those tests.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tjones: Yeah committed offset is Snowflake terminology.

sfc-gh-tjones (approved):
Fantastic change! LGTM barring Jay's comments.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: > I can see a potential exception happening here from kafka side which somwhat says offset not present when we try to reset an offset which is not present. Lets say for example a topic is deleted.

Good point, any suggestion on what we should do here? Without my change, we will just accept whatever kafka sends us?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: I want to keep this as warn since I would imagine most of the corner cases would come from the no offset token situation, keep this as warn makes a bit easier to search

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -361,13 +362,12 @@ public Map<TopicPartition, OffsetAndMetadata> preCommit(
...
           (topicPartition, offsetAndMetadata) -> {
             long offSet = sink.getOffset(topicPartition);
-            if (offSet != 0) {
+            if (offSet != NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE) {
               committedOffsets.put(topicPartition, new OffsetAndMetadata(offSet));
             }
           });
     } catch (Exception e) {
       this.DYNAMIC_LOGGER.error("PreCommit error: {} ", e.getMessage());
-      return new HashMap<>();
```
sfc-gh-tzhang, line 370: This is a new change, i think we should still return the offsets for the partitions that don't have an issue

sfc-gh-japatel: i think we should still return the offsets for the partitions that don't have an issue
> 
But you are returning empty here. Am I missing?

sfc-gh-japatel: ahh sorry.

test/test_suit/test_schema_evolution_drop_table.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # Recreate the table
+        self.driver.snowflake_conn.cursor().execute(
+            "Create or replace table {} (PERFORMANCE_STRING STRING, RECORD_METADATA VARIANT)".format(self.table))
+        self.driver.snowflake_conn.cursor().execute(
+            "alter table {} set ENABLE_SCHEMA_EVOLUTION = true".format(self.table))
+
+        # Ingest another set of rows
+        sleep(30)
+        self.driver.sendBytesData(self.topic, value, key)
```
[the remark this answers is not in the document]
sfc-gh-tzhang: No, we're NOT resetting the offset to 0, the offset token is null after the drop table so we will just accept what KC sends us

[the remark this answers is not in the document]
sfc-gh-japatel: ahh yes, forgot you did create or replace.  thanks for explanation. Low on caffeine today!

PR #568, opened 2023-03-10
Use MDC context instead of kc and logger task

sfc-gh-rcheng:
- Move connector start log to after the connector instance id was set
- Change log levels of certain logs to reduce log size
- Task logging tag was harder to understand than I thought when trying to debug it, so this PR reduces the complexity


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
```diff
@@ -79,18 +77,15 @@ public SnowflakeSinkConnector() {
...
+    LOGGER.info("SnowflakeSinkConnector:starting...");
 
     Utils.checkConnectorVersion();
 
-    LOGGER.info("SnowflakeSinkConnector:start");
     setupComplete = false;
     connectorStartTime = System.currentTimeMillis();
-
-    // initialize logging with global instance Id
-    LoggerHandler.setConnectGlobalInstanceId(this.getKcInstanceId(this.connectorStartTime));
```
sfc-gh-rcheng, line 92: 'SnowflakeSinkConnector:start' log should have the current kc instanceid, so move the instanceid set logic above this log

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
```diff
@@ -307,13 +302,4 @@ public Config validate(Map<String, String> connectorConfigs) {
   public String version() {
     return Utils.VERSION;
   }
-
-  // returns the instance id as a combo of a random uuid and the current time
-  private String getKcInstanceId(long currTime) {
```
sfc-gh-rcheng, line 312: moved to loggerHandler as static method

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -52,18 +52,9 @@
  * Snowflake via Sink service
  */
 public class SnowflakeSinkTask extends SinkTask {
-  // SfTask[ID:taskId.taskOpenCount, #totalTaskCreationCount]
```
sfc-gh-rcheng, line 55: Got feedback and also personally was confused by this tag.... so making it just [TASK:taskid.startTimeHashCode]

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -279,9 +257,6 @@ public void stop() {
    */
   @Override
   public void open(final Collection<TopicPartition> partitions) {
-    this.taskOpenCount++;
```
sfc-gh-rcheng, line 282: While this is definitely useful to keep track of, it was confusing for the tag. Easier to 'grep' on the KC instanceid and the taskid, then diff on how many different hashcodes are returned to see the open count

src/main/java/com/snowflake/kafka/connector/internal/LoggerHandler.java
```diff
@@ -61,8 +73,8 @@ public static void setConnectGlobalInstanceId(String id) {
   public LoggerHandler(String name) {
     this.logger = LoggerFactory.getLogger(name);
 
-    META_LOGGER.info(
```
sfc-gh-rcheng, line 64: reduced the log level from info to trace (and below from info to debug) since they were flooding the logs with not a loooot of usefulness

sfc-gh-rcheng (commented):
This change isn't that urgent, so I'm happy to push this after I return from vacation. Just planning on using this change to help debug the one client restart issue

src/test/java/com/snowflake/kafka/connector/SinkTaskIT.java
```diff
@@ -204,25 +203,17 @@ public void testMultipleSinkTasksWithLogs() throws Exception {
...
 
     String task1Id = "1";
-    int task1OpenCount = 0;
     Map<String, String> task1Config = TestUtils.getConf();
     SnowflakeSinkConnectorConfig.setDefaultValues(task1Config);
     task1Config.put(Utils.TASK_ID, task1Id);
 
-    // set up task1 logging tag
-    String expectedTask1Tag =
-        TestUtils.getExpectedLogTagWithoutCreationCount(task1Id, task1OpenCount);
```
sfc-gh-rcheng, line 214: not necessary to track anymore because there is no open count

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-japatel: In the comment you mentioned 
[Task:taskId.startTime]

I am okay with keeping startTime instead of `startTimeHashCode`

src/main/java/com/snowflake/kafka/connector/internal/LoggerHandler.java
```diff
@@ -37,19 +27,41 @@ public static void setConnectGlobalInstanceId(UUID id) {
    *
```
sfc-gh-japatel, line 27: nit: `remove If an invalid id is given...`

src/main/java/com/snowflake/kafka/connector/internal/LoggerHandler.java
```diff
@@ -37,19 +27,41 @@ public static void setConnectGlobalInstanceId(UUID id) {
...
   }
 
+  /**
+   * Returns the instance id as the hashcode of the kc start time
+   *
+   * @param startTime the start time
+   * @return the formatted instance id
+   */
+  public static String getFormattedKcGlobalInstanceId(long startTime) {
+    return Utils.formatString(KC_GLOBAL_INSTANCEID_FORMAT, Math.abs(("" + startTime).hashCode()));
```
sfc-gh-japatel, line 50: Just thinking out loud, cant we keep startTime instead of its hashcode?

src/main/java/com/snowflake/kafka/connector/internal/LoggerHandler.java
sfc-gh-japatel: 👍

src/main/java/com/snowflake/kafka/connector/internal/LoggerHandler.java
sfc-gh-japatel: Also do we even need to keep startTime? The logger already prints the timestamp.

sfc-gh-japatel:
> This change isn't that urgent, so I'm happy to push this after I return from vacation. Just planning on using this change to help debug the one client restart issue

Thanks for doing this change, I was having hard time reading the tags as well. I am still unsure about the startTime addition in tag if it doesnt provide value. 
May be asking fundamental question will help us design better. 
1. Why we need tags? For debuggability. 
2. What are the important variables. - Tasks, connector name, snowpipe vs snowpipe streaming (I remember you asking me but now I think about it, it might be useful, no need to add here though just putting it out) 
3. Thanks for removing the count, since it wasnt needed for every log. 
4. May be just taskId is enough..

Happy to discuss this more.

src/main/java/com/snowflake/kafka/connector/internal/LoggerHandler.java
sfc-gh-rcheng: I had the same thought - I don't think we need startTime since it has the timestamp. However it does easily provide us with an unique ID. I hashed it because the hash is shorter than the timestamp, to save space in the logs

sfc-gh-japatel: And that unique Id can also be used as a taskId right?

sfc-gh-rcheng:
For the KC instance id:
I do strongly think we need a connector instanceid to help us distinguish between multiple running connectors - I believe that was a large part of the fidelity issue? Start time was simply an easy way to get an unique instanceid for that tag.

For the Task tag:
Yes the task id with the connector instanceid may be enough, however adding the hash at the end will help us determine if a task has been reopened multiple times. The task id does not change if the task or connector has been restarted.

Happy to discuss further toooo

src/main/java/com/snowflake/kafka/connector/internal/LoggerHandler.java
```diff
@@ -1,31 +1,21 @@
...
 import org.slf4j.LoggerFactory;
 
 /** Attaches additional fields to the logs */
 public class LoggerHandler {
   // static properties and methods
   private static final Logger META_LOGGER = LoggerFactory.getLogger(LoggerHandler.class.getName());
-  private static final String EMPTY_ID = new UUID(0L, 0L).toString();
 
-  private static String kcGlobalInstanceId = EMPTY_ID;
+  // [KC:instanceid] where instanceid is a hash of a random uuid and the current time
```
sfc-gh-japatel, line 12: is it UUID + current time or just current time?

sfc-gh-japatel:
> For the KC instance id: I do strongly think we need a connector instanceid to help us distinguish between multiple running connectors - I believe that was a large part of the fidelity issue? Start time was simply an easy way to get an unique instanceid for that tag.
> 
> For the Task tag: Yes the task id with the connector instanceid may be enough, however adding the hash at the end will help us determine if a task has been reopened multiple times. The task id does not change if the task or connector has been restarted.
> 
> Happy to discuss further toooo

Sounds good!

sfc-gh-japatel:
Kafka has MDC for connector context however it needs to be manually set. https://www.confluent.io/blog/kafka-connect-improvements-in-apache-kafka-2-3/

Should we try to do something similar? 
connector-name | task-{ID_NO}-hash

I like `-` as a separator instead of `.` - But thats just me

sfc-gh-japatel:
And since this is done at kafka level, may be we dont need to re-invent the wheel but just call the API: 
MDC is available in slf4j
String connectorContext = MDC.get("connector.context");

sfc-gh-rcheng:
> And since this is done at kafka level, may be we dont need to re-invent the wheel but just call the API: MDC is available in slf4j String connectorContext = MDC.get("connector.context");

Thanks for this! simplified a lot with this. The following is a sample log with the MDC context. Looks like it is just a [<connectorName>|task-<taskid>]. Since connector names are unique (used to http post/delete connectors), this is ok. The main issue is if the customer already has the MDC context setup, it will print this twice in the logs

`[2023-03-28 00:10:25,087] INFO [SF_KAFKA_CONNECTOR] [revi_kc|task-0] Opening a channel with name:revi_tmp_0 for table name:revi_tmp (com.snowflake.kafka.connector.internal.streaming.TopicPartitionChannel)`

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -102,6 +103,8 @@ public class Utils {
 
   private static final KCLogger LOGGER = new KCLogger(Utils.class.getName());
 
+  public static final String MDC_CONN_CTX_KEY = "connector.context";
```
sfc-gh-japatel, line 106: Lets add a comment here about this change and is available only after 2.3.0 
https://cwiki.apache.org/confluence/display/KAFKA/KIP-449%3A+Add+connector+contexts+to+Connect+worker+logs

sfc-gh-japatel:
> > And since this is done at kafka level, may be we dont need to re-invent the wheel but just call the API: MDC is available in slf4j String connectorContext = MDC.get("connector.context");
> 
> Thanks for this! simplified a lot with this. The following is a sample log with the MDC context. Looks like it is just a [|task-]. Since connector names are unique (used to http post/delete connectors), this is ok. The main issue is if the customer already has the MDC context setup, it will print this twice in the logs
> 
> `[2023-03-28 00:10:25,087] INFO [SF_KAFKA_CONNECTOR] [revi_kc|task-0] Opening a channel with name:revi_tmp_0 for table name:revi_tmp (com.snowflake.kafka.connector.internal.streaming.TopicPartitionChannel)`

+1 to double printing, I dont have a great answer to this but is there a way to verify if it is set? (like an API?)

sfc-gh-rcheng:
> +1 to double printing, I dont have a great answer to this but is there a way to verify if it is set? (like an API?) Otherwise I am okay printing it twice, since having two is better than 0? And they can always turn off their own context. WDYT?

I'm also ok with double printing since the param is set at the connector level. Perhaps we can add something in our documentation about this? If we have a section about understanding/debugging your snowflake KC logs.

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-japatel: Lets add a comment here.

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -653,6 +656,14 @@ public static void updateConfigErrorMessage(Config result, String key, String ms
    * @return log message wrapped by snowflake tag
    */
   public static String formatLogMessage(String format, Object... vars) {
+    // note this MDC context is only available for apache kafka versions after 2.3.0, more
+    // information here:
+    // https://cwiki.apache.org/confluence/display/KAFKA/KIP-449%3A+Add+connector+contexts+to+Connect+worker+logs
+    String connCtx = MDC.get(MDC_CONN_CTX_KEY);
+    if (connCtx != null && !connCtx.isEmpty()) {
+      return SF_LOG_TAG + " " + connCtx + formatString(format, vars);
```
sfc-gh-japatel, line 664: now that I think more about this, I do feel about adding this connector context behind a param, with default being set.
Couple of reasons: 
1. They can turn it off at SF KC level. 
2. If it's not set at Kafka Connector level, our default true will help them. 

They might not like the default true option to begin with but if they are super worried about redundant logs (from KC and SF KC connector context) they will switch it off. Turning off at KC level would not be a great option for them since it will affect other connectors. 

WDYT?

src/test/java/com/snowflake/kafka/connector/internal/KCLoggerTest.java
```diff
@@ -37,96 +37,22 @@ public class KCLoggerTest {
 
   @InjectMocks private KCLogger kcLogger = new KCLogger(this.name);
 
-  @After
+  @Before
   public void close() {
```
sfc-gh-japatel, line 41: rename this function to before?

sfc-gh-japatel (commented):
Left a comment about adding this context behind a param set to true. Let me know what you think. The code overall lgtm

src/main/java/com/snowflake/kafka/connector/internal/LoggerHandler.java
sfc-gh-rcheng: removed logging logic, resolving any related comments

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
```diff
@@ -92,6 +92,11 @@ public void start(final Map<String, String> parsedConfig) {
 
     Utils.validateConfig(config);
 
+    // enable mdc logging if needed
+    KCLogger.enableGlobalMdcLoggingContext(
+        Boolean.parseBoolean(
+            config.getOrDefault(SnowflakeSinkConnectorConfig.ENABLE_MDC_LOGGING_CONFIG, "false")));
```
sfc-gh-japatel, line 98: default false makes sense! Thanks

sfc-gh-japatel (approved):
lgtm! Thanks
Can you also create a documentation JIRA for @sfc-gh-lema (Customers have two options, either upgrade to later Kafka and set it at connector level or just use our config)

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-rcheng: added param that defaults to false. thanks for the suggestion!

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
sfc-gh-rcheng: agree - updated the default to be false

PR #569, opened 2023-03-11
SNOW-758111 Use telemetry service for all exceptions wherever possible

sfc-gh-japatel:
- Why as possible
  - It is possible we dont have connection object everywhere, so in that case, we will not be able to send telemetry to Snowflake

Part 1:
- Use telemetry Service wherever possible to send exceptions to Snowflake

Future PRs:
- We will annotate whether Snowpipe Streaming is used or Snowpipe to distinguish. 
- Channel usage telemetry - For example send data for channel open/close.
- Add alerts later on Snowflake Side.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
@@ -739,7 +739,7 @@ public class SnowflakeConnectionServiceV1 extends EnableLogging
       try {
         file = sfconn.downloadStream(stageName, name, true);
       } catch (Exception e) {
-        throw SnowflakeErrors.ERROR_2002.getException(e);
+        throw SnowflakeErrors.ERROR_2002.getException(e, this.telemetry);
       }
       // put
       try {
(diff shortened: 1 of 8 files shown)


sfc-gh-tzhang:
> * It is possible we dont have connection object everywhere, so in that case, we will not be able to send telemetry to Snowflake

This is not true anymore since I added the support without a session, please check how the telemetry works in the Ingest SDK :)

sfc-gh-japatel:
> > * It is possible we dont have connection object everywhere, so in that case, we will not be able to send telemetry to Snowflake
> 
> 
> 
> This is not true anymore since I added the support without a session, please check how the telemetry works in the Ingest SDK :)

Ack, thanks for reminding that, I will take a look how sessionless works. but I dont plan to add it in this PR, will modify description and send another PR for that

sfc-gh-tzhang (approved):
LGTM, one thing I suggest is that we don't have to send all exceptions to snowflake since some of them are not useful as they're human errors (e.g. invalid private key), so please use your best judgement to see whether an exception is needed at server side

sfc-gh-japatel:
> LGTM, one thing I suggest is that we don't have to send all exceptions to snowflake since some of them are not useful as they're human errors (e.g. invalid private key), so please use your best judgement to see whether an exception is needed at server side

Ack, verified they are all either API exceptions or unreachable/should not have happened metrics.

PR #573, opened 2023-03-14
SNOW-760171 Use pom_confluent.xml in running confluent tests

sfc-gh-japatel:
### Benefits:
- Catch issues before releasing a zip in Confluent Hub
- Use pom_confluent.xml in our tests. We only used pom.xml until now and no way to check the diff between those two. 
- For private key which are encrypted, we have to deploy fips jar to confluen setup, this test would have caught that. 

### Details
- Rename build_apache.jar with build_runtime_jar
- Use pom_confluent.xml in Running End to end test against Confluent
- Pass in confluent argument when we want to use confluent as kafka runtime. 
- Add description in sh script. 
- pom_confluent uses https://docs.confluent.io/platform/current/connect/kafka-connect-maven-plugin/site/plugin-info.html plugin to create a zip file which gets released to Confluent Hub. 

Update: [For confluent tests]
- For all versions, we point plugin path to the zip file we extract after running build_runtime_jar.sh which runs the connect maven plugin. 

### Next:
- Some of the End2EndFullTest runs both Apache and Confluent on a weekly basis in the same workflow and is also running older version. 
- I will fix those tests to run apache and confluent only.

### DIFF
--- a/test/build_runtime_jar.sh
+++ b/test/build_runtime_jar.sh
@@ -110,12 +110,9 @@ cp $SNOWFLAKE_PLUGIN_PATH/$SNOWFLAKE_PLUGIN_NAME $KAFKA_CONNECT_PLUGIN_PATH || t
 echo -e "copied SF Plugin Connector to $KAFKA_CONNECT_PLUGIN_PATH"
 
 if [[ $BUILD_FOR_RUNTIME == "confluent" ]]; then
-    echo "For confluent RUNTIME: Copying Lib directory of the Confluent Generated Zip package to $KAFKA_CONNECT_PLUGIN_PATH"
-    ls $SNOWFLAKE_PLUGIN_PATH/components/packages/*/*/lib/*
-    echo "list KAFKA_CONNECT_PLUGIN_PATH"
-    ls $KAFKA_CONNECT_PLUGIN_PATH
-    echo "Performing Recursive Copy"
-    cp -R $SNOWFLAKE_PLUGIN_PATH/components/packages/*/*/lib/* $KAFKA_CONNECT_PLUGIN_PATH
+    echo "For confluent RUNTIME: Copying Kafka Connect Maven Generated Zip file to a temporary location"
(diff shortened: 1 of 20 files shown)


sfc-gh-tjones (approved):

test/build_runtime_jar.sh
```diff
@@ -87,13 +107,15 @@ echo -e "\nbuilt connector name: $SNOWFLAKE_PLUGIN_NAME"
...
+fi
 
 KAFKA_CONNECT_DOCKER_JAR_PATH="$SNOWFLAKE_CONNECTOR_PATH/docker-setup/snowflake-kafka-docker/jars"
 mkdir -m 777 -p $KAFKA_CONNECT_DOCKER_JAR_PATH
 cp $SNOWFLAKE_PLUGIN_PATH/$SNOWFLAKE_PLUGIN_NAME $KAFKA_CONNECT_DOCKER_JAR_PATH || true
-echo -e "copied connector to $KAFKA_CONNECT_DOCKER_JAR_PATH for docker"
-
-echo -e "\n===================================== WARNING ====================================="
-echo -e "not recommended to share this jar with customer unless necessary (PrPr feature, etc)"
-echo -e "absolutely do not let the customer run this jar on their production environment\n"
```
sfc-gh-tzhang, line 99: I thought we added this for a reason?

sfc-gh-tzhang (approved):
Thanks Jay. Is the goal here to support running end2end tests against Confluent? Could you update the description to make it more clear?

test/build_runtime_jar.sh
sfc-gh-japatel: > Thanks Jay. Is the goal here to support running end2end tests against Confluent? Could you update the description to make it more clear?

Ooops, will make it clear.. No, that is not the goal, but this will help us test how our jar/release will behave if we release it on confluent hub.

sfc-gh-japatel: > I thought we added this for a reason?

Not needed anymore, it is not in PrPr..

sfc-gh-tjones (approved):

PR #579, opened 2023-03-17
NO-SNOW Use maven Cache - 4X improvement in runtime of Github Actions for builds

sfc-gh-japatel:
Also upgrade maven plugin versions to a more recent versions to reduce flakiness in Github Action

- Helps reduce the build time of our library on Github Action. 4x improvement
  - Master run for build: https://github.com/snowflakedb/snowflake-kafka-connector/actions/runs/4451294967/jobs/7817781250 (12 mins)
  - Current run for build: https://github.com/snowflakedb/snowflake-kafka-connector/actions/runs/4451509883/jobs/7818250333?pr=579 (3 mins)
  - Stress tests took 1hr to run before, now only 30 mins. 
  - Faster response time for merge gates and pushing the PR to master.

### DIFF
--- a/.github/workflows/End2EndTestApache.yml
+++ b/.github/workflows/End2EndTestApache.yml
@@ -21,6 +21,13 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
+    - name: "Cache local Maven repository"
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2/repository
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          ${{ runner.os }}-maven-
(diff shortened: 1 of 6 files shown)


sfc-gh-tjones (approved):

sfc-gh-tzhang (approved):
Thanks Jay. For my own preference, is there any issue of using the Maven cache?

sfc-gh-japatel:
> Thanks Jay. For my own preference, is there any issue of using the Maven cache?

AFAIK, there is no issue, in case there is cache miss, we will update the cache using key provided in workflow.yml file. Overall I have seen increase in build time almost everytime. 
I am guessing we will see cache miss only for the first time we add new dependency. Chances of that happening is very less compared to build time saved in every other scenario.

PR #583, opened 2023-03-22
SNOW-756851 Part 1: Refactor code to create telemetry service for Snowpipe Streaming

sfc-gh-japatel:
- EnableLogging is kept as an abstract class which is not needed. Java doesnt allow multiple inheritance. (Ahh same old diamond problem). I wanted to turn TelemetryService to abstract class and most of the other classes are inheriting from `EnableLogging`
  - Ideally this can be converted to a class and hence changed it to class. Best practices is to have Logger as a class. 
  - We dont an additional wrapper class, removed this class altogether. Instead rename `LoggerHandler` to `SFLogger`
- Change TelemetryService to abstract class since that can be implemented to have Snowpipe streaming telemetry. 
  - Can also use session less telemetry to not depend on snowpipe's telemetry. 

- Pom changes:
  - Without that, javadocs was failing the build. 
  - Will fix javadocs error in another PR.


src/main/java/com/snowflake/kafka/connector/internal/SFLogger.java
```diff
@@ -0,0 +1,108 @@
+/*
+ * Copyright (c) 2022 Snowflake Inc. All rights reserved.
```
sfc-gh-tzhang, line 2: nit: 2023 :)

src/main/java/com/snowflake/kafka/connector/internal/SFLogger.java
```diff
@@ -0,0 +1,108 @@
...
+  public SFLogger(Class<?> clazz) {
+    this.logger = new LoggerHandler(clazz.getName());
+  }
+
+  /**
+   * Logs an info-level message.
+   *
+   * @param format the message format
+   */
+  public void LOG_INFO_MSG(String format) {
```
sfc-gh-tzhang, line 38: _MSG looks redundant, could we remove it on all functions?

sfc-gh-tzhang: Also LOG_, should we just do INFO, WARN, etc.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeTelemetryServiceV2.java
```diff
@@ -0,0 +1,14 @@
+package com.snowflake.kafka.connector.internal.streaming;
```
sfc-gh-tzhang, line 1: copyright

src/main/java/com/snowflake/kafka/connector/internal/EnableLogging.java
```diff
@@ -1,63 +0,0 @@
...
- * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
- * KIND, either express or implied.  See the License for the
- * specific language governing permissions and limitations
- * under the License.
- */
-
-package com.snowflake.kafka.connector.internal;
-
-// The purpose of this class is for non-static inheritors to easily log information
-public abstract class EnableLogging {
```
sfc-gh-tzhang, line 21: IIRC, this was added by Revi, could you check with her to see why we decide to go this way?

sfc-gh-tzhang (approved):
LGTM, please ask @sfc-gh-rcheng to take a look when she comes back

src/main/java/com/snowflake/kafka/connector/internal/SFLogger.java
sfc-gh-japatel: makes sense for both!

src/main/java/com/snowflake/kafka/connector/internal/EnableLogging.java
sfc-gh-rcheng: I didn't add this class, just added to it as it was there already. I also am not sure why we went this way, so fully support removing it

src/main/java/com/snowflake/kafka/connector/internal/SFLogger.java
```diff
@@ -0,0 +1,108 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.snowflake.kafka.connector.internal;
+
+/** This class provides an easy way for non-static inheritors to log information. */
```
sfc-gh-rcheng, line 20: I don't fully understand the purpose of this class. Could we just use LoggerHandler? Happy to change the name if that is the issue. If this was because of the EnableLogging class, we can remove this.

sfc-gh-rcheng (changes requested):
lgtm on the telemetry changes, added comments about enableLogging

src/main/java/com/snowflake/kafka/connector/internal/EnableLogging.java
sfc-gh-japatel: Ack

src/main/java/com/snowflake/kafka/connector/internal/SFLogger.java
sfc-gh-japatel: Yeah, we dont need this, I wanted to make sure I dont miss anything. I will also change LoggerHandler class.

sfc-gh-rcheng (approved):

PR #585, opened 2023-03-23
SNOW-760171 Part 1: Test to verify DLQ results 

sfc-gh-japatel:
- Insert bad data in value
- Watch it going into DLQ
- Poll DLQ and verify results. 

Next: Will think of more test cases.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -255,6 +255,9 @@
                         <goals>
                             <goal>jar</goal>
                         </goals>
+                        <configuration>
+                            <failOnError>false</failOnError>
+                        </configuration>
                     </execution>
                 </executions>
             </plugin>
--- a/test/rest_request_template/snowpipe_streaming_string_json_dlq.json
+++ b/test/rest_request_template/snowpipe_streaming_string_json_dlq.json
@@ -1,0 +1,27 @@
+{
+  "name": "SNOWFLAKE_CONNECTOR_NAME",
+  "config": {
+    "connector.class": "com.snowflake.kafka.connector.SnowflakeSinkConnector",
+    "topics": "SNOWFLAKE_TEST_TOPIC",
+    "tasks.max": "1",
+    "buffer.flush.time": "10",
+    "buffer.count.records": "100",
+    "buffer.size.bytes": "5000000",
+    "snowflake.url.name": "SNOWFLAKE_HOST",
+    "snowflake.user.name": "SNOWFLAKE_USER",
+    "snowflake.private.key": "SNOWFLAKE_PRIVATE_KEY",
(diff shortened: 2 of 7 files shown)


pom.xml
```diff
@@ -255,6 +255,9 @@
                         <goals>
                             <goal>jar</goal>
                         </goals>
+                        <configuration>
```
sfc-gh-tzhang, line 258: Could you explain why we need this? I think I removed it because it was not consistent between pom and pom_confluent

test/rest_request_template/snowpipe_streaming_string_json_dlq.json
```diff
@@ -0,0 +1,27 @@
+{
```
sfc-gh-tzhang, line 1: Could we have one for schematization as well? Basically all schema difference error shouldn't move rows to DLQ

test/test_verify.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, kafkaAddress, schemaRegistryAddress, kafkaConnectAddress, cre
         self.httpHeader = {'Content-type': 'application/json', 'Accept': 'application/json'}
 
         self.SEND_INTERVAL = 0.01  # send a record every 10 ms
-        self.VERIFY_INTERVAL = 60  # verify every 60 secs
+        self.VERIFY_INTERVAL = 10  # verify every 60 secs
```
sfc-gh-tzhang, line 50: I think 10sec is too short and might cause more flakiness, I suggest we keep at 60

sfc-gh-tzhang (approved):
Thanks Jay. Left some comments, otherwise lgtm

test/test_verify.py
sfc-gh-japatel: ahh yes, this was intended for testing but thanks for the catch!

pom.xml
sfc-gh-japatel: Without this, it is failing on github action that it is missing some return params and other javadocs errors. 
I want to get this in and in separate PR fix all those javadoc errors. 

Basically this config will allow build to pass even if there are javadoc errors.

test/rest_request_template/snowpipe_streaming_string_json_dlq.json
sfc-gh-japatel: Sounds good, can add that. 
Just clarifying, it shouldn't move to DLQ? 
mind sharing an example?

sfc-gh-japatel: Added DLQ for schema mapping.

test/test_verify.py
sfc-gh-japatel: reverted.

Issue #442, opened 2022-06-13
Log name of the config parameter along with "Invalid input connector configuration"


dlg99:
Now I have to debug through the connector to understand what it didn't like:

```
com.snowflake.kafka.connector.internal.SnowflakeKafkaConnectorException:
[SF_KAFKA_CONNECTOR] Exception: Invalid input connector configuration
[SF_KAFKA_CONNECTOR] Error Code: 0001
[SF_KAFKA_CONNECTOR] Detail: input kafka connector configuration is null, missing required values, or wrong input value
	at com.snowflake.kafka.connector.internal.SnowflakeErrors.getException(SnowflakeErrors.java:289) ~[?:?]
	at com.snowflake.kafka.connector.internal.SnowflakeErrors.getException(SnowflakeErrors.java:256) ~[?:?]
	at com.snowflake.kafka.connector.Utils.validateConfig(Utils.java:469) ~[?:?]
	at com.snowflake.kafka.connector.SnowflakeSinkConnector.start(SnowflakeSinkConnector.java:87) ~[?:?]
```

sfc-gh-rcheng:
Requested change released in [1.9.2](https://github.com/snowflakedb/snowflake-kafka-connector/releases/tag/v1.9.2). Thanks for opening this issue

PR #586, opened 2023-03-27
[SNOW-659469] List all invalid parameters 

sfc-gh-rcheng:
Collect all invalid parameters and error messages to log and throw an exception at the end

Changes: 
- Replaced isConfigValid boolean with invalidConfigParam hashmap
- Changed most of the 'validateX' methods to return the list of invalid params and messages
- Added additional assert to config tests to check for the required invalid params

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/Utils.java
+++ b/src/main/java/com/snowflake/kafka/connector/Utils.java
@@ -22,6 +22,7 @@ import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.DELIVER
 import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.INGESTION_METHOD_OPT;
 import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.JMX_OPT;
 
+import com.google.common.collect.ImmutableMap;
 import com.snowflake.kafka.connector.internal.BufferThreshold;
 import com.snowflake.kafka.connector.internal.LoggerHandler;
 import com.snowflake.kafka.connector.internal.SnowflakeErrors;
@@ -216,7 +217,7 @@ public class Utils {
    *
    * @param config connector configuration
    */
-  static Map<String, String> validateProxySetting(Map<String, String> config) {
+  static ImmutableMap<String, String> validateProxySettings(Map<String, String> config) {
     Map<String, String> invalidConfigParams = new HashMap<String, String>();
 
     String host =
@@ -252,7 +253,7 @@ public class Utils {
       }
     }
 
-    return invalidConfigParams;
+    return ImmutableMap.copyOf(invalidConfigParams);
   }
 
   /**
@@ -427,7 +428,7 @@ public class Utils {
           Utils.formatString("{} cannot be empty.", SnowflakeSinkConnectorConfig.SNOWFLAKE_URL));
     }
     // jvm proxy settings
-    invalidConfigParams.putAll(validateProxySetting(config));
+    invalidConfigParams.putAll(validateProxySettings(config));
 
     // set jdbc logging directory
     Utils.setJDBCLoggingDirectory();
@@ -482,21 +483,8 @@ public class Utils {
     // Check all config values for ingestion method == IngestionMethodConfig.SNOWPIPE_STREAMING
     invalidConfigParams.putAll(StreamingUtils.validateStreamingSnowpipeConfig(config));
 
-    // log all invalid params and throw exception
-    if (!invalidConfigParams.isEmpty()) {
-      String invalidParamsMessage = "";
-
-      for (String invalidKey : invalidConfigParams.keySet()) {
-        String invalidValue = invalidConfigParams.get(invalidKey);
-        String errorMessage =
-            Utils.formatString(
-                "Config value '{}' is invalid. Error message: '{}'", invalidKey, invalidValue);
-        invalidParamsMessage += errorMessage + "\n";
-      }
-
-      LOGGER.error("Invalid config: " + invalidParamsMessage);
-      throw SnowflakeErrors.ERROR_0001.getException(invalidParamsMessage);
-    }
+    // logs and throws exception if there are invalid params
+    handleInvalidParameters(ImmutableMap.copyOf(invalidConfigParams));
 
     return connectorName;
   }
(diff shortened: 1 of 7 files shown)


src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -217,20 +216,24 @@ public static void setJDBCLoggingDirectory() {
    *
    * @param config connector configuration
    */
-  static void validateProxySetting(Map<String, String> config) {
+  static Map<String, String> validateProxySetting(Map<String, String> config) {
```
sfc-gh-rcheng, line 219: this method is used in SnowflakeSinkConnector (above file change) and in Utils. The error it throws is always caught, so I believe it is better to just return a list of the invalid params

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -329,7 +335,7 @@ static boolean isValidSnowflakeTableName(String tableName) {
    * @return connector name
    */
   static String validateConfig(Map<String, String> config) {
-    boolean configIsValid = true; // verify all config
+    Map<String, String> invalidConfigParams = new HashMap<String, String>(); // verify all config
```
sfc-gh-rcheng, line 338: this is the crux of the change, replacing the configIsValid boolean with a map of invalid params

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -454,16 +473,29 @@ && parseTopicToTableMap(config.get(SnowflakeSinkConnectorConfig.TOPICS_TABLES_MA
...
+        String invalidValue = invalidConfigParams.get(invalidKey);
+        String errorMessage =
+            Utils.formatString(
+                "Config value '{}' is invalid. Error message: '{}'", invalidKey, invalidValue);
+        invalidParamsMessage += errorMessage + "\n";
+      }
 
-    if (!configIsValid || !isStreamingConfigValid) {
-      throw SnowflakeErrors.ERROR_0001.getException();
+      LOGGER.error("Invalid config: " + invalidParamsMessage);
```
sfc-gh-rcheng, line 497: Here is where we both log the invalid params and throw an exception. Potentially not necessary to do both, however I wanted to retain the previous behavior (logs invalid param and also changes boolean to false)

src/main/java/com/snowflake/kafka/connector/internal/BufferThreshold.java
```diff
@@ -123,19 +125,26 @@ public long getFlushTimeThresholdSeconds() {
...
+    Map<String, String> invalidConfigParams = new HashMap();
+
     if (!providedSFConnectorConfig.containsKey(
         SnowflakeSinkConnectorConfig.BUFFER_FLUSH_TIME_SEC)) {
-      LOGGER.error("Config {} is empty", SnowflakeSinkConnectorConfig.BUFFER_FLUSH_TIME_SEC);
-      return false;
+      invalidConfigParams.put(
+          SnowflakeSinkConnectorConfig.BUFFER_FLUSH_TIME_SEC,
+          Utils.formatString(
+              "Config {} is empty", SnowflakeSinkConnectorConfig.BUFFER_FLUSH_TIME_SEC));
```
sfc-gh-rcheng, line 147: Given previous behavior, could technically return a string instead of a map here. However I think it is better to do a map in case we have further param testing in the future. Open to changing it though

sfc-gh-rcheng (commented):
Note that parseTopicToTableMap still throws a snowflake exception instead of returning the invalid configs. This is because it already returns a Map, so it gets messy to return an invalid param map. Open to suggestions!

But I think it is ok to throw an exception if we have an invalid topic/table for now.

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-japatel, line 219: nit: `validateProxySettings`

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-japatel, line 497: nit: extract it into a different method?

src/main/java/com/snowflake/kafka/connector/internal/BufferThreshold.java
```diff
@@ -123,19 +125,26 @@ public long getFlushTimeThresholdSeconds() {
    * @param ingestionMethodConfig ingestion method used. Check {@link IngestionMethodConfig}
    * @return true if all thresholds are valid.
```
sfc-gh-japatel, line 126: nit: change `return`?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java
```diff
@@ -203,48 +198,53 @@ public static boolean isStreamingSnowpipeConfigValid(final Map<String, String> i
...
-        LOGGER.error("Kafka config:{} error:{}", INGESTION_METHOD_OPT, exception.getMessage());
-        configIsValid = false;
+        invalidParams.put(
+            INGESTION_METHOD_OPT,
+            Utils.formatString(
+                "Kafka config:{} error:{}", INGESTION_METHOD_OPT, exception.getMessage()));
       }
     }
-    return configIsValid;
+    return invalidParams;
```
sfc-gh-japatel, line 210: lets return an immutable map here: 
https://snowflakecomputing.atlassian.net/wiki/spaces/EN/pages/1960706106/Effective+Java+Guidelines+at+Snowflake

sfc-gh-japatel (approved):
minor suggestions but LGTM! Thanks, looks better.
Please add JIRA or associated issue/PR for tracking purposes. 
Integration tests are failing, please fix it before merging.

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-rcheng: moved to private method at bottom of file

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -216,7 +217,7 @@ public static void setJDBCLoggingDirectory() {
    *
    * @param config connector configuration
    */
-  static Map<String, String> validateProxySetting(Map<String, String> config) {
+  static ImmutableMap<String, String> validateProxySettings(Map<String, String> config) {
```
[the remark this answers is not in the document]
sfc-gh-rcheng: Yup didn't need to change it, but just preferred to 😅

sfc-gh-rcheng:
For this issue: https://github.com/snowflakedb/snowflake-kafka-connector/issues/442

PR #587, opened 2023-03-28
SNOW-768488: Accept any offset from kafka if InvalidOffsetException is thrown during startup

sfc-gh-tzhang:
It's possible that customer stops their KC and the data stored in kafka expired, when the KC is restarted, it will use the channel's offset token and resetting the offset will result in an InvalidOffsetException, this PR handles this case and we will accept what kafka sends us instead


sfc-gh-tjones (approved):

sfc-gh-japatel (approved):
lgtm, thanks Toby!

PR #590, opened 2023-03-28
One client change 2/3: Add client provider

sfc-gh-rcheng:


src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+  }
+
+  /**
+   * Initialize the streaming client
+   *
+   * @param connectorConfig The connector config required to create the client
+   * @return An initialized client
+   */
+  private SnowflakeStreamingIngestClient initStreamingClient(Map<String, String> connectorConfig) {
```
sfc-gh-rcheng, line 168: This is basically the existing behavior copy-pasted from: https://github.com/snowflakedb/snowflake-kafka-connector/blob/1d12734da8b15c0c4ba846b86e902338a68f8827/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java#L472

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,208 @@
...
+  private SnowflakeStreamingIngestClient streamingIngestClient;
+
+  // private constructor for singleton
+  private StreamingClientProvider() {
+    this.createdClientId = 0;
+    this.connectorConfig = new HashMap<>();
+  }
+
+  @VisibleForTesting
+  // ONLY FOR TESTING - private constructor to inject properties for testing
```
sfc-gh-tzhang, line 82: nit: comment should be above the annotation

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,208 @@
...
+   *
+   * @param connectorConfig The connector config to define the client
+   */
+  public void createClient(Map<String, String> connectorConfig) {
+    // replace previous connector config and client if applicable
+    if (!this.connectorConfig.isEmpty()) {
+      LOGGER.warn("Overriding previous connector config");
+    }
+
+    if (isClientValid(this.streamingIngestClient)) {
```
sfc-gh-tzhang, line 103: looks like we already do client validation check inside `closeClient`

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,208 @@
...
+   *     invalid
+   * @return The current or newly created streaming client
+   */
+  public SnowflakeStreamingIngestClient getClient(Map<String, String> connectorConfig) {
+    if (Boolean.parseBoolean(
+        connectorConfig.get(
+            SnowflakeSinkConnectorConfig.ENABLE_STREAMING_CLIENT_OPTIMIZATION_CONFIG))) {
+      // recreate streaming client if needed
+      if (!isClientValid(this.streamingIngestClient)) {
+        LOGGER.error("Current streaming client is invalid, recreating client");
```
sfc-gh-tzhang, line 153: Why this is an error?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-tzhang: So the code in SnowflakeSinkServiceV2 will be removed later?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,208 @@
...
+  /**
+   * Creates a new streaming client, will replace the existing one if necessary
+   *
+   * @param connectorConfig The connector config to define the client
+   */
+  public void createClient(Map<String, String> connectorConfig) {
+    // replace previous connector config and client if applicable
+    if (!this.connectorConfig.isEmpty()) {
+      LOGGER.warn("Overriding previous connector config");
+    }
```
sfc-gh-tzhang, line 101: Is this a valid code path? If yes, then we should use INFO?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: moved, thanks

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: Good point, removed the check in favor of more descriptive INFO logging

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: It is not a valid code path, but I removed this warning as part of the close client validation comment earlier

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: yup, that will come in the next PR

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,201 @@
...
+   */
+  @VisibleForTesting
+  public static StreamingClientProvider injectStreamingClientProviderForTests(
+      int createdClientId,
+      Map<String, String> connectorConfig,
+      SnowflakeStreamingIngestClient client) {
+    return new StreamingClientProvider(createdClientId, connectorConfig, client);
+  }
+
+  private static final KCLogger LOGGER = new KCLogger(Utils.class.getName());
```
sfc-gh-japatel, line 68:
```suggestion
  private static final KCLogger LOGGER = new KCLogger(StreamingClientProvider.class.getName());
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.Properties;
+import net.snowflake.ingest.streaming.SnowflakeStreamingIngestClient;
+import net.snowflake.ingest.streaming.SnowflakeStreamingIngestClientFactory;
+import net.snowflake.ingest.utils.SFException;
+import org.apache.kafka.connect.errors.ConnectException;
+
+/** Singleton that provides the streaming client */
+public class StreamingClientProvider {
+  public static final StreamingClientProvider streamingClientProvider =
+      new StreamingClientProvider();
```
sfc-gh-japatel, line 40: I am personally not a fan of public static instance. 
I would rather make this a private and have a getter which returns the singleton `StreamingClientProvider`. This in true sense will make this a singleton. 


```suggestion
private static class Holder {
    private static final StreamingClientProvider INSTANCE = new StreamingClientProvider();
  }

  public static StreamingClientProvider getStreamingClientProviderInstance() {
    return Holder.INSTANCE;
  }

```

Also add docstring to public method.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,201 @@
...
+      int createdClientId,
+      Map<String, String> connectorConfig,
+      SnowflakeStreamingIngestClient client) {
+    return new StreamingClientProvider(createdClientId, connectorConfig, client);
+  }
+
+  private static final KCLogger LOGGER = new KCLogger(Utils.class.getName());
+  private static final String STREAMING_CLIENT_PREFIX_NAME = "KC_CLIENT_";
+
+  private int createdClientId;
```
sfc-gh-japatel, line 71: Should this be threadsafe? 
Use `AtomicInteger` instead.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Properties;
+import net.snowflake.ingest.streaming.SnowflakeStreamingIngestClient;
+import net.snowflake.ingest.streaming.SnowflakeStreamingIngestClientFactory;
+import net.snowflake.ingest.utils.SFException;
+import org.apache.kafka.connect.errors.ConnectException;
+
+/** Singleton that provides the streaming client */
```
sfc-gh-japatel, line 37: Lets clarify here that `StreamingClientProvider` is Singleton and there can be multiple streaming clients.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,201 @@
...
+
+  /**
+   * Gets the current client or creates a new one from the given connector config If client
+   * optimization is not enabled, just create a new streaming client
+   *
+   * @param connectorConfig The connector config, given as a backup in case the current client is
+   *     invalid
+   * @return The current or newly created streaming client
+   */
+  public SnowflakeStreamingIngestClient getClient(Map<String, String> connectorConfig) {
```
sfc-gh-japatel, line 140: I think this method should be threadsafe. We should also make all create and close client method thread safe.

sfc-gh-japatel: It didnt need to be threadsafe because every task would spin up a SnowflakeSinkServiceV2 instance which init a client before. 

If we are doing this at a singleton level, it needs to be threadsafe.

WDYT? 

Can you also add a test where in two separate threads call the `getClient` method?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,201 @@
...
+    this.connectorConfig = connectorConfig;
+    this.streamingIngestClient = client;
+  }
+
+  /**
+   * Creates a new streaming client, will replace the existing one if necessary
+   *
+   * @param connectorConfig The connector config to define the client
+   */
+  public void createClient(Map<String, String> connectorConfig) {
```
sfc-gh-japatel, line 97: lets rename this `updateOrCreateClient`

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,201 @@
...
+    // replace previous connector config and client if applicable
+    LOGGER.info("Creating new client, this will replace the old client and config if exists");
+    this.closeClient();
+
+    this.connectorConfig = connectorConfig;
```
sfc-gh-japatel, line 102: Do we need `connectorConfig`?

sfc-gh-japatel (changes requested):
Left few suggestions about thread safety. Let me know what you think.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: yup, we need it to get the streaming properties. its used in initStreamingClient()

sfc-gh-japatel: I meant it setting it at instance level? `this.connectorConfig`

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: thanks for the catch, updated!

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: I changed it to a public static method that will return the instance that is stored inside a private static subclass

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: Good call, using AtomicInteger

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: I agree it should be threadsafe, I've added a reentrant lock to ensure that it is. Will add a couple unit tests next

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: renamed to createOrReplaceClient

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: Ah I see, it is used soley for the getClientName method, which is used mostly for logging. Alternatively we could save just the clientName instead of the config. Open to either

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: Added unit tests, let me know how it looks

sfc-gh-rcheng:
recheck

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,236 @@
...
+
+  /**
+   * Creates a new streaming client, will replace the existing one if necessary
+   *
+   * @param connectorConfig The connector config to define the client
+   */
+  public void createOrReplaceClient(Map<String, String> connectorConfig) {
+    // replace previous connector config and client if applicable
+    this.clientLock.lock();
+    LOGGER.info("Creating new client, this will replace the old client and config if exists");
```
sfc-gh-japatel, line 118: this can be confusing, we are loggin creating new client and below in closeClient we are printing `Closing...`
Do we need a log line here?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-japatel: Lets pass in client name only. 
Setting connectorConfig for every createOrReplace call seems redundant and can have consequences in future because it is a mutable object. 


Do we really need this instance variable? Cant we just pass around without setting it at provider level?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,236 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+import net.snowflake.ingest.streaming.SnowflakeStreamingIngestClient;
+import net.snowflake.ingest.streaming.SnowflakeStreamingIngestClientFactory;
+import net.snowflake.ingest.utils.SFException;
+import org.apache.kafka.connect.errors.ConnectException;
+
+/**
+ * Singleton that provides the streaming client(s). There should only be one provider, but it may
+ * provide multiple clients
+ */
```
sfc-gh-japatel, line 43: I know I recommended but bringing this up again. It seems we are just replacing clients every time in the same provider instance since it is singleton. So if you are closing a client, only the last client is closed? 

Lets chat offline, it is also very much possible I am mixing things here.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,236 @@
...
+  /**
+   * Gets the current client or creates a new one from the given connector config If client
+   * optimization is not enabled, just create a new streaming client
+   *
+   * @param connectorConfig The connector config, given as a backup in case the current client is
+   *     invalid
+   * @return The current or newly created streaming client
+   */
+  public SnowflakeStreamingIngestClient getClient(Map<String, String> connectorConfig) {
+    if (Boolean.parseBoolean(
```
sfc-gh-japatel, line 164: lets add info about param is enabled/disabled..inside if and else case.

src/test/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProviderTest.java
```diff
@@ -0,0 +1,363 @@
...
+    SnowflakeStreamingIngestClient client2 = getClientRunnable2.getClient();
+    SnowflakeStreamingIngestClient client3 = getClientRunnable3.getClient();
+
+    assert client1.getName().contains(clientName);
+    assert client1.getName().equals(client2.getName());
+    assert client2.getName().equals(client3.getName());
+  }
+
+  @Test
+  public void testMultiThreadGetDisabledParam() {
```
sfc-gh-japatel, line 272: can you call closeClient API here in case optimization is false? I think it will throw NPE.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,236 @@
...
+      if (!isClientValid(this.streamingIngestClient)) {
+        LOGGER.error("Current streaming client is invalid, recreating client");
+        this.createOrReplaceClient(connectorConfig);
+      }
+
+      this.clientLock.unlock();
+
+      return this.streamingIngestClient;
+    } else {
+      return this.initStreamingClient(connectorConfig);
```
sfc-gh-japatel, line 179:
```suggestion
     this.streamingIngestClient = this.initStreamingClient(connectorConfig);
     return this.streamingIngestClient;
```

sfc-gh-japatel: Umm okay, ignore the suggestion, I dont think you can do this since this will have a side effect of getting a stale instance. (if you ever introduce a `getStreamingIngestClient`)

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: Removed the instance variable, but still requires a connectorConfig to get a client, since ive updated the getClient to also create a client if needed

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: removed force replace logic

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: discussed offline. after some additional thinking, i think its better to let the caller manage the client instance. The task already has threading considerations, so lets let the task take care of closing and managing the client

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: added as a debug log

src/test/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientConcurrencyTest.java
```diff
@@ -0,0 +1,434 @@
...
+import net.snowflake.ingest.streaming.SnowflakeStreamingIngestClient;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.mockito.Mockito;
+
+@RunWith(Parameterized.class)
+public class StreamingClientConcurrencyTest {
+  // note: these tests will leak clients, however the clients should autoclose, so it should be ok
```
sfc-gh-rcheng, line 36: The client extends Autocloseable, so I'm assuming its ok to leak a couple clients in this test?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -679,6 +684,19 @@ public static String formatString(String format, Object... vars) {
...
+  public static String safeGetExceptionMessage(String customMessage, Exception ex) {
+    String message =
+        ex.getMessage() == null || ex.getMessage().isEmpty()
+            ? "missing exception message"
+            : ex.getMessage();
+    String cause =
+        ex.getCause() == null || ex.getCause().getStackTrace() == null
+            ? "missing exception cause"
+            : Arrays.toString(ex.getCause().getStackTrace());
+
```
sfc-gh-japatel, line 696: lets use the private final you have declared above.

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -679,6 +684,19 @@ public static String formatString(String format, Object... vars) {
     return format;
   }
 
+  public static String safeGetExceptionMessage(String customMessage, Exception ex) {
```
sfc-gh-japatel, line 687: nit: Lets put javadocs on public method

sfc-gh-japatel, line 687: whats the reason to use `safe`? Can this be renamed to `generateExceptionMessage`? 
If you are worried about sending null, lets use Optional

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
```diff
@@ -0,0 +1,121 @@
...
+   * @return If the client is valid
+   */
+  public static boolean isClientValid(SnowflakeStreamingIngestClient client) {
+    return client != null && !client.isClosed() && client.getName() != null;
+  }
+
+  private static final KCLogger LOGGER = new KCLogger(StreamingClientHandler.class.getName());
+  private static final String STREAMING_CLIENT_PREFIX_NAME = "KC_CLIENT_";
+
+  private AtomicInteger createdClientId = new AtomicInteger(0);
```
sfc-gh-japatel, line 50: nit: lets move this to top of the class. (Keeping public and private methods/fields together)

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
```diff
@@ -0,0 +1,121 @@
...
+
+  /**
+   * Closes the given client. Swallows any exceptions
+   *
+   * @param client The client to be closed
+   */
+  public void closeClient(SnowflakeStreamingIngestClient client) {
+    // don't do anything if client is already invalid
+    if (!isClientValid(client)) {
+      return;
```
sfc-gh-japatel, line 103: lets add log line which will help debugging.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    LOGGER.info("Closing Streaming Client:{}", client.getName());
+    try {
+      client.close();
+    } catch (Exception e) {
+      // the client should auto close, so don't throw an exception here
```
sfc-gh-japatel, line 110: Is this true? 

The client will auto close only if its within a try with resources [block](https://www.baeldung.com/java-try-with-resources#replacing).

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
```diff
@@ -0,0 +1,121 @@
...
+      LOGGER.error(Utils.safeGetExceptionMessage("Failure closing Streaming client", e));
+    }
+  }
+
+  private String getNewClientName(Map<String, String> connectorConfig) {
+    return STREAMING_CLIENT_PREFIX_NAME
+        + connectorConfig.getOrDefault(Utils.NAME, "DEFAULT")
```
sfc-gh-japatel, line 117: nit: extract DEFAULT to a private static final variable. 
Also, this shouldnt happen since name is a required field, but just in case should we rename it to something more meaningful?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,114 @@
...
+  public static StreamingClientProvider injectStreamingClientProviderForTests(
+      SnowflakeStreamingIngestClient parameterEnabledClient,
+      StreamingClientHandler streamingClientHandler) {
+    return new StreamingClientProvider(parameterEnabledClient, streamingClientHandler);
+  }
+
+  private static final KCLogger LOGGER = new KCLogger(StreamingClientProvider.class.getName());
+  private SnowflakeStreamingIngestClient parameterEnabledClient;
+  private StreamingClientHandler streamingClientHandler;
+  private ReentrantLock providerLock;
```
sfc-gh-japatel, line 57: nit: 
```suggestion
  private Lock providerLock;
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,114 @@
...
+
+  // private constructor for singleton
+  private StreamingClientProvider() {
+    this.streamingClientHandler = new StreamingClientHandler();
+    providerLock = new ReentrantLock();
```
sfc-gh-japatel, line 62:
```suggestion
    providerLock = new ReentrantLock(true);
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,114 @@
...
+
+  // ONLY FOR TESTING - private constructor to inject properties for testing
+  @VisibleForTesting
+  private StreamingClientProvider(
+      SnowflakeStreamingIngestClient parameterEnabledClient,
+      StreamingClientHandler streamingClientHandler) {
+    this();
+    this.parameterEnabledClient = parameterEnabledClient;
+    this.streamingClientHandler = streamingClientHandler;
+  }
```
sfc-gh-japatel, line 73: keep all visible for testing annotated ctor's together

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,114 @@
...
+  private ReentrantLock providerLock;
+
+  // private constructor for singleton
+  private StreamingClientProvider() {
+    this.streamingClientHandler = new StreamingClientHandler();
+    providerLock = new ReentrantLock();
+  }
+
+  // ONLY FOR TESTING - private constructor to inject properties for testing
+  @VisibleForTesting
```
sfc-gh-japatel, line 66: do we need visible for testing for this? Its private

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,114 @@
...
+   * optimization is not enabled, it will create a new streaming client and the caller is
+   * responsible for closing it
+   *
+   * @param connectorConfig The connector config
+   * @return A streaming client
+   */
+  public SnowflakeStreamingIngestClient getClient(Map<String, String> connectorConfig) {
+    if (Boolean.parseBoolean(
+        connectorConfig.get(
+            SnowflakeSinkConnectorConfig.ENABLE_STREAMING_CLIENT_OPTIMIZATION_CONFIG))) {
```
sfc-gh-japatel, line 86: lets use getOrDefault for boolean.parseBoolean because ENABLE_STREAMING_CLIENT_OPTIMIZATION_CONFIG may or may not be present. 
Or use BooleanUtils from apache commons if we have that in our dependency which accepts null I believe. https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/BooleanUtils.html#isTrue-java.lang.Boolean-

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,114 @@
...
+import com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig;
+import com.snowflake.kafka.connector.internal.KCLogger;
+import java.util.Map;
+import java.util.concurrent.locks.ReentrantLock;
+import net.snowflake.ingest.streaming.SnowflakeStreamingIngestClient;
+
+/**
+ * Singleton/factory that provides the streaming client(s). There should only be one provider, but
+ * it may provide multiple clients
+ */
```
sfc-gh-japatel, line 30: nit: Singleton and Factory are two different concepts. (Not an or)

```suggestion
/**
 * Factory that provides the streaming client(s). There should only be one provider, but
 * it may provide multiple clients if optimizations are disabled. (Look at ENABLE_STREAMING_CLIENT_OPTIMIZATION_CONFIG)
 */
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -0,0 +1,114 @@
...
+package com.snowflake.kafka.connector.internal.streaming;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig;
+import com.snowflake.kafka.connector.internal.KCLogger;
+import java.util.Map;
+import java.util.concurrent.locks.ReentrantLock;
+import net.snowflake.ingest.streaming.SnowflakeStreamingIngestClient;
+
+/**
```
sfc-gh-japatel, line 27: Overall I like the distinct purposes of these two classes. Looks and reads so much better. Thank you!

src/test/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandlerTest.java
```diff
@@ -0,0 +1,142 @@
...
+    Mockito.verify(closedClient, Mockito.times(0)).getName();
+
+    // invalid client - no name
+    SnowflakeStreamingIngestClient unnamedClient =
+        Mockito.mock(SnowflakeStreamingIngestClient.class);
+    Mockito.when(unnamedClient.isClosed()).thenReturn(false);
+    Mockito.when(unnamedClient.getName()).thenReturn(null);
+    assert !StreamingClientHandler.isClientValid(unnamedClient);
+    Mockito.verify(unnamedClient, Mockito.times(1)).isClosed();
+    Mockito.verify(unnamedClient, Mockito.times(1)).getName();
```
sfc-gh-japatel, line 140: extract this into invalidClientTest? Or rename method to `testClientValidity`?

src/test/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientConcurrencyTest.java
```diff
@@ -0,0 +1,389 @@
...
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.mockito.Mockito;
+
+@RunWith(Parameterized.class)
+public class StreamingClientConcurrencyTest {
```
sfc-gh-japatel, line 39: Love the addition of these tests but this can be simplified if we use `CountDownLatch` and `ExecutorService`. 

A lof of ProviderClass can be removed and you can just focus on overall logic behind the test. 

https://www.baeldung.com/java-countdown-latch

src/test/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientConcurrencyTest.java
```diff
@@ -0,0 +1,389 @@
...
+    this.getClientStart(getClientCaller3);
+
+    // should create just once if param is enabled
+    this.getClientStop(this.getCallers)
+        .forEach(client -> Assert.assertTrue(StreamingClientHandler.isClientValid(client)));
+    Mockito.verify(
+            this.streamingClientHandler,
+            Mockito.times(this.enableClientOptimization ? 1 : this.getCallers.size()))
+        .createClient(Mockito.anyMap());
+  }
```
sfc-gh-japatel, line 216: Here is how countdownlatch would work. (Just an example)


```suggestion
  @Test
    public void testGetClientConcurrency() throws Exception {
        CountDownLatch latch = new CountDownLatch(10);

        // Create and submit tasks
        List<Future<SnowflakeStreamingIngestClient>> futures = new ArrayList<>();
        for (int i = 0; i < 10; i++) {
            Future<SnowflakeStreamingIngestClient> future = executorService.submit(() -> {
                SnowflakeStreamingIngestClient client = streamingClientProvider.getClient(clientConfig1);
                latch.countDown();
                return client;
            });
            futures.add(future);
        }

        // Waits for all tasks to complete
        latch.await();

        // Verify that clients are valid
        for (Future<SnowflakeStreamingIngestClient> future : futures) {
            Assert.assertTrue(StreamingClientHandler.isClientValid(future.get()));
        }

        // Verify that createClient() was called the expected number of times
        Mockito.verify(streamingClientHandler, Mockito.times(enableClientOptimization ? 1 : 10))
            .createClient(....);
    }
```

You can repeat this for all other tests and this doesnt require manual handling of threads/joining etc. (Makes your life much easier)

sfc-gh-japatel (commented):
Left few comments/suggestions and best practices. Let me know what you think.

Also, overall I like this decoupled logic a lot more. I will wait for one more pass before approving this.

sfc-gh-japatel:
Also, I am guessing we will add end to end test with optimization config and without and also add some IT tests?

sfc-gh-rcheng:
> Also, I am guessing we will add end to end test with optimization config and without and also add some IT tests?

yup! it is coming in this PR: https://github.com/snowflakedb/snowflake-kafka-connector/pull/609

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
sfc-gh-rcheng: Fair - I'm not sure its true, I can remove that message. I was looking into it as well, and we don't ever use a try with resources block, nor do we do anything additional if the close fails, so I'm not sure why it extends autocloseable if we never use the 'autoclosing'. I don't think it is possible for us to use the client in just a try with resources block

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
sfc-gh-rcheng: I added the getOrDefault because tests were failing there. Thanks for pointing out its a required field - I've changed DEFAULT to TEST_CLIENT to better signal that it should only happen in test cases

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: Thats true, the StreamingClientProvider is more of the factory that provides the StreamingClientProviderSingleton, since it can return any number of clients

src/test/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientConcurrencyTest.java
sfc-gh-rcheng: THAT IS VERY HELPFUL

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
sfc-gh-japatel: Yes we dont need to use it with try with resources and is not possible here because the client is long lived. Auto closeable is useful to not leak resources. (For instance filereader/writer)

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
sfc-gh-japatel: sounds good!

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: thanks for pushing on it, esp the concurrency stuff! it is a lot cleaner

sfc-gh-japatel (approved):
lgtm! Thank you for adding all tests and being patient on this one. 

Are we ready to push it to next release? If not, may be lets wait before we push it to master. 

Although I am okay to try it out with fallback to multiple clients since that is a config now. 

WDYT?

sfc-gh-rcheng:
> lgtm! Thank you for adding all tests and being patient on this one.
> 
> Are we ready to push it to next release? If not, may be lets wait before we push it to master.
> 
> Although I am okay to try it out with fallback to multiple clients since that is a config now.
> 
> WDYT?

I agree, lets not push the one client optimization in the next release. However I'd like to merge this PR since there is a lot of logic here I'd like to manage anymore, and it shouldn't break anything since it is not called in prod code. The next PR https://github.com/snowflakedb/snowflake-kafka-connector/pull/609 will actually call the one client optimization, and it can wait for the next release after I do more testing on it.

PR #593, opened 2023-03-30
SNOW-480523 Upgrade JDBC to 3.13.25 to use downscope gcs token instea…

sfc-gh-japatel:
…d of presigned URL

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -323,11 +323,24 @@
             </exclusions>
         </dependency>
 
+        <!--JDBC driver for building connection with Snowflake-->
+        <dependency>
+            <groupId>net.snowflake</groupId>
+            <artifactId>snowflake-jdbc</artifactId>
+            <version>3.13.25</version>
+        </dependency>
+
         <!-- Ingest SDK for copy staged file into snowflake table -->
         <dependency>
             <groupId>net.snowflake</groupId>
             <artifactId>snowflake-ingest-sdk</artifactId>
             <version>1.1.0</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>net.snowflake</groupId>
+                    <artifactId>snowflake-jdbc</artifactId>
+                </exclusion>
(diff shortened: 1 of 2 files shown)


test/test_verify.py
```diff
@@ -399,7 +399,16 @@ def createConnector(self, fileName, nameSalt):
 
         print("Post HTTP request to Create Connector:{0}".format(post_url))
         r = requests.post(post_url, json=json.loads(fileContent), headers=self.httpHeader)
-        print(datetime.now().strftime("%H:%M:%S "), r.status_code)
+        print("Connector Name:{0} POST Response:{1}".format(snowflake_connector_name, r.status_code), datetime.now().strftime("%H:%M:%S "))
```
sfc-gh-japatel, line 402: @sfc-gh-rcheng

pom.xml
```diff
@@ -323,11 +323,24 @@
             </exclusions>
         </dependency>
 
+        <!--JDBC driver for building connection with Snowflake-->
+        <dependency>
+            <groupId>net.snowflake</groupId>
+            <artifactId>snowflake-jdbc</artifactId>
+            <version>3.13.25</version>
```
sfc-gh-tzhang, line 330: What's the reason of using 3.13.25 instead of other high versions? Do you know which version other SDKs/connectors are using?

sfc-gh-japatel: https://snowflake.slack.com/archives/C04URTJMC77/p1680152983458869

sfc-gh-japatel: On top of that
My philosophy:
1. We have seen issues in past by upgrading to the latest versions. 
2. The lesser the code changes lesser the probability of bugs. 3.13.25 is the min version we will have to support, so lets go with that. 


But again, I am also okay with upgrading to the latest and greatest.

sfc-gh-tzhang (approved):
lgtm, thanks

PR #598, opened 2023-03-31
NO-SNOW Refactor Test suites  + Disable failing tests to unblock other PRs

sfc-gh-japatel:
- Refactor test suites to make it easier for enabling and disabling a test. 
  - There are 30 tests right now and I cant tell whether a test is enabled or disabled in confluent and apache. 
- Disable two failing tests - SNOW-774533 for re-enabling them
  - One is DLQ test which suddenly started failing.
    - I dont have ideas on why this is failing. Will fix that later since other PRs are blocked. 
    - It is very much possible server side we changed something because ingest SDK is unchanged. 
  - Other is a test containing decryption of fips
   - My suspicion is maven caching has issues because fips library does weird checksum checks of jar files. 
   - Even after deletion of cache, I saw failures so something very weird is happening.
- Add while loop ending method in DLQ consumption. 
- Disable maven caching to eliminate why tests failing for encryption of private key keys.

### DIFF
--- a/test/run_test_confluent.sh
+++ b/test/run_test_confluent.sh
@@ -188,7 +188,5 @@ if [ $testError -ne 0 ]; then
     NC='\033[0m' # No Color
     echo -e "${RED} There is error above this line ${NC}"
     cat $APACHE_LOG_PATH/kc.log
-    echo "Output kafka logs for debugging"
-    cat $APACHE_LOG_PATH/kafka.log
     error_exit "=== test_verify.py failed ==="
 fi
--- a/test/test_suites.py
+++ b/test/test_suites.py
@@ -120,7 +120,7 @@ def create_test_suites(driver, nameSalt, schemaRegistryAddress, testSet):
             test_instance=TestSchemaMapping(driver, nameSalt), clean=True, run_in_confluent=True,run_in_apache=True
         )),
         ("TestSnowpipeStreamingSchemaMappingDLQ", TestSuite(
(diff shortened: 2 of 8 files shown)


sfc-gh-tjones (approved):

test/run_test_confluent.sh
```diff
@@ -188,7 +188,5 @@ if [ $testError -ne 0 ]; then
     NC='\033[0m' # No Color
     echo -e "${RED} There is error above this line ${NC}"
     cat $APACHE_LOG_PATH/kc.log
-    echo "Output kafka logs for debugging"
-    cat $APACHE_LOG_PATH/kafka.log
```
sfc-gh-rcheng, line 192: thank you, the logs bloated the github actions a lot

test/test_suites.py
```diff
@@ -0,0 +1,170 @@
...
+
+from test_suit.test_schema_evolution_nonnullable_json import TestSchemaEvolutionNonNullableJson
+
+from test_suit.test_schema_not_supported_converter import TestSchemaNotSupportedConverter
+
+from test_suit.test_schema_evolution_drop_table import TestSchemaEvolutionDropTable
+
+from test_suit.test_snowpipe_streaming_schema_mapping_dlq import TestSnowpipeStreamingSchemaMappingDLQ
+
+class EndToEndTestSuit:
```
sfc-gh-rcheng, line 46: nit: EndToEndTestSuite

sfc-gh-japatel: Ack, changed it to Suite which is more suit'able!

PR #601, opened 2023-04-05
Add telemetry for fatal errors in TopicPartitionChannel

sfc-gh-japatel:
- Also add default ingestion method in all msg in telemetry. Using ordinal to save some bytes. 
- Sending fatal errors in TopicPartitionChannel. 
  - Not sending all channel reopen events since that data is already present in Snowflake EventLogger.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/IngestionMethodConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/IngestionMethodConfig.java
@@ -55,7 +55,8 @@ public enum IngestionMethodConfig {
   /**
    * Returns the Ingestion Method found in User Configuration for Snowflake Kafka Connector.
    *
-   * <p>Default is always {@link IngestionMethodConfig#SNOWPIPE}.
+   * <p>Default is always {@link IngestionMethodConfig#SNOWPIPE} unless an invalid value is passed
+   * which results in exception.
    */
   public static IngestionMethodConfig determineIngestionMethod(Map<String, String> inputConf) {
     if (inputConf == null
@@ -69,7 +70,7 @@ public enum IngestionMethodConfig {
                 .get(SnowflakeSinkConnectorConfig.INGESTION_METHOD_OPT)
                 .toUpperCase(Locale.ROOT));
       } catch (IllegalArgumentException ex) {
-        return IngestionMethodConfig.SNOWPIPE;
+        throw ex;
       }
     }
   }
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/IngestionMethodConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/IngestionMethodConfig.java
@@ -53,15 +53,16 @@ public enum IngestionMethodConfig {
   }
 
   /**
-   * Returns the Ingestion Method found in User Configuration for Snowflake Kafka Connector. Default
-   * is always {@link IngestionMethodConfig#SNOWPIPE}.
+   * Returns the Ingestion Method found in User Configuration for Snowflake Kafka Connector.
+   *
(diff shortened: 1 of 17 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/IngestionMethodConfig.java
```diff
@@ -48,6 +52,28 @@ public static String[] allIngestionTypes() {
...
 
+  /**
+   * Returns the Ingestion Method found in User Configuration for Snowflake Kafka Connector.
+   *
+   * <p>Default is always {@link IngestionMethodConfig#SNOWPIPE}.
+   */
+  public static IngestionMethodConfig determineIngestionMethod(Map<String, String> inputConf) {
+    if (inputConf == null
+        || inputConf.isEmpty()
+        || !inputConf.containsKey(SnowflakeSinkConnectorConfig.INGESTION_METHOD_OPT)) {
```
sfc-gh-tzhang, line 63: do we need to ignore the cases in this check?

src/main/java/com/snowflake/kafka/connector/internal/streaming/IngestionMethodConfig.java
```diff
@@ -48,6 +52,28 @@ public static String[] allIngestionTypes() {
...
+      return IngestionMethodConfig.SNOWPIPE;
+    } else {
+      try {
+        return IngestionMethodConfig.valueOf(
+            inputConf
+                .get(SnowflakeSinkConnectorConfig.INGESTION_METHOD_OPT)
+                .toUpperCase(Locale.ROOT));
+      } catch (IllegalArgumentException ex) {
+        return IngestionMethodConfig.SNOWPIPE;
```
sfc-gh-tzhang, line 72: Should we error out in this case? This could lead to customer confusions

src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryService.java
```diff
@@ -151,15 +161,17 @@ public void reportKafkaPartitionStart(final SnowflakeTelemetryBasicInfo pipeCrea
...
    * </pre>
    *
    * @return An ObjectNode which is by default always created with certain defined properties in it.
    */
-  protected ObjectNode getObjectNode() {
+  protected ObjectNode getDefaultObjectNode(IngestionMethodConfig ingestionMethodConfig) {
     ObjectNode msg = MAPPER.createObjectNode();
     msg.put(APP_NAME, getAppName());
     msg.put(TASK_ID, getTaskID());
+    msg.put(SnowflakeSinkConnectorConfig.INGESTION_METHOD_OPT, ingestionMethodConfig.ordinal());
```
sfc-gh-tzhang, line 174: Is this possible we put the string instead of ordinal? Ordinal is not straight forward and we have to remember the mappings

sfc-gh-tzhang (approved):
Left some suggestions, otherwise LGTM!

src/main/java/com/snowflake/kafka/connector/internal/streaming/IngestionMethodConfig.java
sfc-gh-japatel: This shouldnt happen as such since we do validations on connector start but I can see this could be confusing if used elsewhere, let me throw an error here.

src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryService.java
sfc-gh-japatel: I had kept string originally but I guess we only have two ingestion types(right now) so its not terrible to remember this. I was thinking we could save some storage since these are constants and gets added to all telemetry.

PR #604, opened 2023-04-07
SNOW-774533 Fix DLQ bug, re-enable test

sfc-gh-japatel:
- Well, well, well, guess DLQ was not working. (When we manually try to insert it from Kafka Connect)
- Works if serialization/deserialization issue. 
- Fixing the bug and re-enabling the test. 

- Was working un-expectedly because the test was disabled in the previous refactoring of confluent and apache runs. 
 - Now we know which tests are running and disabled/enabled. Less noise, less flaky.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
@@ -204,7 +204,7 @@ public class SnowflakeSinkTask extends SinkTask {
     enableRebalancing =
         Boolean.parseBoolean(parsedConfig.get(SnowflakeSinkConnectorConfig.REBALANCING));
 
-    KafkaRecordErrorReporter kafkaRecordErrorReporter = noOpKafkaRecordErrorReporter();
+    KafkaRecordErrorReporter kafkaRecordErrorReporter = createKafkaRecordErrorReporter();
 
     // default to snowpipe
     IngestionMethodConfig ingestionType = IngestionMethodConfig.SNOWPIPE;
@@ -450,6 +450,10 @@ public class SnowflakeSinkTask extends SinkTask {
               (record, error) -> {
                 try {
                   // Blocking this until record is delivered to DLQ
+                  DYNAMIC_LOGGER.debug(
+                      "Sending Sink Record to DLQ with recordOffset:{}, partition:{}",
+                      record.kafkaOffset(),
+                      record.kafkaPartition());
                   errantRecordReporter.report(record, error).get();
                 } catch (InterruptedException | ExecutionException e) {
                   final String errMsg = "ERROR reporting records to ErrantRecordReporter";
@@ -465,6 +469,8 @@ public class SnowflakeSinkTask extends SinkTask {
         this.DYNAMIC_LOGGER.info(
             "Kafka versions prior to 2.6 do not support the errant record reporter.");
       }
+    } else {
+      DYNAMIC_LOGGER.warn("SinkTaskContext is not set");
     }
     return result;
   }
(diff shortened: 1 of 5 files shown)


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -450,6 +450,10 @@ private KafkaRecordErrorReporter createKafkaRecordErrorReporter() {
               (record, error) -> {
                 try {
                   // Blocking this until record is delivered to DLQ
+                  DYNAMIC_LOGGER.debug(
```
sfc-gh-tzhang, line 453: Looks like we don't have any tests that go into this branch?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -491,6 +497,9 @@ private String getTaskLoggingTag() {
    */
   @VisibleForTesting
   static KafkaRecordErrorReporter noOpKafkaRecordErrorReporter() {
-    return (record, e) -> {};
+    return (record, e) -> {
+      STATIC_LOGGER.warn(
```
sfc-gh-tzhang, line 501: same here, is it easy to add a test?

test/test_suites.py
```diff
@@ -1,6 +1,5 @@
 from collections import OrderedDict
 from test_suit.test_string_json import TestStringJson
-from test_suit.test_string_json_proxy import TestStringJsonProxy
```
sfc-gh-tzhang, line 3: Any reason this is removed?

sfc-gh-tzhang (approved):
LGTM, please add a test to coverage the changes if possible

test/test_suites.py
sfc-gh-japatel: Unused import.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-japatel: We have an in memory queue implementation which is tested. 

It's hard to test this in Java. But let me see if we can add more coverage for this.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-japatel: Added a test, a lot of plumbing but I would say worth it. We now have unit test, IT and E2E test coverage.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-japatel: Yes, this is essentially a no-op. E2E tests covers this.

sfc-gh-japatel: Added coverage but only through E2E test.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-japatel: Yes, this is essentially a no-op. E2E tests covers this.
Added a coverage through an in memory implementation of the error reporter (which can be thought of as a DLQ)

test/test_suites.py
sfc-gh-japatel: proxy test is added in test_verify towards the end which I recently enabled.

sfc-gh-japatel:
I have read the CLA Document and I hereby sign the CLA

PR #614, opened 2023-04-12
[SNOW-726924] Add New jvm.nonProxy.hosts Parameter - Undo Revert 

sfc-gh-japatel:
Readding #533 

The jvm.nonProxy.hosts parameter can be added to the connector's configuration file and is the equivalent to the JVM argument -Dhttp.nonProxyHosts. If a user provides that parameter and the connector runs in a JVM where -Dhttp.nonProxyHosts was set, then we will merge the two lists of hosts. In the future, we'll want another method of adding proxy configurations without resorting to System.setProperty() calls

- Jay - Rebasing master, picking 3.13.25 and other changes.


sfc-gh-tzhang (approved):
Thanks Jay, could you help with adding back the change in the SDK as well?

sfc-gh-japatel:
> Thanks Jay, could you help with adding back the change in the SDK as well?

Yes, that PR is next!

PR #617, opened 2023-04-18
Add resilience testing

sfc-gh-rcheng:
Test create, resume, delete, pause and restart connector in different method combinations, with/without pressure (data being sent to the connector) while the method is executing


test/test_suites.py
```diff
@@ -80,8 +93,9 @@ def create_end_to_end_test_suites(driver, nameSalt, schemaRegistryAddress, testS
     :return:
     '''
     test_suites = OrderedDict([
+        # Disable Failing tests only in confluent because of fips error, re-enable: SNOW-774533
```
sfc-gh-japatel, line 96: nit: rebase master, this will disable the test, I enabled it in more recent commits.

test/test_suit/resilience_tests/test_kc_delete_create_pressure.py
```diff
@@ -0,0 +1,76 @@
+import datetime
+
+from test_suit.test_utils import RetryableError, NonRetryableError, ResetAndRetry
+import json
+from time import sleep
+
+class TestKcDeleteCreatePressure:
```
sfc-gh-japatel, line 7: whats the difference between this and `TestKcDeleteCreate`? 
Pressure test as far as I know does a copy unload from a table and reloads it into topic. I see only 1k records. I was wondering if we could combine those two? (Just to reduce the e2e test and noise if there is something wrong)

sfc-gh-rcheng: "pressure" is just a test that sends data between the two methods. I made the distinction because the customer's issue with v1.9.0 only happens if there is data ingesting while the connector is paused then resumed. Just pausing/resuming the connector does not raise an error

test_kc_delete_create_pressure: delete connector, send data, create connector
test_kc_delete_create: delete connector, create connector

sfc-gh-japatel: Got it, thanks for explaning! 
Lets add a description on each file in python class.

sfc-gh-japatel:
> can we also add a test to restart a task? `POST /connectors/(string:name)/tasks/(int:taskid)/restart` https://docs.confluent.io/platform/current/connect/references/restapi.html#topics
> 
> I think this handles different interfaces and code path. (Similar to rebalance)

Just found out about this: https://kafka.apache.org/documentation/#connect_rest
You can restart tasks and connector both in one API. 
`POST /connectors/{name}/restart?includeTasks=<true|false>&onlyFailed=<true|false> - restart a connector and its tasks instances.`

test/test_suit/resilience_tests/test_kc_delete_create.py
```diff
@@ -4,6 +4,11 @@
 import json
 from time import sleep
 
+# sends data 1/2
```
sfc-gh-japatel, line 7: 💯 thanks

test/test_suit/resilience_tests/test_kc_delete_create.py
```diff
@@ -0,0 +1,74 @@
+import datetime
+
+from test_suit.test_utils import RetryableError, NonRetryableError, ResetAndRetry
+import json
+from time import sleep
+
+class TestKcDeleteCreate:
```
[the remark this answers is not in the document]
sfc-gh-rcheng: added execution steps and expected result for each test

test/test_suites.py
sfc-gh-rcheng: reverted, unsure how that snuck in 😅

sfc-gh-rcheng:
> Also, just FYI, I am super happy we are adding a lot of tests but this is increasing overall runtime of end to end tests. From 11 mins (check latest master branch) to 19 mins now in this PR.
> 
> If possible lets combine most of the tests together. WDYT?

We can potentially combine the _pressure tests, however not every test ends in a happy state (delete_resume will have latent unsent data, so simply recreating the connector between the nonpressure and pressure tests will not accurately test the functionality). 

I'd also like to combine the tests, but I'm not sure how to since these files do test different behavior. Perhaps we'd need to do a larger refactoring to make each test setup lighter/more customized for smaller tests like these, however that would be out of scope.

test/test_verify.py
```diff
@@ -316,6 +316,11 @@ def restartConnector(self, connectorName):
         r = requests.post(requestURL, headers=self.httpHeader)
         print(datetime.now().strftime("%H:%M:%S "), r, " restart connector")
 
+    def restartConnectorAndTasks(self, connectorName):
```
sfc-gh-japatel, line 319: thanks!

sfc-gh-japatel:
> > Also, just FYI, I am super happy we are adding a lot of tests but this is increasing overall runtime of end to end tests. From 11 mins (check latest master branch) to 19 mins now in this PR.
> > If possible lets combine most of the tests together. WDYT?
> 
> We can potentially combine the _pressure tests, however not every test ends in a happy state (delete_resume will have latent unsent data, so simply recreating the connector between the nonpressure and pressure tests will not accurately test the functionality).
> 
> I'd also like to combine the tests, but I'm not sure how to since these files do test different behavior. Perhaps we'd need to do a larger refactoring to make each test setup lighter/more customized for smaller tests like these, however that would be out of scope.

Sounds good, lets track this in a jira somewhere to reduce the runtime of all of these tests.

sfc-gh-japatel (approved):
lgtm! Thanks
(Should we instead use `Chaos` in place of `Pressure` in all tests?)

PR #625, opened 2023-04-28
SNOW-791549: Infer NULL to VARCHAR instead of VARIANT since ingestion timestamp into VARIANT column is not allowed

sfc-gh-tzhang:
when the column data is NULL, we don't know which data to infer, so we use VARIANT currently, but VARIANT doesn't work with timestamp using the current SDK, this PR will infer the data type to VARCHAR instead of VARIANT.


sfc-gh-japatel (approved):
lgtm!

PR #628, opened 2023-05-10
Enable topic regexes for table2topic mapping

sfc-gh-rcheng:
Currently each topic is specifically mapped to each table, but we've had multiple customer asks to add regexes so that they can map many topics to one table. 

Notes:
- `.*` is the regex that's being tested, though technically other regexes could work since the code uses string.match(). 
- `.*` is also the 'catch all regex' that has been special cased. Any topic that doesn't have a table will point to the 'catch all table' if there is a catch all regex

parseTopic2TableMap:
- validates that no topic regexes overlaps (matches with each other) except the catch all regex

generateValidName:
- tries matching the topicPartition name to the topic regexes in the table
- returns the catchall table if exists and cannot match to another table


src/test/java/com/snowflake/kafka/connector/SinkTaskIT.java
```diff
@@ -323,4 +328,126 @@ public void testMultipleSinkTasksWithLogs() throws Exception {
     assert offsetMap1.get(topicPartitions0.get(0)).offset() == BUFFER_COUNT_RECORDS_DEFAULT;
     assert offsetMap0.get(topicPartitions1.get(0)).offset() == BUFFER_COUNT_RECORDS_DEFAULT;
   }
+
+  @Test
+  public void testTopicToTableMapParseAndCreation() {
+    // TODO @rcheng question: should i duplicate this test for streaming?
```
sfc-gh-rcheng, line 334: Should I duplicate this test in SnowflakeSinkTaskForStreamingIt? I didn't because i don't think this parsing and table name generation will differ across streaming vs snowpipe.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -124,6 +124,15 @@ public SnowflakeSinkTask(
...
 
+  @VisibleForTesting
+  public SnowflakeSinkTask(
+      SnowflakeSinkService service,
+      SnowflakeConnectionService connectionService,
+      Map<String, String> topic2table) {
+    this(service, connectionService);
+    this.topic2table = topic2table;
+  }
+
```
sfc-gh-tjones, line 135: FYI Codecov is claiming lines 134-135 are not covered via tests.

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -102,6 +102,8 @@ public class Utils {
 
   private static final KCLogger LOGGER = new KCLogger(Utils.class.getName());
 
+  private static final String CATCH_ALL_REGEX = ".*";
```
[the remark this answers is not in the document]
sfc-gh-tjones: I think to make things simper we can preventing the overlap. If we see demand for a catch-all case we can add it back in.

src/test/java/com/snowflake/kafka/connector/SinkTaskIT.java
sfc-gh-japatel: Might be worth adding it to only `SnowflakeSinkTaskForStreamingIt` but this is fine as well. probably in future we might get away with snowpipe code but not streaming and it is easy enough to forget adding it in streaming but removing it from here.

test/rest_request_template/test_regex_topic_to_table.json
```diff
@@ -0,0 +1,27 @@
...
+    "buffer.flush.time": "10",
+    "buffer.count.records": "100",
+    "buffer.size.bytes": "5000000",
+    "snowflake.url.name": "SNOWFLAKE_HOST",
+    "snowflake.user.name": "SNOWFLAKE_USER",
+    "snowflake.private.key": "SNOWFLAKE_PRIVATE_KEY",
+    "snowflake.database.name": "SNOWFLAKE_DATABASE",
+    "snowflake.schema.name": "SNOWFLAKE_SCHEMA",
+    "snowflake.role.name": "SNOWFLAKE_ROLE",
+    "snowflake.ingestion.method": "SNOWPIPE",
```
sfc-gh-japatel, line 17: lets add for streaming too?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-rcheng: Thanks, trying out a couple things to ignore those lines

test/rest_request_template/test_regex_topic_to_table.json
sfc-gh-rcheng: Duplicated the e2e test and IT for streaming as well

src/test/java/com/snowflake/kafka/connector/SinkTaskIT.java
```diff
@@ -323,4 +328,135 @@ public void testMultipleSinkTasksWithLogs() throws Exception {
...
+
+  @Test
+  public void testTopicToTableRegex() {
+    Map<String, String> config = TestUtils.getConf();
+    SnowflakeSinkConnectorConfig.setDefaultValues(config);
+
+    testTopicToTableRegexMain(config);
+  }
+
+  // runner for topic to table regex testing, used to test both streaming and snowpipe scenarios.
```
sfc-gh-rcheng, line 340: This is a little hacky, but I'd rather do this than copy-paste the test twice. I tried moving it to testUtils, however the scope of SnowflakeSinkConnectorConfig.TOPICS_TABLES_MAP and a couple other variables were not accessible. 

Any thoughts on how to improve this?

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskForStreamingIT.java
```diff
@@ -303,4 +307,135 @@ public void testSinkTaskWithMultipleOpenClose() throws Exception {
...
   }
+
+  @Test
+  public void testTopicToTableRegex() {
+    Map<String, String> config = TestUtils.getConfForStreaming();
+
+    testTopicToTableRegexMain(config);
+  }
+
+  // runner for topic to table regex testing, used to test both streaming and snowpipe scenarios.
```
sfc-gh-rcheng, line 318: This is a little hacky, but I'd rather do this than copy-paste the test twice. I tried moving it to testUtils, however the scope of SnowflakeSinkConnectorConfig.TOPICS_TABLES_MAP and a couple other variables were not accessible.

Any thoughts on how to improve this?

src/test/java/com/snowflake/kafka/connector/SinkTaskIT.java
sfc-gh-rcheng: Thats a good point, thanks! I created a static runner inside streamingIt that is called from sinkTaskIt. Opened a different thread about best practices for that, so resolving this thread

src/main/java/com/snowflake/kafka/connector/Utils.java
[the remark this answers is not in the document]
sfc-gh-rcheng: removed the catchall case, made the tests a lot more straightforward. thanks!

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-tjones: Now it says lines 135-136 :)

sfc-gh-tjones (approved):

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-rcheng: code coverage check isn't complaining anymore, resolving

PR #631, opened 2023-05-12
NO-SNOW: fix commit offset behavior for Snowpipe

sfc-gh-tzhang:
The change introduced in https://github.com/snowflakedb/snowflake-kafka-connector/commit/afd3588a96b82216db62c91cfbbd3c62fb33e130 was not meant for Snowpipe and it's causing reset of offset to 0 during KC restart, this PR contains the change to revert back to the old behavior for Snowpipe


sfc-gh-japatel (approved):
lgtm! Thanks for quick fix

PR #632, opened 2023-05-12
Release version 1.9.3

sfc-gh-tzhang:
- Upgrades the version to 1.9.3
- Release notes will be available in release tag.


sfc-gh-tjones (approved):

sfc-gh-japatel (approved):
lgtm! Lets do a release on monday, I do want to spend time on channel close issue today and see if we can release it altogether. 

If you are worried about usage/impact of 1.9.2, we can quickly find out how many customers are using it from client telemetry view.

PR #634, opened 2023-05-16
SNOW-480523 Fix Downscope URL for GCS

sfc-gh-japatel:
***TLDR:***
- Downscope URL in GCS doesnt work with version 1.9.2 (Thanks to @sfc-gh-tzhang for identifying the issue)

***Fix:***
- Add destFileName in uploadWithoutConnection API
- Released version 1.1.4 for ingest sdk which brings in jdbc 3.13.30
- Upgraded jdbc version in KC to 3.13.30
- Why not simply upgrade it here and exclude it from ingest-sdk? That's actually what I tried but thanks to shading. 
  - Shading in ingesy sdk brings complication about whats being chosen by the kafka connect during runtime. 
- Another workaround was to use a unshaded `1.1.3-unshaded` but it has its own downsides. (Jar size and more vulnerabilities because of relocation of jars)
- We decided to release 1.1.4 in ingest sdk and simply use it here. 

Few things that helped me debug:
- 3.13.29 class files were used by kafka connect environment. 
- You can enable JDBC tracing and looking at the output. 
```
export JDBC_TRACE=true
KAFKA_OPTS="-Djava.io.tmpdir=/tmp/jdbclogs/" 
```
Lookout for this debug statement. (Prints out version, for me it always used 3.13.29 and then this https://github.com/snowflakedb/snowflake-ingest-java/issues/356 made me think about all shading fuzziness)
```Trying to establish session, JDBC driver version: 3.13.30```

### DIFF
--- a/test/test_verify.py
+++ b/test/test_verify.py
@@ -109,7 +109,7 @@ class KafkaTest:
                   self.VERIFY_INTERVAL), flush=True)
         sleep(self.VERIFY_INTERVAL)
 
-    def verifyWithRetry(self, func, round):
+    def verifyWithRetry(self, func, round, configFileName):
         retryNum = 0
         while retryNum < self.MAX_RETRY:
             try:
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: test/test_verify.py [hunk 1]
```diff
@@ -109,33 +109,33 @@ def verifyWaitTime(self):
                   self.VERIFY_INTERVAL), flush=True)
         sleep(self.VERIFY_INTERVAL)
 
-    def verifyWithRetry(self, func, round):
+    def verifyWithRetry(self, func, round, configFileName):
```
sfc-gh-tzhang, line 112: Can we add a test to verify both should work with and without the downscoped token enabled?

sfc-gh-japatel: I cant think of an easier way to do it over a public internet. The only way I know is to enable/disable it is do it from snowflake account and run this tests again. Or, just set up another user for our tests and make it enable so we can run two tests in parallel. It's a lot of plumbing and I don't think it worth the efforts. 
https://github.com/snowflakedb/snowflake-kafka-connector/pull/634#issuecomment-1553576742

sfc-gh-tzhang (approved):
LGTM, could you work with @sfc-gh-rcheng to see if the one client change could be included in 1.9.3?

sfc-gh-japatel:
> LGTM, could you work with @sfc-gh-rcheng to see if the one client change could be included in 1.9.3?

Yes, @sfc-gh-rcheng and I have already communicated about bringing her one client change first. I will push this after that is pushed.

sfc-gh-tzhang:
> Tests are passing with downscope parameter enabled. I am going to turn off the parameter and run the tests again to make sure there is no regression since DOWNSCOPE URL is not by default turned on for GCS.

Please also make sure that we don't get the presigned url when the downscope parameter is enabled, thanks!

sfc-gh-japatel:
> > Tests are passing with downscope parameter enabled. I am going to turn off the parameter and run the tests again to make sure there is no regression since DOWNSCOPE URL is not by default turned on for GCS.
> 
> Please also make sure that we don't get the presigned url when the downscope parameter is enabled, thanks!

Yes, I had added that while I was debugging: You can take a look at history of the commits: 4fa4840e1254aaca4c500ee64a8d519508bd7892

PR #637, opened 2023-05-24
Revert kafka put and precommit log message to info

sfc-gh-rcheng:
Most SnowflakeSinkTask from debug to info for debugging purposes. Also update the precommit log to include the number of partitions instead of record count. Note that this may increase log size, however they are very nice to have for debugging.

Example log:
`[SF_KAFKA_CONNECTOR] Successfully called PUT with 10000 records, executionTime: 1643 ms`
`[SF_KAFKA_CONNECTOR] Successfully called PRECOMMIT on all 1 partitions, safe to commit 1 partitions, executionTime: 744 ms`

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
@@ -406,11 +406,11 @@ public class SnowflakeSinkTask extends SinkTask {
           size,
           executionTime);
     } else {
-      this.DYNAMIC_LOGGER.debug(
+      this.DYNAMIC_LOGGER.info(
           "successfully called {} with {} records, execution time: {} seconds",
           apiName,
           size,
-          getExecutionTimeSec(startTime, System.currentTimeMillis()));
+          executionTime);
     }
   }
 
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
@@ -395,23 +395,22 @@ public class SnowflakeSinkTask extends SinkTask {
 
   void logWarningForPutAndPrecommit(long startTime, int size, String apiName) {
     long executionTimeMs = getDurationFromStartMs(startTime);
-    long executionTimeSec = executionTimeMs / 1000;
-    if (executionTimeSec > 300) {
+    if (executionTimeMs > 300000) {
       // This won't be frequently printed. It is vary rare to have execution greater than 300
       // seconds.
       // But having this warning helps customer to debug their Kafka Connect config.
       this.DYNAMIC_LOGGER.warn(
-          "{} {}. Time: {} milliseconds = {} seconds > 300 seconds. If there is"
+          "{} {}. Time: {} ms = {} seconds > 300 seconds. If there is"
               + " CommitFailedException in the log or there is duplicated records, refer to this"
               + " link for solution: "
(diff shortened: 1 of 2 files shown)


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -388,29 +388,33 @@ private static void waitFor(Supplier<Boolean> func)
...
+  private static long getDurationFromStartMs(long startTime) {
+    final long currTime = System.currentTimeMillis();
+    return currTime - startTime;
   }
 
   void logWarningForPutAndPrecommit(long startTime, int size, String apiName) {
-    long executionTime = getExecutionTimeSec(startTime, System.currentTimeMillis());
-    if (executionTime > 300) {
+    long executionTimeMs = getDurationFromStartMs(startTime);
+    long executionTimeSec = executionTimeMs / 1000;
```
sfc-gh-tzhang, line 398: why do we need the sec here as well?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -388,29 +388,33 @@ private static void waitFor(Supplier<Boolean> func)
...
           apiName,
           size,
-          executionTime);
+          executionTimeMs,
+          executionTimeSec);
     } else {
-      this.DYNAMIC_LOGGER.debug(
-          "successfully called {} with {} records, execution time: {} seconds",
+      this.DYNAMIC_LOGGER.info(
+          "successfully called {} with {} records, execution time: {} milliseconds",
```
sfc-gh-tzhang, line 414: Maybe try to reduce the words here if the concern is length

sfc-gh-tzhang (approved):

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-rcheng: I retained the sec because 300 sec is more user friendly than a 300,000 milisecond timeout

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-rcheng: changed to ms

sfc-gh-tzhang (approved):

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -388,29 +388,31 @@ private static void waitFor(Supplier<Boolean> func)
...
-          executionTime);
+          executionTimeMs,
+          executionTimeMs / 1000);
     } else {
-      this.DYNAMIC_LOGGER.debug(
-          "successfully called {} with {} records, execution time: {} seconds",
+      this.DYNAMIC_LOGGER.info(
+          "successfully called {} with {} records, execution time: {} ms",
           apiName,
           size,
```
sfc-gh-japatel, line 414: can you fix the records thing as well? just may be check apiName and if it is precommit, print partitions. (It's hacky but works for now)

sfc-gh-rcheng: Yup, it now prints the total number of offsets and the number of commitable offsets. updated the description with examples of the logs

sfc-gh-japatel (approved):
lgtm!

sfc-gh-japatel (approved):
lgtm!

sfc-gh-tjones (approved):

PR #639, opened 2023-05-26
[Buffer Flush 1/2] Refactor buffer threshold variables and methods to be more consistent

sfc-gh-rcheng:
The existing variable and method names in bufferthreshold are not consistent. Yes this is a very nitpicky PR, but I would like to make these more consistent for the changes ill be making here in the future.

https://github.com/snowflakedb/snowflake-kafka-connector/pull/638 uses the changes

### Variable Changes
flushTimeThresholdSeconds -> bufferFlushTimeThreshold
bufferSizeThresholdBytes -> bufferByteSizeThreshold
bufferKafkaRecordCountThreshold ->  bufferRecordCountThreshold

### Log Changes
Changing the variable names also changes the log message as follows

What is in master now:
`StreamingBufferThreshold{flushTimeThresholdSeconds=10, bufferSizeThresholdBytes=20000000, bufferKafkaRecordCountThreshold=1}`

After this PR
`StreamingBufferThreshold{bufferFlushTimeThreshold=10, bufferByteSizeThreshold=20000000, bufferRecordCountThreshold=1}`

Full log message:
`26-05-2023 12:55:20 main DEBUG TopicPartitionChannel:70 - [SF_KAFKA_CONNECTOR] Flush based on buffered bytes or buffered number of records for channel:TESTDB_KAFKA.KAFKA_TEST.KAFKA_CONNECTOR_TEST_TABLE_6292311087018711202.KAFKA_CONNECTOR_TEST_TABLE_6292311087018711202_0,currentBufferSizeInBytes:357, currentBufferedRecordCount:1, connectorBufferThresholds:StreamingBufferThreshold{bufferFlushTimeThreshold=10, bufferByteSizeThreshold=20000000, bufferRecordCountThreshold=1}`

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/BufferThreshold.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/BufferThreshold.java
@@ -1,6 +1,8 @@
 package com.snowflake.kafka.connector.internal;
 
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.BUFFER_COUNT_RECORDS_MIN;
 import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.BUFFER_FLUSH_TIME_SEC_MIN;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.BUFFER_SIZE_BYTES_MIN;
 import static com.snowflake.kafka.connector.internal.streaming.StreamingUtils.STREAMING_BUFFER_FLUSH_TIME_MINIMUM_SEC;
 
 import com.google.common.base.MoreObjects;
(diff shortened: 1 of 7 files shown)


src/main/java/com/snowflake/kafka/connector/internal/BufferThreshold.java
```diff
@@ -256,9 +247,9 @@ private static ImmutableMap<String, String> verifyBufferCountThreshold(
   @Override
   public String toString() {
```
sfc-gh-japatel, line 248: umm, can you verify if this class is serialized somewhere?

src/main/java/com/snowflake/kafka/connector/internal/BufferThreshold.java
```diff
@@ -24,108 +23,100 @@ public abstract class BufferThreshold {
...
+        >= (this.bufferFlushTimeThreshold * SECOND_TO_MILLIS);
   }
 
   /** @return Get flush time threshold in seconds */
   public long getFlushTimeThresholdSeconds() {
-    return flushTimeThresholdSeconds;
+    return this.bufferFlushTimeThreshold;
   }
 
-  /**
```
sfc-gh-japatel, line 119: why are we removing this?
public methods are recommended to have javadocs

src/main/java/com/snowflake/kafka/connector/internal/BufferThreshold.java
sfc-gh-japatel: like in telemetry? 
If yes, can you call that out as well..

sfc-gh-japatel (approved):
minor questions. Apart form that LGTM. please verify if the changed fields are serialized anywhere.

src/main/java/com/snowflake/kafka/connector/internal/BufferThreshold.java
sfc-gh-rcheng: good point. BufferThreshold is only logged and serialized in a log message indicating that flush has occurred. I've updated the description with that information

sfc-gh-rcheng: The buffer thresholds are also logged in telemetry, however it is logged as part of the larger Map<String, String> config, rathe than as a BufferThreshold which would be serialized here

src/main/java/com/snowflake/kafka/connector/internal/BufferThreshold.java
sfc-gh-rcheng: oh good call out, removed by mistake on my end

src/main/java/com/snowflake/kafka/connector/internal/BufferThreshold.java
sfc-gh-japatel: Can you make changes to the wiki regarding the same! Where we write helpful debug queries. 
For instance we will have two types of telemetries based on versions. 

Thanks

PR #649, opened 2023-06-13
SNOW-758492: Upgrade Ingest SDK to 2.0.0 and remove thread sleep

sfc-gh-tzhang:


src/test/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandlerTest.java
```diff
@@ -63,7 +63,7 @@ public void testCreateClientException() {
   @Test
   public void testCreateClientOverrideBdecVersion() {
```
sfc-gh-japatel, line 64: remove this test?

sfc-gh-japatel (approved):
lgtm!

sfc-gh-tjones (approved):
LOVE IT!

src/test/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandlerTest.java
sfc-gh-tzhang: I think we still need this in the future since we need a new BDEC version for supporting Iceberg or anything in case we need to update the format

sfc-gh-japatel: Ack

sfc-gh-tzhang:
Will close this and see how we can fix the missing data issue

PR #650, opened 2023-06-13
SNOW-826772 Unification of Snowflake role and user for Snowpipe Streaming

sfc-gh-japatel:
- Here is what the connection object is used for: [Only for Snowpipe Streaming]
  - DB existence checker. 
  - Schema existence checker.
  - CreateTable
  - Alter table 

Do you need test for this change?
- Existing tests uses a user and a uses a default role. [Check profile.json]
  - Verifying existing tests passes is enough. 
  - ConnectionServiceIT is one test which verifies this code path. 
- It's not worth adding the another profile.json and modifying the existing tests to use a new user/role. 
- Manual tests: Ran a manual test in sfctest0 account with a new role and a new user. Also ran a test which revoked an insert privilege from this role and it resulted into channel open error. 

Can anything go wrong with this change?
- Yes, if the default role of the user has more privileges than the role supplied. The role being used for Snowpipe Streaming needs atleast an insert privilege. But most of the times, an insert priv is accompanied by usage priv as well.
  - We will be adding this change in a new 2.0.0 version of Kafka Connector, so this will be advertised/documented. 
  - Only affects Snowpipe Streaming.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
@@ -109,7 +109,7 @@ class InternalUtils {
 
   /**
    * Use this function if you want to create properties from user provided Snowflake kafka connector
-   * config.
+   * config. It assumes the caller wants to use this Property for Snowpipe based Kafka Connector.
    *
    * @param conf User provided kafka connector config
    * @param sslEnabled is sslEnabled?
--- a/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
@@ -1,7 +1,9 @@
 package com.snowflake.kafka.connector.internal;
 
+import com.google.common.base.Strings;
 import com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig;
 import com.snowflake.kafka.connector.Utils;
+import com.snowflake.kafka.connector.internal.streaming.IngestionMethodConfig;
 import com.snowflake.kafka.connector.internal.telemetry.SnowflakeTelemetryService;
 import java.security.KeyFactory;
 import java.security.PrivateKey;
@@ -18,6 +20,7 @@ import net.snowflake.client.core.SFSessionProperty;
 import net.snowflake.client.jdbc.internal.apache.commons.codec.binary.Base64;
 import net.snowflake.client.jdbc.internal.org.bouncycastle.jce.provider.BouncyCastleProvider;
 import net.snowflake.ingest.connection.IngestStatus;
+import org.graalvm.compiler.core.common.util.Util;
 
 class InternalUtils {
   // JDBC parameter list
@@ -105,14 +108,20 @@ class InternalUtils {
     return date;
   }
 
+  static Properties createProperties(Map<String, String> conf, boolean sslEnabled) {
+    return createProperties(conf, sslEnabled, IngestionMethodConfig.SNOWPIPE);
+  }
+
   /**
    * create a properties for snowflake connection
    *
    * @param conf a map contains all parameters
    * @param sslEnabled if ssl is enabled
+   * @param ingestionMethodConfig which ingestion method is provided.
    * @return a Properties instance
    */
-  static Properties createProperties(Map<String, String> conf, boolean sslEnabled) {
+  static Properties createProperties(
+      Map<String, String> conf, boolean sslEnabled, IngestionMethodConfig ingestionMethodConfig) {
     Properties properties = new Properties();
 
     // decrypt rsa key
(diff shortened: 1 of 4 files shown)


src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
```diff
@@ -170,6 +186,17 @@ static Properties createProperties(Map<String, String> conf, boolean sslEnabled)
      */
     properties.put(SFSessionProperty.ALLOW_UNDERSCORES_IN_HOST.getPropertyKey(), "true");
 
+    if (ingestionMethodConfig == IngestionMethodConfig.SNOWPIPE_STREAMING) {
+      final String providedSFRoleInConfig = conf.get(Utils.SF_ROLE);
+      if (!Strings.isNullOrEmpty(providedSFRoleInConfig)) {
+        LOGGER.debug("Using provided role {} for JDBC connection.", providedSFRoleInConfig);
```
sfc-gh-tzhang, line 192: Could we combine the log here with the one in the else branch?

sfc-gh-tzhang: And I suggest to use INFO since DEBUG logging at this point is useless due to the large number of debug logging in the ingest SDK

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
```diff
@@ -105,14 +107,28 @@ static String timestampToDate(long time) {
     return date;
   }
 
+  /**
+   * Use this function if you want to create properties from user provide Snowflake kafka connector
+   * config.
+   *
+   * @param conf User provided kafka connector configu
```
sfc-gh-tzhang, line 114:
```suggestion
   * @param conf User provided kafka connector config
```

sfc-gh-tzhang (approved):
lgtm!

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
```diff
@@ -105,14 +107,28 @@ static String timestampToDate(long time) {
     return date;
   }
 
+  /**
+   * Use this function if you want to create properties from user provide Snowflake kafka connector
```
sfc-gh-asen, line 111:
```suggestion
   * Use this function if you want to create properties from user provided Snowflake kafka connector
```

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
```diff
@@ -170,6 +186,17 @@ static Properties createProperties(Map<String, String> conf, boolean sslEnabled)
...
     properties.put(SFSessionProperty.ALLOW_UNDERSCORES_IN_HOST.getPropertyKey(), "true");
 
+    if (ingestionMethodConfig == IngestionMethodConfig.SNOWPIPE_STREAMING) {
+      final String providedSFRoleInConfig = conf.get(Utils.SF_ROLE);
+      if (!Strings.isNullOrEmpty(providedSFRoleInConfig)) {
+        LOGGER.debug("Using provided role {} for JDBC connection.", providedSFRoleInConfig);
+        properties.put(SFSessionProperty.ROLE, providedSFRoleInConfig);
+      } else {
+        LOGGER.debug(
+            "No role is provided, default role for the user will be used for JDBC connection.");
```
sfc-gh-xhuang, line 196: should messages be consistent?

https://github.com/snowflakedb/snowflake-ingest-java/pull/507/files#diff-f74efdef79a994538152e7cb1924dc3620e994c9fbc6e4652e480b7bf022cdc6R135

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
sfc-gh-japatel: Yes, makes sense. no harm in using info.

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
sfc-gh-japatel: oops

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
sfc-gh-japatel: ack

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
sfc-gh-japatel: Ack, will do that.

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
```diff
@@ -105,14 +107,28 @@ static String timestampToDate(long time) {
...
 
+  /**
+   * Use this function if you want to create properties from user provided Snowflake kafka connector
+   * config.
+   *
+   * @param conf User provided kafka connector config
+   * @param sslEnabled is sslEnabled?
+   * @return Properties object which will be passed down to JDBC connection
+   */
+  static Properties createProperties(Map<String, String> conf, boolean sslEnabled) {
```
sfc-gh-rcheng, line 118: can we add snowpipe something to this method name (createPropertiesForSnowpipe) or mention in the comments its for snowpipe?

sfc-gh-rcheng (approved):

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
sfc-gh-japatel: I didnt want make number of changes in test classes to reflect this change. I can add comment about using default ingestmethodconfig

PR #651, opened 2023-06-14
SNOW-811265 Do not close channel on rebalance

sfc-gh-japatel:
Benefits:
- Performance improvement since we will not be waiting for channels to get closed. Currently, channel.close().get() is a blocking call. 
- We anyways reopen the channel during open partition. 
- I have seen significant improvements in closing of partitions in terms of latency. This will be visible when more partitions are assigned to a task. (Happens when customers uses a lot of topics in their connector)
- Less number of exceptions in close API call if previous call to insert rows resulted in failure. 

Tests:
- Existing tests are enough. 

Here are few examples of what we gain:
- Test setup: Create a single snowflake connector starting with one topic and 4 tasks. Keep on adding 10 datagen connector adding data to 10 topics and also modify the existing snowflake connector with the newly added topic. 
- In the end we will have 11 connectors - 10 Datagen and 1 SF connector with 10 topics. 

- As you can see from below image, partitions get reassigned when you add more topics (with 4 partitions each)
- The time taken to close the channel grows. 
- Removing the close call has no affect on behavior. 
- During open partitions, we reopen channel, fetch the offsetToken and reset a non-null offset to kafka. 

Screenshot 2023-06-23 at 10 12 25 AM

Screenshot 2023-06-23 at 10 09 11 AM

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
@@ -302,14 +302,15 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
   /**
    * This function is called during rebalance.
    *
-   * <p>All the channels are closed. The client is still active. Upon rebalance, (inside {@link
-   * com.snowflake.kafka.connector.SnowflakeSinkTask#open(Collection)} we will reopen the channel.
+   * <p>We don't close the channels. Upon rebalance, (inside {@link
+   * com.snowflake.kafka.connector.SnowflakeSinkTask#open(Collection)} we will reopen the channel
+   * anyways. [Check c'tor of {@link TopicPartitionChannel}]
    *
    * <p>We will wipe the cache partitionsToChannel so that in {@link
    * com.snowflake.kafka.connector.SnowflakeSinkTask#open(Collection)} we reinstantiate and fetch
    * offsetToken
    *
-   * @param partitions a list of topic partition
+   * @param partitions a list of topic partitions to close/shutdown
    */
   @Override
   public void close(Collection<TopicPartition> partitions) {
@@ -321,15 +322,16 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
               partitionsToChannel.get(partitionChannelKey);
           // Check for null since it's possible that the something goes wrong even before the
           // channels are created
-          if (topicPartitionChannel != null) {
-            topicPartitionChannel.closeChannel();
-          }
           LOGGER.info(
-              "Closing partitionChannel:{}, partition:{}, topic:{}",
+              "Removing partitionChannel:{}, partition:{}, topic:{} from map(partitionsToChannel)",
               topicPartitionChannel == null ? null : topicPartitionChannel.getChannelName(),
               topicPartition.topic(),
               topicPartition.partition());
         });
+    LOGGER.info(
+        "Closing {} partitions and Clearing partitionsToChannel Map of size:{}",
+        partitions.size(),
+        partitionsToChannel.size());
     partitionsToChannel.clear();
   }
 


src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -321,15 +322,16 @@ public void close(Collection<TopicPartition> partitions) {
               partitionsToChannel.get(partitionChannelKey);
           // Check for null since it's possible that the something goes wrong even before the
           // channels are created
-          if (topicPartitionChannel != null) {
-            topicPartitionChannel.closeChannel();
-          }
           LOGGER.info(
-              "Closing partitionChannel:{}, partition:{}, topic:{}",
+              "Removing partitionChannel:{}, partition:{}, topic:{} from map(partitionsToChannel)",
```
sfc-gh-rcheng, line 326: Are we relying on the channel cleaner to close the channel then? Also it is slightly concerning that we can make this behavior change and not break any of our existing tests

sfc-gh-japatel: Channel cleaner doesnt close the channel. Channel cleaner cleans up inactive channels. (Those who dont have any commit activity in last 30 days)

Valid concern regarding tests not failing, I will take a look.

sfc-gh-japatel: On the other hand, its kind of safe to remove it since none of our tests are failing. 
I checked our existing test and it kind of makes sense to remove it since there was no behavior change:
Checkout `testRebalanceOpenCloseIngestion` and `testChannelCloseIngestion`

It would have failed if I had checked for channel .isClosed() but that wasnt present in any test.

sfc-gh-rcheng: Yeah there are pros and cons to having our tests be fine grained enough to catch behavior like this. 

I am confused about the channel cleaner - does it close the channel as part of cleaning up inactive channels? I'm mostly concerned that we'll be leaving behind a lot of orphaned channels, especially as we currently have this bug with cleaning up old inactive channels 

https://snowflakecomputing.atlassian.net/browse/SNOW-800927

sfc-gh-japatel: `StreamingIngestChannelCleaner` is the class if you want to go through the logic. 
But essentially, it removes the channels from FDB. If you are seeing old channels, it might be worth investigating why this is happening. - might be a bug somewhere. Also note that it scans the lastCommitTime, not createdOn. 


> I am confused about the channel cleaner - does it close the channel as part of cleaning up inactive channels?
There is no callback but lets say from client side if you still have a channel instance and channel gets wiped off from server side, you will start getting exception. (Note that, you will have to reopen the channel)

> I'm mostly concerned that we'll be leaving behind a lot of orphaned channels
Umm, we should investigate it why is this the case because this seems to be a bug then.

sfc-gh-rcheng: Sounds good! thanks for the explanation. Its something on my backlog, just havent had time to look into it yet

sfc-gh-japatel: No problem, this is something I have looked on, going to discuss this on monday meeting, i can take a look at the jira and gs logs!

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -302,14 +302,15 @@ public void closeAll() {
   /**
    * This function is called during rebalance.
    *
-   * <p>All the channels are closed. The client is still active. Upon rebalance, (inside {@link
-   * com.snowflake.kafka.connector.SnowflakeSinkTask#open(Collection)} we will reopen the channel.
+   * <p>We don't close the channels. Upon rebalance, (inside {@link
```
sfc-gh-tzhang, line 305: I'm a bit concern about this change, did the customer test this and verify that this actually fixes their issue? Or they just reduced the number of topics and the issue is gone. If the theory is that multiple channel open compete with each other then this change won't help, right?

Two area of concerns I have:
- This means that we're not verifying that everything is committed before rebalance and if there is a reported customer issue, we won't know that the issue is because of channel invalidation by reading through the KC log
- Resources are not released and keeps increasing, we have a cache per client and the flush service will iterate over all active channels to see if there is anything to flush, the number of active channels will just keep increasing

sfc-gh-tzhang (approved):
I'm ok with this change but left some concerns, PTAL and see what you want to do, thanks!

sfc-gh-tzhang:
> * less number of channel open/close calls.

The number of open calls should be the same

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: These are some valid concerns and I will come back with more findings and whether we "really" need this change. Thanks Toby!

sfc-gh-japatel:
Based on customer's and my updated description, I am inclined towards pushing this PR. 
CC: @sfc-gh-tzhang @sfc-gh-rcheng

PR #654, opened 2023-06-16
[SNOW-840882] Disable flakey streaming client test

sfc-gh-rcheng:
testMultipleGetAndClose() passes locally however fails on aws and azure. Most likely some concurrency issue


sfc-gh-japatel (approved):
lgtm

PR #658, opened 2023-06-22
SNOW-850959: Fix a wrong result issue that offsets are skipped when schematization is enabled

sfc-gh-tzhang:
Looks like we have a gap in KC that may skip ingesting some offsets, consider this case where you have two topics with different schemas trying to ingest into the same table, internally KC will create two channels (channel A and channel B) with offset_token=NULL, then both channels start to buffer data and flush files, but channel A fails committing the first batch because the file schema doesn't match the latest table schema due to schema evolution, then channel A will be invalidated and reopened but we won't reset the consumer offset because the offset_token for channel A is still NULL, we say that we will rely on Kafka to send us the correct data when the offset_token is NULL, so in this case Kafka will continue send us the next batch and we will accept it, this means that the first batch for channel A will be skipped forever. We need to rethink about what we need to do when the offset_token for a channel is NULL and I don't think we can purely rely on Kafka to resend us the correct offset.

The fix is to manage the Kafka consumer offset in the connector as well and use that to reset Kafka when the offset token for a channel is NULL instead of relying on Kafka to send us the correct offset


src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -345,7 +359,8 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
    */
   private boolean shouldIgnoreAddingRecordToBuffer(
       SinkRecord kafkaSinkRecord, long currentProcessedOffset) {
-    if (!isOffsetResetInKafka) {
+    if (!isOffsetResetInKafka
+        || currentProcessedOffset == NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE) {
```
sfc-gh-japatel, line 363: I am confused why is this needed. Mind adding a comment here?

test/test_suites.py
```diff
@@ -54,13 +11,47 @@
...
+from test_suit.test_multiple_topic_to_one_table_snowpipe_streaming import \
+    TestMultipleTopicToOneTableSnowpipeStreaming
+from test_suit.test_native_complex_smt import TestNativeComplexSmt
+from test_suit.test_native_string_avrosr import TestNativeStringAvrosr
+from test_suit.test_native_string_json_without_schema import TestNativeStringJsonWithoutSchema
+from test_suit.test_native_string_protobuf import TestNativeStringProtobuf
+from test_suit.test_schema_evolution_avro_sr import TestSchemaEvolutionAvroSR
+from test_suit.test_schema_evolution_drop_table import TestSchemaEvolutionDropTable
+from test_suit.test_schema_evolution_json import TestSchemaEvolutionJson
+# res tests
```
sfc-gh-japatel, line 30: nit: what is `res?`

test/test_suit/test_schema_evolution_drop_table.py
```diff
@@ -46,7 +46,7 @@ def send(self):
         self.driver.sendBytesData(self.topic, value, key)
 
         # Sleep for some time and then verify the rows are ingested
-        sleep(60)
+        sleep(120)
```
sfc-gh-japatel, line 49: Curious why it's taking more time. I see you have tried multiple times.

sfc-gh-japatel (approved):
lgtm! Thanks a lot! 
Some minor comments to clarify.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -94,6 +94,10 @@ public class TopicPartitionChannel {
   private final AtomicLong processedOffset =
       new AtomicLong(NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE);
 
+  // The in-memory consumer offset managed by the connector, we need this to tell Kafka which
+  // offset to resend when the channel offset token is NULL
+  private long latestConsumerOffset = NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE;
```
sfc-gh-rcheng, line 99: the above `offsetPersistedInSnowflake` and `processedOffset` are AtomicLongs, should this also be an AtomicLong?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -94,6 +94,10 @@ public class TopicPartitionChannel {
   private final AtomicLong processedOffset =
       new AtomicLong(NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE);
 
+  // The in-memory consumer offset managed by the connector, we need this to tell Kafka which
```
sfc-gh-rcheng, line 97: we now have three offsets, can we unify the names or clarify the different uses in the comments? From what I can tell, latestConsumerOffset is either the first KC offset (initial processedOffset) or the snowflake (offsetPersistedInSnowflake). If so, can we comment something like:

offsetPersistedInSnowflake: This offset represents the data persisted in Snowflake. More specifically it is the Snowflake offset determined from the insertRows API call. It is set after calling the fetchOffsetToken API for this channel

processedOffset: This offset represents the data buffered in KC. More specifically it is the KC offset to ensure exactly once functionality. On creation it is set to the latest committed token in Snowflake (see offsetPersistedInSnowflake) and updated on each new row from KC.

latestConsumerOffset:  This offset is a fallback to represent the data buffered in KC. It is similar to processedOffset, however it is only used to resend the offset when the channel offset token is NULL. It is updated to the first offset sent by KC (see processedOffset) or the offset persisted in snowflake (see offsetPersistedInSnowflake)

test/test_suit/test_schema_evolution_drop_table.py
sfc-gh-rcheng: ^ as well as why we're removing the 30 sec sleep below

test/test_suites.py
sfc-gh-rcheng: I added that for 'resilience tests', we can add the full word there. Can we move the comment back above the rest of the resilience tests?

test/test_suites.py
```diff
@@ -112,91 +104,114 @@ def create_end_to_end_test_suites(driver, nameSalt, schemaRegistryAddress, testS
             test_instance=TestAvrosrAvrosr(driver, nameSalt), clean=True, run_in_confluent=True, run_in_apache=False
         )),
         ("TestNativeStringAvrosr", EndToEndTestSuite(
-            test_instance=TestNativeStringAvrosr(driver, nameSalt), clean=True, run_in_confluent=True, run_in_apache=False
+            test_instance=TestNativeStringAvrosr(driver, nameSalt), clean=True, run_in_confluent=True,
```
sfc-gh-rcheng, line 107: nit: why are all these 'run_in_apache' lines pushed down a line? i don't see any other differences, is there a formatting change?

sfc-gh-rcheng (approved):
lgtm, just want more clarification on the different offsets

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -266,7 +266,9 @@ public long getOffset(TopicPartition topicPartition) {
     String partitionChannelKey =
         partitionChannelKey(topicPartition.topic(), topicPartition.partition());
     if (partitionsToChannel.containsKey(partitionChannelKey)) {
-      return partitionsToChannel.get(partitionChannelKey).getOffsetSafeToCommitToKafka();
+      long offset = partitionsToChannel.get(partitionChannelKey).getOffsetSafeToCommitToKafka();
```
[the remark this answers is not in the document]
sfc-gh-tzhang: I think this will be optimized by the complier so it's usually a matter of style?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: thanks for the suggestion, updated

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: I believe everything should be long given that there is no concurrent logic?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: We don't want to skip any rows when the offset token for a channel is NULL, think about the case when the data is expired so a row with higher offset is sent by Kafka upon restarting

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -954,26 +966,15 @@ private long getRecoveredOffsetFromSnowflake(
...
-        return NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE;
-      } else {
-        long latestCommittedOffsetInSnowflake = Long.parseLong(offsetToken);
-        LOGGER.info(
-            "Fetched offsetToken:{} for channelName:{}",
-            latestCommittedOffsetInSnowflake,
-            this.channel.getFullyQualifiedName());
-        return latestCommittedOffsetInSnowflake;
-      }
+      LOGGER.info(
```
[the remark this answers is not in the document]
sfc-gh-tzhang: 1. I don't think warn makes sense because there are valid cases that a offset token could be NULL
2. We want to unify the format for both cases so it's easier to search and debug

test/test_suit/test_schema_evolution_drop_table.py
sfc-gh-tzhang: good question, this solution is not perfect, if we don't wait for more than 60 seconds than the in-memory offset is not updated and we will restart from the beginning, I put 120 seconds to make sure it's not flaky

test/test_suites.py
sfc-gh-tzhang: all the formatting is done automatically so I have no control, I simply remove it

test/test_suites.py
sfc-gh-tzhang: all the formatting is done automatically so I have no control

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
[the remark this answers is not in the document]
sfc-gh-rcheng: oh good to know!

PR #660, opened 2023-06-22
NO-SNOW Too many info logs for successfully called put and precommit,…

sfc-gh-japatel:
… this will just blow up

- It's okay for testing but this will blow up the disk space. 
- For reference:
Screenshot 2023-06-22 at 2 44 20 PM

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
@@ -416,7 +416,7 @@ public class SnowflakeSinkTask extends SinkTask {
           logExecutionContent,
           executionTimeMs);
     } else {
-      this.DYNAMIC_LOGGER.info("Successfully " + logExecutionContent);
+      this.DYNAMIC_LOGGER.debug("Successfully " + logExecutionContent);
     }
   }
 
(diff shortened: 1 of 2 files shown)


sfc-gh-asen (approved):

sfc-gh-tzhang:
> … this will just blow up
> 
> * It's okay for testing but this will blow up the disk space.
> * For reference:
> 
> Screenshot 2023-06-22 at 2 44 20 PM

I suggest we keep this as INFO, understand the concern here and there're customers that may complain, but this is really important for perf regression investigation and we can't ask customer to simply do DEBUG since there will be too many logs for the SDK

sfc-gh-tzhang:
> > … this will just blow up
> > 
> > * It's okay for testing but this will blow up the disk space.
> > * For reference:
> > 
> > Screenshot 2023-06-22 at 2 44 20 PM
> 
> I suggest we keep this as INFO, understand the concern here and there're customers that may complain, but this is really important for perf regression investigation and we can't ask customer to simply do DEBUG since there will be too many logs for the SDK

Or we could do some sort of sampling so that we could at least know the trends on how long it would take for the PUT to finish

sfc-gh-japatel:
> > … this will just blow up
> > 
> > * It's okay for testing but this will blow up the disk space.
> > * For reference:
> > 
> > Screenshot 2023-06-22 at 2 44 20 PM
> 
> I suggest we keep this as INFO, understand the concern here and there're customers that may complain, but this is really important for perf regression investigation and we can't ask customer to simply do DEBUG since there will be too many logs for the SDK

SDK and KC logging are different and can be set at a class level. 
```
com.snowflake=DEBUG,net.snowflake=ERROR
```
For aws msk it was a bit difficult but we dont have too many customers on it.

sfc-gh-rcheng (approved):

PR #665, opened 2023-07-06
[SNOW-857447] Enable one client optimization by default

sfc-gh-rcheng:


sfc-gh-asen (approved):

sfc-gh-tzhang (approved):
lgtm, can we update our doc as well?

sfc-gh-rcheng:
> lgtm, can we update our doc as well?

ill follow up with lei

PR #672, opened 2023-07-24
[Streaming JMX #1] Small changes to Snowpipe Telemetry and JMX

sfc-gh-rcheng:
Planning to add telemetry and JMX metrics to streaming snowpipe. This PR makes small fixes to the existing Snowpipe telemetry and JMX code:
- import each variable explicitly, instead of using *
- remove stageName property from SnowflakeTelemetryBasicInfo. It will be null for streaming, so it doesn't make sense to have a nullable required property in the constructor


src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryPipeStatus.java
```diff
@@ -266,6 +283,7 @@ public void dumpTo(ObjectNode msg) {
     msg.put(STAGE_NAME, stageName);
     msg.put(PIPE_NAME, pipeName);
 
+    // TODO @rcheng: changed from metricsutil to telemetry?
```
sfc-gh-rcheng, line 286: Most of these properties referred to the metricsutil property instead of the telemetry property. Does this difference matter? i can change the telemetry property to match the metrics utils property since I'm worried this change might break some dashboards

ex: PROCESSED_OFFSET
`processed-offset` would become -> `processed_offset`

sfc-gh-japatel: yeah lets not change the metrics name, it will break scripts and dashboard.

sfc-gh-japatel (approved):
lgtm!

src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryPipeStatus.java
sfc-gh-rcheng: Changed telemetry property to match the metric property

src/main/java/com/snowflake/kafka/connector/internal/telemetry/TelemetryConstants.java
```diff
@@ -9,21 +9,21 @@ public final class TelemetryConstants {
   public static final String STAGE_NAME = "stage_name";
   public static final String PIPE_NAME = "pipe_name";
 
-  public static final String PROCESSED_OFFSET = "processed_offset";
-  public static final String FLUSHED_OFFSET = "flushed_offset";
-  public static final String COMMITTED_OFFSET = "committed_offset";
-  public static final String PURGED_OFFSET = "purged_offset";
+  public static final String PROCESSED_OFFSET = "processed-offset";
```
sfc-gh-rcheng, line 12: these were not used previously, changed to match the metricsUtil properties as we were accidentally using those previously

Issue #441, opened 2022-06-07
SnowflakeSinkServiceV1: log "task is already registered" as info instead of an error


dlg99:
As I understand this is perfectly normal ad requires no attention:
https://github.com/snowflakedb/snowflake-kafka-connector/blob/ed1b37a3554be84892e99e165023d0b856ecd24f/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java#L119-L121

Yet this logged as an error which raises questions/investigation in whether some configuration has to be corrected.
I think the message can be logged as info or warning instead of an error.

sfc-gh-rcheng:
Changed merged, thanks for the suggestion

PR #679, opened 2023-08-01
Change pipe already registered log from error to warn level

sfc-gh-rcheng:
Title. See this [issue](https://github.com/snowflakedb/snowflake-kafka-connector/issues/441)

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
@@ -108,7 +108,7 @@ class SnowflakeSinkServiceV1 implements SnowflakeSinkService {
     String stageName = Utils.stageName(conn.getConnectorName(), tableName);
     String nameIndex = getNameIndex(topicPartition.topic(), topicPartition.partition());
     if (pipes.containsKey(nameIndex)) {
-      LOGGER.warn("task is already registered, name: {}", nameIndex);
+      LOGGER.warn("task is already registered with {} partition", nameIndex);
     } else {
       String pipeName =
           Utils.pipeName(conn.getConnectorName(), tableName, topicPartition.partition());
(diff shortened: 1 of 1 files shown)


sfc-gh-xhuang (approved):

sfc-gh-xhuang:
Nice small fix :)

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -108,7 +108,7 @@ public void startTask(final String tableName, final TopicPartition topicPartitio
     String stageName = Utils.stageName(conn.getConnectorName(), tableName);
     String nameIndex = getNameIndex(topicPartition.topic(), topicPartition.partition());
     if (pipes.containsKey(nameIndex)) {
-      LOGGER.error("task is already registered, name: {}", nameIndex);
+      LOGGER.warn("task is already registered, name: {}", nameIndex);
```
sfc-gh-japatel, line 111: this log line actually doesnt make sense, `task is already registered with {} partition` makes more sense.

sfc-gh-japatel (approved):
lgtm! see my comment about the wordings.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-tzhang: Not familiar with this code path, but seems like the expectation is that this shouldn't happen and that's why it was logged as an error?

sfc-gh-japatel: that's true, it shouldnt happen. will only happen if close didnt wipe off map. so keeping it error makes sense.

sfc-gh-rcheng: This was opened as an issue.

Since having a task already registered is not a breaking scenario, i think it does make sense that it should be a warning instead of an error?

PR #683, opened 2023-08-01
SNOW-858328 Add cache for schema evolution permission query

sfc-gh-alhuang:
According to [SNOW-858328](https://snowflakecomputing.atlassian.net/browse/SNOW-858328), add a cache for schema evolution permission query to pull up query to table level and increase the performance.

[SNOW-858328]: https://snowflakecomputing.atlassian.net/browse/SNOW-858328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -176,7 +177,33 @@ public void startTask(String tableName, TopicPartition topicPartition) {
...
+        tp -> {
+          String tableName = Utils.tableName(tp.topic(), topic2Table);
+
+          if (!tableName2SchemaEvolutionPermission.containsKey(tableName)) {
+            createTableIfNotExists(tableName);
+            tableName2SchemaEvolutionPermission.put(
+                tableName,
+                this.conn.hasSchemaEvolutionPermission(
+                    tableName, connectorConfig.get(SNOWFLAKE_ROLE)));
+          }
```
sfc-gh-japatel, line 203:
```suggestion
          tableName2SchemaEvolutionPermission.putIfAbsent(
        tableName,
        this.conn.hasSchemaEvolutionPermission(tableName, connectorConfig.get(SNOWFLAKE_ROLE))
    );
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -194,6 +194,7 @@ public TopicPartitionChannel(
         topicPartition,
         channelName,
         tableName,
+        null,
```
sfc-gh-japatel, line 197: add a comment here saying what is null
like 
```
null /**schemaEvolutionPermission*/
```

sfc-gh-japatel (approved):
this is a great improvement, lgtm! Lets get blessing from @sfc-gh-tzhang as well!

sfc-gh-japatel:
Would be great to test it on a db schema with more than one partitions. you can compare two results and see the improvements. 
It would be great if you could also post the latency improvement. Not a blocker but would be good to show.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -176,7 +177,33 @@ public void startTask(String tableName, TopicPartition topicPartition) {
...
+  /**
+   * Initializes multiple Channels and partitionsToChannel maps with new instances of {@link
+   * TopicPartitionChannel}
+   *
+   * @param partitions collection of topic partition
+   * @param topic2Table map of topic to table name
+   */
+  @Override
+  public void startTasks(Collection<TopicPartition> partitions, Map<String, String> topic2Table) {
+    Map<String, Boolean> tableName2SchemaEvolutionPermission = new HashMap<>();
```
sfc-gh-tzhang, line 192: @sfc-gh-japatel @sfc-gh-rcheng Is there another place we could put this so that the map doesn't have to be repopulated for every rebalance? Could you move it to the place where we create the client? We should have all the table info from the configuration?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -257,11 +261,15 @@ public TopicPartitionChannel(
...
-        this.enableSchematization
-            && this.conn != null
-            && this.conn.hasSchemaEvolutionPermission(
-                tableName, sfConnectorConfig.get(SNOWFLAKE_ROLE));
+
+    if (hasSchemaEvolutionPermission == null) {
+      hasSchemaEvolutionPermission =
+          this.conn != null
+              && this.conn.hasSchemaEvolutionPermission(
+                  tableName, sfConnectorConfig.get(SNOWFLAKE_ROLE));
```
sfc-gh-tzhang, line 269: nit: these three condition checks could be combined?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -176,7 +177,33 @@ public void startTask(String tableName, TopicPartition topicPartition) {
...
+    partitions.forEach(
+        tp -> {
+          String tableName = Utils.tableName(tp.topic(), topic2Table);
+
+          if (!tableName2SchemaEvolutionPermission.containsKey(tableName)) {
+            createTableIfNotExists(tableName);
```
sfc-gh-tzhang, line 198: Looks like the startTask could be removed now given that we're not using it anymore?

sfc-gh-tzhang (commented):
Sorry about the delay, left some comments, PTAL! My main comment is about whether the map could be put in a different place so that there is no need to be repopulated in every rebalance.

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelTest.java
```diff
@@ -218,6 +218,7 @@ public void testCloseChannelException() throws Exception {
             topicPartition,
             TEST_CHANNEL_NAME,
             TEST_TABLE_NAME,
+            null,
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Added [here](https://github.com/snowflakedb/snowflake-kafka-connector/blob/e63dbfc17ddf0e045a38b9b19ed6295ce34a7877/src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java#L284).

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelTest.java
```diff
@@ -606,6 +608,7 @@ public void testInsertRows_ValidationResponseHasErrors_NoErrorTolerance() throws
...
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Removed.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-alhuang: The boolean expression might looks like this:
```
hasSchemaEvolutionPermission = 
    (hasSchemaEvolutionPermission != null && hasSchemaEvolutionPermission) ||
    (hasSchemaEvolutionPermission == null &&
     this.conn != null && 
     this.conn.hasSchemaEvolutionPermission(tableName, sfConnectorConfig.get(SNOWFLAKE_ROLE))
    );
```
I think it's quit complicate, should I use ternary instead?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-alhuang: Still using it [here](https://github.com/snowflakedb/snowflake-kafka-connector/blob/e63dbfc17ddf0e045a38b9b19ed6295ce34a7877/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceFactory.java#L70)

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-rcheng: The client is lazily created in the [SnowflakeSinkTask](https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java#L158) since we do need the configuration. It would not be affected by rebalance because the [singleton provider](https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java)

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkService.java
```diff
@@ -22,6 +22,15 @@ public interface SnowflakeSinkService {
...
   void startTask(String tableName, TopicPartition topicPartition);
 
+  /**
+   * Start a collection of TopicPartition. This should handle any configuration parsing and one-time
+   * setup of the task.
+   *
+   * @param partitions collection of topic partitions
+   * @param topic2Table a mapping from topic to table
+   */
+  void startTasks(Collection<TopicPartition> partitions, Map<String, String> topic2Table);
```
sfc-gh-japatel, line 32: I would not call this `startTasks`

SnowflaekSinkService is in instance associated to the task already. 

Lets call this `startPartitions()`

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -119,6 +119,17 @@ public void startTask(final String tableName, final TopicPartition topicPartitio
     }
   }
 
+  /**
+   * Create multiple ingesting tasks
+   *
+   * @param partitions collection of topic partitions
+   * @param topic2Table a mapping from topic to table
+   */
```
sfc-gh-japatel, line 127: you can drop the javadocs here since it is already present in interface

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: even if we move it to connector level, it is very much possible it needs to repopulated there. that's what is happening with client creation as well.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: IMO it could be combined but it is doesn't read well. current way @sfc-gh-alhuang has is more readable.

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelTest.java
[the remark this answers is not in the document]
sfc-gh-japatel: Nice, thanks @sfc-gh-alhuang

sfc-gh-japatel (approved):
apart from few naming conventions, LGTM!

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkService.java
sfc-gh-alhuang: Fixed.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-alhuang: Removed.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-tzhang: Not sure if I understand, so do we have a better place to put this map so that it won't be repopulated every time after rebalance?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: ok, I think the way we're doing this is wried, we should pass a boolean instead of Boolean. It's unexpected that the map doesn't have the table. Then this check could be removed completely, wdyt?

sfc-gh-alhuang: @sfc-gh-tzhang make sense. Removed.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -194,6 +193,7 @@ public TopicPartitionChannel(
         topicPartition,
         channelName,
         tableName,
+        false,
```
sfc-gh-japatel, line 196: add a single line comment here saying what this false represents

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -119,6 +119,12 @@ public void startTask(final String tableName, final TopicPartition topicPartitio
     }
   }
 
+  @Override
+  public void startPartitions(
+      Collection<TopicPartition> partitions, Map<String, String> topic2Table) {
+    partitions.forEach(tp -> this.startTask(Utils.tableName(tp.topic(), topic2Table), tp));
```
sfc-gh-japatel, line 125: we should rename this to `startPartition` instead of `startTask`

Let's create a jira and add a TODO here. You can do this if you have time left after your current PRs are merged.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: I see your concern, valid point. 

I think this can be moved up a level inside a constructor of SnowflakeSinkServiceV2 and we can pass down the map. Is that what you are thinking?

sfc-gh-japatel: What you can do here is call `hasSchemaEvolutionPermission` from inside the Ctor of SnowflakeSinkServiceV2 for every table from topicToTable Map 

pass down the cache in startTask and do a get or if there is cache miss, populate the cache.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-alhuang: Renamed

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-alhuang: Added.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -172,17 +189,13 @@ public SnowflakeSinkServiceV2(
...
     createTableIfNotExists(tableName);
 
     // Create channel for the given partition
     createStreamingChannelForTopicPartition(
-        tableName,
-        topicPartition,
-        this.conn != null
-            && this.conn.hasSchemaEvolutionPermission(
-                tableName, connectorConfig.get(SNOWFLAKE_ROLE)));
+        tableName, topicPartition, tableName2SchemaEvolutionPermission.get(tableName));
```
sfc-gh-japatel, line 200: what if `tableName2SchemaEvolutionPermission.get(tableName)` is null? have we handled that case?

sfc-gh-alhuang: `createTableIfNotExists` always make sure `tableName` exist. [Ref](https://github.com/snowflakedb/snowflake-kafka-connector/blob/ad5a34544ef706c3e7d9ec2247f057c226312dc6/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java#L559C2-L559C2)

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -159,6 +165,20 @@ public SnowflakeSinkServiceV2(
...
+      this.topicToTableMap.forEach(
+          (topic, tableName) -> {
+            if (!this.tableName2SchemaEvolutionPermission.containsKey(tableName)) {
+              this.tableName2SchemaEvolutionPermission.put(
+                  tableName,
+                  this.conn != null
+                      && this.conn.hasSchemaEvolutionPermission(
+                          tableName, connectorConfig.get(SNOWFLAKE_ROLE)));
+            }
+          });
```
sfc-gh-japatel, line 180: this should also be inside a an uber check of 
`enableSchematization` i.e populate the map only if `enableSchematization` = true

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -122,6 +126,8 @@ public SnowflakeSinkServiceV2(
             .getClient(this.connectorConfig);
 
     this.partitionsToChannel = new HashMap<>();
+
+    this.tableName2SchemaEvolutionPermission = new HashMap<>();
```
sfc-gh-japatel, line 130: why not add the same logic here?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel, line 180: lets extract this out to a new function which can be called `populateSchemaEvolutionPermissions`

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: I dont see `null` handling. 
So here is what I think we should do:
1. if table is present in cache, do a get. 
2. If not, call permission function and populate it in cache and do a get again. 
3. pass the boolean down from above fetches `createStreamingChannelForTopicPartition`. It is then guaranteed to be non null since your cache is <String, Boolean> and your parameter in function is boolean. 

Let me know if this makes sense.

sfc-gh-japatel (commented):
left some comments. PTAL. 
Lets also add a case where in the beginning topic2table has x tables and later in startPartition you add a table which was not part of topic2table. 

Essentially, we are trying to mimick a rebalance case where a table partition was not part of assigned partitions. this will kick in cache miss and we will populate new cache key.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkService.java
```diff
@@ -20,7 +20,7 @@ public interface SnowflakeSinkService {
    * @param tableName destination table name
    * @param topicPartition TopicPartition passed from Kafka
    */
-  void startTask(String tableName, TopicPartition topicPartition);
+  void startPartition(String tableName, TopicPartition topicPartition);
```
sfc-gh-japatel, line 23: thanks for the refactor @sfc-gh-alhuang

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: that check is fine. I am more concerned about the existence of tableName in hashmap. Not in db schema in snowflake. 

there is possibility that a new table/partition gets assigned to a task after rebalance. even though table must have created, your hashmap might not have that tableName.

sfc-gh-japatel: Talked offline, Alec is populating the cache in `createTableIfNotExists`. 
I will wait for the test to be added as mentioned earlier but otherwise looks very promising. Thanks

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-alhuang: Extracted.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-alhuang: The table2topic is empty in this function.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-alhuang: Added.

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -369,6 +369,111 @@ public void testStreamingIngest_multipleChannelPartitions() throws Exception {
     service.closeAll();
   }
 
+  @Test
+  public void testStreamingIngest_multipleChannelPartitionsWithTopic2Table() throws Exception {
```
sfc-gh-alhuang, line 373: This test cover `startPartition()` with `topic2Table`

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -370,7 +370,59 @@ public void testStreamingIngest_multipleChannelPartitions() throws Exception {
...
   @Test
-  public void testStreamingIngest_startTasksWithMultipleChannelPartitions() throws Exception {
+  public void testStreamingIngest_multipleChannelPartitionsWithTopic2Table() throws Exception {
+    final int partitionCount = 5;
+    final int recordsInEachPartition = 2;
+
+    Map<String, String> config = TestUtils.getConfForStreaming();
+    SnowflakeSinkConnectorConfig.setDefaultValues(config);
+    Map<String, String> topic2Table = new HashMap<>();
+    topic2Table.put(topic, table);
```
sfc-gh-japatel, line 380: can you put multiple topics here and then insert records for a topic partition not present in topic2table.

sfc-gh-alhuang: Sounds good, added.

sfc-gh-japatel:
LGTM, lets get a signoff from @sfc-gh-tzhang

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -94,6 +95,9 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
    */
   private final Map<String, TopicPartitionChannel> partitionsToChannel;
 
+  // Cache for schema evolution
+  private Map<String, Boolean> tableName2SchemaEvolutionPermission = new HashMap<>();
```
sfc-gh-tzhang, line 99: You don't need to initialize here if you choose to do it in the constructor

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -257,11 +260,8 @@ public TopicPartitionChannel(
     /* Schematization related properties */
     this.enableSchematization =
         this.recordService.setAndGetEnableSchematizationFromConfig(sfConnectorConfig);
-    this.enableSchemaEvolution =
-        this.enableSchematization
-            && this.conn != null
-            && this.conn.hasSchemaEvolutionPermission(
-                tableName, sfConnectorConfig.get(SNOWFLAKE_ROLE));
+
+    this.enableSchemaEvolution = this.enableSchematization && hasSchemaEvolutionPermission;
```
sfc-gh-tzhang, line 264: Much better now, thanks for the refactor!

sfc-gh-tzhang (approved):
LGTM! Left a minor comments, thanks!

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-alhuang: Modified

Issue #599, opened 2023-03-31
SnowflakeAvroConverter formats bytes as ISO_8859_1 string


tom-j-irvine:
When using the `SnowflakeAvroConverter`, records that contain binary values (`BYTES`) end up formatted as ISO_8859_1 strings.  This is because the `parseAvroWithSchema()` method simply calls `toString()` on the `GenericRecord`.

This is an issue because there is no way in Snowflake to convert this back to binary.  The `TO_BINARY()` function only accepts formats of HEX, BASE64, or UTF-8.  If the string was Base64 encoded (like other converters are doing), it could easily be converted with `BASE64_DECODE_BINARY()`

sfc-gh-japatel:
Can you confirm the behavior with Community/Confluent Avro converter? `io.confluent.connect.avro.AvroConverter`

tom-j-irvine:
With the community converter, you get base64 - this was my initial workaround, but I also need to allow tombstone messages, so that wouldn't allow me to use it.  Another difference between the two would be that some of the date/time logical types are handled correctly (back to actual dates or times) by the confluent converter, but the SnowflakeAvroConverter leaves this values "raw" (so, we get int values, not dates/times).

tom-j-irvine:
To add a bit more detail comparing the `SnowflakeAvroConverter` with the confluent `AvroConverter`

With `bytes`:
* Confluent encodes as base64, so I can do `BASE64_DECODE_BINARY(RECORD_CONTENT:SomeField)`
* Snowflakes encodes as a string, so I tried `TO_BINARY(CAST(RECORD_CONTENT:SomeField AS STRING), 'UTF-8')` but this only worked on some values - any time the ISO_8859_1 string deviated from UTF-8, everything went to mush

With logical type `date`:
* Confluent turns the int into a date, so I can do `TO_DATE(RECORD_CONTENT:SomeField)`
* Snowflake keeps the int value, so I have to do `TO_DATE(DATEADD('dd', RECORD_CONTENT:SomeField, '1970-01-01'))`

With logical type `time-millis`:
* Confluent turns the int into a time, so I can do `TO_TIME(RECORD_CONTENT:SomeField, 'HH24:MI:SS.FF3+TZHTZM')`
* Snowflake keeps the int value, so I have to do `TO_TIME(TIMEADD('ms', RECORD_CONTENT:SomeField, '1970-01-01'))`

Other logical types that I haven't used that would likely behave similarly:
* time-micros
[...]

Issue #481, opened 2022-09-02
Tombstone handling for Json schema data from schema registry


mounikamuddasani1:
We are trying to handle tombstones in topic with Jsonschema using schemaregistr. 
When the Kafka connector encounters a tombstone record, it needs to insert an empty JSON string in the record content column in snowflake.

Connector version used: 1.7.0 
When using behavior.on.null.value:DEFAULT, 

it fails with error

message:
[SF_KAFKA_CONNECTOR] Failed to parse JSON record
[SF_KAFKA_CONNECTOR] java.io.CharConversionException: Invalid UTF-32 character 0xa17a2241 (above 0x0010ffff) at char #1, byte #7)
serviceName:kafka-connect

mounikamuddasani1:
Connector config:

{
        "connector.class": "com.snowflake.kafka.connector.SnowflakeSinkConnector",
        "snowflake.topic2table.map": "TopicName-dev:tablename_dev",
        "topics": "TopicName-dev",
        "tasks.max": "3",
        "snowflake.private.key.passphrase": "${paramstore:snowflake/private.key.passphrase}",
        "buffer.flush.time": "1800",
        "snowflake.url.name": "${paramstore:snowflake/url.name}",
        "snowflake.database.name": "${paramstore:snowflake/database.name}",
        "snowflake.schema.name": "${paramstore:snowflake/schema.name}",
        "value.converter.schema.registry.url": "${paramstore:schemaRegistryUrl}",
        "buffer.count.records": "10000",
        "snowflake.private.key": "${paramstore:snowflake/private.key}",
        "snowflake.user.name": "${paramstore:snowflake/user.name}",
        "name": "Connector_dev",
        "value.converter.schemas.enable": "false",
        "value.converter.break.on.schema.registry.error": "false",
        "value.converter": "com.snowflake.kafka.connector.records.SnowflakeJsonConverter",
        "key.converter":"org.apache.kafka.connect.storage.StringConverter",
        "behavior.on.null.values": "DEFAULT",
        "consumer.override.auto.offset.reset": "latest"
    }

sfc-gh-xhuang:
We recommend using Snowpipe Streaming and version 2.1.0 instead of version 1.7.0
V 1.7.0 is unsupported

Issue #416, opened 2022-03-25
Error Code: 5016 SinkRecord.value and SinkRecord.valueSchema cannot be null


alexnikitchuk:
According to [docs](https://docs.snowflake.com/en/user-guide/kafka-connector-install.html) default setting of `behavior.on.null.values` is `DEFAULT`  which means when the Kafka connector encounters a tombstone record, it inserts an empty JSON string in the content column.
But in our case we receive the error instead (see below).
We use protobuf serialization.
Sink connector v.1.5.5
Can someone clarify please?

```
[SF_KAFKA_CONNECTOR] Exception: Invalid SinkRecord received
[SF_KAFKA_CONNECTOR] Error Code: 5016
[SF_KAFKA_CONNECTOR] Detail: SinkRecord.value and SinkRecord.valueSchema cannot be null (org.apache.kafka.connect.runtime.WorkerSinkTask)
com.snowflake.kafka.connector.internal.SnowflakeKafkaConnectorException:
[SF_KAFKA_CONNECTOR] Exception: Invalid SinkRecord received
[SF_KAFKA_CONNECTOR] Error Code: 5016
[SF_KAFKA_CONNECTOR] Detail: SinkRecord.value and SinkRecord.valueSchema cannot be null
	at com.snowflake.kafka.connector.internal.SnowflakeErrors.getException(SnowflakeErrors.java:281)
	at com.snowflake.kafka.connector.internal.SnowflakeErrors.getException(SnowflakeErrors.java:248)
	at com.snowflake.kafka.connector.records.RecordService.processRecord(RecordService.java:91)
	at com.snowflake.kafka.connector.internal.SnowflakeSinkServiceV1$ServiceContext$PartitionBuffer.insert(SnowflakeSinkServiceV1.java:982)
	at com.snowflake.kafka.connector.internal.SnowflakeSinkServiceV1$ServiceContext$PartitionBuffer.access$700(SnowflakeSinkServiceV1.java:950)
	at com.snowflake.kafka.connector.internal.SnowflakeSinkServiceV1$ServiceContext.insert(SnowflakeSinkServiceV1.java:591)
	at com.snowflake.kafka.connector.internal.SnowflakeSinkServiceV1$ServiceContext.access$300(SnowflakeSinkServiceV1.java:335)
	at com.snowflake.kafka.connector.internal.SnowflakeSinkServiceV1.insert(SnowflakeSinkServiceV1.java:180)
[...]

PR #688, opened 2023-08-18
[SNOW-851840] Enable tombstone record ingestion in Snowpipe Streaming

sfc-gh-rcheng:
Allows tombstone records (records with NULL value and value schema) to be ingested.

- Main change in RecordService creates an empty SnowflakeRecordValue when given a null value or schema

Motivation:
- 6 customers across slack requests and github issues 
  - https://github.com/snowflakedb/snowflake-kafka-connector/issues/481
  - https://github.com/snowflakedb/snowflake-kafka-connector/issues/416
  - https://github.com/snowflakedb/snowflake-kafka-connector/issues/599
- [1 pager](https://docs.google.com/document/d/1pvaQi6NaXGxqom2BrUS2oMsvyh9QTdHPGEUHqRo6aS0/edit?usp=sharing)


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -172,18 +172,6 @@ public void start(final Map<String, String> parsedConfig) {
     final long bufferFlushTime =
         Long.parseLong(parsedConfig.get(SnowflakeSinkConnectorConfig.BUFFER_FLUSH_TIME_SEC));
 
-    // Falling back to default behavior which is to ingest an empty json string if we get null
-    // value. (Tombstone record)
-    SnowflakeSinkConnectorConfig.BehaviorOnNullValues behavior =
-        SnowflakeSinkConnectorConfig.BehaviorOnNullValues.DEFAULT;
-    if (parsedConfig.containsKey(SnowflakeSinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG)) {
```
sfc-gh-rcheng, line 179: BehaviorOnNullValues (aka ingest or don't ingest tombstone records) is only used in RecordService. Thus we don't need `setBehaviorOnNullValuesConfig` and can just pass the config into RecordService

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -924,6 +924,19 @@ public static String getExceptionMessage(String customMessage, Exception ex) {
...
   }
 
+  public static SnowflakeSinkConnectorConfig.BehaviorOnNullValues getBehaviorOnNullValuesEnum(
+      Map<String, String> config) {
+    return Arrays.stream(SnowflakeSinkConnectorConfig.BehaviorOnNullValues.values())
+        .filter(
+            behavior ->
+                behavior
+                    .toString()
+                    .equalsIgnoreCase(
```
sfc-gh-rcheng, line 934: needed because we need to convert the config param to the enum with an `equalsIgnoreCase`.

Alternatively I can do something like the following since we only have two options in the enum, however it's not optimal if we decide to add any more behaviors

```
return config.get(BEHAVIOR_ON_NULL_VALUES_CONFIG).equalsIgnoreCase(SnowflakeSinkConnectorConfig.BehaviorOnNullValues.IGNORE.toString()) ?
        SnowflakeSinkConnectorConfig.BehaviorOnNullValues.IGNORE :
        SnowflakeSinkConnectorConfig.BehaviorOnNullValues.DEFAULT;
```
return

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -89,60 +86,47 @@ public class RecordService {
...
 
   /**
    * process records output JSON format: { "meta": { "offset": 123, "topic": "topic name",
    * "partition": 123, "key":"key name" } "content": "record content" }
    *
    * <p>create a JsonRecordService instance
-   *
-   * @param telemetryService Telemetry Service Instance. Can be null.
    */
-  public RecordService(SnowflakeTelemetryService telemetryService) {
```
sfc-gh-rcheng, line 104: telemetryService is never used in RecordService, so I removed it

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -152,19 +136,25 @@ public void setEnableSchematization(final boolean enableSchematization) {
...
     if (record.value() == null || record.valueSchema() == null) {
-      throw SnowflakeErrors.ERROR_5016.getException();
-    }
-    if (!record.valueSchema().name().equals(SnowflakeJsonSchema.NAME)) {
-      throw SnowflakeErrors.ERROR_0009.getException();
-    }
-    if (!(record.value() instanceof SnowflakeRecordContent)) {
-      throw SnowflakeErrors.ERROR_0010.getException(
-          "Input record should be SnowflakeRecordContent object");
+      if (this.ingestTombstoneRecords) {
```
sfc-gh-rcheng, line 142: This is the crux of the PR - if we want to ingestTombstoneRecords, create an empty record

src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
```diff
@@ -146,173 +148,6 @@ public void testIngestion() throws Exception {
     //    assert !conn.pipeExist(pipe);
   }
 
-  @Test
-  public void testTombstoneRecords_DEFAULT_behavior_ingestion_SFJsonConverter() throws Exception {
```
sfc-gh-rcheng, line 150: moved to TombstoneRecordIngestionIT.java

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -108,13 +109,12 @@ public RecordService(SnowflakeTelemetryService telemetryService) {
   /** Record service with null telemetry Service, only use it for testing. */
   @VisibleForTesting
   public RecordService() {
-    this.telemetryService = null;
+    this(null);
```
sfc-gh-japatel, line 112: set default here?

sfc-gh-rcheng: I don't think that is necessary as it its set to default on line 75?

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -89,7 +91,6 @@ public class RecordService {
 
   // This class is designed to work with empty metadata config map
   private SnowflakeMetadataConfig metadataConfig = new SnowflakeMetadataConfig();
-
```
sfc-gh-tzhang, line 92: Is this change intended?

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -108,13 +109,12 @@ public RecordService(SnowflakeTelemetryService telemetryService) {
...
   @VisibleForTesting
   public RecordService() {
-    this.telemetryService = null;
+    this(null);
   }
 
   public void setMetadataConfig(SnowflakeMetadataConfig metadataConfigIn) {
     metadataConfig = metadataConfigIn;
   }
-
```
sfc-gh-tzhang, line 117: same here

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -297,7 +316,7 @@ public SnowflakeTableRow(SnowflakeRecordContent content, JsonNode metadata) {
   }
 
   void putKey(SinkRecord record, ObjectNode meta) {
-    if (record.key() == null) {
+    if (record.key() == null || record.keySchema() == null) {
```
sfc-gh-tzhang, line 319: why we can't put the key when the keySchema is null but not the key? This change might be risky and introduce a behavior change for both Snowpipe and Snowpipe Streaming

sfc-gh-tzhang (approved):
LGTM, only comment is about a potential risky change which I'm not sure if it's necessary

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-rcheng: reverted, thanks for the catch

src/test/java/com/snowflake/kafka/connector/internal/SinkServiceIT.java
```diff
@@ -749,19 +579,23 @@ public void testNativeNullIngestion() throws Exception {
     service.closeAll();
   }
 
-  @Test(expected = SnowflakeKafkaConnectorException.class)
```
sfc-gh-rcheng, line 752: We need to allow null value ingestion because community converters create null values instead of empty SnowflakeRecordContents. Currently Snowpipe fails to ingest null value records with community converters.

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-rcheng: a null keySchema will NPE on the next if statement, however we should never get into this situation due to previous handleNativeRecord() calls. I'll update the logic here to both put the key and handle the NPE

sfc-gh-rcheng: the current code will NPE on the next if statement if keySchema is null. 

handleNativeRecord() prevents null keySchemas from being passed into putKey, however we should throw a better error if we ever get a null keySchema, so I added a null check.

This shouldn't be risky because we are changing the error thrown from NPE to a Snowflake error.

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -301,6 +322,12 @@ void putKey(SinkRecord record, ObjectNode meta) {
       return;
     }
 
+    if (record.keySchema() == null) {
```
sfc-gh-japatel, line 325: why do we need this?

src/test/java/com/snowflake/kafka/connector/ConnectorConfigTest.java
```diff
@@ -16,11 +16,59 @@
...
+      this.name = name;
+      this.converter = converter;
+    }
+
+    public String toString() {
+      return this.name;
+    }
+  };
+
+  public enum CustomSfConverter {
```
sfc-gh-japatel, line 48: This is fine but we do have
` public static final Set<String> CUSTOM_SNOWFLAKE_CONVERTERS ` which can be resued.

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-japatel: I am concerned about this change.

sfc-gh-rcheng: discussed in this comment - ensuring we throw a snowflake error instead of NPE. We should never get into this situation, but i want to add this just in case

src/test/java/com/snowflake/kafka/connector/ConnectorConfigTest.java
sfc-gh-rcheng: i considered using that, but i preferred having the enum

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-japatel: keyschemas are often null
I dont think we should do this. this seems like a behavior change.

sfc-gh-rcheng: keyschemas are often null, we update it in handleNativeRecords earlier in the call stack. if the key schema is null in this method the next if statement throws an NPE. 

The only behavior change here is to change the exception thrown from an NPE to a Snowflake Error

sfc-gh-japatel: I See, thanks. Sounds like a harmless change but its recommended to not have refactoring/fixes in the same PR as features. 

This is a small enough change to go through but it just raises more questions from reviewers POV

sfc-gh-japatel (approved):
lgtm

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-rcheng: I see, it was caught during testing that method since i wanted to give all possible null scenarios

sfc-gh-rcheng:
@aidan-melen this should go out in release 2.0.2, which should be released in the next couple weeks

PR #693, opened 2023-08-24
SNOW-808908: Handle Reserved Keywords and Special Characters in Schematization

sfc-gh-tzhang:
To Handle Reserved Keywords and Special Characters in Schematization, we will convert all names to uppercase and then add double quotes, except for the names that have double quotes already. By doing this, we tell compiler to bypass the reserved keyword or special characters check, and we’re also preserving the old behavior for customers
Examples:

- start -> “START”
- a@b -> “A@B”
- c1/C1 -> c1, C1, “C1” would all work
- “ab” -> “ab”


src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -137,6 +138,94 @@ public class Utils {
 
   private static final KCLogger LOGGER = new KCLogger(Utils.class.getName());
 
+  private static final Pattern unquotedIdentifierPattern =
+      Pattern.compile("[$][0-9]+|[_a-zA-Z][_a-zA-Z0-9]*[$]?[_a-zA-Z0-9]*");
+
+  private static final String[] reservedKeywords = {
```
sfc-gh-xhuang, line 144: Do we need to put these reservedKeywords here? It will not always be kept in sync / be the full ist

sfc-gh-tzhang:
ping, can I get a review on this one?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -137,6 +137,18 @@ public class Utils {
 
   private static final KCLogger LOGGER = new KCLogger(Utils.class.getName());
 
+  /**
+   * Quote the column name if needed: When there is quote already, we do nothing; otherwise we
+   * convert the name to upper case and add quote
+   */
+  public static String quoteNameIfNeeded(String name) {
+    int length = name.length();
+    if (name.charAt(0) == '"' && (length >= 2 && name.charAt(length - 1) == '"')) {
```
sfc-gh-japatel, line 146: can use startswith and endswith?

src/test/java/com/snowflake/kafka/connector/internal/TestUtils.java
```diff
@@ -354,8 +355,16 @@ static Map<String, String> getConfWithEncryptedKey() {
   }
 
   public static Map<String, String> getConfWithOAuth() {
+    // Return null if we don't have the corresponding OAuth required configuration in the config
+    // file
+    Map<String, String> conf = getConf();
```
sfc-gh-japatel, line 360: can you make oauth changes in separate PRs?

src/test/java/com/snowflake/kafka/connector/records/RecordContentTest.java
```diff
@@ -271,6 +271,6 @@ public void testColumnNameFormatting() throws JsonProcessingException {
     Map<String, Object> got = service.getProcessedRecordForStreamingIngest(record);
 
     assert got.containsKey("\"NaMe\"");
-    assert got.containsKey("AnSwEr");
+    assert got.containsKey("\"ANSWER\"");
```
sfc-gh-japatel, line 274: to understand again, is this because name is double quoted but answer is not?

sfc-gh-japatel (commented):
does any of the test we have connects to JDBC with this column name changes and verifies the behavior?
I also think this deserves a new End to end test. WDYT

Overall lgtm! Thank you Toby!

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -137,6 +137,18 @@ public class Utils {
 
   private static final KCLogger LOGGER = new KCLogger(Utils.class.getName());
 
+  /**
+   * Quote the column name if needed: When there is quote already, we do nothing; otherwise we
+   * convert the name to upper case and add quote
+   */
+  public static String quoteNameIfNeeded(String name) {
```
sfc-gh-rcheng, line 144: add unit tests for this method

sfc-gh-rcheng (commented):
lgtm, need unit tests for the quoteNameIfNeeded method

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-tzhang: No strong preference here, but charAt should performance better than startswith? I see startswith is usually used for matching a string instead of a character

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-tzhang: added

src/test/java/com/snowflake/kafka/connector/records/RecordContentTest.java
sfc-gh-tzhang: correct

sfc-gh-tzhang:
> does any of the test we have connects to JDBC with this column name changes and verifies the behavior? I also think this deserves a new End to end test. WDYT
> 
> Overall lgtm! Thank you Toby!

We many both IT and end2end testing for this. For IT, one example is testSchematizationWithTableCreationAndAvroInput where we insert the rows and compare the result by scanning the table. For end2end, one example is TestSchemaEvolutionJson where we test names with and without double quotes, and compare it with gold_type in the end. Let me know if you think anything is missing here.

sfc-gh-rcheng (approved):

PR #694, opened 2023-08-25
[Streaming Telemetry 2][SNOW-899866] Enables reportKafkaPartitionStart/Usage telemetry for Streaming

sfc-gh-rcheng:
- SnowflakeTelemetryChannelStatus becomes a telemetry object, similar to SnowflakeTelemetryPipeUsage
- Creates SnowflakeTelemetryChannelCreation, similar to SnowflakeTelemetryPipeCreation
- Unifies reportKafkaPartitionStart/Usage code paths and enables telemetry for Streaming
- Adds additional code coverage to the TelemetryService
https://snowflakecomputing.atlassian.net/browse/SNOW-899866


src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -271,11 +272,21 @@ public TopicPartitionChannel(
 
     this.enableSchemaEvolution = this.enableSchematization && hasSchemaEvolutionPermission;
 
+    // telemetry and metrics
+    this.snowflakeTelemetryChannelStatus =
```
sfc-gh-rcheng, line 276: we create this outside of the telemetryServiceV2 null check because JMX metrics relies on this status

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -284,12 +294,6 @@ public TopicPartitionChannel(
               + " correct offset instead",
           this.getChannelName());
     }
-
-    if (enableCustomJMXMonitoring) {
```
sfc-gh-rcheng, line 288: moved to inside SnowflakeTelemetryChannelStatus constructor creation

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1095,10 +1132,9 @@ protected SnowflakeTelemetryService getTelemetryServiceV2() {
     return this.telemetryServiceV2;
   }
 
-  protected void setLatestConsumerOffset(long consumerOffset) {
```
sfc-gh-rcheng, line 1098: moved above with the other offset setters

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1317,45 +1353,4 @@ public String toString() {
...
   }
-
-  /**
-   * Registers all the Metrics inside the metricRegistry.
-   *
-   * @param channelName channelName
-   * @param metricsJmxReporter wrapper class for registering all metrics related to above connector
-   *     and channel
-   */
-  public void registerChannelJMXMetrics(
```
sfc-gh-rcheng, line 1328: moved into SnowflakeTelemetryChannelStatus

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryServiceV2.java
```diff
@@ -38,10 +40,9 @@ public SnowflakeTelemetryServiceV2(Connection conn) {
     this.telemetry = TelemetryClient.createTelemetry(conn);
   }
 
-  @Override
```
sfc-gh-rcheng, line 41: moved reportKafkaPartitionUsage to the parent SnowflakeTelemetryService

src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryBasicInfo.java
```diff
@@ -7,19 +7,22 @@
...
+  public static final KCLogger LOGGER = new KCLogger(SnowflakeTelemetryBasicInfo.class.getName());
 
   /**
    * Base Constructor. Accepts a tableName and StageName.
    *
    * @param tableName Checks for Nullability
    */
-  public SnowflakeTelemetryBasicInfo(final String tableName) {
+  public SnowflakeTelemetryBasicInfo(
+      final String tableName, SnowflakeTelemetryService.TelemetryType telemetryType) {
```
sfc-gh-rcheng, line 21: Pass telemetry type here to simplify reportKafkaPartitionUsage/Creation code paths across snowpipe and streaming

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -289,6 +290,8 @@ public TopicPartitionChannel(
             this.offsetPersistedInSnowflake,
             this.processedOffset,
             this.latestConsumerOffset);
+    this.telemetryServiceV2.reportKafkaPartitionStart(
+        new SnowflakeTelemetryChannelCreation(this.tableName, this.channelName));
```
sfc-gh-tzhang, line 294: Is this the only place that we're creating channel? If no, we need to either cover all cases or update the metric name to something like partition start

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
```diff
@@ -18,23 +18,32 @@
...
+ *
+ * <p>Most of the data sent to Snowflake is an aggregated data.
+ */
+public class SnowflakeTelemetryChannelStatus extends SnowflakeTelemetryBasicInfo {
   public static final long NUM_METRICS = 3; // update when new metrics are added
 
   // channel properties
   private final String channelName;
   private final MetricsJmxReporter metricsJmxReporter;
+  private final AtomicLong startTime;
```
sfc-gh-tzhang, line 46: why we need AtomicLong here?

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
```diff
@@ -75,6 +87,28 @@ public SnowflakeTelemetryChannelStatus(
     }
   }
 
+  @Override
+  public boolean isEmpty() {
```
sfc-gh-tzhang, line 91: I assume this is used to not send data when everything is default? If yes, I think we should always send since it can help us to know a restart or a table is dropped so everything is reset, wdyt?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1044,6 +1047,9 @@ private SnowflakeStreamingIngestChannel openChannelForTable() {
   public void closeChannel() {
     try {
       this.channel.close().get();
+
+      // telemetry and metrics
+      this.telemetryServiceV2.reportKafkaPartitionUsage(this.snowflakeTelemetryChannelStatus, true);
```
sfc-gh-tzhang, line 1052: Should we send this somewhere periodically instead of just during close?

sfc-gh-tzhang (commented):
Left some comments, PTAL. This is super helpful for us to debug issues without asking for client logs, thanks!

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-rcheng: That would be ideal, but currently we don't have anything running periodically in the background. InsertRows is an option, but I'm worried about overloading the telemetry service since we have no control over how often the customer flushes their buffer

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-rcheng: Yes this is the only place we create a channel. There are two constructors for TopicPartitionChannel. The test constructor calls into this constructor, so all TopicPartitionChannel object creations will go through this codepath

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
sfc-gh-rcheng: Correct that is what it is used for. Makes sense to send this regardless of state since this is the 'close' event corresponding with the reportKafkaPartitionStart event. 

Made the change in SnowflakeTelemetryService.java, this means we're making a behavior change to mirror this behavior into Snowpipe

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
sfc-gh-rcheng: I'll change it to a normal final long.

startTime represents when the SnowflakeTelemetryChannelStatus object started tracking data. it gets updated to the current time every time we send telemetry data, these were behaviors I took over from the corresponding Snowpipe object.

Currently since we only send telemetry data once, it can be a final long, however in the future if we want to send multiple SnowflakeTelemetryChannelStatus's, an AtomicLong may make more sense for threading.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: I thought we also reopen/create channel when the channel gets invalidated? Not sure if this is the same code path though

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: Is it something we plan to add in the future? I think we send the usage for Snowpipe periodically. Otherwise I feel like this is not very useful since there're no usage reported if there is no open/close on the channel

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-rcheng: The TopicPartitionChannel object does not get recreated when the channel is invalidated, so the current reportKafkaPartitionStart tracks when a new TopicPartitionChannel is created.

Alternatively we could emit a reportKafkaPartitionStart with each channel open, however I don't think this is a good idea because we can track this server side and we already have multiple channel open/close issues that may overload the telemetry service with events. 

There are two code paths for opening channels: [TopicPartitionChannel constructor](https://github.com/snowflakedb/snowflake-kafka-connector/blob/bc7af5e54774697d297eb31b4a6980c98cca7588/src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java#L277) and when we [recover offset from snowflake](https://github.com/snowflakedb/snowflake-kafka-connector/blob/bc7af5e54774697d297eb31b4a6980c98cca7588/src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java#L977) (channel invalidation).

I will add the a channel open count to the reportPartitionUsage event as a way to track the sequencer

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-rcheng: We do not have any plans to add that in the future. It could potentially be a part of the flush service.

Currently we not periodically send Snowpipe usage. `reportKafkaPartition(`) is called on on [close](https://github.com/snowflakedb/snowflake-kafka-connector/blob/bc7af5e54774697d297eb31b4a6980c98cca7588/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java#L964) and when the [cleaner runs](https://github.com/snowflakedb/snowflake-kafka-connector/blob/bc7af5e54774697d297eb31b4a6980c98cca7588/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java#L505), which only runs on the first insert to initialize a pipe

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelCreation.java
```diff
@@ -0,0 +1,62 @@
...
+import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.node.ObjectNode;
+
+/**
+ * This object is sent only once when a channel starts. No concurrent modification is made on this
+ * object, thus no lock is required.
+ */
+public class SnowflakeTelemetryChannelCreation extends SnowflakeTelemetryBasicInfo {
+  private final long startTime; // start time of the channel
+  private final String channelName;
+  private boolean isReuseTable = false; // is the channel reusing existing table
```
sfc-gh-japatel, line 36: why is this needed?
we already have channel sequencers to help us with that

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelCreation.java
```diff
@@ -0,0 +1,62 @@
...
+import com.snowflake.kafka.connector.internal.telemetry.SnowflakeTelemetryBasicInfo;
+import com.snowflake.kafka.connector.internal.telemetry.SnowflakeTelemetryService;
+import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.node.ObjectNode;
+
+/**
+ * This object is sent only once when a channel starts. No concurrent modification is made on this
+ * object, thus no lock is required.
+ */
+public class SnowflakeTelemetryChannelCreation extends SnowflakeTelemetryBasicInfo {
+  private final long startTime; // start time of the channel
```
sfc-gh-japatel, line 34: is this open channel start time or time when first row was ingested?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1034,6 +1040,7 @@ private SnowflakeStreamingIngestChannel openChannelForTable() {
             .build();
     LOGGER.info(
         "Opening a channel with name:{} for table name:{}", this.channelName, this.tableName);
+    this.channelTryOpenCount.incrementAndGet();
```
sfc-gh-japatel, line 1043: Sorry I dont understand use of this metrics, this will always be 1 right? 

Right place would be to increment in `openChannelForTable` right?

sfc-gh-japatel, line 1043: we are incrementing it here but is this sent to snowflake in telemetry?

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
```diff
@@ -18,33 +18,45 @@
...
   private final MetricsJmxReporter metricsJmxReporter;
+  private final long startTime;
 
   // offsets
   private final AtomicLong offsetPersistedInSnowflake;
   private final AtomicLong processedOffset;
   private final AtomicLong latestConsumerOffset;
 
+  // channel metrics
+  private final AtomicLong channelTryOpenCount;
```
sfc-gh-japatel, line 54: Umm, should we rename it something else? this is more of a channelReopenCount in current partition usage..
This will be super confusing with our channel sequencers.. at any given point in time channelSequencers >= channelTryOpenCount

This value will be reset to 0 when there is a rebalance right?
I do think this will be useful in cases during rebalance..
May be rename this to `channelReopenCountInActiveOpenTask` 

I cant think of a better name though, open to suggestions.

src/main/java/com/snowflake/kafka/connector/internal/telemetry/TelemetryConstants.java
```diff
@@ -8,11 +8,14 @@ public final class TelemetryConstants {
...
   public static final String STAGE_NAME = "stage_name";
   public static final String PIPE_NAME = "pipe_name";
+  public static final String CHANNEL_NAME = "channel_name";
 
   public static final String PROCESSED_OFFSET = "processed-offset";
   public static final String FLUSHED_OFFSET = "flushed-offset";
   public static final String COMMITTED_OFFSET = "committed-offset";
   public static final String PURGED_OFFSET = "purged-offset";
+  public static final String OFFSET_PERSISTED_IN_SNOWFLAKE = "persisted-in-snowflake-offset";
+  public static final String LATEST_CONSUMER_OFFSET = "latest-consumer-offset";
```
sfc-gh-japatel, line 18: may be a move everything to a section which says snowpipe streaming?

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelCreation.java
sfc-gh-rcheng: This was something I added in response to this [comment](https://github.com/snowflakedb/snowflake-kafka-connector/pull/694#discussion_r1323337498). I don't feel strongly about keeping it, so i'll remove it

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelCreation.java
sfc-gh-japatel: this is more of a start time in current open partition call. 
I understand we are following the pattern from snowpipe but lets only add if it is useful or doesnt create more confusion.

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelCreation.java
```diff
@@ -0,0 +1,62 @@
...
+  public SnowflakeTelemetryChannelCreation(final String tableName, String channelName) {
+    super(tableName, SnowflakeTelemetryService.TelemetryType.KAFKA_CHANNEL_START);
+    this.channelName = channelName;
+    this.startTime = System.currentTimeMillis();
+  }
+
+  @Override
+  public void dumpTo(ObjectNode msg) {
+    msg.put(TABLE_NAME, this.tableName);
+    msg.put(CHANNEL_NAME, this.channelName);
```
sfc-gh-japatel, line 47: should we add `CONNECTOR_NAME` too? until we get connector name in our channel name? WDYT?

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelCreation.java
sfc-gh-rcheng: This is open channel start time - when TopicPartitionChannel was created

sfc-gh-japatel (commented):
mostly looks good, left some questions and suggestions. PTAL and let me know what you think. Thanks!

sfc-gh-japatel:
Lets also verify the data we send to snowflake and confirm we dont change anything in snowpipe telemetry. You can pick any of the above gh action runs and verify the data in telemetry view in snowhouse. (I dont think anything has changed but just to be double sure)

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelCreation.java
sfc-gh-japatel: sounds good to remove, thanks!

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelCreation.java
sfc-gh-rcheng: Good point, added connector name to telemetry. Didn't add to metrics because it is surfaced in metric registry already

src/main/java/com/snowflake/kafka/connector/internal/telemetry/TelemetryConstants.java
sfc-gh-rcheng: Added a Streaming subclass

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
sfc-gh-rcheng: removed variable, n/a

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelCreation.java
sfc-gh-rcheng: updated the constant to `TP_CHANNEL_CREATION_TIME = "tp_channel_creation_time"` to clarify what the timestamp means. Prepended 'TP' to clarify TopicPartitionChannel in comparison to a streaming channel

src/main/java/com/snowflake/kafka/connector/internal/telemetry/TelemetryConstants.java
```diff
@@ -46,4 +47,12 @@ public final class TelemetryConstants {
   public static final String IS_REUSE_PIPE = "is_reuse_pipe";
   public static final String FILE_COUNT_RESTART = "file_count_restart";
   public static final String FILE_COUNT_REPROCESS_PURGE = "file_count_reprocess_purge";
+
+  public class Streaming {
+    public static final String OFFSET_PERSISTED_IN_SNOWFLAKE = "persisted-in-snowflake-offset";
```
sfc-gh-japatel, line 52: 1. remove inner class. previously recommendation was to have a separate section/heading in this class which distinguishes between snowpipe and streaming. 
2. lets clarify what tp here means. it is not intuitive but hopefully they read code and understand what tp means.

sfc-gh-rcheng: 1. discussed offline, added as a comment section
2. I'll expand TP -> TopicPartition

sfc-gh-rcheng:
Telemetry in snowhouse
image
image

src/main/java/com/snowflake/kafka/connector/internal/metrics/MetricsUtil.java
```diff
@@ -91,6 +83,16 @@ public class MetricsUtil {
 
   public static final String LATENCY_SUB_DOMAIN = "latencies";
 
+  // ************ Streaming Constants ************//
+  /**
+   * See {@link com.snowflake.kafka.connector.internal.streaming.TopicPartitionChannel} for offset
+   * description
+   */
+  public static final String OFFSET_PERSISTED_IN_SNOWFLAKE = "persisted-in-snowflake-offset";
```
sfc-gh-japatel, line 91: 👍

sfc-gh-japatel (approved):
Follow up:
1. merge gates are failing, please ensure they are passing before pushing
2. doc changes - please create a jira for @sfc-gh-lema (for jmx)

LGTM, otherwise! Good stuff! Thank you

sfc-gh-rcheng:
[Jira](https://snowflakecomputing.atlassian.net/browse/SNOW-870373) for documentation changes @sfc-gh-lema

PR #696, opened 2023-08-30
[Streaming Telemetry 1] Prepares TopicPartitionChannel for telemetry 

sfc-gh-rcheng:
- Requires TelemetryService not null in TopicPartitionChannel
- Adds offset setters in TopicPartitionChannel to update offset and SnowflakeTelemetryChannelStatus offsets at the same time
- Moves JMX into new SnowflakeTelemetryChannelStatus 

Changes coming in [part 2 PR](https://github.com/snowflakedb/snowflake-kafka-connector/pull/694)
- SnowflakeTelemetryChannelStatus becomes a telemetry object
- Enables reportKafkaPartitionStart/Usage telemetry


src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -284,12 +291,6 @@ public TopicPartitionChannel(
               + " correct offset instead",
           this.getChannelName());
     }
-
-    if (enableCustomJMXMonitoring) {
```
sfc-gh-rcheng, line 288: moved into snowflakeTelemetryChannelStatus constructor

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1095,10 +1127,9 @@ protected SnowflakeTelemetryService getTelemetryServiceV2() {
     return this.telemetryServiceV2;
   }
 
-  protected void setLatestConsumerOffset(long consumerOffset) {
```
sfc-gh-rcheng, line 1098: moved up to the other offset setters

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
```diff
@@ -0,0 +1,142 @@
...
+import static com.snowflake.kafka.connector.internal.metrics.MetricsUtil.constructMetricName;
+
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.MetricRegistry;
+import com.snowflake.kafka.connector.internal.KCLogger;
+import com.snowflake.kafka.connector.internal.metrics.MetricsJmxReporter;
+import com.snowflake.kafka.connector.internal.metrics.MetricsUtil;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class SnowflakeTelemetryChannelStatus {
```
sfc-gh-rcheng, line 29: Will extend SnowflakeTelemetryBasicInfo and become a proper telemetry object in the next PR

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryServiceV2.java
sfc-gh-rcheng, line 1: moved into streaming/telemetry folder

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelTest.java
```diff
@@ -817,27 +835,52 @@ public void testBigAvroBufferBytesThreshold() throws Exception {
   }
 
   @Test
-  public void testRegisterJmxMetrics() {
```
sfc-gh-rcheng, line 820: moved into SnowflakeTelemetryChannelStatusTest

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -128,6 +131,13 @@ public SnowflakeSinkServiceV2(
     this.partitionsToChannel = new HashMap<>();
 
     this.tableName2SchemaEvolutionPermission = new HashMap<>();
+
+    // jmx
+    String connectorName =
+        conn == null || Strings.isNullOrEmpty(this.conn.getConnectorName())
+            ? "default_connector"
```
sfc-gh-japatel, line 138: Ideally this is not possible since name is a requirement but good to check.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1065,6 +1068,24 @@ public String getChannelName() {
...
+    this.snowflakeTelemetryChannelStatus.setOffsetPersistedInSnowflake(offsetPersistedInSnowflake);
+  }
+
+  private void setProcessedOffset(long processedOffset) {
+    this.processedOffset.set(processedOffset);
+    this.snowflakeTelemetryChannelStatus.setProcessedOffset(processedOffset);
+  }
+
+  protected void setLatestConsumerOffset(long consumerOffset) {
+    if (consumerOffset > this.latestConsumerOffset) {
```
sfc-gh-japatel, line 1083: we arent checking this in original code, so lets not add this

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
sfc-gh-japatel, line 29: nit: javadocs fine if coming in next PR

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
```diff
@@ -0,0 +1,142 @@
...
+      new KCLogger(SnowflakeTelemetryChannelStatus.class.toString());
+
+  public static final long NUM_METRICS = 3; // update when new metrics are added
+  private static final long INITIAL_OFFSET = -1;
+
+  // channel properties
+  private final String channelName;
+  private final MetricsJmxReporter metricsJmxReporter;
+
+  private final AtomicLong startTime;
```
sfc-gh-japatel, line 40: this seems unused, remove if not needed

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
```diff
@@ -0,0 +1,142 @@
...
+
+  // offsets
+  private AtomicLong offsetPersistedInSnowflake;
+  private AtomicLong processedOffset;
+  private long latestConsumerOffset;
```
sfc-gh-japatel, line 45: do we need atomic long?

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelTest.java
```diff
@@ -817,27 +835,52 @@ public void testBigAvroBufferBytesThreshold() throws Exception {
...
-    MetricRegistry metricRegistry = Mockito.spy(MetricRegistry.class);
-    MetricsJmxReporter metricsJmxReporter =
-        Mockito.spy(new MetricsJmxReporter(metricRegistry, TEST_CONNECTOR_NAME));
+  public void testTopicPartitionChannelInvalidJmxReporter() throws Exception {
```
sfc-gh-japatel, line 838: lets also add a test where there are >1 channels and we verify the metrics values after inserting few records in each channel.

sfc-gh-japatel: check out `SnowflakeTelemetryPipeStatusMetricsIT`

sfc-gh-japatel (approved):
overall lgtm, couple of questions and one request to add a new test.

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
sfc-gh-rcheng: SnowflakeTelemetryPipeStatus uses AtomicLongs, so I figured this should follow precedent

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
sfc-gh-rcheng: Added javadocs

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelTest.java
sfc-gh-rcheng: thanks for the suggestion, there's a similar test in the next PR but it only has one channel. 

Added metrics verification to testStreamingIngest_multipleChannelPartitions_withMetrics in SnowflakeSinkServiceV2IT.java

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
sfc-gh-rcheng: removed

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-tzhang: My two cents: we should not do this if the assumption is connector name should never be null, you can add an assert instead

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1036,6 +1037,8 @@ private SnowflakeStreamingIngestChannel openChannelForTable() {
   public void closeChannel() {
     try {
       this.channel.close().get();
+
```
sfc-gh-tzhang, line 1040: nit: extra line?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1065,6 +1068,24 @@ public String getChannelName() {
     return this.channel.getFullyQualifiedName();
   }
 
+  // ------ SETTERS ------ //
+  private void setOffsetPersistedInSnowflake(long offsetPersistedInSnowflake) {
+    this.offsetPersistedInSnowflake.set(offsetPersistedInSnowflake);
+    this.snowflakeTelemetryChannelStatus.setOffsetPersistedInSnowflake(offsetPersistedInSnowflake);
```
sfc-gh-tzhang, line 1074: Instead of getting the offset from here, why we decide to let the telemetry service maintain a copy of all three offsets?

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
```diff
@@ -0,0 +1,183 @@
...
+        this.metricsJmxReporter.getMetricRegistry().getMetrics().keySet().toString());
+    this.metricsJmxReporter.removeMetricsFromRegistry(this.channelName);
+
+    MetricRegistry currentMetricRegistry = this.metricsJmxReporter.getMetricRegistry();
+
+    try {
+      // offsets
+      currentMetricRegistry.register(
+          constructMetricName(
+              this.channelName,
```
sfc-gh-tzhang, line 89: Not familiar with JMX, but can a channel name be uniquely identify a metric? What if we have two channels with the same name but different client sequencer?

sfc-gh-tzhang (commented):
left some comments, PTAL, thanks!

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
```diff
@@ -0,0 +1,183 @@
...
+import com.snowflake.kafka.connector.internal.KCLogger;
+import com.snowflake.kafka.connector.internal.metrics.MetricsJmxReporter;
+import com.snowflake.kafka.connector.internal.metrics.MetricsUtil;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class SnowflakeTelemetryChannelStatus {
+  private static final KCLogger LOGGER =
+      new KCLogger(SnowflakeTelemetryChannelStatus.class.toString());
+
+  public static final long NUM_METRICS = 3; // update when new metrics are added
```
sfc-gh-tzhang, line 33: Why is this needed? I only see it being used in test, I don't think people will remember update this every time...

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
```diff
@@ -0,0 +1,183 @@
...
+  private static final long INITIAL_OFFSET = -1;
```
sfc-gh-tzhang, line 34: I believe we have a constant for this already

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
sfc-gh-rcheng: Channel names are meant to be[ unique identifiers](https://github.com/snowflakedb/snowflake-kafka-connector/blob/3879f8a3c2349b1ab3955c0bb22e601e08ba35da/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java#L507) for that particular topic and partition. 

From what I understand, client sequencers are bumped up when a channel is reopened, which means the registry should replace the previous channel name's metrics. This should be ok behavior, wdyt?

Another potential issue is two connectors listening to the same topic, however this is not an issue because the MetricRegistry is created off the unique connector name.

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
sfc-gh-rcheng: I think this is important to keep to ensure the IT test using the parameter verifies each new metric added. [SnowflakeTelemetryPipeStatusMetricsIT](https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/src/test/java/com/snowflake/kafka/connector/internal/SnowflakeTelemetryPipeStatusMetricsIT.java#L71) does something similar except it hard codes 14 instead of using a variable, which I think is more confusing

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-rcheng: you're right, the SnowflakeTelemetryPipeStatus creates a copy, however that is not strictly necessary, so i've passed the offsets in TopicPartitionChannel creation

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
sfc-gh-rcheng: removed

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelTest.java
```diff
@@ -97,14 +95,15 @@ public void setupEachTest() {
...
+    mockSnowflakeConnectionService = Mockito.mock(SnowflakeConnectionService.class);
     mockTelemetryService = Mockito.mock(SnowflakeTelemetryService.class);
     Mockito.when(mockStreamingClient.isClosed()).thenReturn(false);
     Mockito.when(mockStreamingClient.openChannel(ArgumentMatchers.any(OpenChannelRequest.class)))
         .thenReturn(mockStreamingChannel);
     Mockito.when(mockStreamingChannel.getFullyQualifiedName()).thenReturn(TEST_CHANNEL_NAME);
     this.topicPartition = new TopicPartition(TOPIC, PARTITION);
     this.sfConnectorConfig = TestUtils.getConfig();
-    this.streamingBufferThreshold = new StreamingBufferThreshold(10, 10_000, 1);
+    this.streamingBufferThreshold = new StreamingBufferThreshold(1, 10_000, 1);
```
sfc-gh-rcheng, line 106: reduced the time flush to 1 second to replicate our recommended settings

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
sfc-gh-rcheng: removed, register the guages with TopicPartitionChannel offsets directly instead

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
sfc-gh-tzhang: what about during rebalance? Any concern about emit metrics from two nodes?

sfc-gh-tzhang (approved):
LGTM, ship it!

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
```diff
@@ -0,0 +1,132 @@
...
+import static com.snowflake.kafka.connector.internal.metrics.MetricsUtil.constructMetricName;
+
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.MetricRegistry;
+import com.snowflake.kafka.connector.internal.KCLogger;
+import com.snowflake.kafka.connector.internal.metrics.MetricsJmxReporter;
+import com.snowflake.kafka.connector.internal.metrics.MetricsUtil;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class SnowflakeTelemetryChannelStatus {
```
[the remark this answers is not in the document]
sfc-gh-rcheng: will be added in the next PR

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
sfc-gh-rcheng: There shouldn't be rebalance issues because JmxReporters are created at the serviceSink level. When registering new metrics for a TopicPartitionChannel during rebalance, we remove all existing metrics and then register new ones.

PR #700, opened 2023-09-11
Fix tombstone ingestion with schematization

sfc-gh-rcheng:
- Enables schematization tombstone ingestion
- Adds comment about a [bug](https://snowflakecomputing.atlassian.net/browse/SNOW-916052) in ignoring snowpipe tombstone records

Test changes
- Adds IT for snowpipe, streaming and schematization ingestion methods
- Alters existing json e2e tests to ingest tombstone records by default
- Adds new e2e tests to ignore tombstone records
- Special cases Kafka version 2.5.1 due to a bug with tombstone records. Will need to make a note of this in the documentation


src/main/java/com/snowflake/kafka/connector/internal/streaming/SchematizationUtils.java
```diff
@@ -152,7 +152,7 @@ static Map<String, String> getColumnTypes(SinkRecord record, List<String> column
   private static Map<String, String> getSchemaMapFromRecord(SinkRecord record) {
     Map<String, String> schemaMap = new HashMap<>();
     Schema schema = record.valueSchema();
-    if (schema != null) {
+    if (schema != null && schema.fields() != null) {
```
sfc-gh-rcheng, line 155: added additional null check to ensure we don't NPE in the next for loop

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -275,6 +275,13 @@ public Map<String, Object> getProcessedRecordForStreamingIngest(SinkRecord recor
   private Map<String, Object> getMapFromJsonNodeForStreamingIngest(JsonNode node)
       throws JsonProcessingException {
     final Map<String, Object> streamingIngestRow = new HashMap<>();
+
+    // return empty if tombstone record
+    if (node.size() == 0
+        && this.behaviorOnNullValues == SnowflakeSinkConnectorConfig.BehaviorOnNullValues.DEFAULT) {
```
sfc-gh-rcheng, line 281: crux of the schematization tombstone change

test/rest_request_template/test_string_json_ignore_tombstone.json
```diff
@@ -0,0 +1,17 @@
...
+    "connector.class":"com.snowflake.kafka.connector.SnowflakeSinkConnector",
+    "topics":"SNOWFLAKE_TEST_TOPIC",
+    "snowflake.url.name":"SNOWFLAKE_HOST",
+    "snowflake.user.name":"SNOWFLAKE_USER",
+    "snowflake.private.key":"SNOWFLAKE_PRIVATE_KEY",
+    "snowflake.private.key.passphrase":"${file:./apache_properties/file-secrets.txt:PASSPHRASE}",
+    "snowflake.database.name":"SNOWFLAKE_DATABASE",
+    "snowflake.schema.name":"SNOWFLAKE_SCHEMA",
+    "key.converter":"org.apache.kafka.connect.storage.StringConverter",
+    "value.converter": "org.apache.kafka.connect.json.JsonConverter",
```
sfc-gh-rcheng, line 13: Using community JsonConverter instead of Snowflake customer converter due to this [bug](https://snowflakecomputing.atlassian.net/browse/SNOW-916052)

test/test_suit/test_snowpipe_streaming_string_json.py
```diff
@@ -31,10 +31,19 @@ def send(self):
             print("Sending in Partition:" + str(p))
             key = []
             value = []
-            for e in range(self.recordNum):
+            for e in range(self.recordNum - 1):
```
sfc-gh-japatel, line 34: why did we change this?

sfc-gh-japatel:
Can you add a description on what bug is fixed?

test/test_suit/test_snowpipe_streaming_string_json.py
sfc-gh-rcheng: we are sending the tombstone record as the last record. The schema evolution tests rely on exactly recordNum rows being sent (throws [NonRetryableError](https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/test/test_suit/test_schema_evolution_json.py#L84)). Additionally a majority of the tests rely on row count as a buffer threshold, so sending the tombstone record as an additional row would increase test runtime

sfc-gh-rcheng:
> Can you add a description on what bug is fixed?

Updated the description. The fixed bug is to enable tombstone ingestion with schematization

test/test_suit/test_snowpipe_streaming_string_json.py
sfc-gh-japatel: I see, thanks. 
why do we have to make changes in normal snowpipe streaming tests? we can do this change in tombstone specific tests right?

sfc-gh-rcheng: yes i could add a tombstone specific test, but i didn't want to extend the e2e runtime. plus tombstone ingestion is enabled by default, so i think it makes sense to put it in the normal e2e tests

happy to copy this into a new test if you prefer

sfc-gh-japatel: Oh I see, thanks for explanation, i missed it is default enabled. please add a comment wherever possible. Thanks

sfc-gh-japatel (approved):
lgtm! 
Left a suggestion for adding comment about tombstone since it wont be obvious by reading test code why we expect one less record. But thanks for adding all those tests. Good stuff. 
Lets also get an approval from @sfc-gh-tzhang related to schematization

test/test_suit/test_snowpipe_streaming_string_json.py
sfc-gh-rcheng: Good point it is confusing, added comments to the tests

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -275,6 +275,13 @@ public Map<String, Object> getProcessedRecordForStreamingIngest(SinkRecord recor
   private Map<String, Object> getMapFromJsonNodeForStreamingIngest(JsonNode node)
       throws JsonProcessingException {
     final Map<String, Object> streamingIngestRow = new HashMap<>();
+
+    // return empty if tombstone record
```
sfc-gh-tzhang, line 279: You probably has tested this, but all columns will be NULL for that row?

sfc-gh-tzhang (approved):
LGTM, thanks!

sfc-gh-tzhang:
> * Adds comment about a [bug](https://snowflakecomputing.atlassian.net/browse/SNOW-916052) in ignoring snowpipe tombstone records

I think we want to fix this as well

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-rcheng: yup tombstone records only have a RECORD_METADATA column and the rest are all null. The key is in the RECORD_METADATA
image

sfc-gh-rcheng:
> > * Adds comment about a [bug](https://snowflakecomputing.atlassian.net/browse/SNOW-916052) in ignoring snowpipe tombstone records
> 
> I think we want to fix this as well

I'll do this in a followup PR, I want to reduce existing tombstone behavior changes

Issue #702, opened 2023-09-13
Prefix the channel name


BDeus:
Hi,
in rare scenario when we want to use two differents connectors from the same topic (with different snowflake table), 
the channel name disable this possibility because the channel name is set from the [source topic and partition](https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java#L238).
Is it possible to have a prefix option in connector or use the (connector name hash + topic/partition) as channel name to avoid this issue ?

Furthermore, i read that is it possible [to reset offset from a channel](https://github.com/snowflakedb/snowflake-kafka-connector/issues/690#issuecomment-1691131315), but i cannot find how (changing connector name is not enough due to previous topic/partition channel name restriction).

Thanks for advice

sfc-gh-japatel:
This is a legit feature request and we will work on it. Ideally, connector name and topic partition name should be enough..

sfc-gh-xhuang:
> Furthermore, i read that is it possible https://github.com/snowflakedb/snowflake-kafka-connector/issues/690#issuecomment-1691131315, but i cannot find how (changing connector name is not enough due to previous topic/partition channel name restriction).

Here is the documentation
https://docs.snowflake.com/en/sql-reference/functions/system_snowpipe_streaming_update_channel_offset_token

PR #703, opened 2023-09-13
[NO-SNOW] Prefix with channel name with connector name

sfc-gh-japatel:
#702 
- We need this when they want to use two connectors with same topic name to ingest into two different tables. 
- Current logic will ping pong the open-close channel API between two connectors. 
- Nothing changes in pipe name in Snowpipe Kafka

Fixed tests.


sfc-gh-rcheng (approved):
lgtm!

PR #705, opened 2023-09-20
[SNOW-916052] Fix null value bug in Snowpipe Tombstone ingestion

sfc-gh-rcheng:
- Fixes bug in shouldSkipNullValue() by simplifying the logic
- Unifies the e2e tests testStringJsonIgnoreTombstone and testStringJson by using the same converter
- Adds additional null records to the tombstone IT

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/records/RecordService.java
+++ b/src/main/java/com/snowflake/kafka/connector/records/RecordService.java
@@ -567,15 +567,16 @@ public class RecordService {
     boolean isRecordValueNull =
         // skip when record value or schema is null
         (record.value() == null || record.valueSchema() == null)
-        ||
-        // skip when snowflake converter and record is null
-        (record.valueSchema() instanceof SnowflakeJsonSchema
-        && record.value() instanceof  SnowflakeRecordContent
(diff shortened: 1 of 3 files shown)


src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -559,51 +559,31 @@ else if (value instanceof ByteBuffer) {
   public boolean shouldSkipNullValue(
       SinkRecord record,
       final SnowflakeSinkConnectorConfig.BehaviorOnNullValues behaviorOnNullValues) {
+    // do not skip when config is DEFAULT
     if (behaviorOnNullValues == SnowflakeSinkConnectorConfig.BehaviorOnNullValues.DEFAULT) {
       return false;
-    } else {
```
sfc-gh-rcheng, line 564: the current logic in this method is as follows:
- do not skip record if the config allows null records
- skip record if the record value schema is a Snowflake converter schema AND the record value has isRecordContentValueNull 
- otherwise skip record if record value is null

The new logic in the method is similar, however we do the record value null check first

PR #709, opened 2023-09-25
[SNOW-916052] Use python None instead of empty string for tombstone e2e tests

sfc-gh-rcheng:
Snowflake converters and community converters convert empty string records different. Snowflake converters treat "" as a valid record, community converters treat "" as a tombstone record. Use python None (null) instead of empty string to accurately test tombstone ingestion

Discussed offline with @sfc-gh-tzhang to use None instead

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/records/RecordService.java
+++ b/src/main/java/com/snowflake/kafka/connector/records/RecordService.java
@@ -566,7 +566,6 @@ public class RecordService {
       // get valueSchema
       Schema valueSchema = record.valueSchema();
       if (valueSchema instanceof SnowflakeJsonSchema) {
-        // TODO SNOW-916052: will not skip if record.value() == null
         // we can conclude this is a custom/KC defined converter.
         // i.e one of SFJson, SFAvro and SFAvroWithSchemaRegistry Converter
         if (record.value() instanceof SnowflakeRecordContent) {
--- a/test/rest_request_template/test_string_json_ignore_tombstone.json
+++ b/test/rest_request_template/test_string_json_ignore_tombstone.json
@@ -10,8 +10,7 @@
     "snowflake.database.name":"SNOWFLAKE_DATABASE",
     "snowflake.schema.name":"SNOWFLAKE_SCHEMA",
     "key.converter":"org.apache.kafka.connect.storage.StringConverter",
-    "value.converter": "org.apache.kafka.connect.json.JsonConverter",
-    "value.converter.schemas.enable": "false",
+    "value.converter":"com.snowflake.kafka.connector.records.SnowflakeJsonConverter",
     "behavior.on.null.values": "IGNORE"
   }
 }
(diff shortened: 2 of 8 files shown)


src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -566,7 +566,6 @@ public boolean shouldSkipNullValue(
       // get valueSchema
       Schema valueSchema = record.valueSchema();
       if (valueSchema instanceof SnowflakeJsonSchema) {
-        // TODO SNOW-916052: will not skip if record.value() == null
```
sfc-gh-rcheng, line 569: valueschema will never be null because snowflake json converter always [returns a value schema
](https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/src/main/java/com/snowflake/kafka/connector/records/SnowflakeJsonConverter.java#L31)

test/rest_request_template/test_string_json_ignore_tombstone.json
```diff
@@ -10,8 +10,7 @@
     "snowflake.database.name":"SNOWFLAKE_DATABASE",
     "snowflake.schema.name":"SNOWFLAKE_SCHEMA",
     "key.converter":"org.apache.kafka.connect.storage.StringConverter",
-    "value.converter": "org.apache.kafka.connect.json.JsonConverter",
```
sfc-gh-rcheng, line 13: the corresponding default tombstone e2e ingestion test uses SnowflakeJsonConverter

sfc-gh-japatel: what will be the behavior if it is org.apache.kafka.connect.json.JsonConverter? Do we have a test for that?

sfc-gh-rcheng: JsonConverter treats empty string as a tombstone record whereas SnowflakeJsonConverter treats it as a normal record.

We don't have snowpipe e2e tests with JsonConverter, but we do have IT tests that [test all converters](https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/src/test/java/com/snowflake/kafka/connector/internal/TombstoneRecordIngestionIT.java#L140) with snowpipe

test/test_suit/test_string_json.py
```diff
@@ -24,7 +24,7 @@ def send(self):
                 {'numbernumbernumbernumbernumbernumbernumbernumbernumbernumbernumbernumber': str(100)}
             ).encode('utf-8'))
         else:
-            value.append('')
```
sfc-gh-tzhang, line 27: I understand we need use None for the tombstone tests, but should we have a test to test the empty string behavior instead of updating everything to None?

sfc-gh-tzhang (approved):
Left one comment, otherwise LGTM!

test/test_suit/test_string_json.py
sfc-gh-rcheng: added empty string test and comment to explain the differences between community vs custom converter record behavior

test/rest_request_template/test_string_json_ignore_tombstone.json
sfc-gh-japatel: Please double confirm that we dont have any behavior changes in snowpipe. Thanks

sfc-gh-japatel (approved):
thanks for fixing and improving test! Left a comment about being cautious in snowpipe code path.

PR #722, opened 2023-10-11
SNOW-938038: Support AVRO Logical Types

sfc-gh-tzhang:
There're 10 AVRO logical types listed in https://avro.apache.org/docs/1.11.0/spec.html#Logical+Types and we're able to support 4 of them, the mapping is as below:

- date -> DATE
- time-mills -> TIME(6)
- timestamp-mills -> TIMESTAMP_NTZ(6)
- decimal -> VARCHAR (We can't do NUMBER because we could have precision bigger than 36)

We can't do for the rest of 6 types because that's not supported by ConnectSchema, see [code](https://github.com/a0x8o/kafka/blob/54eff6af115ee647f60129f2ce6a044cb17215d0/connect/api/src/main/java/org/apache/kafka/connect/data/ConnectSchema.java#L68) for more detail. We need to find another way to support other logical types or any sources (like Debezium).

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SchematizationUtils.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SchematizationUtils.java
@@ -158,8 +158,13 @@ public class SchematizationUtils {
     Schema schema = record.valueSchema();
     if (schema != null && schema.fields() != null) {
       for (Field field : schema.fields()) {
-        schemaMap.put(
-            field.name(), convertToSnowflakeType(field.schema().type(), field.schema().name()));
+        String snowflakeType = convertToSnowflakeType(field.schema().type(), field.schema().name());
+        LOGGER.info(
+            "Got the snowflake data type for field:{}, schema_name:{}, type:{}",
+            field.name(),
+            field.schema().name(),
+            snowflakeType);
+        schemaMap.put(field.name(), snowflakeType);
       }
     }
     return schemaMap;
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SchematizationUtils.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SchematizationUtils.java
@@ -160,9 +160,11 @@ public class SchematizationUtils {
       for (Field field : schema.fields()) {
         String snowflakeType = convertToSnowflakeType(field.schema().type(), field.schema().name());
         LOGGER.info(
-            "Got the snowflake data type for field:{}, schema_name:{}, type:{}",
+            "Got the snowflake data type for field:{}, schema_name:{}, kafkaType:{},"
+                + " snowflakeType:{}",
             field.name(),
             field.schema().name(),
+            field.schema().type(),
             snowflakeType);
         schemaMap.put(field.name(), snowflakeType);
       }
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SchematizationUtils.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SchematizationUtils.java
@@ -4,11 +4,29 @@
 
 package com.snowflake.kafka.connector.internal.streaming;
 
+import static org.apache.kafka.connect.data.Schema.Type.ARRAY;
+import static org.apache.kafka.connect.data.Schema.Type.BOOLEAN;
+import static org.apache.kafka.connect.data.Schema.Type.BYTES;
+import static org.apache.kafka.connect.data.Schema.Type.FLOAT32;
+import static org.apache.kafka.connect.data.Schema.Type.FLOAT64;
+import static org.apache.kafka.connect.data.Schema.Type.INT16;
+import static org.apache.kafka.connect.data.Schema.Type.INT32;
+import static org.apache.kafka.connect.data.Schema.Type.INT64;
+import static org.apache.kafka.connect.data.Schema.Type.STRING;
+import static org.apache.kafka.connect.data.Schema.Type.STRUCT;
+
 import com.snowflake.kafka.connector.Utils;
 import com.snowflake.kafka.connector.internal.SnowflakeConnectionService;
 import com.snowflake.kafka.connector.internal.SnowflakeErrors;
 import com.snowflake.kafka.connector.internal.SnowflakeKafkaConnectorException;
 import com.snowflake.kafka.connector.records.RecordService;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import javax.annotation.Nonnull;
 import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.JsonNode;
 import org.apache.kafka.connect.data.Date;
 import org.apache.kafka.connect.data.Decimal;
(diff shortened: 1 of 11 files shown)


src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -87,7 +87,7 @@ public class RecordService {
           });
 
   public static final ThreadLocal<SimpleDateFormat> TIME_FORMAT =
-      ThreadLocal.withInitial(() -> new SimpleDateFormat("HH:mm:ss.SSSZ"));
+      ThreadLocal.withInitial(() -> new SimpleDateFormat("HH:mm:ss.SSSXXX"));
```
sfc-gh-tzhang, line 90: This needs to be updated for Snowpipe Streaming but I don't see a good way to distinguish Snowpipe VS Snowpipe Streaming in RecordService, happy to hear any suggestions

sfc-gh-japatel: simplest way to do it would be in TopicPartitionChannel ctor. since TopicPartitionChannel is only used in Snowpipe Streaming, just set a private field in RecordService to true. this will not change anything in snowpipe. Something like

```
public void setEnableSchematization(final boolean enableSchematization) {
    this.enableSchematization = enableSchematization;
  }
```

sfc-gh-tzhang: I thought about that, but `convertToJson` is a static function :)

sfc-gh-tzhang: Basically we need to add a boolean to `convertToJson` and update all the references, I'm debating on whether this is really needed. The difference between Z and XXX is that there will be a colon with XXX for the timezone, for example, Z = `02:24:00.000+0000` and XXX = `02:24:00.000+00:00`. In fact, it will fail if you try to insert `02:24:00.000+0000` to a TIME column so this is technically a bug. I'm not sure why there is no complains from customer though.

sfc-gh-tzhang:
ping, can I get a review on this? Thanks!

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-rcheng: out of scope of this PR, but RecordService does have a lot of overlap between snowpipe and streaming, might be useful to refactor or split it into two files for simplicity

sfc-gh-rcheng (approved):
lgtm after e2e tests pass. Do we need doc changes as well?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SchematizationUtils.java
```diff
@@ -154,7 +158,15 @@ private static Map<String, String> getSchemaMapFromRecord(SinkRecord record) {
     Schema schema = record.valueSchema();
     if (schema != null && schema.fields() != null) {
       for (Field field : schema.fields()) {
-        schemaMap.put(field.name(), convertToSnowflakeType(field.schema().type()));
+        String snowflakeType = convertToSnowflakeType(field.schema().type(), field.schema().name());
+        LOGGER.info(
+            "Got the snowflake data type for field:{}, schema_name:{}, kafkaType:{},"
```
sfc-gh-japatel, line 163: nit: to match with other fields `schemaName`

src/main/java/com/snowflake/kafka/connector/internal/streaming/SchematizationUtils.java
```diff
@@ -167,7 +179,7 @@ private static String inferDataTypeFromJsonObject(JsonNode value) {
       // only when the type of the value is unrecognizable for JAVA
       throw SnowflakeErrors.ERROR_5021.getException("class: " + value.getClass());
     }
-    return convertToSnowflakeType(schemaType);
+    return convertToSnowflakeType(schemaType, null);
```
sfc-gh-japatel, line 182: nit: lets add a single line description here about what null is

src/main/java/com/snowflake/kafka/connector/internal/streaming/SchematizationUtils.java
```diff
@@ -201,16 +213,26 @@ private static Type convertJsonNodeTypeToKafkaType(JsonNode value) {
...
       case INT16:
         return "SMALLINT";
       case INT32:
-        return "INT";
+        if (Date.LOGICAL_NAME.equals(schemaName)) {
+          return "DATE";
+        } else if (Time.LOGICAL_NAME.equals(schemaName)) {
+          return "TIME(6)";
+        } else {
+          return "INT";
```
sfc-gh-japatel, line 228: thinking out loud here: feel free to ignore. 
will this be a behavior change? before this, we converted all INT32 to INT, now we are introducing new snowflake data types- which essentially mean new columns?

I also understand this is a good distinction for them but will this haunt us?

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-japatel: > Basically we need to add a boolean to convertToJson and update all the references, I'm debating on whether this is really needed. The difference between Z and XXX is that there will be a colon with XXX in the timezone, for example, Z = 02:24:00.000+0000 and XXX = 02:24:00.000+00:00. In fact, it will fail if you try to insert 02:24:00.000+0000 to a TIME column so this is technically a bug. I'm not sure why there is no complains from customer, probably because not many people is using logical types.

Thanks for detailed explanation. 
I would be careful to make changes in snowpipe code. 
Most folks havent complained because we dont have schematization in snowpipe. they have base table and then more tables on top of that. This change will break their downstream pipeline right?

sfc-gh-japatel: You might have to take a giant step and add another arg in `convertToJson` :( 
@sfc-gh-rcheng 's suggestion is the only way forward.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SchematizationUtils.java
```diff
@@ -201,16 +213,26 @@ private static Type convertJsonNodeTypeToKafkaType(JsonNode value) {
...
 
   /** Convert the kafka data type to Snowflake data type */
-  private static String convertToSnowflakeType(Type kafkaType) {
+  private static String convertToSnowflakeType(Type kafkaType, String schemaName) {
     switch (kafkaType) {
       case INT8:
         return "BYTEINT";
       case INT16:
         return "SMALLINT";
       case INT32:
```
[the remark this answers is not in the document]
sfc-gh-tzhang: I did add it to the PR description, do you want to take a look to see what's missing? We will definitely update our online doc to reflect this change.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SchematizationUtils.java
sfc-gh-tzhang: Not really, not sure if you remember, but before the insert will fail because it's inserting VARCHAR into a INT column

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-tzhang: Added as an argument, PTAL

src/main/java/com/snowflake/kafka/connector/internal/streaming/SchematizationUtils.java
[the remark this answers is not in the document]
sfc-gh-rcheng: pr description is great, but adding it could help with readability, but i dont feel too strongly about it, up to you

src/main/java/com/snowflake/kafka/connector/internal/streaming/SchematizationUtils.java
```diff
@@ -220,7 +243,11 @@ private static String convertToSnowflakeType(Type kafkaType) {
       case STRING:
         return "VARCHAR";
       case BYTES:
-        return "BINARY";
+        if (Decimal.LOGICAL_NAME.equals(schemaName)) {
+          return "VARCHAR";
+        } else {
+          return "BINARY";
+        }
```
sfc-gh-japatel, line 250: it would be good to mention in description that it is not decimal to varchar but the bytes that represents decimal. Is that correct understanding?

sfc-gh-japatel (approved):
lgtm, thanks!

Issue #724, opened 2023-10-16
Struct array elements are being serialized before writing to Array column


cchandurkar:
I have a field in avro schema that is an Array of Struct type. Table has an existing column that's of Array type. I'm seeing array elements being serialized (Seeing array of stringified JSON in the SF column). Since the column is Array, it can take an array of object (struct/variant). Is there any way to disable this serialization and treat the elements as a variant instead?

Here's a quick comparison:
Expected value in column:
```json
[
  null,
  {
    "xCategory": null,
    "xEntityId": null,
    "xEntityName": null,
    "xId": "89asda9s0a"
  }
]
```

Actual:
```json
[
  "null",
  "{\"xCategory\":null,\"xEntityId\":null,\"xId\":\"89asda9s0a\",\"xEntityName\":null}"
]
```

cchandurkar:
Was digging into the codebase. It looks like this [RecordService#getMapFromJsonNodeForStreamingIngest()](https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/src/main/java/com/snowflake/kafka/connector/records/RecordService.java#L302) does not handle Struct and just stringifies it. Is this intentional? What if you recursively convert Struct into an ObjectNode?

sfc-gh-azagrebin:
This is indeed not clear why to stringify array elements. SDK would indeed treat it as a string.
@sfc-gh-japatel @sfc-gh-tzhang wdyt?

@sfc-gh-zefan is it expected this way for Snowpipe?

cchandurkar:
Closing at this is fixed in https://github.com/snowflakedb/snowflake-kafka-connector/pull/730.

PR #727, opened 2023-10-17
Fix array of objects serialization

cchandurkar:
Currently, array of structs/objects are being written as an array of stringified json objects.

Here's a quick comparison:
Expected value in ARRAY column:
```json
[
  null,
  {
    "xCategory": null,
    "xEntityId": null,
    "xEntityName": null,
    "xId": "89asda9s0a"
  }
]
```

Actual value:
```json
[
  "null",
  "{\"xCategory\":null,\"xEntityId\":null,\"xId\":\"89asda9s0a\",\"xEntityName\":null}"
]
```

 I think it's happening because those objects/structs are getting serialized twice. A simple fix here would be to serialize an entire ArrayNode as is instead of creating a `List<String>` out of it. This way, the value (Stringified array) is a valid JSON that is accepted by `net.snowflake.ingest.streaming.SnowflakeStreamingIngestChannel.insertRow()`

Issue: https://github.com/snowflakedb/snowflake-kafka-connector/issues/724

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/records/RecordService.java
+++ b/src/main/java/com/snowflake/kafka/connector/records/RecordService.java
@@ -287,14 +287,7 @@ public class RecordService {
       String columnName = columnNames.next();
       JsonNode columnNode = node.get(columnName);
       Object columnValue;
-      if (columnNode.isArray()) {
-        List<String> itemList = new ArrayList<>();
-        ArrayNode arrayNode = (ArrayNode) columnNode;
-        for (JsonNode e : arrayNode) {
-          itemList.add(e.isTextual() ? e.textValue() : MAPPER.writeValueAsString(e));
-        }
-        columnValue = itemList;
-      } else if (columnNode.isTextual()) {
+      if (columnNode.isTextual()) {
         columnValue = columnNode.textValue();
       } else if (columnNode.isNull()) {
         columnValue = null;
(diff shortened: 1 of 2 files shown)


src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -287,14 +287,7 @@ private Map<String, Object> getMapFromJsonNodeForStreamingIngest(JsonNode node)
       String columnName = columnNames.next();
       JsonNode columnNode = node.get(columnName);
       Object columnValue;
-      if (columnNode.isArray()) {
```
sfc-gh-tzhang, line 290: This is an interesting observation, I assume this was added for a reason, we will evaluate it to see if it will break anything, thanks!

cchandurkar: Looks like it was introduced in [PR](https://github.com/snowflakedb/snowflake-kafka-connector/pull/467/files#diff-7920dca33f32a43f73302b546c7877c26ceeeed3fd555a7dfb110d77b42658f7R251-R259) that enabled schematization and hasn't been updating since.

cchandurkar: If you serialize an entire array node, it's JSON elements won't be stringified. They will appear as "expected" in description. And that won't require to use `parse_json()`

sfc-gh-xhuang:
Closing as https://github.com/snowflakedb/snowflake-kafka-connector/pull/730 has been merged

PR #729, opened 2023-10-18
[SNOW-943288] Do not skip records when we're expecting the offset to be reset

sfc-gh-rcheng:
https://snowflakecomputing.atlassian.net/browse/SNOW-943288

Do not skip files when there is no kafka offset reset, remove the processedOffset check

## Context
Customer noticed first 500 rows are missing when using schema evolution. This is because we do not reset the kafka offset correctly when we fail in the middle of a KC batch.

Current Behavior, with buffer flush size 500:
1. KC sends 36 records (offset 0-35), inserted into buffer
2. KC sends 500 records (offset 36-535)
    a. 464 (offset 36-500) records inserted into buffer and triggers flush, schema alter column, reopen channel and reset kafka offset
    b. Remaining 36 records (offset 500-535) inserted into buffer _**<- These records should be skipped**_ 
3. KC resends initial 536 records, however the processedOffset has already been set to 535 due to the last 36 records (offset 500-535)

So this PR alters the logic to skip the remaining 36 records (offset 500-535)


src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -124,7 +126,7 @@ public class TopicPartitionChannel {
    * <p>This boolean is used to indicate that we reset offset in kafka and we will only buffer once
    * we see the offset which is one more than an offset present in Snowflake.
    */
-  private boolean isOffsetResetInKafka;
+  private boolean isOffsetResetInKafka = false; // TODO @rcheng question: atomic?
```
sfc-gh-rcheng, line 129: The offsets (offsetPersistedInSnowflake, processedOffset, latestConsumerOffset) are AtomicLongs implying that there is some parallelism considerations. Should this isOffsetResetInKafka be an AtomicBoolean as well? I think it may be overkill, but all other variables are Atomic, final or uses a lock (StreamingBuffer)

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -323,28 +325,26 @@ public TopicPartitionChannel(
    * @param kafkaSinkRecord input record from Kafka
    */
   public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
-    final long currentOffsetPersistedInSnowflake = this.offsetPersistedInSnowflake.get();
-    final long currentProcessedOffset = this.processedOffset.get();
```
sfc-gh-rcheng, line 327: Crux of the issue: shouldIgnoreAddingRecordToBuffer modifies processedOffset, but this currentProcessedOffset ignores the change

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -391,14 +391,19 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
...
    * @param kafkaSinkRecord Record to check for above condition only in case of failures
    *     (isOffsetResetInKafka = true)
+   * @param currentProcessedOffset The current processed offset
    * @return true if this record can be skipped to add into buffer, false otherwise.
    */
   private boolean shouldIgnoreAddingRecordToBuffer(
-      SinkRecord kafkaSinkRecord, long currentProcessedOffset) {
-    // Don't skip rows if there is no offset reset or there is no offset token information in the
+      SinkRecord kafkaSinkRecord, final long currentProcessedOffset) {
+    // Don't skip rows if there is no offset reset and there is no offset token information in the
```
sfc-gh-rcheng, line 399: Also a major change, discussed in the description

test/test_suit/test_schema_evolution_w_auto_table_creation_avro_sr.py
```diff
@@ -78,11 +83,22 @@ def getConfigFileName(self):
...
-            for _ in range(self.recordNum):
+            for _ in range(self.initialRecordCount):
                 value.append(self.records[i])
             self.driver.sendAvroSRData(topic, value, self.valueSchema[i], key=[], key_schema="", partition=0)
 
+            sleep(2)
+
+            # send second batch that should flush
+            value = []
+            for _ in range(self.flushRecordCount):
```
sfc-gh-rcheng, line 96: confirmed that these tests repro the issue without the changes

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-tzhang: I don't think we run anything in parallel, do we?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -391,14 +392,19 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
...
     // channel
     if (!isOffsetResetInKafka
-        || currentProcessedOffset == NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE) {
+        && currentProcessedOffset == NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE) {
```
sfc-gh-tzhang, line 403: Should we just remove the check on currentProcessedOffset? I don't understand what this will give us now

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -391,14 +392,19 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
...
+      LOGGER.debug(
+          "No offset registered in Snowflake and offset is not being reset, we can add this offset"
+              + " to buffer for channel:{}",
+          currentProcessedOffset);
```
sfc-gh-tzhang, line 407: This will add a log for almost every row, any concern?

test/test_suit/test_schema_evolution_w_auto_table_creation_json.py
```diff
@@ -13,7 +14,11 @@ def __init__(self, driver, nameSalt):
         self.fileName = "travis_correct_schema_evolution_w_auto_table_creation_json"
         self.topics = []
         self.table = self.fileName + nameSalt
-        self.recordNum = 100
+
+        # records
+        self.initialRecordCount = 12
```
sfc-gh-tzhang, line 19: Could we add a new test will random row count?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-rcheng: We shouldn't and I can't think of a scenario where it would apply so I removed it. Just was wondering because we have other parallelization considerations

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-rcheng: Yeah lets do that, removed

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-rcheng: Thats true, added this as debugging, but will remove now

sfc-gh-japatel:
I will take a look but can we postpone merging this fix until we have a new release?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-asen: What does this change? I don't see any assignments in this method and since it is primitive long is pass by value

sfc-gh-rcheng: not the crux of the change, just something i thought affected the issue. i thought i'd resolved this comment already

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -488,7 +488,8 @@ public boolean hasSchemaEvolutionPermission(String tableName, String role) {
   public void appendColumnsToTable(String tableName, Map<String, String> columnToType) {
     checkConnection();
     InternalUtils.assertNotEmpty("tableName", tableName);
-    StringBuilder appendColumnQuery = new StringBuilder("alter table identifier(?) add column if not exists ");
+    StringBuilder appendColumnQuery =
```
sfc-gh-rcheng, line 491: required for formatter to pass for some reason. not relevant to this PR, just a small format change

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -364,7 +366,6 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
       } finally {
         bufferLock.unlock();
       }
-
```
sfc-gh-tzhang, line 367: is this intended?

sfc-gh-tzhang (approved):
Left some comments, otherwise LGTM!

sfc-gh-tzhang:
This is a behavior change, should we follow the new BCR release process for drivers/SDKs? cc: @sfc-gh-xhuang

test/test_suit/test_schema_evolution_w_random_row_count.py
```diff
@@ -84,25 +74,23 @@ def verify(self, round):
...
 
-        for columnName in gold_columns_copy:
+        for columnName in self.gold_columns:
             raise NonRetryableError("Column {} was not created".format(columnName))
 
         res = self.driver.snowflake_conn.cursor().execute(
             "SELECT count(*) FROM {}".format(self.table)).fetchone()[0]
-        if res != self.recordNum * len(self.topics):
-            print("Number of record expected: {}, got: {}".format(self.recordNum * len(self.topics), res))
+        if res != len(self.topics) * self.recordNum:
```
sfc-gh-japatel, line 92: great test, thanks!

sfc-gh-japatel (approved):
lgtm! 
Thanks for the work.. Talked offline and I think this is a good short term solution. Please see if we can re-add those tests as Toby mentioned. For long term:

1. We do need to remove buffer in KC. 
2. if we cant remove bufffer from KC, we should have an interface between KC buffer and the logic which returns the insertRowsResponse. Here we can discard the batch only after the bad row and not necessarily entire batch. Of course this needs to be thought through since we are talking about discarding the incoming batch from Kafka which spans across multiple partitions. (but you can route those records to individual partitions, and discard partition batches) but I do feel this is a good short term solution not relying on removal of buffer..

sfc-gh-xhuang:
Let's discuss more on whether this is a behavior change or a bug fix but following the BCR process for drivers to simple enough too.

test/test_suites.py
```diff
@@ -235,9 +235,5 @@ def create_end_to_end_test_suites(driver, nameSalt, schemaRegistryAddress, testS
         ("TestKcRestart", EndToEndTestSuite(
             test_instance=TestKcRestart(driver, nameSalt), clean=True, run_in_confluent=True, run_in_apache=True
         )),
-        ("TestSchemaEvolutionMultiTopicDropTable", EndToEndTestSuite(
```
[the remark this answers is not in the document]
sfc-gh-rcheng: added a test for the random row count, drop table e2e discussed in the other thread, resolving this one

test/test_suites.py
```diff
@@ -194,10 +198,6 @@ def create_end_to_end_test_suites(driver, nameSalt, schemaRegistryAddress, testS
             test_instance=TestSchemaNotSupportedConverter(driver, nameSalt), clean=True, run_in_confluent=True,
             run_in_apache=True
         )),
-        ("TestSchemaEvolutionDropTable", EndToEndTestSuite(
```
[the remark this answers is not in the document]
sfc-gh-rcheng: Good point, i'll add the tests back in but disabled. They fail now, but would be a good test for the long term solution

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-rcheng: nope, added new line back in, thanks for the catch

PR #732, opened 2023-10-22
PROD-39429 Add parameter for connector name in channel name:

sfc-gh-japatel:
PROD-39429

## Description
This Commit is adding a parameter to https://github.com/snowflakedb/snowflake-kafka-connector/commit/3bf9106b22510c62068f7d2f7137b9e57989274c and is also reverting the behavior. 
i.e If this commit is pushed and is in release branch, the behavior of determining the `channelName` is reverted. (Contains only topic name and partition number)

Users will have to enable the parameter `snowflake.enable.new.channel.name.format` (set to true) to have connector name in their channel name. 

## Scenarios for users reading this commit:
1. For customers on 2.1.0 where [this](3bf9106b22510c62068f7d2f7137b9e57989274c) change was introduced, we are asking you all to be cautious and look for duplicates. Chances of duplicates are very slim but still possible. We would encourage you all to upgrade to next patch version which has this revert released and set parameter to true to continue with your 2.1.0's behavior. 
2. For users who are below 2.1.0, you dont have to do anything except not using 2.1.0 :)

## Why
TLDR:
Commit 3bf9106b22510c62068f7d2f7137b9e57989274c which was release in v2.1.0 is not compatible with old versions upgrading to 2.1.0 because it changes the channel name and we lose offset information in new channels.  
Longer Description:
```
Assumptions: TopicName = HALLOWEEN, partitions 2, connector_name = connector_name1
1. Channel name HALLOWEEN_1, HALLOWEEN_2 -> offset token 10, 12 respectively in snowflake but last committed offset in kafka is 8.10. 
3. We turned off connector. Upgraded jar and used 2.1.0 kc version. (One with the commit)
4. This new version will create  connector_name1_HALLOWEEN_1, connector_name1_HALLOWEEN_2. 
5. These two channels will have null offset token and since last committed in kafka is 8, 10, we accept that in kafka connector since channel offsets are null on snowflake. 
6. This has potential to have duplicates.

```

## Plan
- Patch release a new version with this commit. 
- Existing 2.1.0 users will have to enable param to true. (We will reach out to all those users)
- Default param is false. 
- Long term fix: Smooth transition between old version and new version to transfer offsets from old channelNames to new channelNames

## Tests
Added tests using ParameterizedTest feature.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -168,6 +168,18 @@ public class SnowflakeSinkConnectorConfig {
       "Whether to optimize the streaming client to reduce cost. Note that this may affect"
           + " throughput or latency and can only be set if Streaming Snowpipe is enabled";
 
+  public static final String ENABLE_CONNECTOR_NAME_IN_STREAMING_CHANNEL_NAME =
+      "enable.connector_name.in.streaming_channel_name";
+  public static final boolean ENABLE_CONNECTOR_NAME_IN_STREAMING_CHANNEL_NAME_DEFAULT = false;
+
+  public static final String ENABLE_CONNECTOR_NAME_IN_STREAMING_CHANNEL_NAME_DISPLAY =
+      "Enable Connector Name in Snowpipe Streaming Channel Name";
+
+  public static final String ENABLE_CONNECTOR_NAME_IN_STREAMING_CHANNEL_NAME_DOC =
+      "Whether to use connector name in streaming channels. If it is set to false, we will not use"
+          + " connector name in channel name. Note: Please use this config cautiously and it is not"
+          + " advised to use this if you are coming from old Snowflake Kafka Connector Version. ";
+
   // MDC logging header
   public static final String ENABLE_MDC_LOGGING_CONFIG = "enable.mdc.logging";
   public static final String ENABLE_MDC_LOGGING_DISPLAY = "Enable MDC logging";
@@ -591,7 +603,17 @@ public class SnowflakeSinkConnectorConfig {
             CONNECTOR_CONFIG,
             8,
             ConfigDef.Width.NONE,
-            ENABLE_MDC_LOGGING_DISPLAY);
+            ENABLE_MDC_LOGGING_DISPLAY)
+        .define(
+            ENABLE_CONNECTOR_NAME_IN_STREAMING_CHANNEL_NAME,
+            Type.BOOLEAN,
+            ENABLE_CONNECTOR_NAME_IN_STREAMING_CHANNEL_NAME_DEFAULT,
+            Importance.LOW,
+            ENABLE_CONNECTOR_NAME_IN_STREAMING_CHANNEL_NAME_DOC,
+            CONNECTOR_CONFIG,
+            9,
+            ConfigDef.Width.NONE,
+            ENABLE_CONNECTOR_NAME_IN_STREAMING_CHANNEL_NAME_DISPLAY);
   }
 
   public static class TopicToTableValidator implements ConfigDef.Validator {
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -168,17 +168,19 @@ public class SnowflakeSinkConnectorConfig {
       "Whether to optimize the streaming client to reduce cost. Note that this may affect"
           + " throughput or latency and can only be set if Streaming Snowpipe is enabled";
 
-  public static final String ENABLE_CONNECTOR_NAME_IN_STREAMING_CHANNEL_NAME =
-      "enable.connector_name.in.streaming_channel_name";
-  public static final boolean ENABLE_CONNECTOR_NAME_IN_STREAMING_CHANNEL_NAME_DEFAULT = false;
+  public static final String SNOWFLAKE_ENABLE_STREAMING_CHANNEL_FORMAT_V2 =
+      "snowflake.enable.streaming.channel.format.v2";
+  public static final boolean SNOWFLAKE_ENABLE_STREAMING_CHANNEL_FORMAT_V2_DEFAULT = false;
 
-  public static final String ENABLE_CONNECTOR_NAME_IN_STREAMING_CHANNEL_NAME_DISPLAY =
-      "Enable Connector Name in Snowpipe Streaming Channel Name";
+  public static final String SNOWFLAKE_ENABLE_STREAMING_CHANNEL_FORMAT_V2_DISPLAY =
+      "Enable Connector Name in Snowpipe Streaming Channel Name - V2 of Channel Name";
 
-  public static final String ENABLE_CONNECTOR_NAME_IN_STREAMING_CHANNEL_NAME_DOC =
+  public static final String SNOWFLAKE_ENABLE_STREAMING_CHANNEL_FORMAT_V2_DOC =
       "Whether to use connector name in streaming channels. If it is set to false, we will not use"
-          + " connector name in channel name. Note: Please use this config cautiously and it is not"
-          + " advised to use this if you are coming from old Snowflake Kafka Connector Version. ";
+          + " connector name in channel name(Which is version 2 of Channel Name). Note: Please use"
+          + " this config cautiously and it is not advised to use this if you are coming from old"
+          + " Snowflake Kafka Connector Version where Default Channel Name doesnt contain Connector"
+          + " Name, contains Topic Name and Partition # only.";
 
   // MDC logging header
   public static final String ENABLE_MDC_LOGGING_CONFIG = "enable.mdc.logging";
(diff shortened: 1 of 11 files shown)


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -168,6 +168,18 @@ public class SnowflakeSinkConnectorConfig {
       "Whether to optimize the streaming client to reduce cost. Note that this may affect"
           + " throughput or latency and can only be set if Streaming Snowpipe is enabled";
 
+  public static final String ENABLE_CONNECTOR_NAME_IN_STREAMING_CHANNEL_NAME =
+      "enable.connector_name.in.streaming_channel_name";
```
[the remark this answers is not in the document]
sfc-gh-tzhang: And I suggest we make it a more general config, something like `snowflake.enable.new.channel.format`

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -138,6 +143,11 @@ public SnowflakeSinkServiceV2(
             ? "default_connector"
             : this.conn.getConnectorName();
     this.metricsJmxReporter = new MetricsJmxReporter(new MetricRegistry(), connectorName);
+    this.shouldUseConnectorNameInChannelName =
```
sfc-gh-tzhang, line 146: please make sure we don't run into the same issue as the one client optimization :)

sfc-gh-tzhang (commented):
Left some comments, PTAL, and thanks for the quick change!

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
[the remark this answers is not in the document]
sfc-gh-japatel: Ack, like that suggestion, let me do that!

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: Ack.. yes kept that in mind but still double confirmed!

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -168,6 +168,20 @@ public class SnowflakeSinkConnectorConfig {
       "Whether to optimize the streaming client to reduce cost. Note that this may affect"
           + " throughput or latency and can only be set if Streaming Snowpipe is enabled";
 
+  public static final String SNOWFLAKE_ENABLE_STREAMING_CHANNEL_FORMAT_V2 =
+      "snowflake.enable.streaming.channel.format.v2";
+  public static final boolean SNOWFLAKE_ENABLE_STREAMING_CHANNEL_FORMAT_V2_DEFAULT = false;
```
sfc-gh-xhuang, line 173: are we doing default false or default true?

sfc-gh-xhuang: default false ensures 2.0.1 or below upgrade is safe.
default true ensures 2.1.0 upgrade is safe. 

default false makes sense to me but we need to make sure the scenario for 2.1.0 that upgrade are well tested.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -168,6 +168,20 @@ public class SnowflakeSinkConnectorConfig {
...
+
+  public static final String SNOWFLAKE_ENABLE_STREAMING_CHANNEL_FORMAT_V2_DISPLAY =
+      "Enable Connector Name in Snowpipe Streaming Channel Name - V2 of Channel Name";
+
+  public static final String SNOWFLAKE_ENABLE_STREAMING_CHANNEL_FORMAT_V2_DOC =
+      "Whether to use connector name in streaming channels. If it is set to false, we will not use"
```
sfc-gh-xhuang, line 179: Whether to include connector name in streaming channels names. If set to false, channel name will not be included connect name. Note: exercise caution around offset handling when setting this config to true as previous version of the Snowflake connector for Kafka only contained Topic Name and Partition #";

sfc-gh-xhuang: How is that note instead?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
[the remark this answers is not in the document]
sfc-gh-xhuang: I like it too. We should have done the same for one client optimization but too late now

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -527,11 +556,20 @@ public Optional<MetricRegistry> getMetricRegistry(String partitionChannelKey) {
    *     or PROD)
    * @param topic topic name
    * @param partition partition number
+   * @param shouldUseConnectorNameInChannelName If true, use connectorName, else not
```
sfc-gh-rcheng, line 559: nit: add jira number or pr number in case we plan to revert this change in the future

sfc-gh-rcheng (approved):
Overall lgtm, lets wait on toby and xin's approvals too.

Have you gotten a repro of the upgrade issue and confirmed that this change fixes it for them?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-tzhang: default should be false, and we will include the communication on how 2.1.0 customers will do the upgrade (they have the choice of making it true, or using false with their own risk (stop the ingestion and waiting for everything to be committed)

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java
```diff
@@ -1,14 +1,6 @@
...
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.BOOLEAN_VALIDATOR;
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.CUSTOM_SNOWFLAKE_CONVERTERS;
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.ERRORS_DEAD_LETTER_QUEUE_TOPIC_NAME_CONFIG;
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.ERRORS_LOG_ENABLE_CONFIG;
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.ERRORS_TOLERANCE_CONFIG;
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.ErrorTolerance;
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.INGESTION_METHOD_OPT;
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.KEY_CONVERTER_CONFIG_FIELD;
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.VALUE_CONVERTER_CONFIG_FIELD;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.*;
```
sfc-gh-tzhang, line 3: nit: No star import

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-japatel: Sounds good, let me modify it to include offset information.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java
sfc-gh-japatel: Ack!

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-japatel: We dont have plans to revert it for now!

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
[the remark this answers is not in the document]
sfc-gh-japatel: I have used `snowflake.enable.streaming.channel.format.v2`
Please let me know if you all like this instead! 
(Chances are we will not change it ever and hence I went with v2)

sfc-gh-japatel:
> Overall lgtm, lets wait on toby and xin's approvals too.
> 
> Have you gotten a repro of the upgrade issue and confirmed that this change fixes it for them?

Repro is not done yet since it is non-trivial and also this fix doesn't confirm it will fix issue for existing 2.1.0 customers. 
This is to reduce the blast radius and going into an unwanted territory. (I will modify description)

Repro'ing this along with blast radius analysis is going to be my next priority.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-tjones: Default should be false.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -527,11 +556,20 @@ public Optional<MetricRegistry> getMetricRegistry(String partitionChannelKey) {
...
+  public static String partitionChannelKey(
+      String connectorName,
+      String topic,
+      int partition,
+      final boolean shouldUseConnectorNameInChannelName) {
+    if (shouldUseConnectorNameInChannelName) {
+      return connectorName + "_" + topic + "_" + partition;
+    } else {
+      return topic + "_" + partition;
+    }
```
sfc-gh-tjones, line 572: Nit, this can be reduced to:
```
return shouldUseConnectorNameInChannelName ? 
    connectorName + "_" + topic + "_" + partition :
    topic + "_" + partition;
```

sfc-gh-tjones (approved):
LGTM barring the outstanding comments others have made.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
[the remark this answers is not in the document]
sfc-gh-japatel: Alright, this proves again that naming is the toughest task in SWE, so living upto that. 
We have agreed on `snowflake.enable.new.channel.name.format` 

Will make changes in description.

PR #737, opened 2023-10-26
PROD-39429 PART-2 Make REST API(/channels/status) call which acts as Channel Exi…

sfc-gh-japatel:
…stence Checker

## What
- A skeleton code, to find out whether a Streaming Channel exists for a table. 
- This is a backdoor way to get Channel Status (If client sequencers match) and to find out if a channel exists without opening a channel or without needing to have a StreamingIngestChannel Object
- It has a lot of repeated code from Ingest SDK. 
- We would want to move away from KC code to ingest SDK. More on that will come later SNOW-951262

## Why
- We could have always opened a new Channel, find offsetToken and assign it to oldChannel. 
- But this has a risk of opening a new channel everytime and we would want to go away from that. This could also have a lot of channels on table and we could reach a limit. 
- Doing `show channels iliike '<channelName>' in table <tableName>` is an expensive operation and is not optimal and also doesnt get away from opening a channel. 
- This will also help us to have our future code (PR no 3, more on it later) backward compatible. 

If interested, take a look at this [algorithm](https://docs.google.com/document/d/1kXY_wXKoEYXu-fDP4whQzhYCKLuxXzIf0iw2AwMOvwo/edit#heading=h.ckf809ta3j2k):  (Internal employees only)

## Tests
Added tests.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
@@ -1,3 +1,6 @@
+/*
+ * Copyright (c) 2023 Snowflake Computing Inc. All rights reserved.
+ */
 package com.snowflake.kafka.connector.internal.streaming;
 
 import static net.snowflake.ingest.streaming.internal.StreamingIngestResponseCode.ERR_CHANNEL_DOES_NOT_EXIST_OR_IS_NOT_AUTHORIZED;
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
@@ -1,0 +1,290 @@
+package com.snowflake.kafka.connector.internal.streaming;
+
+import static net.snowflake.ingest.streaming.internal.StreamingIngestResponseCode.ERR_CHANNEL_DOES_NOT_EXIST_OR_IS_NOT_AUTHORIZED;
+import static net.snowflake.ingest.streaming.internal.StreamingIngestResponseCode.ERR_CHANNEL_HAS_INVALID_CLIENT_SEQUENCER;
+import static net.snowflake.ingest.streaming.internal.StreamingIngestResponseCode.SUCCESS;
+
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.annotations.VisibleForTesting;
+import com.snowflake.kafka.connector.Utils;
+import com.snowflake.kafka.connector.internal.KCLogger;
+import java.io.IOException;
+import java.security.NoSuchAlgorithmException;
+import java.security.PrivateKey;
+import java.security.spec.InvalidKeySpecException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+import java.util.UUID;
+import net.snowflake.client.core.SFSessionProperty;
+import net.snowflake.client.jdbc.internal.apache.http.HttpEntity;
+import net.snowflake.client.jdbc.internal.apache.http.HttpResponse;
+import net.snowflake.client.jdbc.internal.apache.http.HttpStatus;
+import net.snowflake.client.jdbc.internal.apache.http.StatusLine;
+import net.snowflake.client.jdbc.internal.apache.http.client.methods.CloseableHttpResponse;
+import net.snowflake.client.jdbc.internal.apache.http.client.methods.HttpUriRequest;
+import net.snowflake.client.jdbc.internal.apache.http.impl.client.CloseableHttpClient;
+import net.snowflake.client.jdbc.internal.apache.http.util.EntityUtils;
+import net.snowflake.ingest.connection.OAuthCredential;
+import net.snowflake.ingest.connection.RequestBuilder;
+import net.snowflake.ingest.connection.ServiceResponseHandler;
+import net.snowflake.ingest.streaming.SnowflakeStreamingIngestChannel;
+import net.snowflake.ingest.utils.BackOffException;
+import net.snowflake.ingest.utils.Constants;
+import net.snowflake.ingest.utils.ErrorCode;
+import net.snowflake.ingest.utils.HttpUtil;
+import net.snowflake.ingest.utils.SFException;
+import net.snowflake.ingest.utils.SnowflakeURL;
+
+/**
+ * Class which handles REST API call directly for a Streaming channel and obtains a response of its
+ * existence.
+ *
+ * <p>This is a bit different from what we are doing in {@link
+ * SnowflakeStreamingIngestChannel#getLatestCommittedOffsetToken()} wherein we first have to Open A
+ * Channel and then make another call for getOffsetToken.
+ *
+ * <p>A method (checkChannelExistence) in this class bypasses that open Channel call and directly
+ * creates the payload for Get Status {@link
+ * StreamingChannelExistenceChecker#CHANNEL_STATUS_ENDPOINT} API.
+ */
+public class StreamingChannelExistenceChecker {
+  private static final KCLogger LOGGER =
+      new KCLogger(StreamingChannelExistenceChecker.class.getName());
+
+  private static final String DEBUG_PREFIX = "Channel Existence Checker - Kafka Connector";
+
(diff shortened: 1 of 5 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
```diff
@@ -0,0 +1,293 @@
...
+ *
+ * <p>A method (checkChannelExistence) in this class bypasses that open Channel call and directly
+ * creates the payload for Get Status {@link
+ * StreamingChannelExistenceChecker#CHANNEL_STATUS_ENDPOINT} API.
+ */
+public class StreamingChannelExistenceChecker {
+  private static final KCLogger LOGGER =
+      new KCLogger(StreamingChannelExistenceChecker.class.getName());
+
+  private static final String DEBUG_PREFIX = "Channel Existence Checker - Kafka Connector";
```
sfc-gh-tzhang, line 60:
```suggestion
  private static final String DEBUG_PREFIX = "StreamingChannelExistenceChecker";
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
```diff
@@ -0,0 +1,293 @@
...
+            tableName,
+            channelName,
+            this.clientSequencer);
+    channelExistenceCheckerRequest.setChannels(Collections.singletonList(requestDTO));
+    channelExistenceCheckerRequest.setRole(connectorConfig.get(Utils.SF_ROLE));
+    channelExistenceCheckerRequest.setRequestId(UUID.randomUUID().toString());
+
+    try {
+      ObjectMapper objectMapper = new ObjectMapper();
+      objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
sfc-gh-tzhang, line 132: what is this?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
```diff
@@ -0,0 +1,293 @@
...
+
+      HttpUriRequest request =
+          requestBuilder.generateStreamingIngestPostRequest(
+              requestPayload,
+              CHANNEL_STATUS_ENDPOINT,
+              DEBUG_PREFIX /* Message passed on to Ingest SDK */);
+
+      ChannelExistenceCheckerResponse channelExistenceCheckerResponse;
+      try (CloseableHttpResponse closeableHttpResponse = httpClient.execute(request)) {
+
```
sfc-gh-tzhang, line 143: nit: extra line

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
```diff
@@ -0,0 +1,293 @@
...
+        if (closeableHttpResponse == null) {
+          LOGGER.warn(
+              "Null response obtained in {} for channel:{}", CHANNEL_STATUS_ENDPOINT, channelName);
```
sfc-gh-tzhang, line 146: why we need a log if we're going to throw?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
```diff
@@ -0,0 +1,293 @@
...
+                .getStatusCode()
+                .equals(ERR_CHANNEL_HAS_INVALID_CLIENT_SEQUENCER.getStatusCode())
+            || channelExistenceCheckerResponse
+                .getChannels()
+                .get(0)
+                .getStatusCode()
+                .equals(SUCCESS.getStatusCode())) {
+          // We check two status codes, one for invalid client sequencers and one for a valid client
+          // Sequencer
+          LOGGER.debug(String.format("%s Channel:%s does exist", DEBUG_PREFIX, channelName));
```
sfc-gh-tzhang, line 189: duplicate with above?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
```diff
@@ -0,0 +1,293 @@
...
+          return true;
+        }
+      }
+    } catch (Exception e) {
+      LOGGER.error(
+          "{} Error Getting a valid response for ChannelName:{} error:{}",
+          DEBUG_PREFIX,
+          channelName,
+          e.getMessage());
```
sfc-gh-tzhang, line 198: any issue with returning here instead of thrown an error?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
sfc-gh-japatel: ack!

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
sfc-gh-japatel: good catch, not needed!

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
sfc-gh-japatel: to know which channel name had this issue.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
sfc-gh-japatel: it's Does and doesn't for channel.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
```diff
@@ -0,0 +1,293 @@
...
+    this(tableName, connectorConfig, 0L);
+  }
+
+  /**
+   * Determines if a Channel Exists or not for a table defined in this class's ctor.
+   *
+   * @param channelName Name of Channel to Check for Existence
+   * @return True if it exists, false if it doesn't.
+   */
+  public boolean checkChannelExistence(final String channelName) {
```
sfc-gh-rcheng, line 101: nit: checkChannelExists so that it's easier to intuit that returning TRUE means the channel exists

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
sfc-gh-rcheng: I think that's ok, being unable to get a channel basically means it doesnt exist. We should add that behavior to the javadoc though: @return True if it exists, false if it doesn't or failed to get status

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
```diff
@@ -0,0 +1,293 @@
...
+          "{} Status hit from {}, requestId:{}",
+          statusLine.getStatusCode(),
+          apiName,
+          requestId == null ? "" : requestId.toString());
+
+      // if we have a 503 exception throw a backoff
+      switch (statusLine.getStatusCode()) {
+          // If we have a 503, BACKOFF
+        case HttpStatus.SC_SERVICE_UNAVAILABLE:
+          throw new BackOffException();
```
sfc-gh-rcheng, line 224: Do we want to retry for this BackOffException? We don't retry in checkChannelExistance

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
sfc-gh-japatel: makes sense, will do!

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingChannelExistenceChecker.java
sfc-gh-japatel: good catch, I copied this function from sdk and I do see we dont have retries in channelexistence, I plan to add retries, but let me check this..

PR #738, opened 2023-10-26
PROD-39429 PART-3 Implement System Function for Update Channel OffsetToken

sfc-gh-japatel:
## What
- A simple implementation of updating Channel with an OffsetToken
- fetchOffsetTokenWithRetry which calls the channel's getOffsetToken is now checking for an explicit `NULL` offsetToken. This is because, no matter what I set through system function (null, `NULL`) I always get NULL from channel API and not null string. 

 
## Why
- We will set NULL to new channels which were created by the commit 3bf9106b22510c62068f7d2f7137b9e57989274c after we have fetched the offsetToken from it and reset those to Kafka committed offset. 
- Idea behind it is next time we reopen channelV2 (new format) we see their offset NULL and hence we will continue to use old ones. 
- We ideally want to also drop this new channels but that is not possible today. 

## Tests
Added tests.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
@@ -2,6 +2,8 @@ package com.snowflake.kafka.connector.internal;
 
 import static com.snowflake.kafka.connector.Utils.TABLE_COLUMN_CONTENT;
 import static com.snowflake.kafka.connector.Utils.TABLE_COLUMN_METADATA;
+import static com.snowflake.kafka.connector.internal.streaming.StreamingUtils.CHANNEL_UPDATE_RESPONSE_CHANNEL_DOESNT_EXIST_SYSTEM_FUNC_MSG;
+import static com.snowflake.kafka.connector.internal.streaming.StreamingUtils.SUCCESSFUL_CHANNEL_UPDATE_RESPONSE_SYSTEM_FUNC_MSG;
 
 import com.snowflake.kafka.connector.Utils;
 import com.snowflake.kafka.connector.internal.streaming.IngestionMethodConfig;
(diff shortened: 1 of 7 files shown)


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -1005,4 +1007,67 @@ public Connection getConnection() {
...
     return this.internalStage;
   }
+
+  @Override
+  public boolean updateStreamingChannelOffsetToken(
+      String tableName, String channelName, String offsetToken)
+      throws SnowflakeKafkaConnectorException {
+    InternalUtils.assertNotEmpty("tableName", tableName);
+    InternalUtils.assertNotEmpty("channelName", channelName);
+    String fullyQualifiedTableName =
```
sfc-gh-tzhang, line 1017: do we handle case sensitive names here?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -1005,4 +1007,67 @@ public Connection getConnection() {
   public SnowflakeInternalStage getInternalStage() {
     return this.internalStage;
   }
+
+  @Override
+  public boolean updateStreamingChannelOffsetToken(
```
sfc-gh-tzhang, line 1012: Actually, can we just use this function to check whether the new channel exists instead of making the REST API call? If yes, then everything in PART-2 can be discard

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeErrors.java
```diff
@@ -196,6 +196,11 @@ public enum SnowflakeErrors {
       "2017",
       "Failed to check schema evolution permission",
       "Failed to check schema evolution permission"),
+
```
sfc-gh-tzhang, line 199: nit: extra line

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1010,7 +1010,7 @@ private long fetchLatestCommittedOffsetFromSnowflake() {
       offsetToken = this.channel.getLatestCommittedOffsetToken();
       LOGGER.info(
           "Fetched offsetToken for channelName:{}, offset:{}", this.getChannelName(), offsetToken);
-      return offsetToken == null
+      return offsetToken == null || offsetToken.equalsIgnoreCase("NULL")
```
sfc-gh-tzhang, line 1013: this is a behavior change, right?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -1005,4 +1007,67 @@ public Connection getConnection() {
...
+      throws SnowflakeKafkaConnectorException {
+    InternalUtils.assertNotEmpty("tableName", tableName);
+    InternalUtils.assertNotEmpty("channelName", channelName);
+    String fullyQualifiedTableName =
+        prop.getProperty(InternalUtils.JDBC_DATABASE)
+            + "."
+            + prop.getProperty(InternalUtils.JDBC_SCHEMA)
+            + "."
+            + tableName;
+    String query = "select system$snowpipe_streaming_update_channel_offset_token((?), (?), (?));";
```
sfc-gh-tzhang, line 1023: @sfc-gh-asen Are we always going to set the offset token for the latest client sequencer? What if there is a concurrent open channel call?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: That's a good point but I dont know how else we would identify NULL set by our sys function. Somehow I wasnt able to get a null offsetToken (like a new channel) after setting in sys function. (I tried NULL, null, empty strings but with no luck)

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-japatel: Thats a good point.. we can use sys function to check if channel exist or not too. 

so then it looks like we will use two sys calls ..

sfc-gh-japatel: actually I just thought about this. We have to have part-2. you cant use this sys func to check if channel exists or not. in doing so you will lose a legit offset token in case a channel exists.

sfc-gh-rcheng: Isn't the new offset token be passed in by Kafka valid? so we can throw away the legit offset token of the existing channel anyways

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-japatel: Good question, for instance I tried both 
select system$snowpipe_streaming_update_channel_offset_token('TESTDB_KAFKA.KAFKA_TEST.kafka_connector_test_table_3347672768933248582','test_connector_kafka_connector_test_table_3347672768933248582_2','');

and 
select system$snowpipe_streaming_update_channel_offset_token('TESTDB_KAFKA.KAFKA_TEST.KAFKA_CONNECTOR_TEST_TABLE_3347672768933248582','TEST_CONNECTOR_KAFKA_CONNECTOR_TEST_TABLE_3347672768933248582_2','');

It works for both. so case insensitive i guess.

PR #740, opened 2023-11-07
SNOW-913746 upgrade JDBC to 3.14.3

sfc-gh-japatel:
We need this for a customer who wants to use s3 + regional URL. 

Here are supporting PRs which this change was originally dependent on:
1. Server side: https://github.com/snowflakedb/snowflake/commit/208702b0a6bdc3639c46f0c4b8ef4375824456b6
2. JDBC: https://github.com/snowflakedb/snowflake-jdbc/commit/14c13183df076cc55cf0ab0663756d43ab7be448

Originally, it was released in 3.14.2 but 3.14.2 is backward incompatible change which impacts telemetry constructor from IngestSDK. Hence using 3.14.3

TLDR;
- We use a response string from configure client API for snowpipe streaming. 
- We send this to get SnowflakeFileTransferMetadata Object. 
- JDBC needs to honor this variable and use the s3 regional URL. 
- This still needs the param set at an account level. (Please look at server side change for that)

We will push this to master as well as to a release branch for a requested customer.

### DIFF
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -28,9 +28,9 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-maven-
     - name: Install Python
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v4
       with:
-        python-version: '3.6'
+        python-version: '3.9'
         architecture: 'x64'
     - name: Decrypt profile.json in Snowflake Cloud ${{ matrix.snowflake_cloud }}
       run: ./.github/scripts/decrypt_secret.sh ${{ matrix.snowflake_cloud }}
--- a/.github/workflows/PerfTest.yml
+++ b/.github/workflows/PerfTest.yml
@@ -22,9 +22,9 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-maven-
     - name: Install Python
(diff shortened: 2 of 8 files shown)


sfc-gh-rcheng (approved):

sfc-gh-xhuang (approved):

sfc-gh-tzhang (approved):
Want to make sure this is on top of 2.0.0? And where is the change to upgrade the KC version?

sfc-gh-japatel:
> Want to make sure this is on top of 2.0.0? And where is the change to upgrade the KC version? 

This is not on top of 2.0.0 because there was a bug fix in 2.0.1

I want to also push this to master! Upgrade will come later!

sfc-gh-japatel:
FYI, all apache tests are failing since our regex 
snowflake-kafka-connector-[0-9]*\.[0-9]*\.[0-9]*\.jar$

doesnt match with snowflake-kafka-connector-2.0.2-rc
It passes confluent because we copy entire zip file and dont do regex match. https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/test/build_runtime_jar.sh#L109

I will ignore those errors for now since only bump up version commits are failing.

sfc-gh-japatel:
Closing this PR, merged this change to master and also release 2.0.2-rc.

Issue #571, opened 2023-03-14
[question] Why does Snowpipe Streaming require a user role to be specified?


JonathonO:
Hi,

Currently using v1.9.1, and I expected the user's default role to be used when no `snowflake.role.name` is specified, but, that's not the case.

Based on documentation and this old [comment](https://github.com/snowflakedb/snowflake-kafka-connector/issues/200#issuecomment-694332962), it seems to be supported for `snowpipe` ingestion mode at least, although I've not tested this myself.

Is this on the roadmap at all?

sfc-gh-japatel:
Hi @JonathonO yes, role is required for Snowpipe Streaming. 
Thanks for bringing this up to us, we will take this as a feature request.

sfc-gh-xhuang:
We have eliminated this requirement in the java sdk and have it on our backlog to add the same support for kafka connector though there are some dependencies for roles in our kafka connector such as schema evolution

sfc-gh-gjachimko:
closed due to inactivity and submitted code changes.

PR #744, opened 2023-11-10
[SNOW-954150] Use map of clients with different configurations instead of one client for multiple connectors configurations

sfc-gh-rcheng:
### Summary
If two sink connectors with the one client optimization and different connection properties such as the user's role, ingestion is stopped on the connector that was last added. The fix is to map client properties to existing client in the one client optimization.

### Changes
1. Register multiple clients in StreamingClientProvider with a LoadingCache backed with Caffeine
2. Added a StreamingClientProperties object that is converted from a subset of the given connector configurations. This is used in StreamingClientProvider to determine client equality (if a new client needs to be created)
    1. Multiple connectors that have the same client properties (authentication, target db and schema, role, etc) can use the same client. This means that connectors with different configs like buffer flush thresholds, target table name or JMX configurations can use the same client.
    2. This should not leak any important configurations across connectors. 
4. StreamingClientProvider does its best to provide valid clients by recreating clients when the registered client is invalid.
3. Refactored StreamingClientProviderTest to be UT and mock the StreamingClientHandler entirely and split off a StreamingClientProviderIT

### Initial Issue Investigation
ConnectorA with the admin role and ConnectorB with the public role
1. ConnectorA(admin) is ingesting data successfully
4. ConnectorB(public) is added
  a. If the one client optimization is enabled, ConnectorB(public) tries using ConnectorA(admin)'s client which has the admin role. As a result ConnectorB(public) fails with 400 unauthorized and does not begin ingestion.

#### Testing
- Got repro and confirmed fix manually and with the [KC release testing suite](https://github.com/snowflakedb/snowflake-kafka-connector-testing/tree/main/reltests/test/test_suites/kcauto_twosink_sp/configs)
  - Added Streaming, Schematization and Snowpipe ingestion to the release test suite to use two sink connectors at the same time
- Added sink service level IT to test multiple tpChannels with different role configurations
image

Caused by this jira: https://snowflakecomputing.atlassian.net/browse/SNOW-954150


src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -1507,4 +1508,74 @@ private Map<String, String> getConfig() {
       return TestUtils.getConfForStreamingWithOAuth();
     }
   }
+
+  // note this test relies on testrole_kafka and testrole_kafka_1 roles being granted to test_kafka
+  // user
+  @Test
+  public void testStreamingIngest_multipleChannel_distinctClients() throws Exception {
```
sfc-gh-rcheng, line 1515: confirmed this test fails with the same error as jira when run on master, passes with this pr's changes

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
```diff
@@ -50,34 +51,48 @@ public static boolean isClientValid(SnowflakeStreamingIngestClient client) {
...
-      Optional<String> snowpipeStreamingBdecVersion =
-          Optional.ofNullable(connectorConfig.get(SNOWPIPE_STREAMING_FILE_VERSION));
-      snowpipeStreamingBdecVersion.ifPresent(
-          overriddenValue -> {
-            LOGGER.info("Config is overridden for {} ", SNOWPIPE_STREAMING_FILE_VERSION);
-            parameterOverrides.put(BLOB_FORMAT_VERSION, overriddenValue);
-          });
+    // Override only if bdec version is explicitly set in config, default to the version set
+    // inside Ingest SDK
+    Map<String, Object> parameterOverrides = new HashMap<>();
```
sfc-gh-rcheng, line 85: moved otuside of the try-catch because we don't need to catch any exceptions that come from this

sfc-gh-xhuang:
Is the role issue related to this?
https://github.com/snowflakedb/snowflake-kafka-connector/issues/571
https://snowflakecomputing.atlassian.net/browse/SNOW-532834

sfc-gh-rcheng:
> Is the role issue related to this? #571 https://snowflakecomputing.atlassian.net/browse/SNOW-532834

Unrelated, but it could factor into why this wasn't caught until now

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -50,28 +57,31 @@ public static StreamingClientProvider getStreamingClientProviderInstance() {
...
       StreamingClientHandler streamingClientHandler) {
     this();
-    this.parameterEnabledClient = parameterEnabledClient;
+    this.registeredClientMap = registeredClientMap;
     this.streamingClientHandler = streamingClientHandler;
   }
 
   private static final KCLogger LOGGER = new KCLogger(StreamingClientProvider.class.getName());
-  private SnowflakeStreamingIngestClient parameterEnabledClient;
+  private Map<Properties, SnowflakeStreamingIngestClient> registeredClientMap;
```
sfc-gh-japatel, line 77: lets mention what properties might contain.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -84,25 +94,69 @@ private StreamingClientProvider() {
...
+        if (!invalidClientProps.isEmpty()) {
+          String invalidClientNames = "";
+
+          for (Map.Entry<Properties, SnowflakeStreamingIngestClient> entry : invalidClientProps) {
+            SnowflakeStreamingIngestClient invalidClient = entry.getValue();
+            invalidClientNames +=
+                invalidClient != null
+                        && invalidClient.getName() != null
+                        && !invalidClient.getName().isEmpty()
+                    ? invalidClient.getName()
```
sfc-gh-japatel, line 124: should we add comma here somewhere, or consider usinig arraylist for invalidClientNames

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -84,25 +94,69 @@ private StreamingClientProvider() {
...
+        if (this.registeredClientMap.containsKey(inputProps)) {
+          resultClient = this.registeredClientMap.get(inputProps);
+          LOGGER.debug("Using existing streaming client with name: {}", resultClient.getName());
+        } else {
+          Pair<Properties, SnowflakeStreamingIngestClient> propertiesAndClient =
+              this.streamingClientHandler.createClient(connectorConfig);
+          resultClient = propertiesAndClient.getSecond();
+          this.registeredClientMap.put(
+              propertiesAndClient.getFirst(), propertiesAndClient.getSecond());
+          LOGGER.debug("Created and registered new client with name: {}", resultClient.getName());
```
sfc-gh-japatel, line 146: should we print specific properties values here? so that we can debug which properties it is creating for and which not (in case things go wrong?)

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -1507,4 +1508,74 @@ private Map<String, String> getConfig() {
...
+    catService.insert(catRecords);
+    dogService.insert(dogRecords);
+
+    // check data was ingested
+    TestUtils.assertWithRetry(() -> catService.getOffset(catTp) == catRecordCount, 20, 5);
+    TestUtils.assertWithRetry(() -> dogService.getOffset(dogTp) == dogRecordCount, 20, 5);
+
+    // close services
+    catService.closeAll();
+    dogService.closeAll();
```
sfc-gh-japatel, line 1579: lets assert number of clients == 2 or size of map we created. we can create a setter in provider.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -22,15 +22,22 @@
...
 
 /**
- * Factory that provides the streaming client(s). There should only be one provider, but it may
- * provide multiple clients if optimizations are disabled - see
- * ENABLE_STREAMING_CLIENT_OPTIMIZATION_CONFIG in the {@link SnowflakeSinkConnectorConfig }
+ * Static factory that provides streaming client(s). There should only be one provider per KC worker
+ * node, meaning that there may be multiple providers serving one connector and/or multiple
+ * connectors on one provider. If the optimization is disabled, the provider will not reuse old
+ * clients,/ see ENABLE_STREAMING_CLIENT_OPTIMIZATION_CONFIG in the {@link
+ * SnowflakeSinkConnectorConfig }
```
sfc-gh-japatel, line 40: I feel we should mention about map we created here. 
Its true that StreamingClientProvider is going to be per worker but props to client map is what we care more about IMO. worth mentioning that but a minor nit.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -84,25 +94,69 @@ private StreamingClientProvider() {
...
-        this.parameterEnabledClient = this.streamingClientHandler.createClient(connectorConfig);
+      LOGGER.debug(
+          "Streaming client optimization is enabled per worker node. Reusing valid clients when"
+              + " possible");
+
+      try {
+        this.providerLock.lock();
+
+        // remove invalid clients
+        List<Map.Entry<Properties, SnowflakeStreamingIngestClient>> invalidClientProps =
```
sfc-gh-japatel, line 111: do we care about removing invalid clients in get client? feel a little weird we have it here.

sfc-gh-japatel:
Overall lgtm, will wait for one for revision. Also, it might be worth looking into LoadingCache too since we are using map which has loading, eviction technique etc. Not saying we should use either since both has pros and cons.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: good point, added comment above parameter

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: Using array list for the name instead! unsure why i used string to begin with...

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: printing the non sensitive values here

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: thats true, i dont think it necessarily should be here, ill look into the loadingcache or caffeine as an alternative

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: i think the class comment should be more abstracted away from the client map for future proofing. i added a brief comment on the map object itself, what do you think about that as an alternative?

sfc-gh-tzhang:
> The fix is to map client properties to existing client in the one client optimization.

Could you update the description to explain how the map is done?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
```diff
@@ -50,34 +52,61 @@ public static boolean isClientValid(SnowflakeStreamingIngestClient client) {
...
+    if (connectorConfig != null) {
+      streamingClientProps.putAll(
+          StreamingUtils.convertConfigForStreamingClient(new HashMap<>(connectorConfig)));
+    }
+
+    return streamingClientProps;
+  }
+
+  public static String getLoggablePropertyStr(Properties properties) {
+
```
sfc-gh-tzhang, line 73: extra line?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
```diff
@@ -50,34 +52,61 @@ public static boolean isClientValid(SnowflakeStreamingIngestClient client) {
...
+    return properties.entrySet().stream()
+        .filter(
+            propKvp ->
+                !StreamingUtils.SENSITIVE_STREAMING_CONFIG_PROPERTIES.contains(
+                    propKvp.getKey().toString()))
```
sfc-gh-tzhang, line 78: Do we have the issue about lowercase VS uppercase?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
```diff
@@ -50,34 +52,61 @@ public static boolean isClientValid(SnowflakeStreamingIngestClient client) {
...
+        .collect(Collectors.toList())
+        .toString();
+
+    //    while (properties.)
```
sfc-gh-tzhang, line 82: remove?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
```diff
@@ -50,34 +52,61 @@ public static boolean isClientValid(SnowflakeStreamingIngestClient client) {
...
+   *
+   * @param connectorConfig configuration properties for a connector
+   * @return the Properties object needed for client creation
+   */
+  public static Properties getClientProperties(Map<String, String> connectorConfig) {
+    Properties streamingClientProps = new Properties();
+
+    if (connectorConfig != null) {
+      streamingClientProps.putAll(
+          StreamingUtils.convertConfigForStreamingClient(new HashMap<>(connectorConfig)));
```
sfc-gh-tzhang, line 66: Do you know why we need to create a new HashMap?

sfc-gh-tzhang: Also should we just return a Properties from `StreamingUtils.convertConfigForStreamingClient`

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
sfc-gh-tzhang: I don't think it matters much, but why we don't want to catch the exception from this line?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-tzhang: +1, I don't think we want to do this here

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -84,25 +100,75 @@ private StreamingClientProvider() {
...
+              "Found and removed {} invalid clients: {}",
+              invalidClientProps.size(),
+              invalidClientNames.toString());
+        }
+
+        // look for client corresponding to the input properties or create new client
+        Properties inputProps = StreamingClientHandler.getClientProperties(connectorConfig);
+        if (this.registeredClientMap.containsKey(inputProps)) {
+          resultClient = this.registeredClientMap.get(inputProps);
+          LOGGER.debug("Using existing streaming client with name: {}", resultClient.getName());
```
sfc-gh-tzhang, line 145: Should we update this and the one below to INFO? Otherwise it would be hard if we have issues in this area in the future

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java
```diff
@@ -73,6 +75,16 @@ public class StreamingUtils {
   public static final String STREAMING_CONSTANT_OAUTH_CLIENT_SECRET = "oauth_client_secret";
   public static final String STREAMING_CONSTANT_OAUTH_REFRESH_TOKEN = "oauth_refresh_token";
 
+  // contains config properties that should not be exposed to the customer as they may contain PII
+  // or authentication information
+  public static final List<String> SENSITIVE_STREAMING_CONFIG_PROPERTIES =
```
sfc-gh-tzhang, line 80: This looks risky, it's possible that people forgot updating this list and log sensitive data, should we do the other way around? Only log the information that is in the list

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -84,25 +100,75 @@ private StreamingClientProvider() {
...
+
+          LOGGER.error(
+              "Found and removed {} invalid clients: {}",
+              invalidClientProps.size(),
+              invalidClientNames.toString());
+        }
+
+        // look for client corresponding to the input properties or create new client
+        Properties inputProps = StreamingClientHandler.getClientProperties(connectorConfig);
+        if (this.registeredClientMap.containsKey(inputProps)) {
```
sfc-gh-tzhang, line 143: What's this mean here by using all the properties as the key? Like if one connector has buffer flush time to be 1s and the other one is 10s, then a separate client is used? But why it can't use the same client?

sfc-gh-tzhang (commented):
Left some comments, PTAL!

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
sfc-gh-rcheng: yeah it doesn't matter too much. I moved it out because I prefer to minimize scope of the try catch

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: yeah thats true, updating this and the else statement log level to info

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
sfc-gh-rcheng: We don't have a case concern because it is looking at the key which should be the same constant string value. However i will tolowercase all of it just to be safe

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java
sfc-gh-rcheng: Good point, inverted it to LOGGABLE_STREAMING_CONFIG_PROPERTIES

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
sfc-gh-rcheng: Yes that would be cleaner, altered the convertConfigForStreamingClient to return Properties instead of a map. Removed hashmap clone as i don't see why we need that as well

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
```diff
@@ -50,34 +65,42 @@ public static boolean isClientValid(SnowflakeStreamingIngestClient client) {
...
   }
 
+  public static String getLoggableClientProperties(Properties properties) {
+    return properties.entrySet().stream()
+        .filter(
+            propKvp ->
+                LOGGABLE_STREAMING_CONFIG_PROPERTIES.contains(
+                    propKvp.getKey().toString().toUpperCase()))
+        .collect(Collectors.toList())
+        .toString();
```
sfc-gh-japatel, line 75: this string is essentially returning only the keys? is that what you want or you wanted key value pair both?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -84,25 +97,76 @@ private StreamingClientProvider() {
...
+        //                        && !invalidClient.getName().isEmpty()
+        //                    ? invalidClient.getName()
+        //                    : "noClientNameFound");
+        //            this.registeredClientMap.remove(entry.getKey());
+        //          }
+        //
+        //          LOGGER.error(
+        //              "Found and removed {} invalid clients: {}",
+        //              invalidClientProps.size(),
+        //              invalidClientNames.toString());
```
sfc-gh-japatel, line 136: i am guessing you are planning to remove this commented code?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
```diff
@@ -50,34 +65,42 @@ public static boolean isClientValid(SnowflakeStreamingIngestClient client) {
     return client != null && !client.isClosed() && client.getName() != null;
   }
 
+  public static String getLoggableClientProperties(Properties properties) {
```
sfc-gh-japatel, line 68: javadocs on public method

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
sfc-gh-rcheng: No this string returns keys and values. The filter method only looks at the string, but the collect will contain the entire key value pair. Added a test in StreamingClientHandler to confirm this

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
sfc-gh-rcheng: added, thanks

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: yup it has been removed!

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: Removed. Now we use Caffeine as a loading cache to ensure the registered clients don't eat up all the memory

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -47,72 +50,123 @@ public static StreamingClientProvider getStreamingClientProviderInstance() {
...
+      LoadingCache<Map<String, String>, SnowflakeStreamingIngestClient> registeredClients) {
     this.streamingClientHandler = streamingClientHandler;
+    this.registeredClients = registeredClients;
   }
 
   private static final KCLogger LOGGER = new KCLogger(StreamingClientProvider.class.getName());
-  private SnowflakeStreamingIngestClient parameterEnabledClient;
   private StreamingClientHandler streamingClientHandler;
-  private Lock providerLock;
+  private LoadingCache<Map<String, String>, SnowflakeStreamingIngestClient> registeredClients;
```
sfc-gh-rcheng, line 94: keying this cache on the connector config means that we are creating more clients. However I think this is better because connector configs would never be shared between clients. 

We can also use the Properties created convertConfigForStreamingClient which would be a subset of the connector configs. This would reduce the number of clients created, however I worry about accidentally leaking a config cross connector

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
```diff
@@ -50,6 +63,25 @@ public static boolean isClientValid(SnowflakeStreamingIngestClient client) {
...
+    return properties == null
+        ? ""
+        : properties.entrySet().stream()
+            .filter(
+                propKvp ->
+                    LOGGABLE_STREAMING_CONFIG_PROPERTIES.stream()
+                        .anyMatch(propKvp.getKey().toString()::equalsIgnoreCase))
+            .collect(Collectors.toList())
+            .toString();
```
sfc-gh-japatel, line 82: are we fetching corresponding values?

sfc-gh-rcheng: yup, theres a test that confirms this. the filter is only looking at the key but it also returns the value in the .collect

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-japatel: I dont think we should have key as a map in loadingcache,, 
How about we create a CacheKey object in this class with the required properties we care about? 
take a look at this for reference: https://github.com/snowflakedb/snowflake/blob/main/GlobalServices/modules/dbsec/authn-api/src/main/java/com/snowflake/security/ShiroPasswordHashCache.java#L148

Once you have a class as a key, you can then define your own equals and hashcode and that will guarantee to work as expected? 
For instance here you can add all properties as instance variable which you think is helpful for unique creation of a client. 

IMO,
ideally the connector name should be enough to create a client at a worker level. each task which belongs to that connector will get the same instance and all other connectors will create its own client..

src/test/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProviderTest.java
```diff
@@ -101,34 +106,66 @@ public void testFirstGetClient() {
...
+    assert this.invalidClient.getName().contains(invalidClientConfig.get(Utils.NAME));
+    Mockito.verify(injectedStreamingClientHandler, Mockito.times(1))
+        .createClient(invalidClientConfig);
   }
 
   @Test
   public void testGetExistingClient() {
-    // test
+    // setup configs with different roles
+    this.clientConfig1.put(SF_ROLE, "testrole_kafka");
```
sfc-gh-japatel, line 165: lets not reveal our connector role, since this is UT we can use any random name right?

sfc-gh-rcheng: no, this actually creates the client, we are only spy-ing on the streamingHandler so it creates the client. i'll see if i can use a different role

sfc-gh-rcheng: Made the StreamingClientProviderTest file actually UT so we don't use the role anymore. Split off a separate StreamingClientProviderIT that will reveal the role though

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
sfc-gh-japatel: got it thanks!

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -47,72 +56,166 @@ public static StreamingClientProvider getStreamingClientProviderInstance() {
...
+      // Override only if bdec version is explicitly set in config, default to the version set
+      // inside Ingest SDK
+      this.parameterOverrides = new HashMap<>();
+      Optional<String> snowpipeStreamingBdecVersion =
+          Optional.ofNullable(connectorConfig.get(SNOWPIPE_STREAMING_FILE_VERSION));
+      snowpipeStreamingBdecVersion.ifPresent(
+          overriddenValue -> {
+            LOGGER.info("Config is overridden for {} ", SNOWPIPE_STREAMING_FILE_VERSION);
+            parameterOverrides.put(BLOB_FORMAT_VERSION, overriddenValue);
+          });
```
sfc-gh-japatel, line 207: not needed now but I think we can remove this since this is deprecated on server side. 
WDYT @sfc-gh-tzhang

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -47,72 +56,166 @@ public static StreamingClientProvider getStreamingClientProviderInstance() {
...
-    this.providerLock.unlock();
+  }
+
+  // TEST ONLY - return the current state of the registered clients
+  @VisibleForTesting
+  public Map<StreamingClientProperties, SnowflakeStreamingIngestClient> getRegisteredClients() {
+    return this.registeredClients.asMap();
+  }
+
+  public static class StreamingClientProperties {
```
sfc-gh-japatel, line 187: short description here would be useful! but I like the idea.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -47,72 +56,166 @@ public static StreamingClientProvider getStreamingClientProviderInstance() {
...
+
+    @Override
+    public boolean equals(Object other) {
+      return other.getClass().equals(StreamingClientProperties.class)
+          & ((StreamingClientProperties) other).clientProperties.equals(this.clientProperties);
+    }
+
+    @Override
+    public int hashCode() {
+      return this.clientProperties.hashCode();
```
sfc-gh-japatel, line 218: i think i am okay for this being the hashcode and equals implementation but it would be better to use all instances and not only clientprops

sfc-gh-japatel: You could use intellij's feature of creating hashcode and equals..
right click -> generate -> equals and hashcode

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-japatel: This would be fixed by recent change from @sfc-gh-rcheng

sfc-gh-japatel (approved):
lgtm! 
Couple of minor suggestions. We definitely need improved test infra for running multiple connectors and ingesting them at same time. 
Lets create a JIRA for warsaw team.. could be a good task.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: you're right, @sfc-gh-tzhang had the same concern so I've created a StreamingClientProperties class that will take in a connectorConfig and parse out the important properties passed into the client. The LoadingCache uses the StreamingClientProperties to determine if a new client needs to be created. This means connectors with the same configuration except different names or buffer flush thresholds would still use the same client

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: yup! discussed in this [thread](https://github.com/snowflakedb/snowflake-kafka-connector/pull/744#discussion_r1399966870) and offline. thanks for the important call out

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: Split into a separate class with comments, PTAL!

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: I overrode the hashcode and equals here because those methods are what the LoadingCache uses to determine key equality. The clientName is based on the connector name which is unique and the parameter override is not relevant to the client, so I ignored those two properties.

sfc-gh-rcheng:
> We definitely need improved test infra for running multiple connectors and ingesting them at same time. Lets create a JIRA for warsaw team.. could be a good task.
Created [jira](https://snowflakecomputing.atlassian.net/browse/SNOW-976889)!

sfc-gh-rcheng:
Updated description with changes, rerunning release testing framework for e2e test with multiple connectors now

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
```diff
@@ -51,40 +39,26 @@ public static boolean isClientValid(SnowflakeStreamingIngestClient client) {
...
+   * @param streamingClientProperties The properties to create the client
    * @return A newly created client
    */
-  public SnowflakeStreamingIngestClient createClient(Map<String, String> connectorConfig) {
+  public SnowflakeStreamingIngestClient createClient(
+      StreamingClientProperties streamingClientProperties) {
     LOGGER.info("Initializing Streaming Client...");
 
-    // get streaming properties from config
-    Properties streamingClientProps = new Properties();
```
sfc-gh-rcheng, line 63: moved into StreamingClientProperties

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -47,72 +52,137 @@ public static StreamingClientProvider getStreamingClientProviderInstance() {
     return StreamingClientProviderSingleton.streamingClientProvider;
   }
 
-  /** ONLY FOR TESTING - to get a provider with injected properties */
```
sfc-gh-rcheng, line 50: moved these methods to bottom of class for readability

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-japatel: sounds good!

src/test/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProviderTest.java
```diff
@@ -58,139 +42,235 @@ public static Collection<Object[]> input() {
...
-
-    this.streamingClientHandler = Mockito.spy(StreamingClientHandler.class);
-    this.streamingClientProvider =
-        StreamingClientProvider.getStreamingClientProviderForTests(
-            null, this.streamingClientHandler);
-  }
-
-  @After
-  public void tearDown() {
-    this.streamingClientHandler.closeClient(this.client1);
```
sfc-gh-rcheng, line 83: Since this is UT it should not actually create the client, so I refactored this test to use mocks and pulled most of the complexity into the new IT. caught by @sfc-gh-japatel

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  /**
+   * Gets the loggable properties, see {@link
+   * StreamingClientProperties#LOGGABLE_STREAMING_CONFIG_PROPERTIES}
+   *
+   * @return A formatted string with the loggable properties
+   */
+  public String getLoggableClientProperties() {
+    return this.clientProperties == null | this.clientProperties.isEmpty()
```
sfc-gh-tzhang, line 99: can `clientProperties` be null or empty? I thought we have check to make sure some of the configurations are required?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -47,72 +52,137 @@ public static StreamingClientProvider getStreamingClientProviderInstance() {
...
       LOGGER.info(
-          "Streaming client optimization is enabled, returning the existing streaming client if"
-              + " valid");
-      this.providerLock.lock();
-      // recreate streaming client if needed
-      if (!StreamingClientHandler.isClientValid(this.parameterEnabledClient)) {
-        LOGGER.error("Current streaming client is invalid, recreating client");
-        this.parameterEnabledClient = this.streamingClientHandler.createClient(connectorConfig);
+          "Streaming client optimization is enabled per worker node. Reusing valid clients when"
+              + " possible");
```
sfc-gh-tzhang, line 119: nit: should we move this log after the client creation is succeeded? Similar to the logic below when one client optimization is not enabled, then you can combine two log lines together.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -47,72 +52,137 @@ public static StreamingClientProvider getStreamingClientProviderInstance() {
...
+    StreamingClientProperties clientProperties = new StreamingClientProperties(connectorConfig);
+
+    // invalidate cache
+    SnowflakeStreamingIngestClient registeredClient =
+        this.registeredClients.getIfPresent(clientProperties);
+    if (registeredClient != null) {
+      // invalidations are processed on the next get or in the background, so we still need to close
+      // the client here
+      this.registeredClients.invalidate(clientProperties);
+      this.streamingClientHandler.closeClient(registeredClient);
```
sfc-gh-tzhang, line 157: duplicate as line 162?

sfc-gh-tzhang: Also do you need to check whether the client is still valid before closing it?

sfc-gh-tzhang (approved):
Left some minor comments, PTAL, otherwise LGTM, thanks!

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
sfc-gh-rcheng: it shouldn't ever be null or empty, but I added the check just in case

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: We call close on the given client and the client registered in the cache. Technically they should be the same client, however I prefer to call close on both just in case the given client is different or somehow the registered cache was corrupted

The streamingClientHandler will check if the client is valid before calling close, so the extra close call will no-op if it is invalid

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-rcheng: good point. Moved line to end of method so that we only have one log and added a warn log for when the registered client is invalid because ideally it should always be valid

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientHandler.java
```diff
@@ -35,9 +28,6 @@
 /** This class handles all calls to manage the streaming ingestion client */
 public class StreamingClientHandler {
   private static final KCLogger LOGGER = new KCLogger(StreamingClientHandler.class.getName());
-  private static final String STREAMING_CLIENT_PREFIX_NAME = "KC_CLIENT_";
-  private static final String TEST_CLIENT_NAME = "TEST_CLIENT";
-
   private AtomicInteger createdClientId = new AtomicInteger(0);
```
sfc-gh-rcheng, line 31: Added this back (was removed in previous PR iterations) in case removing it causes concurrency issues with client naming

PR #745, opened 2023-11-11
SNOW-913746 upgrade jdbc 3.14.3 

sfc-gh-japatel:
Upgrade JDBC to 3.14.3. 

Modify some tests since v3.13.31 is throwing NPE.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
@@ -299,6 +299,9 @@ public class InternalUtils {
         proxyProperties.put(SFSessionProperty.PROXY_USER.getPropertyKey(), username);
         proxyProperties.put(SFSessionProperty.PROXY_PASSWORD.getPropertyKey(), password);
       }
+      // There is a change in JDBC version 3.13.31 which causes NPE if this is not added.
+      // https://github.com/snowflakedb/snowflake-jdbc/blob/master/src/main/java/net/snowflake/client/jdbc/SnowflakeUtil.java#L614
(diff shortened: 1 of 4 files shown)


src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
```diff
@@ -299,6 +299,9 @@ protected static Properties generateProxyParametersIfRequired(Map<String, String
         proxyProperties.put(SFSessionProperty.PROXY_USER.getPropertyKey(), username);
         proxyProperties.put(SFSessionProperty.PROXY_PASSWORD.getPropertyKey(), password);
       }
+      // There is a change in JDBC version 3.13.31 which causes NPE if this is not added.
```
sfc-gh-tzhang, line 302: Do we have a JIRA created for the owner to fix this? We shouldn't have this assumption, this is a behavior change

sfc-gh-tzhang (approved):

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
sfc-gh-japatel: Will create a issue on JDBC. 
I think this is more of a breaking change than a behavior change. Since we have a workaround, its okay to add this here but I agree it should be fixed by JDBC.

sfc-gh-japatel: Opened a PR in JDBC: https://github.com/snowflakedb/snowflake-jdbc/pull/1559

sfc-gh-japatel:
@sfc-gh-jfan snyk is not running can you check please? thanks

PR #749, opened 2023-11-16
SNOW-532834 snowflake.role.name is optional for Snowpipe Streaming

sfc-gh-mbobowski:
This PR makes 'snowflake.role.name' parameter in Snowpipe Streaming configuration optional.

Additionally it fixes a bug in unit tests so that they can actually fail when behaviour change.


sfc-gh-rcheng (approved):
lgtm, thanks for the fix! lets wait for @sfc-gh-tzhang 's review as well since he has more context on the role removal

sfc-gh-japatel (approved):
lgtm! Thanks
As far as we are above v2.0.2 (ingest sdk) we should be able to use client without the role.

sfc-gh-tzhang (approved):
LGTM, thanks! Like I mentioned, there is a second change required for this JIRA, please make sure to handle that before closing

sfc-gh-xhuang:
Please also test with schematization enabled. I believe Alec mentioned there were some complications with this change and schematization but I cannot find the exact notes

sfc-gh-mbobowski:
PR needs more work to handle schematization. Closing for now to prevent accidental merge.

PR #750, opened 2023-11-16
PROD-39429 Implement migrate sys func from new channel(Format V2) to old channel (V1)

sfc-gh-japatel:
This is a long term implementation for potential data duplication introduced because of a new channel name format. 
- It is about moving away for customers who might have onboarded to new channel format which was introduced in version 2.1.0 (Now de-listed)
- We will stick to old format for new customers running this for the first time. 
- Renamed channelName to channelNameFormatV1 and introduced `channelNameFormatV2`

Added tests in `TopicPartitionChannel` and `TopicPartitionChannel`

Notes:
- This PR is opened against [japatel-SNOW-971412-release-2-1-1-ready](https://github.com/snowflakedb/snowflake-kafka-connector/tree/japatel-SNOW-971412-release-2-1-1-ready) which is going to be the release branch for v2.1.1
- A similar PR will also be opened against master. 

End to End tests:
1. Use 2.1.0
2. Use two connectors:
3. Uses channel Name v2
Screenshot 2023-11-20 at 4 54 59 PM
4. Stop 
5. Replace jar (2.1.1) and restart. 
6. Does the migration and you can only see old channel format
Screenshot 2023-11-20 at 5 17 59 PM

7.  Stop again and restart in 2.1.1
Nothing happens and we do get a valid response saying newchannelFormatV2 doesnt exist
Screenshot 2023-11-20 at 5 18 11 PM

PR #751, opened 2023-11-16
PROD-39429 Implement migrate sys func from new channel(Format V2) to old channel (V1) - Push to Main

sfc-gh-japatel:
PR https://github.com/snowflakedb/snowflake-kafka-connector/pull/750 but pushed in main. 

Copying it as is. 

This is a long term implementation for potential data duplication introduced because of a new channel name format. 
- It is about moving away for customers who might have onboarded to new channel format which was introduced in version 2.1.0 (Now de-listed)
- We will stick to old format for new customers running this for the first time. 
- Renamed channelName to channelNameFormatV1 and introduced `channelNameFormatV2`

Added tests in `TopicPartitionChannel` and `TopicPartitionChannel`

Notes:
- This PR is opened against [japatel-SNOW-971412-release-2-1-1-ready](https://github.com/snowflakedb/snowflake-kafka-connector/tree/japatel-SNOW-971412-release-2-1-1-ready) which is going to be the release branch for v2.1.1
- A similar PR will also be opened against master. 

End to End tests:
1. Use 2.1.0
2. Use two connectors:
3. Uses channel Name v2
Screenshot 2023-11-20 at 4 54 59 PM
4. Stop 
5. Replace jar (2.1.1) and restart. 
6. Does the migration and you can only see old channel format
Screenshot 2023-11-20 at 5 17 59 PM

7.  Stop again and restart in 2.1.1
Nothing happens and we do get a valid response saying newchannelFormatV2 doesnt exist
Screenshot 2023-11-20 at 5 18 11 PM

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -175,7 +175,9 @@ public class SnowflakeSinkConnectorConfig {
   public static final boolean ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_DEFAULT = true;
   public static final String ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_DOC =
       "This config is used to enable/disable streaming channel offset migration logic. If true, we"
-          + " will migrate offset token from channel name format V2 to name format v1.";
+          + " will migrate offset token from channel name format V2 to name format v1. V2 channel"
+          + " format is deprecated and V1 will be used always, disabling this config could have"
+          + " ramifications. Please consult Snowflake support before setting this to false.";
 
   // MDC logging header
   public static final String ENABLE_MDC_LOGGING_CONFIG = "enable.mdc.logging";
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -168,6 +168,15 @@ public class SnowflakeSinkConnectorConfig {
       "Whether to optimize the streaming client to reduce cost. Note that this may affect"
           + " throughput or latency and can only be set if Streaming Snowpipe is enabled";
 
+  public static final String ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_CONFIG =
+      "enable.streaming.channel.offset.migration";
+  public static final String ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_DISPLAY =
+      "Enable streaming Channel Offset Migration";
+  public static final boolean ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_DEFAULT = true;
+  public static final String ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_DOC =
+      "This config is used to enable/disable streaming channel offset migration logic. If true, we"
+          + " will migrate offset token from channel name format V2 to name format v1.";
+
   // MDC logging header
   public static final String ENABLE_MDC_LOGGING_CONFIG = "enable.mdc.logging";
   public static final String ENABLE_MDC_LOGGING_DISPLAY = "Enable MDC logging";
@@ -591,7 +600,17 @@ public class SnowflakeSinkConnectorConfig {
             CONNECTOR_CONFIG,
             8,
             ConfigDef.Width.NONE,
-            ENABLE_MDC_LOGGING_DISPLAY);
+            ENABLE_MDC_LOGGING_DISPLAY)
+        .define(
+            ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_CONFIG,
+            Type.BOOLEAN,
+            ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_DEFAULT,
+            Importance.LOW,
+            ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_DOC,
+            CONNECTOR_CONFIG,
+            9,
+            ConfigDef.Width.NONE,
+            ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_DISPLAY);
   }
 
   public static class TopicToTableValidator implements ConfigDef.Validator {
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -171,7 +171,7 @@ public class SnowflakeSinkConnectorConfig {
   public static final String ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_CONFIG =
       "enable.streaming.channel.offset.migration";
   public static final String ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_DISPLAY =
-      "Enable streaming Channel Offset Migration";
+      "Enable Streaming Channel Offset Migration";
   public static final boolean ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_DEFAULT = true;
   public static final String ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_DOC =
       "This config is used to enable/disable streaming channel offset migration logic. If true, we"
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
@@ -1038,13 +1038,12 @@ public class SnowflakeConnectionServiceV1 implements SnowflakeConnectionService 
         migrateOffsetTokenResultFromSysFunc = resultSet.getString(1 /*Only one column*/);
       }
       if (migrateOffsetTokenResultFromSysFunc == null) {
-        LOGGER.warn(
-            "No result found in Migrating OffsetToken through System Function for tableName:{},"
-                + " sourceChannel:{}, destinationChannel:{}",
-            fullyQualifiedTableName,
-            sourceChannelName,
-            destinationChannelName);
-        return false;
+        final String errorMsg =
+            String.format(
+                "No result found in Migrating OffsetToken through System Function for tableName:%s,"
+                    + " sourceChannel:%s, destinationChannel:%s",
+                fullyQualifiedTableName, sourceChannelName, destinationChannelName);
+        throw SnowflakeErrors.ERROR_5023.getException(errorMsg, this.telemetry);
       }
 
       ChannelMigrateOffsetTokenResponseDTO channelMigrateOffsetTokenResponseDTO =
@@ -1059,12 +1058,15 @@ public class SnowflakeConnectionServiceV1 implements SnowflakeConnectionService 
           channelMigrateOffsetTokenResponseDTO);
       return true;
     } catch (SQLException | JsonProcessingException e) {
-      LOGGER.error(
-          "Migrating OffsetToken for a SourceChannel:{} in table:{} failed due to:{}",
-          sourceChannelName,
-          fullyQualifiedTableName,
-          e.getMessage());
-      return false;
+      final String errorMsg =
+          String.format(
+              "Migrating OffsetToken for a SourceChannel:%s in table:%s failed due to"
+                  + " exceptionMessage:%s and stackTrace:%s",
+              sourceChannelName,
+              fullyQualifiedTableName,
+              e.getMessage(),
+              Arrays.toString(e.getStackTrace()));
+      throw SnowflakeErrors.ERROR_5023.getException(errorMsg, this.telemetry);
     }
   }
 }
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
@@ -3,7 +3,10 @@ package com.snowflake.kafka.connector.internal;
 import static com.snowflake.kafka.connector.Utils.TABLE_COLUMN_CONTENT;
 import static com.snowflake.kafka.connector.Utils.TABLE_COLUMN_METADATA;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.snowflake.kafka.connector.Utils;
+import com.snowflake.kafka.connector.internal.streaming.ChannelMigrateOffsetTokenResponseDTO;
 import com.snowflake.kafka.connector.internal.streaming.IngestionMethodConfig;
 import com.snowflake.kafka.connector.internal.streaming.SchematizationUtils;
 import com.snowflake.kafka.connector.internal.telemetry.SnowflakeTelemetryService;
@@ -55,6 +58,8 @@ public class SnowflakeConnectionServiceV1 implements SnowflakeConnectionService 
 
   // User agent suffix we want to pass in to ingest service
   public static final String USER_AGENT_SUFFIX_FORMAT = "SFKafkaConnector/%s provider/%s";
+
+  private final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
   SnowflakeConnectionServiceV1(
       Properties prop,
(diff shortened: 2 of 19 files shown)


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -1006,4 +1011,60 @@ public Connection getConnection() {
...
+      stmt.setString(3, destinationChannelName);
+      ResultSet resultSet = stmt.executeQuery();
+
+      String migrateOffsetTokenResultFromSysFunc = null;
+      if (resultSet.next()) {
+        migrateOffsetTokenResultFromSysFunc = resultSet.getString(1 /*Only one column*/);
+      }
+      if (migrateOffsetTokenResultFromSysFunc == null) {
+        LOGGER.warn(
+            "No result found in Migrating OffsetToken through System Function for tableName:{},"
```
sfc-gh-tzhang, line 1042: what does it mean by no result found? No destination channel or source channel found? or both?

sfc-gh-tzhang: If this is not expected, should we throw an exception?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -1006,4 +1011,60 @@ public Connection getConnection() {
...
+      LOGGER.info(
+          "Migrate OffsetToken response for table:{}, sourceChannel:{}, destinationChannel:{}"
+              + " is:{}",
+          tableName,
+          sourceChannelName,
+          destinationChannelName,
+          channelMigrateOffsetTokenResponseDTO);
+      return true;
+    } catch (SQLException | JsonProcessingException e) {
+      LOGGER.error(
```
sfc-gh-tzhang, line 1062: What's the reason of only logging here instead of throw an exception?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -1006,4 +1011,60 @@ public Connection getConnection() {
...
+          "Migrating OffsetToken for a SourceChannel:{} in table:{} failed due to:{}",
+          sourceChannelName,
+          fullyQualifiedTableName,
+          e.getMessage());
```
sfc-gh-tzhang, line 1066: getMessage might be NULL for some exceptions, could we do better?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -134,7 +137,10 @@ public class TopicPartitionChannel {
   private final TopicPartition topicPartition;
 
   /* Channel Name is computed from topic and partition */
-  private final String channelName;
+  private final String channelNameFormatV1;
+
+  /* Channel Name format V2 is computed from connector name, topic and partition */
+  private final String channelNameFormatV2;
```
sfc-gh-tzhang, line 143: Why this is needed as a class variable?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -293,24 +308,66 @@ public TopicPartitionChannel(
...
+        Boolean.parseBoolean(
+            sfConnectorConfig.getOrDefault(
+                SnowflakeSinkConnectorConfig.ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_CONFIG,
+                Boolean.toString(ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_DEFAULT)));
+    if (!isEnableChannelOffsetMigration) {
+      LOGGER.info(
+          "Config:{} is disabled for connector:{}",
+          ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_CONFIG,
+          conn.getConnectorName());
     }
```
sfc-gh-tzhang, line 353: @sfc-gh-rcheng is making some changes in this area to log all non-sensitive configurations

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskStreamingTest.java
```diff
@@ -88,17 +88,17 @@ public void testSinkTaskInvalidRecord_InMemoryDLQ() throws Exception {
         new TopicPartitionChannel(
             mockStreamingClient,
             topicPartition,
-            SnowflakeSinkServiceV2.partitionChannelKey(TEST_CONNECTOR_NAME, topicName, partition),
```
sfc-gh-tzhang, line 91: Do we have any tests that can actually test the end2end migration?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-japatel: This is not expected and I thought we decided to not throw any exceptions/swallow all and continue using old channel. 

I can see the concern though but at this point I feel it is about whether we want to halt ingestion or atleast ignore the exception and continue moving forward with old. Halting ingestion could be better and it helps us know something is wrong rather than continuing with old channel with ramifications we dont. 
WDYT Toby?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-japatel: Same reason as above!

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-japatel: might be worth logging stacktrace too. let me add that

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskStreamingTest.java
sfc-gh-japatel: all end to end tests will eventually go through this path right? since it is default to true.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -168,6 +168,15 @@ public class SnowflakeSinkConnectorConfig {
...
           + " throughput or latency and can only be set if Streaming Snowpipe is enabled";
 
+  public static final String ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_CONFIG =
+      "enable.streaming.channel.offset.migration";
+  public static final String ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_DISPLAY =
+      "Enable Streaming Channel Offset Migration";
+  public static final boolean ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_DEFAULT = true;
+  public static final String ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_DOC =
+      "This config is used to enable/disable streaming channel offset migration logic. If true, we"
+          + " will migrate offset token from channel name format V2 to name format v1.";
```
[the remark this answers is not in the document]
sfc-gh-japatel: good point, let me add!

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -278,6 +285,14 @@ public TopicPartitionChannel(
 
     this.enableSchemaEvolution = this.enableSchematization && hasSchemaEvolutionPermission;
 
+    this.channelNameFormatV2 =
```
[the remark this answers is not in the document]
sfc-gh-japatel: it was private final so had to put it outside in ctor. Let me think if this is needed as instance variable. (i had plans to modify it in tests but havent added any IT tests)

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskStreamingTest.java
sfc-gh-japatel: good idea to ingest through topicpartition channel through v2 and then let it go through migration.. I will add IT tests..

sfc-gh-japatel: I think I see the issue, the IT tests are showing up other PR: https://github.com/snowflakedb/snowflake-kafka-connector/pull/750/files
I might have messed up with merge conflicts.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: not needed great catch! fixed..

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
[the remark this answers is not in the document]
sfc-gh-japatel: Done!

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: @sfc-gh-rcheng do you mind waiting for my change to go in first? I think it would be easier for you to rebase.. But let me know if its not possible..

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskStreamingTest.java
sfc-gh-japatel: ported the IT tests here, added few more tests in connectionIT tests.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-japatel: For an unexpected exception, I am throwing a runtime exception. PTAL! It's better to fail here IMO. Thanks folks!

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-japatel: Done!

sfc-gh-japatel:
> overall lgtm, but i would like an IT or e2e test that explictly opens and ingests to a V2 channel and confirms that is migrated later on

Sorry, the merge conflicts didnt show IT tests. 
I ported IT tests from other PR, added few more connectionserviceIT tests. 
e2e test involving two jars is going to be a bit difficult with our current infra, but let me think more on that.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -1006,4 +1011,62 @@ public Connection getConnection() {
...
+      ChannelMigrateOffsetTokenResponseDTO channelMigrateOffsetTokenResponseDTO =
+          OBJECT_MAPPER.readValue(
+              migrateOffsetTokenResultFromSysFunc, ChannelMigrateOffsetTokenResponseDTO.class);
+      LOGGER.info(
+          "Migrate OffsetToken response for table:{}, sourceChannel:{}, destinationChannel:{}"
+              + " is:{}",
+          tableName,
+          sourceChannelName,
+          destinationChannelName,
+          channelMigrateOffsetTokenResponseDTO);
```
sfc-gh-tzhang, line 1058: This log might be confusing to customer, could we only log when there is actually a migration being done? Or do we even need this client side log given that we have a bunch of server side logging in place?

src/main/java/com/snowflake/kafka/connector/internal/streaming/ChannelMigrationResponseCode.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  /**
+   * Given a response code which was received from server side, check if it as successful migration
+   * or a failure.
+   *
+   * @param statusCodeToCheck response from JDBC system function call.
+   * @return true or false
+   */
+  public static boolean isStatusCodeSuccessful(final long statusCodeToCheck) {
```
sfc-gh-tzhang, line 81: private function since this is only used here?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -1006,4 +1011,62 @@ public Connection getConnection() {
...
+      return channelMigrateOffsetTokenResponseDTO;
```
sfc-gh-tzhang, line 1059: Discussed offline, it's better to throw an exception for all invalid cases, and customer could use the configuration if they customer to bypass the exceptions

src/main/java/com/snowflake/kafka/connector/internal/streaming/ChannelMigrationResponseCode.java
```diff
@@ -0,0 +1,101 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.snowflake.kafka.connector.internal.streaming;
+
+import com.google.common.annotations.VisibleForTesting;
+
+/**
```
sfc-gh-tzhang, line 21: Please add a comment or add a prefix if this is test only code

sfc-gh-tzhang (approved):
Left some comments, PTAL, otherwise LGTM!

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-japatel: This should be fine, this is only once per channel during open partition, not very often. I would like to keep this which helps us in debugging any customer issue.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-japatel: +1

src/main/java/com/snowflake/kafka/connector/internal/streaming/ChannelMigrationResponseCode.java
sfc-gh-japatel: using it in migration function

src/main/java/com/snowflake/kafka/connector/internal/streaming/ChannelMigrationResponseCode.java
sfc-gh-japatel: done!

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -1006,4 +1012,69 @@ public Connection getConnection() {
...
+    try {
+      PreparedStatement stmt = conn.prepareStatement(query);
+      stmt.setString(1, fullyQualifiedTableName);
+      stmt.setString(2, sourceChannelName);
+      stmt.setString(3, destinationChannelName);
+      ResultSet resultSet = stmt.executeQuery();
+
+      String migrateOffsetTokenResultFromSysFunc = null;
+      if (resultSet.next()) {
+        migrateOffsetTokenResultFromSysFunc = resultSet.getString(1 /*Only one column*/);
```
sfc-gh-rcheng, line 1039: out of scope for this PR - do we have any guarantees that this DTO won't change on the server side? a comment, test or something server side to warn against changes since we now throw the exception on failure

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-rcheng: yeah no worries, lets go with this change first!

src/test/java/com/snowflake/kafka/connector/internal/ConnectionServiceIT.java
```diff
@@ -411,4 +424,114 @@ public void testConnectionFunction() {
...
+      List<SinkRecord> recordsInChannelFormatV2 =
+          TestUtils.createJsonStringSinkRecords(0, noOfRecords * 2, tableName, 0);
+      recordsInChannelFormatV2.forEach(newChannelFormatV2::insertRecordToBuffer);
+
+      TestUtils.assertWithRetry(
+          () -> newChannelFormatV2.getOffsetSafeToCommitToKafka() == (noOfRecords * 2), 5, 5);
+
+      conn.migrateStreamingChannelOffsetToken(tableName, sourceChannelName, destinationChannelName);
+      TestUtils.assertWithRetry(
+          () -> service.getOffset(new TopicPartition(tableName, 0)) == (noOfRecords * 2), 5, 5);
```
sfc-gh-rcheng, line 530: can we also send some more data after migration? to confirm that ingestion works after as well

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelIT.java
```diff
@@ -562,4 +563,168 @@ public void testPartialBatchChannelInvalidationIngestion_schematization() throws
...
+    ChannelMigrateOffsetTokenResponseDTO channelMigrateOffsetTokenResponseDTO =
+        conn.migrateStreamingChannelOffsetToken(
+            testTableName, channelNameFormatV2, testChannelName);
+    Assert.assertTrue(isChannelMigrationResponseSuccessful(channelMigrateOffsetTokenResponseDTO));
+    Assert.assertEquals(
+        SUCCESS.getStatusCode(), channelMigrateOffsetTokenResponseDTO.getResponseCode());
+
+    // Fetch offsetToken from API should now give you same as other channel
+    TestUtils.assertWithRetry(
+        () -> service.getOffset(new TopicPartition(topic, PARTITION)) == noOfRecords, 5, 5);
```
sfc-gh-rcheng, line 632: ditto on the above - can we try ingestion after migration as well?

sfc-gh-rcheng (approved):
Thanks for the change! The ITs are pretty cool - but small request to try ingestion after the migration as well

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-japatel: yeah, i think the json exception will help with that.. I think I have the comment on server side saying be cautious to make changes in the interface. but let me try to mimick and see what happens if response is changed. I could probably handle new fields in the object mapper.

src/test/java/com/snowflake/kafka/connector/internal/ConnectionServiceIT.java
sfc-gh-japatel: umm, connectionServiceIT should only test the function but let me add it in TopicPartitionChannelIT. good call!

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelIT.java
sfc-gh-japatel: Ahh i realized, the ingestion is happening in testChannelMigrateOffsetTokenSystemFunction_NullOffsetTokenInFormatV2 after migration.

sfc-gh-japatel: But no harm in adding it here, so I have added in recent commit!

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-japatel: moved the method to its own method so that we can test it. If the response changes from server side which shouldnt, this tests methods will start failing

Issue #748, opened 2023-11-16
Using configProviders other than file fails validation


swehner:
When using any configProvider with a name different from file, config validation fails because it receives the raw value and tries to validate the connection.

Example config:

```
connector.class=com.snowflake.kafka.connector.SnowflakeSinkConnector
config.providers=secretManager
config.providers.secretManager.class=com.github.jcustenborder.kafka.config.aws.SecretsManagerConfigProvider
snowflake.user.name=testuser
snowflake.private.key=${secretManager:data-platform-demo/msk_connect_credentials:private_key}
```

Error is 

```
  | 2023-11-16T17:57:16.000+01:00 | [Worker-0a7081b95c2eab512] [2023-11-16 16:57:16,917] ERROR [SF_KAFKA_CONNECTOR] Validate: Error connecting to snowflake:[SF_KAFKA_CONNECTOR] Exception: Invalid private key
  | 2023-11-16T17:57:16.000+01:00 | [Worker-0a7081b95c2eab512] Error Code: 0002
  | 2023-11-16T17:57:16.000+01:00 | [Worker-0a7081b95c2eab512] Detail: private key should be a valid PEM RSA private key
```

Adding some more log statements it receives the uninterpolated string:

```
2023-11-16T17:57:16.000+01:00 | [Worker-0a7081b95c2eab512] [2023-11-16 16:57:16,903] ERROR [SF_KAFKA_CONNECTOR] %%%% got private key ${secretManager:data-platform-demo/msk_connect_credentials:private_key} (com.snowflake.kafka.connector.internal.InternalUtils:94)
```

Opened a pull request to fix this issue here: https://github.com/snowflakedb/snowflake-kafka-connector/pull/747

hitesh-yadav-x:
Assuming that you are trying to pull the credentials from AWS Secrets Manager. May I suggest use the default environment provider that kafka has and inject the variables in the container/run environment.

If using DOCKER set
`ENV CONNECT_CONFIG_PROVIDERS="env"
ENV CONNECT_CONFIG_PROVIDERS_ENV_CLASS="org.apache.kafka.common.config.provider.EnvVarConfigProvider"`

Then in config use below:

`"snowflake.role.name": "${env:SNOWFLAKE_ROLE}",
"snowflake.url.name":"${env:SNOWFLAKE_URL}",
"snowflake.user.name": "${env:SNOWFLAKE_USER}",
[...]

PR #747, opened 2023-11-16
Support other configProviders than file in validation phase

swehner:
I'm trying to use different config providers in my snowflake connector, but get an error during validation because the snowflake private key is not valid.
This is because validation seems to receive the raw config values (not running through configProviders which seems iffy).

In any case there was already a workaround to support `${file:` as a config provider value and to skip validation. I've extended it to include any configProvider with a different name.

PR #753, opened 2023-11-30
[external contributor] Support config providers in validation by @swehner

sfc-gh-rcheng:
Copy @swehner 's [PR](https://github.com/snowflakedb/snowflake-kafka-connector/pull/747) for github actions runs. 
@sfc-gh-xhuang please confirm that CLA was signed

### Original PR description:
I'm trying to use different config providers in my snowflake connector, but get an error during validation because the snowflake private key is not valid.
This is because validation seems to receive the raw config values (not running through configProviders which seems iffy).

In any case there was already a workaround to support ${file: as a config provider value and to skip validation. I've extended it to include any configProvider with a different name.

Fixes issue https://github.com/snowflakedb/snowflake-kafka-connector/issues/748


sfc-gh-xhuang:
CLA was signed and good to merge

PR #754, opened 2023-12-01
Disable jmx by default 

sfc-gh-rcheng:
JMX very slightly affects performance at high throughputs. This PR disables by default for both Snowpipe and Snowpipe Streaming, customers will have to enable it in the configuration.


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -110,7 +110,7 @@ public class SnowflakeSinkConnectorConfig {
 
   // metrics
   public static final String JMX_OPT = "jmx";
-  public static final boolean JMX_OPT_DEFAULT = true;
+  public static final boolean JMX_OPT_DEFAULT = false;
```
sfc-gh-tzhang, line 113: Should we do it for snowpipe streaming only? otherwise we need to go through a BCR process

PR #758, opened 2023-12-06
NO-SNOW: Expose the Ingest SDK MAX_CLIENT_LAG configuration in KC 

sfc-gh-tzhang:
When schematization is enabled, we're inserting row by row in order to preserve the original order of the batch, this leads to an issue if the data validation for each row is slow, we might end up with many small files since the SDK flushes every second. Exposing the MAX_CLIENT_LAG helps solve this issue shoer term. Longer term, it would be good if we can avoid inserting row by row for schematization.

- Support a new configuration `snowflake.streaming.max.client.lag` in KC. Instead of adding a new configuration, i update the blob version config since we don't support ARROW in the new SDK, adding less code is always better
- Extend the `StreamingClientProperties` to use different client if the `parameterOverrides` is also different
- Corresponding test cases added

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -118,7 +118,8 @@ public class SnowflakeSinkConnectorConfig {
       IngestionMethodConfig.SNOWPIPE.toString();
 
   // This is the streaming max client lag which can be defined in config
-  public static final String SNOWPIPE_STREAMING_MAX_LAG = "snowflake.streaming.max.lag";
+  public static final String SNOWPIPE_STREAMING_MAX_CLIENT_LAG =
+      "snowflake.streaming.max.client.lag";
 
   // TESTING
   public static final String REBALANCING = "snowflake.test.rebalancing";
@@ -540,7 +541,7 @@ public class SnowflakeSinkConnectorConfig {
             ConfigDef.Width.NONE,
             INGESTION_METHOD_OPT)
         .define(
-            SNOWPIPE_STREAMING_MAX_LAG,
+            SNOWPIPE_STREAMING_MAX_CLIENT_LAG,
             Type.LONG,
             StreamingUtils.STREAMING_BUFFER_FLUSH_TIME_MINIMUM_SEC,
             ConfigDef.Range.atLeast(StreamingUtils.STREAMING_BUFFER_FLUSH_TIME_MINIMUM_SEC),
@@ -549,7 +550,7 @@ public class SnowflakeSinkConnectorConfig {
             CONNECTOR_CONFIG,
             6,
             ConfigDef.Width.NONE,
-            SNOWPIPE_STREAMING_MAX_LAG)
+            SNOWPIPE_STREAMING_MAX_CLIENT_LAG)
         .define(
             ERRORS_TOLERANCE_CONFIG,
             Type.STRING,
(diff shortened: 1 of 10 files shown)


sfc-gh-rcheng (approved):
lgtm! concerned about if we need to notify customers or change docs if the bdec version override parameter is removed

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -1461,12 +1462,49 @@ public void testSchematizationSchemaEvolutionWithNonNullableColumn() throws Exce
...
+      TestUtils.assertWithRetry(
+          () -> service.getOffset(new TopicPartition(topic, partition)) == noOfRecords, 5, 4);
+      Assert.fail("The rows should not be flushed");
+    } catch (Exception e) {
+      // do nothing
+    }
+
+    // Wait for enough time, the rows should be flushed
+    TestUtils.assertWithRetry(
+        () -> service.getOffset(new TopicPartition(topic, partition)) == noOfRecords, 30, 6);
```
sfc-gh-rcheng, line 1496: your github actions test is failing on this, probably just need to bump the retry count?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -117,9 +117,9 @@ public class SnowflakeSinkConnectorConfig {
   public static final String INGESTION_METHOD_DEFAULT_SNOWPIPE =
       IngestionMethodConfig.SNOWPIPE.toString();
 
-  // This is the streaming bdec file version which can be defined in config
-  // NOTE: Please do not override this value unless recommended from snowflake
```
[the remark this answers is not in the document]
sfc-gh-tjones: This was meant to be internal only so I think it's fine to remove. I believe we track old file versions in Snowflake and alert if we detect arrow files.

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
sfc-gh-tzhang: Yeah, I don't know why, I think 180 seconds should be enough....

sfc-gh-rcheng: I've similar issues (passes locally, consistently fails on github) when i used 100 seconds, had to bump the total retry to 400 seconds

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
```diff
@@ -77,15 +78,15 @@ public StreamingClientProperties(Map<String, String> connectorConfig) {
         STREAMING_CLIENT_PREFIX_NAME
             + connectorConfig.getOrDefault(Utils.NAME, DEFAULT_CLIENT_NAME);
 
-    // Override only if bdec version is explicitly set in config, default to the version set
-    // inside Ingest SDK
+    // Override only if the max client lag is explicitly set in config
     this.parameterOverrides = new HashMap<>();
     Optional<String> snowpipeStreamingBdecVersion =
```
sfc-gh-rcheng, line 83: change: snowpipeStreamingBdecVersion -> snowpipeStreamingMaxClientLag or something similar

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
sfc-gh-tzhang: ok, bumping again...seems very weird, this is something I want to investigate after coming back

sfc-gh-rcheng:
merging for toby - needed to create securonix jar

PR #767, opened 2023-12-20
Remove committed records in buffer instead of emptying on channel reopen

sfc-gh-rcheng:
### Background
Currently when the channel needs to be reopened (invalidated, error, schematization) we clear the buffer and expect kafka to resend the data. This works in most situations, however if kafka sends a batch with a missing offset, ingestion stalls.

### Changes
Do not empty the buffer on channel reopen but to instead clear out any committed records to prevent duplication. 

### Tests
- Added e2e tests for snowpipe, streaming and schematization (though this change only applies to streaming/schematization)
- Added missing offset integration test
- Our schematization tests also rely heavily on hitting the record threshold perfectly. Fixed this by adding an empty insert call that will do a time based flush


src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1000,24 +937,28 @@ private void resetChannelMetadataAfterRecovery(
...
     try {
-      LOGGER.warn(
-          "[RESET_PARTITION] Emptying current buffer:{} for Channel:{} due to reset of offsets in"
-              + " kafka",
-          this.streamingBuffer,
-          this.getChannelNameFormatV1());
-      this.streamingBuffer = new StreamingBuffer();
+      // reset buffer by removing already committed records
+      StreamingBuffer resetBuffer = streamingBufferToReset;
+      if (offsetRecoveredFromSnowflake != NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE) {
```
sfc-gh-rcheng, line 942: this is the crux of the change - instead of emptying the buffer and waiting for record.offset == processedOffset + 1, we remove only the records that have been committed, so we can take any record.offset > processedOffset

sfc-gh-tzhang:
I'm trying to fix these issues in a different way which I feel like it's better than splitting and merging buffers, please take a look at https://github.com/snowflakedb/snowflake-kafka-connector/pull/775 and let me know what you think when you come back.

sfc-gh-tzhang:
Closing this PR, I will create a new PR with integrating the updated tests only

PR #774, opened 2024-01-10
SNOW-989387: Fix for java.lang.NoClassDefFoundError

sfc-gh-wfateem:
java.lang.NoClassDefFoundError: Could not initialize class net.snowflake.client.jdbc.internal.apache.arrow.memory.RootAllocator occurs when the JDBC driver tries to process the result of the offset migration query. This is a known long-standing issue in the JDBC driver so a workaround is introduced with this fix.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
@@ -32,6 +32,7 @@ public class InternalUtils {
   static final String JDBC_WAREHOUSE = "warehouse"; // for test only
   static final String JDBC_AUTHENTICATOR = SFSessionProperty.AUTHENTICATOR.getPropertyKey();
   static final String JDBC_TOKEN = SFSessionProperty.TOKEN.getPropertyKey();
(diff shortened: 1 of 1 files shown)


sfc-gh-tzhang (approved):
LGTM, thanks! I assume existing tests should cover the testing?

sfc-gh-wfateem:
> LGTM, thanks! I assume existing tests should cover the testing?

@sfc-gh-tzhang you know, now that you mention it, I'm not actually sure. If they were then we should have caught this issue, but I don't know if that's because of the lack of coverage or, more likely, because we're using an older JDK than version 16. The problem would only surface on JDK 16 and higher. I'll look into it a bit more.

sfc-gh-wfateem:
@sfc-gh-tzhang the tests in place are sufficient, but it seems like this problem slipped under the radar because it only occurs on JDK runtime environments 16 and higher. We're running all our tests on lower JDK runtime environments. Since we already have plans to add tests on different JDK runtime versions in the future, then there's not much more to do here. I'll go ahead and merge the change as agreed.

PR #775, opened 2024-01-10
SNOW-943288, SNOW-938379: Fix two schema evolution cases that could cause non-exactly once delivery

sfc-gh-tzhang:
Fix two exactly once delivery behavior issues with Snowpipe Streaming:
- When there're gaps between offsets due to records being put into the DLQ or NULL records being skipped, the current logic doesn't work after a channel reopening event (e.g. schema evolution) since it expects continuous offsets in order to guarantee exactly once delivery, causing ingestion to stop. 
-  When a flush is triggered by size or rowcount threshold, it's possible that only partial rows in the buffer are flushed and with a channel reopening event (e.g. schema evolution), currently the leftover rows in the buffer are still being considered which is causing the offset to be advanced and skip some offsets. 

The fix in this PR makes sure we won't flush any leftover rows in the buffer, they will all be skipped and discarded. The next batch should start from the expected offset after the offset reset as part of the channel reopening event.


src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -388,8 +370,13 @@ public void insertRecordToBuffer(SinkRecord kafkaSinkRecord) {
...
 
-    // Ignore adding to the buffer until we see the expected offset value
-    if (shouldIgnoreAddingRecordToBuffer(kafkaSinkRecord, currentProcessedOffset)) {
+    // Reset the value if it's a new batch
+    if (isFirstRowInBatch) {
+      needToSkipCurrentBatch = false;
+    }
+
+    // Simply skip inserting into the buffer if the row should be ignored after channel reset
+    if (needToSkipCurrentBatch) {
```
sfc-gh-tzhang, line 379: TODO: add a log here when we're skipping rows

test/test_suit/test_schema_evolution_json_ignore_tombstone.py
```diff
@@ -65,7 +65,7 @@ def send(self):
                 key.append(json.dumps({'number': str(self.recordNum - 1)}).encode('utf-8'))
                 value.append(None)
                 key.append(json.dumps({'number': str(self.recordNum)}).encode('utf-8'))
-                value.append("") # community converters treat this as a tombstone
+                value.append("")  # community converters treat this as a tombstone
```
sfc-gh-tzhang, line 68: TODO: add a new tests with higher row count so that we can have multiple batches

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkService.java
```diff
@@ -45,6 +45,14 @@ public interface SnowflakeSinkService {
    */
   void insert(final SinkRecord record);
 
+  /**
+   * Insert a JSON record into the buffer
+   *
+   * @param record record content
+   * @param isFirstRowInBatch indicates whether the given record is the first record in a batch
```
sfc-gh-japatel, line 52: its good to indicate if its first record in the batch that kafka sends or first record in insert rows API batch.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-japatel: are you planning to do it here?

test/test_suites.py
```diff
@@ -178,9 +180,10 @@ def create_end_to_end_test_suites(driver, nameSalt, schemaRegistryAddress, testS
             test_instance=TestSchemaEvolutionJson(driver, nameSalt), clean=True, run_in_confluent=True,
             run_in_apache=True
         )),
+        # TODO: tzhang need to root cause
         ("TestSchemaEvolutionJsonIgnoreTombstone", EndToEndTestSuite(
-            test_instance=TestSchemaEvolutionJsonIgnoreTombstone(driver, nameSalt), clean=True, run_in_confluent=True,
-            run_in_apache=True
+            test_instance=TestSchemaEvolutionJsonIgnoreTombstone(driver, nameSalt), clean=True, run_in_confluent=False,
```
sfc-gh-japatel, line 185: planning to enable this again?

sfc-gh-japatel:
@sfc-gh-rcheng has written some useful tests in this PR: https://github.com/snowflakedb/snowflake-kafka-connector/pull/767
Can we merge the two by piggybacking on those tests so that we have enough confidence in this PR.

sfc-gh-japatel (commented):
I left my concerns/questions in the PR. Consider my soft approval on this PR but it would be really great if we can consolidate tests from two PRs. 
Thanks for the work. (To both Revi and Toby)

sfc-gh-tzhang:
> @sfc-gh-rcheng has written some useful tests in this PR: #767 Can we merge the two by piggybacking on those tests so that we have enough coverage in this PR?

I added (and modified) all tests that make sense. All three added end2end tests don't really make sense because even though the test tries to skip some offsets, but that's before adding to Kafka so the records in Kafka will still have continuous offset. I believe the two fixed behaviors described in the description should be covered by existing tests + the added/re-enabled IT/end2end tests.

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelIT.java
```diff
@@ -706,4 +708,61 @@ public void testChannelMigrateOffsetTokenSystemFunction_NullOffsetTokenInFormatV
...
+
+    // Resend a new batch should succeed even if there is an offset gap from the previous committed
+    // offset
+    gapRecords.remove(0);
+    service.insert(gapRecords);
+
+    TestUtils.assertWithRetry(
+        () -> service.getOffset(new TopicPartition(topic, PARTITION)) == 5, 20, 5);
+
+    assert TestUtils.tableSize(testTableName) == 3
```
sfc-gh-rcheng, line 764: shouldn't this == 4?

test/test_suit/test_schema_evolution_json_ignore_tombstone.py
```diff
@@ -65,7 +66,7 @@ def send(self):
...
```
sfc-gh-rcheng, line 69: nit: unneeded space

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelIT.java
```diff
@@ -706,4 +708,61 @@ public void testChannelMigrateOffsetTokenSystemFunction_NullOffsetTokenInFormatV
 
     service.closeAll();
   }
+
+  @Test
+  public void testInsertRowsWithGaps_schematization() throws Exception {
```
sfc-gh-rcheng, line 713: can we add a non schematization IT as well?

sfc-gh-rcheng (changes requested):
You're right that the previous E2E tests don't make sense given the offsets would be reset in Kafka, however I still think it is worth adding tests to cover that area. This issue was caused by Kafka unexpectedly dropping records, so we should at least add tests to cover that. We can try sending some entirely null records and see if the kafka producer accepts/drops them similar to the customer's behavior.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -163,6 +163,12 @@ public void insert(SinkRecord record) {
     pipes.get(nameIndex).insert(record);
   }
 
+  /** Not supported for Snowpipe */
```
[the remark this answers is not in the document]
sfc-gh-tzhang: This is outdated

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelIT.java
sfc-gh-tzhang: Please take a look at the test (I made some modification), the goal here is to test that the ingestion should be able to continue:
- If this is a new batch
- Even if there is an offset gap, or the resend batch's offset is not the same as the previous batch

test/test_suit/test_schema_evolution_json_ignore_tombstone.py
sfc-gh-tzhang: This is done by the auto formatter

sfc-gh-tzhang:
> You're right that the previous E2E tests don't make sense given the offsets would be reset in Kafka, however I still think it is worth adding tests to cover that area. This issue was caused by Kafka unexpectedly dropping records, so we should at least add tests to cover that. We can try sending some entirely null records and see if the kafka producer accepts/drops them similar to the customer's behavior.

Wouldn't this case covered by `testInsertRowsWithGaps_schematization`? Will add a new flavor for non-schematization

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelIT.java
sfc-gh-rcheng: Test looks good with the ==4, thanks!

sfc-gh-rcheng:
> Wouldn't this case covered by `testInsertRowsWithGaps_schematization`? Will add a new flavor for non-schematization

testInsertRowsWithGaps_schematization covers the changes made in this PR, but I don't think it covers the Kafka behavior. Using sinkServiceV2.insert() is not the same as putting data into the Kafka Broker and testing to ensure Kafka sends the correct offset and doesn't duplicate, drop or stop ingestion like the customer's environment did.

also you can close/reopen this PR to get snyk to run

sfc-gh-rcheng (approved):
Discussed offline - we do not have SMT testing to fully replicate the customer's issue, but this context should be added to warsaw's SMT testing jira

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -308,7 +315,8 @@ public void insert(SinkRecord record) {
     }
 
     TopicPartitionChannel channelPartition = partitionsToChannel.get(partitionChannelKey);
-    channelPartition.insertRecordToBuffer(record);
+    boolean isFirstPartitionRowInBatch = channelsVisitedPerBatch.add(partitionChannelKey);
```
sfc-gh-japatel, line 318: if this batch has records containing two different partitions. 
This will return true twice. 

the naming seems misleading IMO, shouldnt it be something like `isFirstRowForPartition`?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -378,8 +359,9 @@ public static String generateChannelNameFormatV2(
    * qualifies for being added into buffer.
    *
    * @param kafkaSinkRecord input record from Kafka
+   * @param isFirstRowInBatch indicates whether the given record is the first record in a batch
```
sfc-gh-japatel, line 362: nit: `first record per partition in batch sent by Kafka`

sfc-gh-japatel:
> You're right that the previous E2E tests don't make sense given the offsets would be reset in Kafka, however I still think it is worth adding tests to cover that area. This issue was caused by Kafka unexpectedly dropping records, so we should at least add tests to cover that. We can try sending some entirely null records and see if the kafka producer accepts/drops them similar to the customer's behavior.

+1

Was there any discussion on this?

sfc-gh-japatel (commented):
I dont have any blocking comments, please take a look at naming conventions. 
I agree with revi regarding adding a null test case. But I will defer to you on that.

sfc-gh-tzhang:
> I agree with revi regarding adding a null test case. But I will defer to you on that.

We can't generate a null record without SMT, added this in the SMT work that owns by the Warsaw team

sfc-gh-tzhang:
Looks like Snyk keeps having issues, close and reopen

PR #787, opened 2024-02-15
SNOW-1049869 Cleanup streaming ingest threads when SinkTask stop() is called

sfc-gh-mbobowski:
https://snowflakecomputing.atlassian.net/browse/SNOW-1049869

We clearly have a thread leak when SinkTask stop() method is called. I am not sure if it is the case for the close() method as well. It seems like the client is preserved on purpose during rebalance but maybe you have more context.

Additionally I added JUnit 5 to the project so we can gradually migrate. Since maven-surefire-plugin 2.2.0 configuration is very simple https://stackoverflow.com/questions/47158583/executing-junit-4-and-junit-5-tests-in-a-same-build

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
@@ -399,7 +399,14 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
   }
 
   @Override
-  public void setIsStoppedToTrue() {}
+  public void stop() {
+    try {
+      streamingIngestClient.close();
+    } catch (Exception e) {
+      LOGGER.warn("Could not close streaming ingest client");
+      throw SnowflakeErrors.ERROR_3009.getException(e);
+    }
+  }
 
   /* Undefined */
   @Override
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
@@ -403,8 +403,7 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
     try {
       streamingIngestClient.close();
     } catch (Exception e) {
-      LOGGER.warn("Could not close streaming ingest client");
-      throw SnowflakeErrors.ERROR_3009.getException(e);
+      LOGGER.warn("Could not close streaming ingest client. {}", e.getMessage());
     }
   }
 
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
@@ -401,7 +401,8 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
   @Override
   public void stop() {
     try {
-      streamingIngestClient.close();
+      StreamingClientProvider.getStreamingClientProviderInstance()
+          .closeClient(this.connectorConfig, this.streamingIngestClient);
     } catch (Exception e) {
       LOGGER.warn("Could not close streaming ingest client. {}", e.getMessage());
     }
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,7 @@
         <skipTests>false</skipTests>
         <skipUnitTests>${skipTests}</skipUnitTests>
         <skipIntegrationTests>${skipTests}</skipIntegrationTests>
+        <junit-jupiter.version>5.10.2</junit-jupiter.version>
     </properties>
 
 
@@ -445,6 +446,19 @@
         </dependency>
 
         <!--junit for unit test-->
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <version>${junit-jupiter.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <!--JUnit Jupiter Engine to depend on the JUnit4 engine and JUnit 4 API  -->
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <version>${junit-jupiter.version}</version>
+        </dependency>
+        <!-- Optional : override the JUnit 4 API version provided by junit-vintage-engine -->
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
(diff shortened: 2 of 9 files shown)


sfc-gh-tzhang:
> We clearly have a thread leak when SinkTask stop() method is called. I am not sure if it is the case for the close() method as well.

Do you mean that we use close() for rebalance and stop() for other cases?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -399,7 +399,14 @@ public void close(Collection<TopicPartition> partitions) {
   }
 
   @Override
-  public void setIsStoppedToTrue() {}
+  public void stop() {
+    try {
+      streamingIngestClient.close();
+    } catch (Exception e) {
+      LOGGER.warn("Could not close streaming ingest client");
```
sfc-gh-tzhang, line 406: Do we still need the log if we're throwing an exception?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -399,7 +399,14 @@ public void close(Collection<TopicPartition> partitions) {
...
+      throw SnowflakeErrors.ERROR_3009.getException(e);
```
sfc-gh-tzhang, line 407: What's the connector's behavior if we throw an exception here? Will the connector stop? Or the task will die and restart?

sfc-gh-tzhang (approved):
Left some minor comments, otherwise LGTM, thanks!

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-mbobowski: Good point.
I don't know what will happen with the connector (no source code available and docs remain silent about it) but it may cause some problems. The task is being stopped in this method so throwing an exception does not change much. I will leave only the warn log here.

sfc-gh-mbobowski:
> > We clearly have a thread leak when SinkTask stop() method is called. I am not sure if it is the case for the close() method as well.
> 
> Do you mean that we use close() for rebalance and stop() for other cases?

Yes. I was wondering if thread leak can happen during rebalance.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-mbobowski: Exception removed.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -399,7 +399,13 @@ public void close(Collection<TopicPartition> partitions) {
   }
 
   @Override
-  public void setIsStoppedToTrue() {}
+  public void stop() {
+    try {
+      streamingIngestClient.close();
```
sudeshwasnik, line 404: QQ - Why not call `closeAll()` here? It closes/clears couple of more resources ...

sfc-gh-japatel, line 404: I think we should use
`StreamingClientProvider.getStreamingClientProviderInstance()
        .closeClient(this.connectorConfig, this.streamingIngestClient);`

sfc-gh-tzhang: Good point! I don't fully understand how this would work with one client optimization, even if it's only used in `start()`, I thought it's possible that the client is used by channels from other tasks, then why we close the client during `start() `or `close()`, + @sfc-gh-rcheng as FYI

sfc-gh-japatel: Thats a good point, closeAll will also close Channels and clear the cache. 
close client is already closing all channels so I dont see any issue with using closeAll here. 
@sfc-gh-tzhang do you see any issues with closing channels first vs closing client directly?

sfc-gh-japatel:
> > > We clearly have a thread leak when SinkTask stop() method is called. I am not sure if it is the case for the close() method as well.
> > 
> > 
> > Do you mean that we use close() for rebalance and stop() for other cases?
> 
> Yes. I was wondering if thread leak can happen during rebalance.

during rebalancing, I dont think we should close client it is un-necessarily using de-allocating and realocating resources which takes up CPU. 
Yes, close is called during rebalance and stop for stopping the task. see https://kafka.apache.org/20/javadoc/org/apache/kafka/connect/sink/SinkTask.html#stop--

sfc-gh-japatel:
Also there seems to be some behavior change because the drop tests are failing everywhere. I think it's worth debugging and understand why. 
`Columns not in table:  ['case_sensitive_PERFORMANCE_CHAR', 'RATING_INT']`

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-tzhang: One extra thing if you call close channel first is that we will check and throw exception if not everything is committed; calling close on client directly will just mark all channels as closed. For KC I think it's fine to just close the client.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -399,7 +399,14 @@ public void close(Collection<TopicPartition> partitions) {
   }
 
   @Override
-  public void setIsStoppedToTrue() {}
+  public void stop() {
+    try {
+      StreamingClientProvider.getStreamingClientProviderInstance()
+          .closeClient(this.connectorConfig, this.streamingIngestClient);
```
sfc-gh-rcheng, line 405: the client can be shared across multiple sink connectors, so calling stop here may break or cause a new client to open on other connectors. 
Please try running this code with multiple connectors pointing to one client to understand the behavior. And/or write a test with multiple sinks pointing to one client as well

sfc-gh-mbobowski:
Closing the PR since it needs more analysis on the client sharing mechanism.

Enigma25:
> Closing the PR since it needs more analysis on the client sharing mechanism.

Hi @sfc-gh-mbobowski , just checking - if it would be too early to request for an ETA for this fix?

sfc-gh-mbobowski:
Hi, @Enigma25! Yes, it is more complicated that we initially expected.

ETA depends on business priorities and I am not the one to decide. Will do my best to push this topic forward.

sfc-gh-nsienkiewicz-1:
@Enigma25 the fix was merged recently in other PR: https://github.com/snowflakedb/snowflake-kafka-connector/pull/800

PR #788, opened 2024-02-16
Run subset of e2e tests

sfc-gh-mbobowski:
This PR enables running subset of e2e tests `./run_test_apache.sh 2.5.1 apache_properties false false --skipTests --tests=TestStringJson,TestAvroAvro` 

There is still huge space for improvement here:
- turn all parameters from non-named into named for both Shell and Python
- add decent named parameters parsing for Shell (like https://unix.stackexchange.com/questions/331522/how-do-i-parse-optional-arguments-in-a-bash-script-if-no-order-is-given)
- handle default arguments values
- skip unnecessary downloading for local development

Anyway let's do it step-by-step. This PR alone will save us hours in long term.

### DIFF
--- a/README-TEST.md
+++ b/README-TEST.md
@@ -58,9 +58,9 @@ Run end to end test on Apache 2.2.0 without stress test, with ssl
 	./run_test_apache.sh 2.2.0 ./apache_properties_ssl false true
 ```
 
-There is also option to run selected tests locally for Apache using additional named parameters:
+There is also option to run selected tests locally using additional named parameters:
 ```
-./run_test_apache.sh <version> <path to apache config folder> <stress> <ssl> --local=true --tests=TestStringJson,TestAvroAvro
+./run_test_<confluent/apache>.sh <version> <path to apache config folder> <stress> <ssl> --local=true --tests=TestStringJson,TestAvroAvro
 ```
 
 Note: if ssl is set to true, path to apache config folder must be `apache_properties_ssl` instead of `apache_properties`
--- a/README-TEST.md
+++ b/README-TEST.md
@@ -51,16 +51,15 @@ Checkout from GitHub master and run integration test:
 The above command build a jar file and put it at `/usr/local/share/kafka/plugins`. Then the following command will spin up Kafka and Kafka Connect cluster and execute the test suits. 
 
 ```
-./run_test_<confluent/apache>.sh <version> <path to apache config folder> <stress> <ssl>
+./run_test_<confluent/apache>.sh <version> <path to apache config folder> <stress> <ssl> [--skipProxy] [--tests=TestStringJson,TestAvroAvro,...]
 Run end to end test on Confluent 5.5.0 with stress test, without ssl
 	./run_test_confluent.sh 5.5.0 ./apache_properties true
 Run end to end test on Apache 2.2.0 without stress test, with ssl
 	./run_test_apache.sh 2.2.0 ./apache_properties_ssl false true
-```
-
-There is also option to run selected tests locally using additional named parameters:
-```
-./run_test_<confluent/apache>.sh <version> <path to apache config folder> <stress> <ssl> --local=true --tests=TestStringJson,TestAvroAvro
+	
+Optional arguments (useful for local development):
+    --skipProxy - skip running proxy tests
+    --tests - comma separated list of test suites to run
 ```
 
 Note: if ssl is set to true, path to apache config folder must be `apache_properties_ssl` instead of `apache_properties`
--- a/README-TEST.md
+++ b/README-TEST.md
@@ -58,6 +58,11 @@ Run end to end test on Apache 2.2.0 without stress test, with ssl
 	./run_test_apache.sh 2.2.0 ./apache_properties_ssl false true
 ```
 
+There is also option to run selected tests locally for Apache using additional named parameters:
+```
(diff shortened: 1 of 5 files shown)


test/run_test_apache.sh
```diff
@@ -32,6 +32,21 @@ if [ "$#" -gt 3 ] ; then
 else
   SSL="false"
 fi
+
+if [ "$#" -gt 4 ] && [[ $5 == "--local=true" ]] ; then
```
sfc-gh-mbobowski, line 36: see_no_evil

sfc-gh-lshcharbaty: Hm. Can we decouple parameters from their order? 
Like checking whether any parameter is `--local=true`, or even better, just `--local`.

test/run_test_apache.sh
```diff
@@ -32,6 +32,21 @@ if [ "$#" -gt 3 ] ; then
...
+  LOCAL=true
+else
+  LOCAL=false
+fi
+
+tests_pattern="[^(--tests=).*]"
+if [ "$#" -gt 5 ] && [[ $6 =~ $tests_pattern ]] ; then
```
sfc-gh-lshcharbaty, line 43: Same here, can we check the first parameter that conforms `tests_pattern`, not the 6th one?

README-TEST.md
```diff
@@ -58,6 +58,11 @@ Run end to end test on Apache 2.2.0 without stress test, with ssl
 	./run_test_apache.sh 2.2.0 ./apache_properties_ssl false true
 ```
 
+There is also option to run selected tests locally using additional named parameters:
+```
+./run_test_<confluent/apache>.sh <version> <path to apache config folder> <stress> <ssl> --local=true --tests=TestStringJson,TestAvroAvro
```
sfc-gh-lshcharbaty, line 63: Can you explain in the README what `--local` is responsible for?

test/run_test_confluent.sh
```diff
@@ -31,7 +31,22 @@ if [ "$#" -gt 3 ] ; then
   SSL=$4
 else
   SSL="false"
-fi 
+fi
+
+if [ "$#" -gt 4 ] && [[ $5 == "--local=true" ]] ; then
```
sfc-gh-lshcharbaty, line 36: The same as for the apache script.

test/run_test_confluent.sh
```diff
@@ -31,7 +31,22 @@ if [ "$#" -gt 3 ] ; then
...
+  LOCAL=true
+else
+  LOCAL=false
+fi
+
+tests_pattern="[^(--tests=).*]"
+if [ "$#" -gt 5 ] && [[ $6 =~ $tests_pattern ]] ; then
```
sfc-gh-lshcharbaty, line 43: The same as for the apache script.

sfc-gh-lshcharbaty (approved):

test/run_test_apache.sh
sfc-gh-mbobowski: Changed to `--skipProxy` to make it more descriptive

README-TEST.md
sfc-gh-mbobowski: Added information about optional flags

test/run_test_apache.sh
```diff
@@ -16,8 +16,8 @@ function random-string() {
 source ./utils.sh
 
 # check argument number
-if [ "$#" -lt 2 ] || [ "$#" -gt 4 ] ; then
-    error_exit "Usage: ./run_test.sh <version> <path to apache config folder> <pressure> <ssl>.  Aborting."
+if [ "$#" -lt 2 ] || [ "$#" -gt 6 ] ; then
```
sfc-gh-achyzy, line 19: I'm not sure you want to do it in this PR, but it would be great to improve the parameters hacking to a more recent and convenient parameter handling. Take a look at https://github.com/RenatGilmanov/shell-script-template for templates.

test/test_verify.py
```diff
@@ -564,10 +569,10 @@ def execution(testSet, testSuitList, testCleanEnableList, testSuitEnableList, dr
 
 
 if __name__ == "__main__":
-    if len(sys.argv) != 9:
+    if len(sys.argv) < 10:
```
sfc-gh-achyzy, line 572: Since `skipProxy` and `allowedTestsCsv` are optional won't it fail when I skip them (I'll have 9 elements in argv).

sfc-gh-achyzy (approved):
That is an excellent step format. 
Please consider going even further by redesigning the scripts to a more convenient format.

test/run_test_apache.sh
sfc-gh-mbobowski: It's doable using this: https://unix.stackexchange.com/questions/331522/how-do-i-parse-optional-arguments-in-a-bash-script-if-no-order-is-given
However it requires all parameters to be named. It would make this PR too big imo, so I would merge this PR first, then switch to named parameters and then add loop to parse it.

test/test_verify.py
sfc-gh-mbobowski: skipProxy is set to default value "false" so it is not optional from the `test_verify.py` perspective. But you could be mislead by comment so I marked `skipProxy` as required here.

sfc-gh-japatel (approved):
lgtm! Thank you

PR #791, opened 2024-02-26
DO NOT MERGE - Expose max channel size in bytes via kafka connector

sfc-gh-kgaputis:
# Overview

SNOW-XXXXX

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -121,6 +121,10 @@ public class SnowflakeSinkConnectorConfig {
   public static final String SNOWPIPE_STREAMING_MAX_CLIENT_LAG =
       "snowflake.streaming.max.client.lag";
 
+  // This is the streaming max client lag which can be defined in config
+  public static final String SNOWPIPE_STREAMING_MAX_CHANNEL_SIZE =
+          "snowflake.streaming.max.channel.size";
+
   // TESTING
   public static final String REBALANCING = "snowflake.test.rebalancing";
   public static final boolean REBALANCING_DEFAULT = false;
@@ -551,6 +555,17 @@ public class SnowflakeSinkConnectorConfig {
             6,
             ConfigDef.Width.NONE,
             SNOWPIPE_STREAMING_MAX_CLIENT_LAG)
+        .define(
+            SNOWPIPE_STREAMING_MAX_CHANNEL_SIZE,
+            Type.LONG,
+            StreamingUtils.STREAMING_BUFFER_MAX_CHANNEL_SIZE_DEFAULT_BYTES,
+            ConfigDef.Range.atLeast(StreamingUtils.STREAMING_BUFFER_MAX_CHANNEL_SIZE_MINIMUM_BYTES),
+            Importance.LOW,
+            "Specify the maximum size (in bytes) of a single channel flush performed by the Ingest SDK",
+            CONNECTOR_CONFIG,
+            7, //TODO verify what this should be
+            ConfigDef.Width.NONE,
+            SNOWPIPE_STREAMING_MAX_CHANNEL_SIZE)
         .define(
             ERRORS_TOLERANCE_CONFIG,
             Type.STRING,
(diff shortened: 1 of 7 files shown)


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -121,6 +121,10 @@ public class SnowflakeSinkConnectorConfig {
   public static final String SNOWPIPE_STREAMING_MAX_CLIENT_LAG =
       "snowflake.streaming.max.client.lag";
 
+  // This is the streaming max client lag which can be defined in config
```
sfc-gh-tjones, line 124: Nit, `streaming max channel size (in bytes)`

sfc-gh-japatel (commented):
hey @sfc-gh-kgaputis thanks a lot for contribution
we would rather want to make this a generic config. 

https://snowflakecomputing.atlassian.net/browse/SNOW-1117446

TLDR:
instead of new KC config for every new/future config in streaming client, we can just use a key value pair 
"snowflake.streaming.client.provider.properties.override":"max_client_lag:5000,key:value"


Key should be present in https://github.com/snowflakedb/snowflake-ingest-java/blob/master/src/main/java/net/snowflake/ingest/utils/ParameterProvider.java

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -551,6 +555,17 @@ static ConfigDef newConfigDef() {
...
             SNOWPIPE_STREAMING_MAX_CLIENT_LAG)
+        .define(
+            SNOWPIPE_STREAMING_MAX_CHANNEL_SIZE,
+            Type.LONG,
+            StreamingUtils.STREAMING_BUFFER_MAX_CHANNEL_SIZE_DEFAULT_BYTES,
+            ConfigDef.Range.atLeast(StreamingUtils.STREAMING_BUFFER_MAX_CHANNEL_SIZE_MINIMUM_BYTES),
+            Importance.LOW,
+            "Specify the maximum size (in bytes) of a single channel flush performed by the Ingest SDK",
+            CONNECTOR_CONFIG,
+            7, //TODO verify what this should be
```
sfc-gh-tzhang, line 566: I'm not sure how important this is, but there is a 7 below already, see ENABLE_STREAMING_CLIENT_OPTIMIZATION_CONFIG, you can put it in the end and use a new number just in case

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java
```diff
@@ -45,6 +46,14 @@ public class StreamingUtils {
   public static final long STREAMING_BUFFER_FLUSH_TIME_DEFAULT_SEC =
       Duration.ofSeconds(10).getSeconds();
 
+
+  //TODO decide if this makes sense
```
sfc-gh-tzhang, line 50: Yeah, let's do 32MB

sfc-gh-tzhang (approved):
LGTM, thanks! Please do a quick test with your change to make sure the file size do increase, you can check the file size in the log

sfc-gh-japatel:
I am working on a PR for a map of key value pair, please hold on to this PR, but we appreciate the contribution!

sfc-gh-japatel (changes requested):
left a comment!

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-kgaputis: The name seemed a bit long but I changed config to `snowflake.streaming.max.channel.size.in.bytes` to be more explicit.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-kgaputis: Updated to 10, which was unused in the `CONNECTOR_CONFIG` group.

sfc-gh-kgaputis:
Thank you for reviewing team.   I like the plan @sfc-gh-japatel  and @sfc-gh-tzhang to make all the low level SDK settings overridable from a single param.

PR #793, opened 2024-02-27
fix: fips compatible version

enzo-cappa:
Avoid using the Ingestion SDK shaded JAR, as it's not FIPS compatible. Also use the FIPS JDBC connector.


enzo-cappa:
Created by mistake

PR #794, opened 2024-02-27
SNOW-1117446 Snowpipe Streaming client provider map

sfc-gh-japatel:
SNOW-1117446

Instead of having a new property in KC config for every overridden key value pair, we input a comma separated KV pair. 

## Pre-review checklist
- [ ] Snowpipe Changes
- [X] Snowpipe Streaming Changes
    - [X] `Yes` - Added end to end and Unit Tests. `snowflake.streaming.client.provider.override.map`

## Urgency
This review is ***URGENT*** priority

## Backward and forward Compatible
Imagine customer upgrading to new version and rolling back to older version. Will there be any concerns?
[X] Backward compatible - They will have to remove the new config since we dont validate this new config in older versions. 
[ ] Forward compatible
-->

Reviewers:  Every review must have at least two reviewers for bug fixes, GA'ed component. One reviewer is enough for test only, doc changes.
- Minimum # of Required Reviewers - **Two** for Improvements and Bugfixes - Toby, Tyler, Revi
- Educational purposes - @ Warsaw Team

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -18,7 +18,6 @@ package com.snowflake.kafka.connector;
 
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableSet;
-import com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.CommaSeparatedKeyValueValidator;
 import com.snowflake.kafka.connector.internal.KCLogger;
 import com.snowflake.kafka.connector.internal.streaming.IngestionMethodConfig;
 import com.snowflake.kafka.connector.internal.streaming.StreamingUtils;
@@ -122,7 +121,7 @@ public class SnowflakeSinkConnectorConfig {
   public static final String SNOWPIPE_STREAMING_MAX_CLIENT_LAG =
       "snowflake.streaming.max.client.lag";
 
-  public static final String SNOWPIPE_STREAMING_CLIENT_PARAMETER_OVERRIDE_MAP =
+  public static final String SNOWPIPE_STREAMING_CLIENT_PROVIDER_OVERRIDE_MAP =
       "snowflake.streaming.client.provider.override.map";
 
   // TESTING
@@ -135,7 +134,7 @@ public class SnowflakeSinkConnectorConfig {
   private static final ConfigDef.Validator topicToTableValidator = new TopicToTableValidator();
   private static final ConfigDef.Validator KAFKA_PROVIDER_VALIDATOR = new KafkaProviderValidator();
 
-  private static final ConfigDef.Validator STREAMING_CLIENT_PROVIDER_PROPERTIES_VALIDATOR =
+  private static final ConfigDef.Validator STREAMING_CLIENT_PROVIDER_OVERRIDE_MAP_VALIDATOR =
       new CommaSeparatedKeyValueValidator();
 
   // For error handling
@@ -559,19 +558,19 @@ public class SnowflakeSinkConnectorConfig {
             ConfigDef.Width.NONE,
             SNOWPIPE_STREAMING_MAX_CLIENT_LAG)
         .define(
-            SNOWPIPE_STREAMING_CLIENT_PARAMETER_OVERRIDE_MAP,
+            SNOWPIPE_STREAMING_CLIENT_PROVIDER_OVERRIDE_MAP,
             Type.STRING,
             "",
-            STREAMING_CLIENT_PROVIDER_PROPERTIES_VALIDATOR,
+            STREAMING_CLIENT_PROVIDER_OVERRIDE_MAP_VALIDATOR,
             Importance.LOW,
             "Map of Key value pairs representing Streaming Client Properties to Override. These are"
                 + " optional and recommended to use ONLY after consulting Snowflake Support. Format"
                 + " : comma-separated tuples, e.g.:"
-                + " MAX_CHANNEL_SIZE_IN_BYTES:10000000,MAX_CLIENT_LAG:5000,ENABLE_SNOWPIPE_STREAMING_JMX_METRICS:false...",
+                + " MAX_CLIENT_LAG:5000,other_key:value...",
             CONNECTOR_CONFIG,
             0,
             ConfigDef.Width.NONE,
-            SNOWPIPE_STREAMING_CLIENT_PARAMETER_OVERRIDE_MAP)
+            SNOWPIPE_STREAMING_CLIENT_PROVIDER_OVERRIDE_MAP)
         .define(
             ERRORS_TOLERANCE_CONFIG,
             Type.STRING,
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -564,9 +564,9 @@ public class SnowflakeSinkConnectorConfig {
             "",
             STREAMING_CLIENT_PROVIDER_PROPERTIES_VALIDATOR,
             Importance.LOW,
-            "Map of Key value pairs representing Streaming Client Properties to Override (optional"
-                + " and recommended to use only after consulting Snowflake Support). Format :"
-                + " comma-separated tuples, e.g.:"
+            "Map of Key value pairs representing Streaming Client Properties to Override. These are"
+                + " optional and recommended to use ONLY after consulting Snowflake Support. Format"
+                + " : comma-separated tuples, e.g.:"
                 + " MAX_CHANNEL_SIZE_IN_BYTES:10000000,MAX_CLIENT_LAG:5000,ENABLE_SNOWPIPE_STREAMING_JMX_METRICS:false...",
             CONNECTOR_CONFIG,
             0,
@@ -847,11 +847,9 @@ public class SnowflakeSinkConnectorConfig {
 
     public void ensureValid(String name, Object value) {
       String s = (String) value;
-      if (s != null && !s.isEmpty()) {
-        // Validate the comma-separated key-value pairs string
-        if (!isValidCommaSeparatedKeyValueString(s)) {
-          throw new ConfigException(name, value, "Format: <key-1>:<value-1>,<key-2>:<value-2>,...");
-        }
+      // Validate the comma-separated key-value pairs string
+      if (s != null && !s.isEmpty() && !isValidCommaSeparatedKeyValueString(s)) {
+        throw new ConfigException(name, value, "Format: <key-1>:<value-1>,<key-2>:<value-2>,...");
       }
     }
 
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -18,6 +18,7 @@ package com.snowflake.kafka.connector;
 
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableSet;
+import com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.CommaSeparatedKeyValueValidator;
 import com.snowflake.kafka.connector.internal.KCLogger;
 import com.snowflake.kafka.connector.internal.streaming.IngestionMethodConfig;
 import com.snowflake.kafka.connector.internal.streaming.StreamingUtils;
@@ -121,6 +122,9 @@ public class SnowflakeSinkConnectorConfig {
   public static final String SNOWPIPE_STREAMING_MAX_CLIENT_LAG =
       "snowflake.streaming.max.client.lag";
 
+  public static final String SNOWPIPE_STREAMING_CLIENT_PARAMETER_OVERRIDE_MAP =
+      "snowflake.streaming.client.provider.override.map";
+
   // TESTING
   public static final String REBALANCING = "snowflake.test.rebalancing";
   public static final boolean REBALANCING_DEFAULT = false;
@@ -130,6 +134,9 @@ public class SnowflakeSinkConnectorConfig {
   private static final ConfigDef.Validator nonEmptyStringValidator = new ConfigDef.NonEmptyString();
   private static final ConfigDef.Validator topicToTableValidator = new TopicToTableValidator();
   private static final ConfigDef.Validator KAFKA_PROVIDER_VALIDATOR = new KafkaProviderValidator();
+
+  private static final ConfigDef.Validator STREAMING_CLIENT_PROVIDER_PROPERTIES_VALIDATOR =
+      new CommaSeparatedKeyValueValidator();
 
   // For error handling
   public static final String ERROR_GROUP = "ERRORS";
@@ -552,6 +559,20 @@ public class SnowflakeSinkConnectorConfig {
             ConfigDef.Width.NONE,
             SNOWPIPE_STREAMING_MAX_CLIENT_LAG)
         .define(
+            SNOWPIPE_STREAMING_CLIENT_PARAMETER_OVERRIDE_MAP,
+            Type.STRING,
+            "",
+            STREAMING_CLIENT_PROVIDER_PROPERTIES_VALIDATOR,
+            Importance.LOW,
+            "Map of Key value pairs representing Streaming Client Properties to Override (optional"
+                + " and recommended to use only after consulting Snowflake Support). Format :"
+                + " comma-separated tuples, e.g.:"
+                + " MAX_CHANNEL_SIZE_IN_BYTES:10000000,MAX_CLIENT_LAG:5000,ENABLE_SNOWPIPE_STREAMING_JMX_METRICS:false...",
+            CONNECTOR_CONFIG,
+            0,
+            ConfigDef.Width.NONE,
+            SNOWPIPE_STREAMING_CLIENT_PARAMETER_OVERRIDE_MAP)
+        .define(
             ERRORS_TOLERANCE_CONFIG,
             Type.STRING,
             ERRORS_TOLERANCE_DEFAULT,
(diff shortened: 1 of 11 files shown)


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -551,6 +558,20 @@ static ConfigDef newConfigDef() {
...
             ConfigDef.Width.NONE,
             SNOWPIPE_STREAMING_MAX_CLIENT_LAG)
+        .define(
+            SNOWPIPE_STREAMING_CLIENT_PARAMETER_OVERRIDE_MAP,
+            Type.STRING,
+            "",
+            STREAMING_CLIENT_PROVIDER_PROPERTIES_VALIDATOR,
+            Importance.LOW,
+            "Map of Key value pairs representing Streaming Client Properties to Override (optional"
+                + " and recommended to use only after consulting Snowflake Support). Format :"
```
sfc-gh-tjones, line 568: Nit, I'd remove the parens and change to `Override. These are optional and recommend to use ONLY after consulting with Snowflake Support.`

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -815,4 +836,47 @@ public void ensureValid(String name, Object value) {
...
+    public CommaSeparatedKeyValueValidator() {}
+
+    public void ensureValid(String name, Object value) {
+      String s = (String) value;
+      if (s != null && !s.isEmpty()) {
+        // Validate the comma-separated key-value pairs string
+        if (!isValidCommaSeparatedKeyValueString(s)) {
+          throw new ConfigException(name, value, "Format: <key-1>:<value-1>,<key-2>:<value-2>,...");
+        }
+      }
```
sfc-gh-tjones, line 855: Nit can this be 
```
if (s != null && !s.isEmpty && !isValidCommaSeparatedKeyValueString(s)) {
    throw new ConfigException(name, value, "Format: <key-1>:<value-1>,<key-2>:<value-2>,...");
}
```

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -727,6 +736,34 @@ public static Map<String, String> parseTopicToTableMap(String input) {
...
+    for (String str : input.split(",")) {
+      String[] tt = str.split(":");
+
+      if (tt.length != 2 || tt[0].trim().isEmpty() || tt[1].trim().isEmpty()) {
+        LOGGER.error(
+            "Invalid {} config format: {}",
+            SnowflakeSinkConnectorConfig.SNOWPIPE_STREAMING_CLIENT_PARAMETER_OVERRIDE_MAP,
+            input);
+        isInvalid = true;
+        break;
```
sfc-gh-tjones, line 757: Isn't this validated above?

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -727,6 +736,34 @@ public static Map<String, String> parseTopicToTableMap(String input) {
...
+      }
+      pairs.put(tt[0].trim(), tt[1].trim());
+    }
+    if (isInvalid) {
+      throw SnowflakeErrors.ERROR_0030.getException();
+    }
```
sfc-gh-tjones, line 763: Why not just throw above? Then you don't need to pass around `isInvalid`

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
```diff
@@ -84,10 +85,48 @@ public StreamingClientProperties(Map<String, String> connectorConfig) {
...
           parameterOverrides.put(MAX_CLIENT_LAG, String.format("%s second", overriddenValue));
         });
+    combineStreamingClientOverriddenProperties(connectorConfig);
+  }
+
+  private void combineStreamingClientOverriddenProperties(Map<String, String> connectorConfig) {
+    // MAX_CLIENT_LAG property can also be present in override map, in that case, we will honor the
+    // value
+    // provided in SNOWPIPE_STREAMING_MAX_CLIENT_LAG since it preceded and was released in earlier
+    // versions.
```
sfc-gh-tjones, line 97: Nit, this looks poorly formatted.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
```diff
@@ -84,10 +85,48 @@ public StreamingClientProperties(Map<String, String> connectorConfig) {
...
+    Optional<String> clientOverrideProperties =
+        Optional.ofNullable(connectorConfig.get(SNOWPIPE_STREAMING_CLIENT_PARAMETER_OVERRIDE_MAP));
+    Map<String, String> clientOverridePropertiesMap = new HashMap<>();
+    clientOverrideProperties.ifPresent(
+        overriddenKeyValuePairs -> {
+          LOGGER.info(
+              "Overridden map provided for streaming client Properties: {}",
```
sfc-gh-tjones, line 104: nit, does `Properties` need to be lower cased here?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
```diff
@@ -84,10 +85,48 @@ public StreamingClientProperties(Map<String, String> connectorConfig) {
...
+              overriddenKeyValuePairs);
+          Map<String, String> overriddenPairs =
+              Utils.parseCommaSeparatedKeyValuePairs(overriddenKeyValuePairs);
+          overriddenPairs.forEach(
+              (overriddenKey, overriddenValue) -> {
+                if (overriddenKey.equalsIgnoreCase(MAX_CLIENT_LAG)) {
+                  clientOverridePropertiesMap.put(
+                      MAX_CLIENT_LAG, String.format("%s second", overriddenValue));
```
sfc-gh-tjones, line 112: What is the significance of `second` here?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
sfc-gh-japatel: oops, java formatter didnt run on these files. will push

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
```diff
@@ -84,10 +85,48 @@ public StreamingClientProperties(Map<String, String> connectorConfig) {
...
+    clientOverrideProperties.ifPresent(
+        overriddenKeyValuePairs -> {
+          LOGGER.info(
+              "Overridden map provided for streaming client Properties: {}",
+              overriddenKeyValuePairs);
+          Map<String, String> overriddenPairs =
+              Utils.parseCommaSeparatedKeyValuePairs(overriddenKeyValuePairs);
+          overriddenPairs.forEach(
+              (overriddenKey, overriddenValue) -> {
+                if (overriddenKey.equalsIgnoreCase(MAX_CLIENT_LAG)) {
```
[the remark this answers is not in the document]
sfc-gh-japatel: because this is already its own property.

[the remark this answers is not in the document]
sfc-gh-tjones: We should probably call out somewhere in the docstring for the new override option that `MAX_CLIENT_LAG` is special cased.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
sfc-gh-tjones: Got it - I was wondering if there was a way then to special case in one place it rather than doing it when iterating over keys and then later after iterating all keys.

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-japatel: Agreed, but both the verifications are in separate code path. And I would like to keep both. 
So two code paths are:
1. connector start
2. task start

Task wont start if connector hasnt started but since we cant control KC framework, its good to have double verification, once during creation of connector and once during creation of streaming client.

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-japatel: will do!

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-tjones: That makes sense - would be great if there was just a single verification step but if its possible for something to fail at a later step for a valid reason we should probably just do it twice

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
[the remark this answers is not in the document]
sfc-gh-japatel: okay, I see you are talking about case sensitivity here. i thought you are talking about MAX_CLIENT_LAG's special `case` here

I checked ingest java, we will have to pass in all keys as lower case. https://github.com/snowflakedb/snowflake-ingest-java/blob/master/src/main/java/net/snowflake/ingest/utils/ParameterProvider.java#L204

I will pass in lowercase here.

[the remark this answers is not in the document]
sfc-gh-japatel: > We should probably call out somewhere in the docstring for the new override option that MAX_CLIENT_LAG is special cased.

This is not needed since we will always pass in a lowercase value in streaming provider regardless of what;s the input key.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
sfc-gh-japatel: umm, so this is a special case because:
1. `MAX_CLIENT_LAG` is a constant fetched from ingest sdk which is lowercase itself. In other words, this is a derived property from input snowflake.streaming.max.client.lag
2. New config added here is all customer controlled and we will always have to parse lowercase. 

Let me know if you want to discuss offline, I can add this in javadocs.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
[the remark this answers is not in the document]
sfc-gh-tjones: Oh I wasn't talking about upper or lower case, I was referring to that `MAX_CLIENT_LAG` is treated differently. We might want to tweak the documentation to reflect that and explain how that parameter and this new one impact each other.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
sfc-gh-tjones: See my comment above but TL;DR: let's add this to the Java Doc

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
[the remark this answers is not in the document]
sfc-gh-japatel: yes, I added javadocs with some examples.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
sfc-gh-japatel: done!

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
```diff
@@ -84,10 +85,76 @@ public StreamingClientProperties(Map<String, String> connectorConfig) {
         Optional.ofNullable(connectorConfig.get(SNOWPIPE_STREAMING_MAX_CLIENT_LAG));
     snowpipeStreamingMaxClientLag.ifPresent(
         overriddenValue -> {
-          LOGGER.info(
-              "Config is overridden for {}={}", SNOWPIPE_STREAMING_MAX_CLIENT_LAG, overriddenValue);
           parameterOverrides.put(MAX_CLIENT_LAG, String.format("%s second", overriddenValue));
         });
+    combineStreamingClientOverriddenProperties(connectorConfig);
```
sfc-gh-tzhang, line 90: Do you know if we log these overrides and send to SF? If not, could we add that?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
```diff
@@ -84,10 +85,76 @@ public StreamingClientProperties(Map<String, String> connectorConfig) {
...
+                "Honoring {} value in MAX_CLIENT_LAG for streaming client provider, since it is"
+                    + " explicitly provided. Using: {}",
+                SNOWPIPE_STREAMING_MAX_CLIENT_LAG,
+                parameterOverrides.get(MAX_CLIENT_LAG));
+            clientOverridePropertiesMap.remove(MAX_CLIENT_LAG);
+          }
+          parameterOverrides.putAll(clientOverridePropertiesMap);
+        });
+    parameterOverrides.forEach(
+        (key, value) -> LOGGER.info("Streaming Client Config is overridden for {}={}", key, value));
```
sfc-gh-tzhang, line 157: If we just log the map then it can be in one log line?

sfc-gh-tzhang (commented):
Left some comments, PTAL! Major comment is about whether we could include the overrides as part of the telemetry send to server side.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -121,6 +122,9 @@ public class SnowflakeSinkConnectorConfig {
   public static final String SNOWPIPE_STREAMING_MAX_CLIENT_LAG =
       "snowflake.streaming.max.client.lag";
 
+  public static final String SNOWPIPE_STREAMING_CLIENT_PARAMETER_OVERRIDE_MAP =
+      "snowflake.streaming.client.provider.override.map";
```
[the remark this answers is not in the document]
sfc-gh-japatel: i actually meant snowflake.streaming.client.provider.override.map but I can change it to `snowflake.streaming.client.override.map`

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -131,6 +135,9 @@ public class SnowflakeSinkConnectorConfig {
   private static final ConfigDef.Validator topicToTableValidator = new TopicToTableValidator();
   private static final ConfigDef.Validator KAFKA_PROVIDER_VALIDATOR = new KafkaProviderValidator();
 
+  private static final ConfigDef.Validator STREAMING_CLIENT_PROVIDER_PROPERTIES_VALIDATOR =
```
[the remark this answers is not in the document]
sfc-gh-japatel: ack, will unify!

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -815,4 +836,45 @@ public void ensureValid(String name, Object value) {
...
+   *
+   * <p>It doesn't validate the type of values, only making sure the format is correct.
+   */
+  public static class CommaSeparatedKeyValueValidator implements ConfigDef.Validator {
+    public CommaSeparatedKeyValueValidator() {}
+
+    public void ensureValid(String name, Object value) {
+      String s = (String) value;
+      // Validate the comma-separated key-value pairs string
+      if (s != null && !s.isEmpty() && !isValidCommaSeparatedKeyValueString(s)) {
```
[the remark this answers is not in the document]
sfc-gh-japatel: well it can, but this order signifies that it can be null or empty. we want to check only if its non null and non empty. i prefer it here for readability.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
sfc-gh-japatel: good point, nope. this wasnt added in max client lag either. 
I can add both. thanks for call out

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
sfc-gh-japatel: yes, let me do that. infact, i have 

`Overridden map provided for streaming client properties: {}`  earlier. so I can remove this.

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -727,6 +736,29 @@ public static Map<String, String> parseTopicToTableMap(String input) {
...
+    Map<String, String> pairs = new HashMap<>();
+    for (String str : input.split(",")) {
+      String[] tt = str.split(":");
+
+      if (tt.length != 2 || tt[0].trim().isEmpty() || tt[1].trim().isEmpty()) {
+        LOGGER.error(
+            "Invalid {} config format: {}",
+            SnowflakeSinkConnectorConfig.SNOWPIPE_STREAMING_CLIENT_PROVIDER_OVERRIDE_MAP,
+            input);
+        throw SnowflakeErrors.ERROR_0030.getException();
```
sfc-gh-tzhang, line 755: I think we should put the map as part of the exception instead of logging it? I remember we have customer complains about the needs of reading through the log to find the input

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
sfc-gh-tzhang: sounds good, please remove

sfc-gh-tzhang (approved):
LGTM, left a minor comment, thanks!

sfc-gh-japatel:
@sfc-gh-tjones Can you take a look again? I would want to get this out before next Kc release! Thanks

sfc-gh-tjones (approved):

PR #795, opened 2024-02-28
SNOW-979849: Add offset verification logic

sfc-gh-tzhang:
# Overview

SNOW-979849

Add offset verification logic as part of channel open. We're verifying that the current start offset = previous end offset +1, if not true, then it potentially means missing or duplicate data. Note that there are some false positives that we can't avoid like when SMT is used, but this is good enough to do blast radius analysis for any corruption issue. 

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java
@@ -1,5 +1,24 @@
 package com.snowflake.kafka.connector.internal.streaming;
 
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
+import com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig;
+import com.snowflake.kafka.connector.Utils;
+import com.snowflake.kafka.connector.internal.BufferThreshold;
+import net.snowflake.ingest.streaming.OffsetTokenVerificationFunction;
+import net.snowflake.ingest.utils.Constants;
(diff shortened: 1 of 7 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java
```diff
@@ -75,6 +76,27 @@ public class StreamingUtils {
...
+        if (prevBatchEndOffset != null) {
+          try {
+            long curStart = Long.parseLong(curBatchStartOffset);
+            long prevEnd = Long.parseLong(prevBatchEndOffset);
+
+            if (curStart != prevEnd + 1) {
+              isMatch = false;
+            }
+          } catch (NumberFormatException ignored) {
+            isMatch = false;
```
sfc-gh-japatel, line 93: do we anticipate this happening? I would rather throw an exception since we know this is not happening. WDYT?

sfc-gh-japatel (commented):
i know we have the exact same function in ingest sdk and may be not having a test is okay, but I would still like to see some basic test.

sfc-gh-tzhang:
> i know we have the exact same function in ingest sdk and may be not having a test is okay, but I would still like to see some basic test.
> 
> Thanks for adding this though, this will help a lot!

It's not exactly the same, I didn't check for the rowcount since i'm afraid we get too many false positives due to SMT. Added a small test, PTAL.

sfc-gh-tzhang:
ping, can I get a review on this? Thanks!

sfc-gh-japatel (approved):
lgtm! Thank you

PR #800, opened 2024-03-08
SNOW-1049869 Cleanup streaming ingest threads when SinkTask stop() is called

sfc-gh-mbobowski:
https://snowflakecomputing.atlassian.net/browse/SNOW-1049869

KC does not call `close()` method on streaming ingest client instance when `stop()` is called on task restart or termination. It results in some internal threads not being closed and probably a memory leak as well. 

If one client optimization is enabled it is not a big deal. Client will be reused on task restart and will leak only on task termination which is rare.  Max 10000 clients are cached so no more threads will leak.

When client optimization is disabled then we will experience a massive thread leak as new streaming ingest client is created on every task restart.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -463,6 +463,7 @@
             <groupId>org.junit.vintage</groupId>
             <artifactId>junit-vintage-engine</artifactId>
             <version>${junit-jupiter.version}</version>
+            <scope>test</scope>
         </dependency>
         <!-- Optional : override the JUnit 4 API version provided by junit-vintage-engine -->
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,7 @@
         <skipTests>false</skipTests>
         <skipUnitTests>${skipTests}</skipUnitTests>
         <skipIntegrationTests>${skipTests}</skipIntegrationTests>
+        <junit-jupiter.version>5.10.2</junit-jupiter.version>
     </properties>
 
 
@@ -445,6 +446,25 @@
         </dependency>
 
         <!--junit for unit test-->
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <version>${junit-jupiter.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <version>${junit-jupiter.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <!--JUnit Jupiter Engine to depend on the JUnit4 engine and JUnit 4 API  -->
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <version>${junit-jupiter.version}</version>
+        </dependency>
+        <!-- Optional : override the JUnit 4 API version provided by junit-vintage-engine -->
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -49,7 +49,6 @@
         <skipTests>false</skipTests>
         <skipUnitTests>${skipTests}</skipUnitTests>
         <skipIntegrationTests>${skipTests}</skipIntegrationTests>
-        <junit-jupiter.version>5.10.2</junit-jupiter.version>
     </properties>
 
 
@@ -449,20 +448,17 @@
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-engine</artifactId>
-            <version>${junit-jupiter.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-params</artifactId>
-            <version>${junit-jupiter.version}</version>
             <scope>test</scope>
         </dependency>
         <!--JUnit Jupiter Engine to depend on the JUnit4 engine and JUnit 4 API  -->
         <dependency>
             <groupId>org.junit.vintage</groupId>
             <artifactId>junit-vintage-engine</artifactId>
-            <version>${junit-jupiter.version}</version>
             <scope>test</scope>
         </dependency>
         <!-- Optional : override the JUnit 4 API version provided by junit-vintage-engine -->
@@ -512,4 +508,16 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.junit</groupId>
+                <artifactId>junit-bom</artifactId>
+                <version>5.10.2</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
 </project>
--- a/pom_confluent.xml
+++ b/pom_confluent.xml
@@ -49,7 +49,6 @@
         <skipTests>false</skipTests>
         <skipUnitTests>${skipTests}</skipUnitTests>
         <skipIntegrationTests>${skipTests}</skipIntegrationTests>
-        <junit-jupiter.version>5.10.2</junit-jupiter.version>
         <license.processing.dependencyJarsDir>${project.build.directory}/dependency-jars
         </license.processing.dependencyJarsDir>
         <license.processing.dependencyListFile>${project.build.directory}/dependency-list.txt
@@ -660,4 +659,16 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.junit</groupId>
+                <artifactId>junit-bom</artifactId>
+                <version>5.10.2</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
 </project>
--- a/pom_confluent.xml
+++ b/pom_confluent.xml
@@ -596,20 +596,17 @@
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-engine</artifactId>
-            <version>${junit-jupiter.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-params</artifactId>
-            <version>${junit-jupiter.version}</version>
             <scope>test</scope>
         </dependency>
         <!--JUnit Jupiter Engine to depend on the JUnit4 engine and JUnit 4 API  -->
         <dependency>
             <groupId>org.junit.vintage</groupId>
             <artifactId>junit-vintage-engine</artifactId>
-            <version>${junit-jupiter.version}</version>
             <scope>test</scope>
         </dependency>
         <!-- Optional : override the JUnit 4 API version provided by junit-vintage-engine -->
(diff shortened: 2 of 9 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -1,7 +1,6 @@
 package com.snowflake.kafka.connector.internal.streaming;
 
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.BUFFER_SIZE_BYTES_DEFAULT;
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWFLAKE_ROLE;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.*;
```
sfc-gh-rcheng, line 3: disable star imports please, we want to import explicitly

sfc-gh-rcheng (approved):

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -399,7 +398,20 @@ public void close(Collection<TopicPartition> partitions) {
...
-  public void setIsStoppedToTrue() {}
+  public void stop() {
+    final boolean isOptimizationEnabled =
+        Boolean.parseBoolean(
+            connectorConfig.getOrDefault(
+                SnowflakeSinkConnectorConfig.ENABLE_STREAMING_CLIENT_OPTIMIZATION_CONFIG,
+                Boolean.toString(ENABLE_STREAMING_CLIENT_OPTIMIZATION_DEFAULT)));
+    if (!isOptimizationEnabled) {
+      try {
+        this.streamingIngestClient.close();
```
sfc-gh-japatel, line 409: lets call the `StreamingClientProvider.getStreamingClientProviderInstance().close()`

sfc-gh-japatel (approved):
lgtm!

Although I think we can still close it even if one client is enabled. I just checked the code, each connector instance will have its own StreamingClientProperties because it will have its own name. The client cache itself will have separate entries. So I think we can still close it one connector client without stepping foot on other one. 

It would be good to test this scenario
1. create two connectors and enable streaming client optimization. - should have two clients.
2. create two connectors and disable streaming client optimization. - should have two clients. 

Unfortunately we dont have a good test setup with multiple connectors.
@sfc-gh-rcheng to confirm the theory. 


not blocking this PR since this should go out but lets file a jira and have uniform behavior. 


Thank you for fixing!

sfc-gh-rcheng:
> each connector instance will have its own StreamingClientProperties because it will have its own name

[StreamingClientProperties](https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java#L94) contains just the authentication information for the client, the[ client name is not used for compariso](https://github.com/snowflakedb/snowflake-kafka-connector/blob/5247f6e2a1f7b869c9328fffd47efb6408908936/src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java#L190)n across StreamingClientProperties. This means that multiple sinks with different names could use the same client. I don't think we can close the client with the optimization enabled because i'm not sure what the behavior would be to close the client during a flush.

But I agree we should test it out, made a couple adjustments to @sfc-gh-japatel 's recommendations:
create two connectors with same StreamingClientProperties and enable streaming client optimization. - should have one clients.
create two connectors and disable streaming client optimization. - should have two clients.

pom.xml
```diff
@@ -445,6 +446,25 @@
...
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <version>${junit-jupiter.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <!--JUnit Jupiter Engine to depend on the JUnit4 engine and JUnit 4 API  -->
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <version>${junit-jupiter.version}</version>
```
sfc-gh-japatel, line 465: scope should be test?

sfc-gh-japatel: snyk merge gates are failing. please make sure they pass before pushing.

pom_confluent.xml
```diff
@@ -593,6 +594,25 @@
...
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <version>${junit-jupiter.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <!--JUnit Jupiter Engine to depend on the JUnit4 engine and JUnit 4 API  -->
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <version>${junit-jupiter.version}</version>
```
sfc-gh-japatel, line 613: same as above

pom.xml
sfc-gh-tzhang: I think the snyk issue is not fixed, we may need to disable it for now

pom_confluent.xml
sfc-gh-tzhang: We have another pom for performance, should we keep them in sync?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -399,7 +398,20 @@ public void close(Collection<TopicPartition> partitions) {
...
+      } catch (Exception e) {
+        LOGGER.warn("Could not close streaming ingest client. {}", e.getMessage());
```
sfc-gh-tzhang, line 411: Do we need the client name?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -537,8 +549,7 @@ public static String partitionChannelKey(String topic, int partition) {
   /* Used for testing */
   @VisibleForTesting
   public SnowflakeStreamingIngestClient getStreamingIngestClient() {
-    return StreamingClientProvider.getStreamingClientProviderInstance()
-        .getClient(this.connectorConfig);
+    return this.streamingIngestClient;
```
sfc-gh-tzhang, line 552: Is this correct? I assume we want to get it from the client provider?

sfc-gh-tzhang (commented):
Left some comments, PTAL!

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-mbobowski: When client optimization is off calling `.getClient()` implicitly creates a new client instance. This is very misleading in tests, because I cannot verify that the client was actually closed.

In general provider abstraction is tricky when optimization is off.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-mbobowski: Done

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-mbobowski: Done

pom.xml
sfc-gh-mbobowski: Good catch! Snyk is passing after this change so probably it ignores test dependencies.

pom.xml
```diff
@@ -445,6 +446,26 @@
         </dependency>
 
         <!--junit for unit test-->
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <version>${junit-jupiter.version}</version>
```
sfc-gh-lshcharbaty, line 452: Can we use [junit bom](https://mvnrepository.com/artifact/org.junit/junit-bom/5.10.2) instead of specifying a version for each dependency?

pom_confluent.xml
```diff
@@ -593,6 +594,26 @@
         </dependency>
 
         <!--junit for unit test-->
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <version>${junit-jupiter.version}</version>
```
sfc-gh-lshcharbaty, line 600: Can we use [junit bom](https://mvnrepository.com/artifact/org.junit/junit-bom/5.10.2) instead of specifying a version for each dependency?

pom_confluent.xml
sfc-gh-mbobowski: Sure, added to all three pom files.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -399,7 +400,24 @@ public void close(Collection<TopicPartition> partitions) {
...
 
   @Override
-  public void setIsStoppedToTrue() {}
+  public void stop() {
+    final boolean isOptimizationEnabled =
+        Boolean.parseBoolean(
+            connectorConfig.getOrDefault(
+                SnowflakeSinkConnectorConfig.ENABLE_STREAMING_CLIENT_OPTIMIZATION_CONFIG,
+                Boolean.toString(ENABLE_STREAMING_CLIENT_OPTIMIZATION_DEFAULT)));
+    if (!isOptimizationEnabled) {
```
sfc-gh-lshcharbaty, line 409: Nit: it's obvious now, but let's add a comment explaining why we close clients only when the optimisation is disabled.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-mbobowski: Thanks, added.

test/perf_test/pom.xml
```diff
@@ -45,6 +45,26 @@
       <version>1.70</version>
     </dependency>
     <!--junit for unit test-->
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId>
+      <version>${junit-jupiter.version}</version>
```
sfc-gh-lshcharbaty, line 51: Can we use [junit bom](https://mvnrepository.com/artifact/org.junit/junit-bom/5.10.2) instead of specifying a version for each dependency?

sfc-gh-lshcharbaty (approved):

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -399,7 +400,24 @@ public void close(Collection<TopicPartition> partitions) {
...
+      try {
+        StreamingClientProvider.getStreamingClientProviderInstance()
+            .closeClient(connectorConfig, this.streamingIngestClient);
+      } catch (Exception e) {
+        LOGGER.warn(
```
sfc-gh-akowalczyk, line 414: can we narrow down the exception type for closing clients?
shouldn't we rethrow the exception?

pom.xml
sfc-gh-mbobowski: Good idea. Perhaps it would be better to post it in a separate PR.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-mbobowski: Comment added

sfc-gh-gjachimko (approved):

sfc-gh-mbobowski:
Will extend one client optimization tests in a separate PR.

PR #801, opened 2024-03-11
SNOW-1057151: Add poc test setup for EmbeddedConnectCluster

sfc-gh-wtrefon:
# Overview

SNOW-1057151

## Pre-review checklist
- [ ] Snowpipe Changes
    - [x] `No` - This change is only in test setup


src/test/java/com/snowflake/kafka/connector/ConnectClusterBaseIT.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Before
+    public void setup() {
+        if (connectCluster == null) {
+            connectCluster = new EmbeddedConnectCluster.Builder().name("kafka-push-connector-connect-cluster").numWorkers(3).build();
+            connectCluster.start();
+            connectCluster.kafka().createTopic(TEST_TOPIC);
+            connectCluster.configureConnector(TEST_CONNECTOR_NAME, createProperties());
+            await().timeout(CONNECTOR_MAX_STARTUP_TIME).until(this::isConnectorRunning);
+        }
```
sfc-gh-lshcharbaty, line 54: Nit: will it be better to move the connector setup into @BeforeClass? Looks like we can reuse the same connector between tests.

src/test/java/com/snowflake/kafka/connector/ConnectClusterBaseIT.java
```diff
@@ -0,0 +1,113 @@
...
+            assertThat(records).hasSize(2);
+            assertThat(records.stream().map(SinkRecord::value)).containsExactly("test1", "test2");
+        });
+    }
+
+    protected Map<String, String> createProperties() {
+        Map<String, String> config = new HashMap<>();
+
+        // kafka connect specific
+        config.put(CONNECTOR_CLASS_CONFIG, SnowflakeFakeSinkConnector.class.getName());
```
sfc-gh-lshcharbaty, line 87: Nit: shall we add a comment that a real connector class will be specified here once we work on [SNOW-1055561](https://snowflakecomputing.atlassian.net/browse/SNOW-1055561)?

src/test/java/com/snowflake/kafka/connector/ConnectClusterBaseIT.java
```diff
@@ -0,0 +1,113 @@
...
+        config.put(TOPICS_CONFIG, TEST_TOPIC);
+        config.put(TASKS_MAX_CONFIG, TASK_NUMBER.toString());
+        config.put(KEY_CONVERTER_CLASS_CONFIG, StringConverter.class.getName());
+        config.put(VALUE_CONVERTER_CLASS_CONFIG, StringConverter.class.getName());
```
sfc-gh-lshcharbaty, line 91: I expect we'll need to test a connector with different event formats. 
Doesn't need to be addressed right now. Just to keep it in mind.

src/test/java/com/snowflake/kafka/connector/fake/SnowflakeFakeSinkConnector.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.snowflake.kafka.connector.SnowflakeSinkConnector;
+import com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig;
+import com.snowflake.kafka.connector.Utils;
+import com.snowflake.kafka.connector.internal.KCLogger;
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.connect.connector.Task;
+import org.apache.kafka.connect.sink.SinkConnector;
+
+import java.util.*;
```
sfc-gh-lshcharbaty, line 11: Wildcard imports [are not welcome here](https://github.com/snowflakedb/snowflake-kafka-connector/pull/800#discussion_r1518163589).

src/test/java/com/snowflake/kafka/connector/fake/SnowflakeFakeSinkTask.java
```diff
@@ -0,0 +1,45 @@
+package com.snowflake.kafka.connector.fake;
+
+import com.snowflake.kafka.connector.SnowflakeSinkConnector;
+import com.snowflake.kafka.connector.Utils;
+import com.snowflake.kafka.connector.internal.KCLogger;
+import org.apache.kafka.connect.sink.SinkRecord;
+import org.apache.kafka.connect.sink.SinkTask;
+
+import java.util.*;
```
sfc-gh-lshcharbaty, line 9: And here too.

pom.xml
```diff
@@ -323,6 +327,12 @@
             </exclusions>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-clients</artifactId>
+            <version>${kafka.version}</version>
```
sfc-gh-lshcharbaty, line 333: Shouldn't it be scoped to test as well?

pom.xml
```diff
@@ -490,5 +500,56 @@
             <version>0.9.0.0</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_2.13</artifactId>
+            <version>${kafka.version}</version>
```
sfc-gh-lshcharbaty, line 507: [Kafka bom](https://issues.apache.org/jira/browse/KAFKA-7955) hasn't been created yet :(

pom.xml
```diff
@@ -323,6 +327,12 @@
             </exclusions>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
```
sfc-gh-lshcharbaty, line 331: There is also a confluent pom present where the dependencies should be added to. Please, check [this pr](https://github.com/snowflakedb/snowflake-kafka-connector/pull/800) as an example.

pom.xml
```diff
@@ -49,6 +49,10 @@
         <skipTests>false</skipTests>
         <skipUnitTests>${skipTests}</skipUnitTests>
         <skipIntegrationTests>${skipTests}</skipIntegrationTests>
+
```
sfc-gh-mbobowski, line 52: There are 3 POM files in the project:
- main (Apache)
- Confluent
- preformance tests

Not sure about the last one, but for sure you need to sync changes between Apache and Confluent

sfc-gh-mbobowski:
Looks promising. Please take care of Confluent pom and run format.sh.

src/test/java/com/snowflake/kafka/connector/ConnectClusterBaseIT.java
sfc-gh-wtrefon: yea, i'm not sure there, it's been a while since i used junit < 5. The problem is with createProperties as it's not static a i thought maybe it's gonna be overridden later but i guess I'm thinking too much here and  this can be refactor later

src/test/java/com/snowflake/kafka/connector/ConnectClusterBaseIT.java
sfc-gh-wtrefon: yup

src/test/java/com/snowflake/kafka/connector/fake/SnowflakeFakeSinkConnector.java
sfc-gh-wtrefon: done

pom.xml
sfc-gh-wtrefon: ahh, i missed that as it does not have maven icon in intellij :D thanks!

pom.xml
sfc-gh-wtrefon: this one is problematic. It comes as transitive dependency from connect-api dependency and it has provided scope. When I import it with test scope it overrides the transitive dependency and compilation fails in some parts

pom.xml
sfc-gh-wtrefon: thanks, didn't know that

sfc-gh-lshcharbaty (approved):
lgtm

scripts/process_licenses.py
```diff
@@ -139,7 +139,8 @@ def main():
                     dependency_without_license_count += 1
                     missing_licenses_str += f"{dependency_lookup_key}: {license_name}\n"
                 else:
-                    raise Exception(f"The dependency {dependency_lookup_key} does not ship a license file, but neither is it not defined in ADDITIONAL_LICENSES_MAP")
+                    err_msg = f"The dependency {dependency_lookup_key} does not ship a license file, but neither is it not defined in ADDITIONAL_LICENSES_MAP"
```
sfc-gh-wtrefon, line 142: FYI: @sfc-gh-mbobowski 
I didn't know as well :D
https://docs.astral.sh/ruff/rules/f-string-in-exception/

pom.xml
sfc-gh-japatel: if it comes at transitive dependency, cant we directly use it?
I am afraid adding more dependency will make this jar bulkier. 

that being said, I like the change, so consider this as a non blocking comment.

sfc-gh-japatel (approved):
Another alternative is https://debezium.io/documentation/reference/stable/integrations/testcontainers.html

lot of other connectors use this and it is an easier way to setup kafka and kc in docker. (test containers)

PR #797, opened 2024-03-01
Resolving Vulnerability CVE-2024-25710 and CVE-2024-26308

bansallovish:
# Overview

SNOW-XXXXX

## Pre-review checklist
- [ ] This change should be part of a Behavior Change Release. See [go/behavior-change](http://go/behavior-change).
- [ ] This change has passed Merge gate tests
- [ ] Snowpipe Changes
- [ ] Snowpipe Streaming Changes
- [ ] This change is TEST-ONLY
- [ ] This change is README/Javadocs only
- [ ] This change is protected by a config parameter <PARAMETER_NAME> eg `snowflake.ingestion.method`.
    - [ ] `Yes` - Added end to end and Unit Tests. 
    - [ ] `No` - Suggest why it is not param protected
- [ ] Is his change protected by parameter <PARAMETER_NAME> on the server side?
    - [ ] The parameter/feature is not yet active in production (partial rollout or PrPr, see [Changes for Unreleased Features and Fixes](http://go/ppp-prpr)).
    - [ ] If there is an issue, it can be safely mitigated by turning the parameter off. This is also verified by a test (See [go/ppp](http://go/ppp)).

PR #802, opened 2024-03-11
Upgrade dependency versions

sfc-gh-xhuang:
Copies https://github.com/snowflakedb/snowflake-kafka-connector/pull/797 - merge on behalf of @bansallovish

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -366,16 +366,28 @@
                 </exclusion>
             </exclusions>
         </dependency>
+
+        <!-- https://mvnrepository.com/artifact/org.apache.commons/commons-compress -->
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-compress</artifactId>
+            <version>1.26.0</version>
+        </dependency>
+        
+        <!-- https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-core -->
         <dependency>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -397,6 +397,10 @@
                     <groupId>io.swagger</groupId>
                     <artifactId>swagger-core</artifactId>
                 </exclusion>
+                <exclusion>
```
sfc-gh-tzhang, line 400: @sfc-gh-lsembera Do you know what's the implication of excluding this in other dependencies? I assume we will use the version defined in the ingest sdk (or any other dependencies that uses this)? If we don't exclude this, we will just use a random version that appears to be the first one?

sfc-gh-tzhang: And what's the best practice here? Should we define the version explicitly?

sfc-gh-tzhang (approved):
LGTM, could we ask @sfc-gh-lsembera to take a quick look to make sure we're following the best practice?

sfc-gh-asen (approved):

pom.xml
sfc-gh-lsembera: If we define the dependency directly, there is no need to exclude it from other dependencies. Maven uses the [_nearest definition_](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#transitive-dependencies) rule, which will pick the version declared directly.

PR #804, opened 2024-03-12
SNOW-1049869 Rewrite client optimization unit tests

sfc-gh-mbobowski:
We lack tests showing explicitly what is the behaviour of one client optimization (follow up to the discussion under [PR](https://github.com/snowflakedb/snowflake-kafka-connector/pull/800) ).

I believe that our north star is to refactor `StreamingClientHandler` to be a constructor parameter of `StreamingClientProvider` and introduce proper fakes in https://snowflakecomputing.atlassian.net/browse/SNOW-1057165. But not everything at once.

As for now I rewrote current unit tests as ITs using JUnit 5. Execution time is not bad (6 seconds locally for the whole suite) and we can switch back to units when we get fakes.


sfc-gh-japatel (approved):
lgtm

sfc-gh-achyzy (approved):
LGTM. 
Thanks for removing those pesky mocks :D

src/test/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProviderIT.java
```diff
@@ -18,168 +18,125 @@
...
+    SnowflakeStreamingIngestClient client2 = streamingClientProvider.getClient(clientConfig);
+
+    // then
+    Assertions.assertNotNull(client);
+    Assertions.assertNotNull(client2);
+    Assertions.assertNotEquals(client, client2);
+  }
+
+  @Test
+  public void getClient_forInvalidClient_returnValidClient() throws Exception {
```
sfc-gh-lshcharbaty, line 69: `getClient_forInvalidClient_returnValidClient` sounds a bit odd :)
Is it intended?

src/test/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProviderIT.java
```diff
@@ -18,168 +18,125 @@
...
+        StreamingClientProvider.getStreamingClientProviderInstance();
+
+    // when
+    SnowflakeStreamingIngestClient client = streamingClientProvider.getClient(clientConfig);
+    client.close();
+    SnowflakeStreamingIngestClient client2 = streamingClientProvider.getClient(clientConfig);
+
+    // then
+    Assertions.assertTrue(client.isClosed());
+    Assertions.assertFalse(client2.isClosed());
```
sfc-gh-lshcharbaty, line 82: Will it make sense to verify that `client` and `client2` is the same client? It can be done by using `assertSame`.

sfc-gh-lshcharbaty (approved):
Nice job!

src/test/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProviderIT.java
sfc-gh-mbobowski: Actually in this case clients are **not** the same. But I switched to `assertSame` in the first test case.

src/test/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProviderIT.java
sfc-gh-mbobowski: Yeah, kind of weird. i renamed to `getClient_forInvalidClient_returnNewInstance`

PR #807, opened 2024-03-15
SNOW-1019628 update file cleaner for snowpipe ingest

sfc-gh-gjachimko:
# Overview

SNOW-1019628

This PR addresses reported issue of cleaner not properly handling file cleanup - in some scenarios, files will be kept in staging area for as long as there is no message in the topic. 
This change modifies the file handling algorithm - it no longer depends on the first record to be received to start - it will start automatically for each partition. 

To enable this new cleaner, provide following configuration entry to your connector config:
`"snowflake.snowpipe.v2CleanerEnabled": true`

Core processing logic happens in ` StageFilesProcessor::trackFiles` 
This method is run in a worker thread. The only way to communicate with the cleaner, is through `ProgressRegister` interface, which allows adding new ingested files for tracking in a thread safe manner. 

## Pre-review checklist
- [ ] Snowpipe Changes
- [x] This change is protected by a config parameter  `snowflake.snowpipe.v2CleanerEnabled`.
    - [x] If there is an issue, it can be safely mitigated by turning the parameter off. This is also verified by a test (See [go/ppp](http://go/ppp)).

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionService.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionService.java
@@ -51,9 +51,20 @@ public interface SnowflakeIngestionService {
   Map<String, IngestedFileStatus> readOneHourHistory(List<String> files, long startTime);
 
   /**
-   * retrieve a collection of all ingested files with their status from load history following
-   * provided historyMarker optionally limited to last N seconds (if provided). the method will
-   * upsert fetched history statuses to the content of provided storage collection.
+   * Retrieve a collection of all ingested files with their status from load history. Fetching can
+   * be further customized by:
+   * <li>history marker - when content is null, entire available history is loaded and upon method
+   *     return, marker will be updated to the last available marker. Subsequent calls will fetch
+   *     only history entries created after the marker.
+   * <li>limit the history depth by querying only the last N seconds since now. NOTE: this is not
+   *     absolute timestamp, so adjust for potential latency by adding some buffer
+   * <li>filter the returned file history by supplying a fileFilter predicate. NOTE: this filtering
+   *     takes place at client side, so first - try limiting the result set size by providing
+   *     history marker / history depth in seconds Important thing - supplied storage container is
+   *     mutated by this call by:
+   * <li>adding new file entry if none exist
+   * <li>updating the state of the file, if file entry exists The method wil never delete any entry
+   *     from the map.
    *
    * @param storage - reference to the map where to store status updates for the tracked files
    * @param fileFilter - filter predicate - if provided will be used to filter only files matching
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionService.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionService.java
@@ -4,6 +4,9 @@ import com.snowflake.kafka.connector.internal.InternalUtils.IngestedFileStatus;
 import com.snowflake.kafka.connector.internal.telemetry.SnowflakeTelemetryService;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Predicate;
+import net.snowflake.ingest.connection.HistoryResponse;
 
 /** Ingestion service manages snow pipe */
 public interface SnowflakeIngestionService {
@@ -46,6 +49,27 @@ public interface SnowflakeIngestionService {
    * @return a map contains all file status
    */
   Map<String, IngestedFileStatus> readOneHourHistory(List<String> files, long startTime);
+
+  /**
+   * retrieve a collection of all ingested files with their status from load history following
+   * provided historyMarker optionally limited to last N seconds (if provided). the method will
+   * upsert fetched history statuses to the content of provided storage collection.
+   *
+   * @param storage - reference to the map where to store status updates for the tracked files
+   * @param fileFilter - filter predicate - if provided will be used to filter only files matching
+   *     the criteria.
+   * @param historyMarker - reference to the history marker - provide null as reference's value for
+   *     the first call, method will update the reference with most recent marker, so the next call
+   *     will resume with that specific marker
+   * @param lastNSeconds - optionally narrow down history to just last N seconds. null for max
+   *     history range
+   * @return number of loaded history entries
+   */
+  int readIngestHistoryForward(
+      Map<String, IngestedFileStatus> storage,
+      Predicate<HistoryResponse.FileEntry> fileFilter,
+      AtomicReference<String> historyMarker,
+      Integer lastNSeconds);
 
   /** close ingest service */
   void close();
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
@@ -10,6 +10,7 @@ import java.util.*;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Predicate;
+import javax.annotation.Nullable;
 import net.snowflake.ingest.SimpleIngestManager;
 import net.snowflake.ingest.connection.HistoryRangeResponse;
 import net.snowflake.ingest.connection.HistoryResponse;
@@ -48,7 +49,8 @@ public class SnowflakeIngestionServiceV1 implements SnowflakeIngestionService {
       String stageName,
       String pipeName,
       PrivateKey privateKey,
-      String userAgentSuffix) {
+      String userAgentSuffix,
+      @Nullable SnowflakeTelemetryService telemetry) {
 
     this(
         stageName,
@@ -61,16 +63,22 @@ public class SnowflakeIngestionServiceV1 implements SnowflakeIngestionService {
             connectionScheme,
             pipeName,
             privateKey,
-            userAgentSuffix));
+            userAgentSuffix,
+            telemetry),
+        telemetry);
     LOGGER.info("initialized the pipe connector for pipe {}", pipeName);
   }
 
   @VisibleForTesting
   SnowflakeIngestionServiceV1(
-      String stageName, String pipeName, SimpleIngestManager ingestManager) {
+      String stageName,
+      String pipeName,
+      SimpleIngestManager ingestManager,
+      @Nullable SnowflakeTelemetryService telemetry) {
     this.stageName = stageName;
     this.pipeName = pipeName;
     this.ingestManager = ingestManager;
+    setTelemetry(telemetry);
   }
 
   private static SimpleIngestManager createOrThrow(
@@ -81,7 +89,8 @@ public class SnowflakeIngestionServiceV1 implements SnowflakeIngestionService {
       String connectionScheme,
       String pipeName,
       PrivateKey privateKey,
-      String userAgentSuffix) {
+      String userAgentSuffix,
+      SnowflakeTelemetryService telemetry) {
     try {
       return new SimpleIngestManager(
           accountName,
@@ -93,7 +102,7 @@ public class SnowflakeIngestionServiceV1 implements SnowflakeIngestionService {
           port,
           userAgentSuffix);
     } catch (Exception e) {
-      throw SnowflakeErrors.ERROR_0002.getException(e, null);
+      throw SnowflakeErrors.ERROR_0002.getException(e, telemetry);
     }
   }
 
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
@@ -3,9 +3,13 @@ package com.snowflake.kafka.connector.internal;
 import static com.snowflake.kafka.connector.internal.InternalUtils.convertIngestStatus;
 import static com.snowflake.kafka.connector.internal.InternalUtils.timestampToDate;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.snowflake.kafka.connector.internal.telemetry.SnowflakeTelemetryService;
 import java.security.PrivateKey;
 import java.util.*;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Predicate;
 import net.snowflake.ingest.SimpleIngestManager;
 import net.snowflake.ingest.connection.HistoryRangeResponse;
 import net.snowflake.ingest.connection.HistoryResponse;
@@ -45,23 +49,52 @@ public class SnowflakeIngestionServiceV1 implements SnowflakeIngestionService {
       String pipeName,
       PrivateKey privateKey,
       String userAgentSuffix) {
+
+    this(
+        stageName,
+        pipeName,
+        createOrThrow(
+            accountName,
+            userName,
+            host,
+            port,
+            connectionScheme,
+            pipeName,
+            privateKey,
+            userAgentSuffix));
+    LOGGER.info("initialized the pipe connector for pipe {}", pipeName);
+  }
+
+  @VisibleForTesting
+  SnowflakeIngestionServiceV1(
+      String stageName, String pipeName, SimpleIngestManager ingestManager) {
     this.stageName = stageName;
     this.pipeName = pipeName;
+    this.ingestManager = ingestManager;
+  }
+
+  private static SimpleIngestManager createOrThrow(
+      String accountName,
+      String userName,
+      String host,
+      int port,
+      String connectionScheme,
+      String pipeName,
+      PrivateKey privateKey,
+      String userAgentSuffix) {
     try {
-      this.ingestManager =
-          new SimpleIngestManager(
-              accountName,
-              userName,
-              pipeName,
-              privateKey,
-              connectionScheme,
-              host,
-              port,
-              userAgentSuffix);
+      return new SimpleIngestManager(
+          accountName,
+          userName,
+          pipeName,
+          privateKey,
+          connectionScheme,
+          host,
+          port,
+          userAgentSuffix);
     } catch (Exception e) {
-      throw SnowflakeErrors.ERROR_0002.getException(e, this.telemetry);
+      throw SnowflakeErrors.ERROR_0002.getException(e, null);
     }
-    LOGGER.info("initialized the pipe connector for pipe {}", pipeName);
   }
 
   @Override
@@ -178,6 +211,45 @@ public class SnowflakeIngestionServiceV1 implements SnowflakeIngestionService {
         });
 
     return result;
+  }
+
+  @Override
+  public int readIngestHistoryForward(
+      Map<String, InternalUtils.IngestedFileStatus> storage,
+      Predicate<HistoryResponse.FileEntry> fileFilter,
+      AtomicReference<String> historyMarker,
+      Integer lastNSeconds) {
+    HistoryResponse response;
+    try {
+      response =
+          (HistoryResponse)
+              InternalUtils.backoffAndRetry(
+                  telemetry,
+                  SnowflakeInternalOperations.INSERT_REPORT_SNOWPIPE_API,
+                  () -> ingestManager.getHistory(null, lastNSeconds, historyMarker.get()));
+    } catch (Exception e) {
+      throw SnowflakeErrors.ERROR_3002.getException(e, this.telemetry);
+    }
+
+    AtomicInteger loadedRecords = new AtomicInteger();
+    if (response != null) {
+      historyMarker.set(response.getNextBeginMark());
+      response.files.stream()
+          .filter(file -> fileFilter == null || fileFilter.test(file))
+          .forEach(
+              historyEntry -> {
+                storage.compute(
+                    historyEntry.getPath(),
+                    (key, status) -> convertIngestStatus(historyEntry.getStatus()));
+                loadedRecords.incrementAndGet();
+              });
+      LOGGER.info(
+          "loaded {} files out of {} in ingest report since marker {}",
+          loadedRecords.get(),
+          response.files.size(),
+          historyMarker.get());
+    }
+    return loadedRecords.get();
   }
 
   /**
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
@@ -78,7 +78,7 @@ public class SnowflakeIngestionServiceV1 implements SnowflakeIngestionService {
     this.stageName = stageName;
     this.pipeName = pipeName;
     this.ingestManager = ingestManager;
-    setTelemetry(telemetry);
+    this.telemetry = telemetry;
   }
 
   private static SimpleIngestManager createOrThrow(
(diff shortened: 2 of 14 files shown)


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -429,25 +432,44 @@ private ServiceContext(
             "Registered {} metrics for pipeName:{}", metricRegistry.getMetrics().size(), pipeName);
       }
 
+      // how to get the configuration here??
```
sfc-gh-mbobowski, line 435: `SnowflakeSinkTask#start()` -> `SnowflakeSinkServiceFactory.builder()` -> `SnowflakeSink ServiceV1`

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
```diff
@@ -0,0 +1,505 @@
...
+        void stopTrackingFiles(List<String> files) {
+            files.forEach(file -> stageFiles.computeIfPresent(file, (key, entry) -> {
+                entry.keepTracking = false;
+                return entry;
+            }));
+        }
+    }
+
+    @VisibleForTesting
+    static class IngestEntry {
```
sfc-gh-lshcharbaty, line 405: It isn't clear what `IngestEntry` really means. Let's name it `IngestedFileState` or somehow similar to it.

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
```diff
@@ -0,0 +1,505 @@
...
+            return pipeTelemetry;
+        }
+
+        @Override
+        public  SnowflakeTelemetryService getTelemetry() {
+            return telemetryService;
+        }
+
+        @Override
+        public void close() {
```
sfc-gh-lshcharbaty, line 456: Can we call `close` on `StageFilesProcessor` directly in `SnowflakeSinkServiceV1` and avoid having this awkward reference to an `owner` of this `ProgressRegisterImpl`?

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
```diff
@@ -0,0 +1,505 @@
...
+            if (owner != null) {
+                owner.close();
+                owner = null;
+            }
+        }
+    }
+
+    static class FilteringPredicates {
```
sfc-gh-lshcharbaty, line 464: :+1:

src/test/java/com/snowflake/kafka/connector/internal/StageFilesProcessorTest.java
```diff
@@ -0,0 +1,249 @@
...
+import org.mockito.Mockito;
+
+import java.time.Duration;
+import java.util.*;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.*;
```
sfc-gh-lshcharbaty, line 20: Nit: wildcard import.

src/test/java/com/snowflake/kafka/connector/internal/StageFilesProcessorTest.java
```diff
@@ -0,0 +1,249 @@
...
+import com.snowflake.kafka.connector.internal.telemetry.SnowflakeTelemetryService;
+import net.snowflake.client.jdbc.internal.joda.time.DateTime;
+import net.snowflake.client.jdbc.internal.joda.time.DateTimeZone;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
+
+import java.time.Duration;
+import java.util.*;
```
sfc-gh-lshcharbaty, line 14: Nit: wildcard import.

src/test/java/com/snowflake/kafka/connector/internal/StageFilesProcessorTest.java
```diff
@@ -0,0 +1,249 @@
...
+        currentTime = new AtomicLong(new DateTime(2000, 1, 10, 12, 0, DateTimeZone.UTC).getMillis());
+
+        conn = Mockito.mock(SnowflakeConnectionService.class);
+        ingestionService = Mockito.mock(SnowflakeIngestionService.class);
+        pipeCreation = new SnowflakeTelemetryPipeCreation(TABLE_NAME, STAGE_NAME, PIPE_NAME);
+        pipeTelemetry = Mockito.mock(SnowflakeTelemetryPipeStatus.class);
+        telemetryService = Mockito.mock( SnowflakeTelemetryService.class);
+        register = new StageFilesProcessor.ProgressRegisterImpl(pipeCreation, pipeTelemetry, telemetryService);
+
+        victim = new StageFilesProcessor(PIPE_NAME, TABLE_NAME, STAGE_NAME, PREFIX, conn, ingestionService, pipeTelemetry, telemetryService, currentTime::get);
```
sfc-gh-lshcharbaty, line 50: I haven't seen anyone naming SUTs victims :)
It wasn't really obvious from the tests what is represented by it.

src/test/java/com/snowflake/kafka/connector/internal/StageFilesProcessorTest.java
```diff
@@ -0,0 +1,249 @@
...
+            try {
+                Thread.sleep(Duration.ofSeconds(2L).toMillis());
+                victim.close();
+            } catch (Exception e) {
+            }
+        });
+        stopper.start();
+        stopper.join(Duration.ofSeconds(5L).toMillis());
+
+        assertThat(completed.get()).isTrue();
```
sfc-gh-lshcharbaty, line 71: We can use awaitility instead.
```java
await().atMost(2, SECONDS).until(completed::get);
```

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
```diff
@@ -0,0 +1,505 @@
...
+
+    /**
+     * trackFiles - actual processor logic with extracted thread management for testing purposes.
+     * @param register - client interface
+     * @param worker - normally - common pool worker thread doing the work; allows providing custom thread for testing
+     * @param canContinue - delegate to Thread.currentThread().isInterrupted(), required for testing
+     * @param awaiter - delegate to Thread.sleep(), required for testing
+     */
+    @VisibleForTesting
+    void trackFiles(ProgressRegisterImpl register, Thread worker, Supplier<Boolean> canContinue, TaskAwaiter awaiter) {
```
sfc-gh-lshcharbaty, line 141: Nit: we always pass `Thread.currentThread()` for the `worker` parameter.

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
```diff
@@ -0,0 +1,505 @@
...
+            TimeSupplier currentTimeSupplier) {
+        this.pipeName = pipeName;
+        this.tableName = tableName;
+        this.stageName = stageName;
+        this.prefix = prefix;
+        this.conn = conn;
+        this.currentTimeSupplier = currentTimeSupplier;
+        this.ingestionService = ingestionService;
+        this.telemetryService = telemetryService;
+        this.cleanerExecutor = ForkJoinPool.commonPool();
```
sfc-gh-lshcharbaty, line 108: Do we really want to use a common `ForkJoinPool`? I assume it can be used by multiple different connectors (not only Kafka ones) if they're running in the same kafka connect instance.

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
```diff
@@ -0,0 +1,505 @@
+package com.snowflake.kafka.connector.internal;
+
+
+import com.google.common.annotations.VisibleForTesting;
+import com.snowflake.kafka.connector.internal.telemetry.SnowflakeTelemetryPipeCreation;
+import com.snowflake.kafka.connector.internal.telemetry.SnowflakeTelemetryPipeStatus;
+import com.snowflake.kafka.connector.internal.telemetry.SnowflakeTelemetryService;
+
+import java.time.Duration;
+import java.util.*;
```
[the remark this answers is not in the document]
sfc-gh-gjachimko: different IJ settings i guess .updated.

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
```diff
@@ -0,0 +1,505 @@
...
+        boolean shouldFetchInitialStageFiles = true;
+        boolean firstRun = true;
+
+        List<String> files = new ArrayList<>();
+        while (canContinue.get()) {
+            try {
+                register.getTelemetry().reportKafkaPartitionUsage(register.getPipeCreation(), false);
+
+                // wait 1 minute
+                awaiter.await(Duration.ofMinutes(1L).toMillis());
```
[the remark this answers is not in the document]
sfc-gh-gjachimko: i don't think this code requries any dedicated scheduler TBH. common pool executor should do the job good enough.

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
```diff
@@ -0,0 +1,505 @@
...
+    private List<String> nextCheck(Collection<String> stageFiles, ProgressRegisterImpl register, boolean firstRun) {
+
+        // make first categorization - split files into these with start offset higher than current
+        FileCategorizer fileCategories = FileCategorizer.build(stageFiles, register.offset.get());
+
+        if (firstRun) {
+            register.getPipeCreation().setFileCountReprocessPurge(fileCategories.dirtyFiles.size());
+            register.getPipeCreation().setFileCountRestart(fileCategories.stageFiles.size());
+            register.getPipeTelemetry().addAndGetFileCountOnStage(fileCategories.stageFiles.size());
+            register.getPipeTelemetry().addAndGetFileCountOnIngestion(fileCategories.stageFiles.size());
```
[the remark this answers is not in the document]
sfc-gh-gjachimko: ok, refactored.

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
```diff
@@ -0,0 +1,505 @@
...
+                        e.getStackTrace());
+                shouldFetchInitialStageFiles = true;
+            }
+            firstRun = false;
+        }
+    }
+
+    @Override
+    public void close() {
+        Thread runner = runnerThread.getAndSet(null);
```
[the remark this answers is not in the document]
sfc-gh-gjachimko: first caller gets the reference and breaks the cleaner loop. all other will always get null and will not proceed. 
yes, it is ok to interrupt them - that's the best signal for the executing in methods like sleep or doing I/O to actually give up on execution and return control. 
i'm handling the interrupt exception in the trackFiles method, so in such case the interrupted flag would be cleared and thread returns to the pool as it would normally complete its job.

src/test/java/com/snowflake/kafka/connector/internal/StageFilesProcessorTest.java
sfc-gh-gjachimko: old habit :)

src/test/java/com/snowflake/kafka/connector/internal/StageFilesProcessorTest.java
sfc-gh-gjachimko: i didn't notice we have that library at hand !

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
sfc-gh-gjachimko: and? i'd rather use the threads which are already there, present and ready for work, rather than spawn new ones. 
the moment thread from the pool is stuck on the I/O or timed wait state and pool is running out of threads - it will automatically add a new one. 
also - the lifetime of the sink service is controlled by kafka - it may go up and down quite frequently - in this case, not managing our own threads is even better idea.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionService.java
```diff
@@ -47,6 +50,27 @@ public interface SnowflakeIngestionService {
    */
   Map<String, IngestedFileStatus> readOneHourHistory(List<String> files, long startTime);
 
+  /**
+   * retrieve a collection of all ingested files with their status from load history following
+   * provided historyMarker optionally limited to last N seconds (if provided). the method will
```
sfc-gh-japatel, line 55: nit: this sentence is reading weird!

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
```diff
@@ -45,23 +49,52 @@ public class SnowflakeIngestionServiceV1 implements SnowflakeIngestionService {
...
+          userName,
+          pipeName,
+          privateKey,
+          connectionScheme,
+          host,
+          port,
+          userAgentSuffix);
     } catch (Exception e) {
-      throw SnowflakeErrors.ERROR_0002.getException(e, this.telemetry);
+      throw SnowflakeErrors.ERROR_0002.getException(e, null);
```
sfc-gh-japatel, line 96: why is this set to null?

sfc-gh-gjachimko: it never had a chance to be initialized to something non null - the object was constructed before this.telemetry is assigned - which is initialized later in the code. 
and i had to pass null explicit here, as i had to make the factory method static (to avoid code duplication)

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
```diff
@@ -0,0 +1,622 @@
...
+import java.util.stream.Stream;
+import net.snowflake.ingest.connection.HistoryResponse;
+
+/**
+ * StageFileProcessor - class responsible for tracking files submitted to snowpipe for processing.
+ * This class is responsible for: - purging files with invalid offset - i.e. crash happened before
+ * offset was committed, but file was uploaded: it will be deleted from stage; - purging files which
+ * have been successfully ingested - i.e. their load status is LOADED - moving failed files (their
+ * status is FAILED or PARTIALLY_LOADED) or old files (1h) where no status is available, to the
+ * table stage.
```
sfc-gh-japatel, line 37: Thanks for javadocs but
minor nit:
can you use `<ul> <li>`, it will be more readable?

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
```diff
@@ -0,0 +1,622 @@
...
+   * trackFiles - actual processor logic with extracted thread management for testing purposes.
+   *
+   * @param register - client interface
+   * @param progressTelemetry - telemetry interface
+   * @param canContinue - delegate to Thread.currentThread().isInterrupted(), required for testing
+   * @param awaiter - delegate to Thread.sleep(), required for testing
+   */
+  @VisibleForTesting
+  void trackFiles(
+      ProgressRegisterImpl register,
```
sfc-gh-japatel, line 157: should we use `ProgressRegister` interface here?

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
```diff
@@ -0,0 +1,622 @@
...
+    private boolean keepTracking = true;
+
+    IngestEntry(InternalUtils.IngestedFileStatus status, long timestamp) {
+      this.status = status;
+      this.timestamp = timestamp;
+    }
+  }
+
+  static class ProgressRegisterImpl implements ProgressRegister {
+    private static final KCLogger LOGGER = new KCLogger(ProgressRegister.class.getName());
```
sfc-gh-japatel, line 481: nit: `ProgressRegisterImpl`

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
sfc-gh-gjachimko: implementation has some internal content - thus - the implementation here. the trackFiles itself should be considered private (shall we put it in dedicated package? - it would be effecitvely private for other callers)

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
sfc-gh-gjachimko: i didn't want to expose implementation details - what client gets - is a interface reference, thus the logger name here

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
sfc-gh-japatel: it will make testing difficult. but I dont have a major concern here.

sfc-gh-japatel (commented):
lets discuss this in our meeting for coming friday. High level idea is good, i just wish the PR was a bit smaller. (Breaking up in smaller chunks would have made reviewer's life easy)

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionService.java
sfc-gh-gjachimko: extended

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
sfc-gh-gjachimko: ok, i've refactrored builder, so the telemetry would be accessible in the ctorl

src/main/java/com/snowflake/kafka/connector/internal/StageFilesProcessor.java
```diff
@@ -0,0 +1,656 @@
...
+  /**
+   * trackFiles - actual processor logic with extracted thread management for testing purposes.
+   *
+   * @param register - client interface
+   * @param progressTelemetry - telemetry interface
+   */
+  @VisibleForTesting
+  void trackFiles(ProgressRegisterImpl register, PipeProgressRegistryTelemetry progressTelemetry) {
+    LOGGER.info(
+        "Starting file cleaner for pipe {} on thread {}...",
```
sfc-gh-lshcharbaty, line 176: Nit: we don't really run the cleaner on the current thread.

src/test/java/com/snowflake/kafka/connector/internal/StageFilesProcessorTest.java
```diff
@@ -0,0 +1,411 @@
...
+  void filesProcessWillTerminateOnStopSignal() {
+    StageFilesProcessor.ProgressRegister client = victim.trackFilesAsync();
+    client.close();
+    verify(scheduledFuture.get(), times(1)).cancel(true);
+  }
+
+  @Test
+  void fileProcessorWillTryToGetInitialStateFromStageExactlyOnce() {
+    // lets simulate 1 hour (60 cleanup cycles)
+    loops.set(60);
```
sfc-gh-lshcharbaty, line 100: Nit: it looks a little bit complex to grasp.
First we create a scheduler with `loops` passed as an argument. Then we alter the `loops` values so that during `victim.trackFiles` it's taken into account since we propagated it earlier.
Will it be simpler to have a method, let's say `createStageFilesProcessor`, that takes number of iterations as an argument? So we create it in each test without any indirection involving atomics?

src/test/java/com/snowflake/kafka/connector/internal/StageFilesProcessorTest.java
```diff
@@ -0,0 +1,411 @@
...
+        .thenReturn(0);
+
+    victim.trackFiles(register, telemetry);
+
+    verify(conn, times(2)).listStage(STAGE_NAME, PREFIX);
+    verify(ingestionService, times(10)).readIngestHistoryForward(anyMap(), any(), any(), anyInt());
+  }
+
+  @Test
+  void fileProcessorWillPurgeLoadedFilesWhenHistoryIsAvailableAfterFilesHaveBeenSubmitted() {
```
sfc-gh-lshcharbaty, line 130: Nit: it's quite hard to read this test name. Adding underscores might help: `fileProcessor_willPurgeLoadedFiles_whenHistoryIsAvailableAfterFilesHaveBeenSubmitted`.
Applies to the other tests too.

sfc-gh-lshcharbaty (approved):

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
```diff
@@ -44,24 +49,61 @@ public class SnowflakeIngestionServiceV1 implements SnowflakeIngestionService {
...
+  @VisibleForTesting
+  SnowflakeIngestionServiceV1(
+      String stageName,
+      String pipeName,
+      SimpleIngestManager ingestManager,
+      @Nullable SnowflakeTelemetryService telemetry) {
     this.stageName = stageName;
     this.pipeName = pipeName;
+    this.ingestManager = ingestManager;
+    setTelemetry(telemetry);
```
sfc-gh-achyzy, line 81: Why use setter in the constructor?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -311,6 +329,17 @@ public void setBehaviorOnNullValuesConfig(
     this.behaviorOnNullValues = behavior;
   }
 
+  // enable use of new stage files processor
+  void enableStageFilesProcessor() {
+    this.useStageFilesProcessor = true;
+    if (null != cleanerServiceExecutor) {
```
sfc-gh-achyzy, line 335: Please use a consistent way of null checking - I think in other places we use variable != null

src/test/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1Test.java
```diff
@@ -0,0 +1,122 @@
...
+import org.mockito.Mockito;
+
+class SnowflakeIngestionServiceV1Test {
+
+  private SimpleIngestManager manager;
+  private SnowflakeIngestionService svc;
+
+  @BeforeEach
+  void setup() {
+    manager = Mockito.mock(SimpleIngestManager.class);
```
sfc-gh-achyzy, line 23: Ideally we have our own fake (and abstraction over it SimpleIngestManager) instead of yet another mock.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -311,6 +329,17 @@ public void setBehaviorOnNullValuesConfig(
...
+      cleanerServiceExecutor.shutdown();
+    }
+    cleanerServiceExecutor =
+        new ScheduledThreadPoolExecutor(
+            Math.max(1, Runtime.getRuntime().availableProcessors() / 2));
```
sfc-gh-achyzy, line 340: Why do you need more than 1 thread? It could impact the customer workload and since customers run multiple kafka connect clusters on the same node it could be too much to use the  availableProcessors

sfc-gh-achyzy (commented):
Overall looks good. I'll wait for comment reply before the approval. 
 
For the future please split the refactoring and the core changes to seprate PRs (or at least provide more details list of going through the changes).

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1.java
sfc-gh-gjachimko: it was configured using builder - i've just reused the setter. i've updated the code to directly assign value.

src/test/java/com/snowflake/kafka/connector/internal/SnowflakeIngestionServiceV1Test.java
sfc-gh-gjachimko: maybe, but there were none so far and I wanted to provide some basic testing for the service itself, without relying on the underlying SDK class

sfc-gh-achyzy (approved):
LGTM. 
Thanks.

PR #808, opened 2024-03-18
SNOW-1055524 E2E test with nullable SMT

sfc-gh-lshcharbaty:
# Overview

SNOW-1055524
Implementing an e2e test with an SMT that can return a nullable value.
To keep things simple, `org.apache.kafka.connect.transforms.ExtractField` has been used. By filling in an optional value for half of the events only we achieve alteration between non-nullable and nullable values.

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/test/test_suit/test_nullable_values_after_smt.py
+++ b/test/test_suit/test_nullable_values_after_smt.py
@@ -1,0 +1,59 @@
+import json
+from snowflake.connector import DictCursor
+from test_suit.test_utils import NonRetryableError, RetryableError
+
+
+class TestNullableValuesAfterSmt:
+
+    def __init__(self, driver, nameSalt):
+        self.driver = driver
+        self.fileName = 'travis_correct_nullable_values_after_smt'
+        self.table = self.fileName + nameSalt
+        self.topic = self.table + '0'
+
+        self.driver.snowflake_conn.cursor().execute(
+            f'create or replace table {self.table} (index number not null)')
+
+        self.driver.snowflake_conn.cursor().execute(
+            f'alter table {self.table} set enable_schema_evolution = true')
+
+    def getConfigFileName(self):
+        return self.fileName + '.json'
+
+    def send(self):
+        value = []
+        for idx in range(200):
+            event = { 'index': idx, 'someKey': 'someValue' }
+
+            if idx % 2 == 0: # Only every other event contains optionalField.
+                optional_value = { 'index': idx, 'from_optional_field': True }
+                event.update({ 'optionalField': optional_value })
+
+            value.append(json.dumps(event).encode('utf-8'))
+
+        self.driver.sendBytesData(self.topic, value)
+
+    def verify(self, round):
+        cur = self.driver.snowflake_conn.cursor(DictCursor)
+        res = cur.execute(f'select index, from_optional_field, record_metadata:offset::number as offset from {self.table}').fetchall()
+
+        if len(res) == 0:
+            raise RetryableError()
+        elif len(res) != 100:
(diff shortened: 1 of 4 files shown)


test/test_suit/test_nullable_values_after_smt.py
```diff
@@ -0,0 +1,59 @@
...
+
+        self.driver.snowflake_conn.cursor().execute(
+            f'alter table {self.table} set enable_schema_evolution = true')
+
+    def getConfigFileName(self):
+        return self.fileName + '.json'
+
+    def send(self):
+        value = []
+        for idx in range(200):
```
sfc-gh-achyzy, line 25: Is there a specific reason why we send 200? Because of buffering? Why not more or less? It would be best to send the minimal required amount to keep the processing short.

sfc-gh-achyzy (approved):
LGTM. 
Can we send fewer events than 200 to speed up the test?

test/test_suit/test_nullable_values_after_smt.py
sfc-gh-lshcharbaty: No, there is no specific reason. However, the number of *ingested* events must be lower or equal to `buffer.count.records` parameter. 
We can decrease a number of events to 100 or even 20, but most of the test time is spent for `=== Sleep 60 secs before verify result in Snowflake DB ===` anyway.

test/test_suit/test_nullable_values_after_smt.py
```diff
@@ -0,0 +1,59 @@
+import json
+from snowflake.connector import DictCursor
+from test_suit.test_utils import NonRetryableError, RetryableError
+
+
+class TestNullableValuesAfterSmt:
```
sfc-gh-japatel, line 6: lets add some docs here.

test/test_suit/test_nullable_values_after_smt.py
```diff
@@ -0,0 +1,59 @@
...
+            if rec['INDEX'] != idx:
+                raise NonRetryableError(f"Invalid index value. Expected: {idx}, got: {rec['INDEX']}")
+
+            if not rec['FROM_OPTIONAL_FIELD']:
+                raise NonRetryableError('Invalid from_optional_field value. Expected to be true')
+
+            if rec['OFFSET'] != idx:
+                raise NonRetryableError(f"Invalid offset value. Expected: {idx}, got: {rec['OFFSET']}")
+
+            idx += 2  # Only every other event is going to be ingested.
```
sfc-gh-japatel, line 56: can we verify number idx's value at the end and no of messages in snowflake table?

sfc-gh-tzhang (approved):
I think there're other things that we want to test with SMT in general, for example:
- The behavior without schematization
- The behavior with Snowpipe based Kafka Connector
- The behavior with schematization that the columns with the correct data type are created?
- etc.

Do you plan to add them in future PRs?

sfc-gh-wtrefon (approved):

test/test_suit/test_nullable_values_after_smt.py
sfc-gh-lshcharbaty: WE do verify a number of ingested rows:
```py
res = cur.execute(f'select index, from_optional_field, record_metadata:offset::number as offset from {self.table}').fetchall()

# ...
elif len(res) != 100:
    raise NonRetryableError('Number of record in table is different from number of expected records')
```

sfc-gh-lshcharbaty:
@sfc-gh-tzhang 
This PR addresses the ingestions issue involving SMTs and nulls produced by that SMTs (see the ticket for more details).
I will add more tests involving SMTs in another PR.

PR #811, opened 2024-03-21
SNOW-1055524 E2E SMT without schema evolution

sfc-gh-lshcharbaty:
# Overview

SNOW-1055524
Implementing an e2e test with an SMT that can return a nullable value.
To keep things simple, org.apache.kafka.connect.transforms.ExtractField has been used. By filling in an optional value for half of the events only we achieve alteration between non-nullable and nullable values.

Also renamed `TestNullableValuesAfterSmt` to `TestSchemaEvolutionNullableValuesAfterSmt` and rewrote assertions in a more compact way.

## Pre-review checklist
- [ ] Snowpipe Changes


test/rest_request_template/schema_evolution_nullable_values_after_smt.json
```diff
@@ -0,0 +1,32 @@
+{
```
sfc-gh-lshcharbaty, line 1: This is a renamed file. No changes were made here.

test/test_suit/test_schema_evolution_nullable_values_after_smt.py
```diff
@@ -0,0 +1,61 @@
+import json
```
sfc-gh-lshcharbaty, line 1: This is a renamed file. No changes were made here.

test/test_suit/test_schema_evolution_nullable_values_after_smt.py
```diff
@@ -0,0 +1,61 @@
...
+
+        self.driver.sendBytesData(self.topic, value)
+
+    def verify(self, round):
+        cur = self.driver.snowflake_conn.cursor(DictCursor)
+        res = cur.execute(f'select index, from_optional_field, record_metadata:offset::number as offset from {self.table}').fetchall()
+
+        if len(res) == 0:
+            raise RetryableError()
+        elif len(res) != 100:
```
sfc-gh-gjachimko, line 44: in send you create a range 0 .. 200, but here you verify if len != 100 - is that correct?

test/test_suit/test_schema_evolution_nullable_values_after_smt.py
```diff
@@ -0,0 +1,61 @@
...
+            if rec['INDEX'] != idx:
+                raise NonRetryableError(f"Invalid index value. Expected: {idx}, got: {rec['INDEX']}")
+
+            if not rec['FROM_OPTIONAL_FIELD']:
+                raise NonRetryableError('Invalid from_optional_field value. Expected to be true')
+
+            if rec['OFFSET'] != idx:
+                raise NonRetryableError(f"Invalid offset value. Expected: {idx}, got: {rec['OFFSET']}")
+
+            idx += 2  # Only every other event is going to be ingested.
```
sfc-gh-gjachimko, line 58: would it make sense to simply fetch only records from test table for which from_optional_field is null ?

test/test_suit/test_schema_evolution_nullable_values_after_smt.py
sfc-gh-lshcharbaty: Yes. After applying SMT half of the messages are nulls that are ignored.

test/test_suit/test_schema_evolution_nullable_values_after_smt.py
sfc-gh-lshcharbaty: `from_optional_field` is present for each entry in the table. 
If a message doesn't have `optionalField`, it will be turned to null after SMT and ignored.

test/test_suit/test_snowpipe_streaming_nullable_values_after_smt.py
```diff
@@ -0,0 +1,54 @@
...
+            value.append(json.dumps(event).encode('utf-8'))
+
+        self.driver.sendBytesData(self.topic, value)
+
+    def verify(self, round):
+        cur = self.driver.snowflake_conn.cursor(DictCursor)
+        res = cur.execute(f'select record_content, record_metadata:offset::number as offset from {self.table}').fetchall()
+
+        if len(res) == 0:
+            raise RetryableError()
```
sfc-gh-akowalczyk, line 37: Will retrying be eventually stopped if anything goes wrong here? After how many attempts?

sfc-gh-akowalczyk (approved):
I do not know whether there are any tools to make the test assertions better. The test scenarios are fine, but I'd like to hear whether the assertions can be more "compact"

test/test_suit/test_snowpipe_streaming_nullable_values_after_smt.py
sfc-gh-lshcharbaty: There are [30 retries](https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/test/test_verify.py#L50) before failure.

test/test_suit/test_snowpipe_streaming_nullable_values_after_smt.py
```diff
@@ -0,0 +1,54 @@
...
+        for rec in res:
+            expected_content = { 'index': idx, 'from_optional_field': True }
+
+            if json.loads(rec['RECORD_CONTENT']) != expected_content:
+                raise NonRetryableError(f"Invalid index value. Expected: {expected_content}, got: {rec['RECORD_CONTENT']}")
+
+            if rec['OFFSET'] != idx:
+                raise NonRetryableError(f"Invalid offset value. Expected: {idx}, got: {rec['OFFSET']}")
+
+            idx += 2  # Only every other event is going to be ingested.
```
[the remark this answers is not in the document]
sfc-gh-lshcharbaty: Good call. I replaced these assertions with more compact ones. Unfortunately, they won't provide a lot of information in case of mismatches, but we can improve it later.

sfc-gh-tzhang (approved):
lgtm, thanks!

PR #809, opened 2024-03-18
SNOW-1057165  - Test-doubles for snowflake-ingest-java

sfc-gh-achyzy:
# Overview

SNOW-1057165 Test-doubles for snowflake-ingest-java

Introduction of test doubles for [snowflake-ingest-java](https://github.com/snowflakedb/snowflake-ingest-java):
- FakeStreamingClientHandler to create clients
- FakeSnowflakeStreamingIngestClient for faking clients
- FakeSnowflakeStreamingIngestChannel for faking channels

This change is test-only except for the automatic extraction of the interface + renaming from the IDE. 

## Pre-review checklist
- [ ] This change should be part of a Behavior Change Release. See [go/behavior-change](http://go/behavior-change).
- [ ] This change has passed Merge gate tests
- [ ] Snowpipe Changes
- [ ] Snowpipe Streaming Changes
- [x] This change is TEST-ONLY
- [ ] This change is README/Javadocs only
- [ ] This change is protected by a config parameter <PARAMETER_NAME> eg `snowflake.ingestion.method`.
    - [ ] `Yes` - Added end to end and Unit Tests. 
    - [ ] `No` - Suggest why it is not param protected
- [ ] Is his change protected by parameter <PARAMETER_NAME> on the server side?
    - [ ] The parameter/feature is not yet active in production (partial rollout or PrPr, see [Changes for Unreleased Features and Fixes](http://go/ppp-prpr)).
    - [ ] If there is an issue, it can be safely mitigated by turning the parameter off. This is also verified by a test (See [go/ppp](http://go/ppp)).

1. Suggested reading order:

Go through the commit list and do them one by one. 
or

2.1 Go through automatic refactoring changes

- StreamingClientHandler -> DirectStreamingClientHandler
- extract interface for StreamingClientHandler
- move method DirectStreamingClientHandler#isClientValid -> StreamingClientHandler#isClientValid
- test rename StreamingClientConcurrencyTest -> DirectStreamingClientConcurrencyTest

2.2 Go through fakes code:

- FakeStreamingClientHandler
- FakeSnowflakeStreamingIngestClient
- FakeSnowflakeStreamingIngestChannel

2.3 Check the test using fakes

- StreamingClientProviderConcurrencyTest

PR #711, opened 2023-09-26
Use .equals() to compare schemas

cchandurkar:
If SMT clones the schema of the object for any reason, it does not have the same "hash" as original schema but structurally it's the same. Using `!.equals()` instead of `!=` will allow SMTs to properly clone the schemas.

PR #816, opened 2024-04-02
SNOW-1055561: Check whether SMT returning null values no longer stops a data ingestion.

sfc-gh-akowalczyk:
# Overview

SNOW-1055561

The scenario being tested is:
1. Configure a a connector with a SMT extracting a single field.
2. Produce a few messages, half of them contain the extracted field while the others do not (returns `null` in that case).
3. Check whether all the messages were successfully ingested.

## Pre-review checklist
- [ ] Snowpipe Changes


src/test/java/com/snowflake/kafka/connector/ConnectClusterBaseIT.java
```diff
@@ -75,33 +66,19 @@ public void after() {
...
-            });
-  }
-
-  protected Map<String, String> createProperties() {
+  final Map<String, String> defaultProperties(String topicName, String connectorName) {
     Map<String, String> config = new HashMap<>();
 
     // kafka connect specific
     // real connector will be specified with SNOW-1055561
     config.put(CONNECTOR_CLASS_CONFIG, SnowflakeFakeSinkConnector.class.getName());
```
sfc-gh-lshcharbaty, line 74: As the comment reads, we should use a real connector implementation here instead of `SnowflakeFakeSinkConnector`.
So that we can verify the whole flow of _a real_ implementation.

sfc-gh-lshcharbaty (changes requested):
Please, use a real `SnowflakeSinkConnector` with a [fake streamingClientHandler](https://github.com/snowflakedb/snowflake-kafka-connector/pull/809#discussion_r1528733824) in tests.
Blocking until then.

sfc-gh-gjachimko (approved):

src/test/java/com/snowflake/kafka/connector/ConnectClusterBaseIT.java
sfc-gh-akowalczyk: I replaced it with the real one, could you please check that?

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -38,18 +40,29 @@
...
 
   /**
    * Gets the current streaming provider
    *
    * @return The streaming client provider
    */
   public static StreamingClientProvider getStreamingClientProviderInstance() {
-    return StreamingClientProviderSingleton.streamingClientProvider;
+    if (streamingClientProvider == null) {
+      streamingClientProvider = new StreamingClientProvider(clientHandlerSupplier.get());
```
sfc-gh-lshcharbaty, line 55: Let's add synchronization here.
```java
if (streamingClientProvider == null) {
    synchronized(StreamingClientProvider.class) {
        if (streamingClientProvider == null) {
            streamingClientProvider = new StreamingClientProvider(clientHandlerSupplier.get());
        }
    }
}

return streamingClientProvider;     
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -92,8 +105,8 @@ public static StreamingClientProvider getStreamingClientProviderInstance() {
    * When a client is evicted, the cache will try closing the client, however it is best to still
    * call close client manually as eviction is executed lazily
    */
-  private StreamingClientProvider() {
-    this.streamingClientHandler = new DirectStreamingClientHandler();
+  public StreamingClientProvider(StreamingClientHandler streamingClientHandler) {
```
sfc-gh-lshcharbaty, line 108: I assume we would like to keep a constructor of singleton private.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -38,18 +40,29 @@
...
   public static StreamingClientProvider getStreamingClientProviderInstance() {
-    return StreamingClientProviderSingleton.streamingClientProvider;
+    if (streamingClientProvider == null) {
+      streamingClientProvider = new StreamingClientProvider(clientHandlerSupplier.get());
+    }
+
+    return streamingClientProvider;
+  }
+
+  public static void overrideStreamingClientHandler(StreamingClientHandler streamingClientHandler) {
```
sfc-gh-lshcharbaty, line 61: Let's add a doc explaining it's used for tests only.

src/test/java/com/snowflake/kafka/connector/ConnectClusterBaseIT.java
sfc-gh-lshcharbaty: Are you planning to remove `SnowflakeFakeSinkConnector` in another PR?
We don't need to have once we have a real test.

src/test/java/com/snowflake/kafka/connector/ProduceConsumeMessageIT.java
```diff
@@ -0,0 +1,45 @@
...
+import static org.awaitility.Awaitility.await;
+
+import com.snowflake.kafka.connector.fake.SnowflakeFakeSinkTask;
+import java.util.List;
+import org.apache.kafka.connect.sink.SinkRecord;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class ProduceConsumeMessageIT extends ConnectClusterBaseIT {
```
sfc-gh-lshcharbaty, line 13: Do we need to keep this test?

src/test/java/com/snowflake/kafka/connector/SmtIT.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Map;
+import java.util.function.UnaryOperator;
+import java.util.stream.Stream;
+import org.apache.kafka.connect.json.JsonConverter;
+import org.apache.kafka.connect.storage.StringConverter;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+public class SmtIT extends ConnectClusterBaseIT {
```
sfc-gh-lshcharbaty, line 29: Let's also add a test with `behavior.on.null.values= IGNORE`.

src/test/java/com/snowflake/kafka/connector/SmtIT.java
```diff
@@ -0,0 +1,84 @@
...
+
+  private static final String SMT_TOPIC = "SMT_TOPIC";
+  private static final String SMT_CONNECTOR = "SMT_CONNECTOR";
+  private static final FakeStreamingClientHandler fakeStreamingClientHandler =
+      new FakeStreamingClientHandler();
+
+  @BeforeAll
+  public void createConnector() {
+    StreamingClientProvider.overrideStreamingClientHandler(fakeStreamingClientHandler);
```
sfc-gh-lshcharbaty, line 38: Ideally we also should to revert this override back as other tests can still depend on a real implementation.
To strengthen test isolation will it make sense to create a new `FakeStreamingClientHandler` for each test?

Also, let's move this logic into `ConnectClusterBaseIT`, as I expect every test will override the handler.

src/test/java/com/snowflake/kafka/connector/SmtIT.java
```diff
@@ -0,0 +1,84 @@
...
+  @Test
+  void testIfSmtRetuningNullsIngestDataCorrectly() {
+    Stream.iterate(0, UnaryOperator.identity())
+        .limit(10)
+        .flatMap(v -> Stream.of("{}", "{\"message\":\"value\"}"))
+        .forEach(message -> connectCluster.kafka().produce(SMT_TOPIC, message));
+
+    await()
+        .timeout(Duration.ofSeconds(60))
+        .untilAsserted(() -> assertThat(fakeStreamingClientHandler.ingestedRows()).hasSize(20));
```
sfc-gh-lshcharbaty, line 82: Will it possible to additionally verify `getLatestCommittedOffsetToken()` for the used channels?

sfc-gh-lshcharbaty (commented):
Nice progress! 
Added a few comments to make the things even neater.

src/test/java/com/snowflake/kafka/connector/SmtIT.java
sfc-gh-mbobowski, line 29: Can we setup logging levels for this test? We are flooded with TRACE right now.

src/test/java/com/snowflake/kafka/connector/SmtIT.java
```diff
@@ -0,0 +1,84 @@
...
+    config.put(TRANSFORMS_CONFIG, "extractField");
+    config.put(
+        "transforms.extractField.type", "org.apache.kafka.connect.transforms.ExtractField$Value");
+    config.put("transforms.extractField.field", "message");
+
+    return config;
+  }
+
+  @Test
+  void testIfSmtRetuningNullsIngestDataCorrectly() {
```
sfc-gh-rcheng, line 74: nit: Returning

src/test/java/com/snowflake/kafka/connector/ConnectClusterBaseIT.java
sfc-gh-akowalczyk: removed, the mentioned class and `FakeSinkTask`

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-akowalczyk: done

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
sfc-gh-akowalczyk: done, added annotation as well

src/test/java/com/snowflake/kafka/connector/ProduceConsumeMessageIT.java
sfc-gh-akowalczyk: As we decided to remove the FakeConnector I removed the test

src/test/java/com/snowflake/kafka/connector/SmtIT.java
sfc-gh-akowalczyk: done

src/test/java/com/snowflake/kafka/connector/SmtIT.java
sfc-gh-akowalczyk: Do you have any particular solution in mind?

src/test/java/com/snowflake/kafka/connector/SmtIT.java
sfc-gh-akowalczyk: good idea, added

sfc-gh-rcheng (approved):
thanks! lgtm, lets get the IT passing

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProvider.java
```diff
@@ -52,12 +52,36 @@ public class StreamingClientProvider {
...
+        if (streamingClientProvider == null) {
+          streamingClientProvider = new StreamingClientProvider(clientHandlerSupplier.get());
+        }
+      }
     }
 
     return streamingClientProvider;
   }
 
+  public static void reset() {
```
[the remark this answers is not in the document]
sfc-gh-akowalczyk: That was the intention of the `reset` method, to revert the state to the pre-initialization stage. I changed its implementation a little to reflect the intention better.

I've added the mentioned annotation.

src/test/java/com/snowflake/kafka/connector/SmtIT.java
```diff
@@ -0,0 +1,69 @@
...
+          .untilAsserted(
+              () -> {
+                assertThat(fakeStreamingClientHandler.ingestedRows()).hasSize(expectedRecordNumber);
+                assertThat(fakeStreamingClientHandler.getLatestCommittedOffsetTokensPerChannel())
+                    .hasSize(1)
+                    .containsValue("19");
+              });
+    } finally {
+      connectCluster.kafka().deleteTopic(smtTopic);
+      connectCluster.deleteConnector(smtConnector);
```
sfc-gh-lshcharbaty, line 66: Nit: we can create a topic and generate a connector name as well as clean them up in @BeforeEach/@AfterEach.

sfc-gh-lshcharbaty (approved):
Nice job!
As discussed offline, in the future we can consider injecting `FakeStreamingClientHandler` as a dependency that will allow us to isolate connectors and parallelize test execution.

sfc-gh-xhuang:
While you're testing SMT, can you also take a look at this PR: https://github.com/snowflakedb/snowflake-kafka-connector/pull/711

and merge as needed?

PR #822, opened 2024-04-17
SNOW-1061822: Add new properties for streaming client

sfc-gh-wtrefon:
# Overview

SNOW-1061822

Changes:
New property `snowflake.streaming.max.memory.limit.bytes` introduced to set explicitly MAX_MEMORY_LIMIT_IN_BYTES client sdk property.

Repurpose `buffer.size.bytes` to set buffer in ingest sdk - MAX_CHANNEL_SIZE_IN_BYTES 

## Pre-review checklist
- [ ] Snowpipe Changes
- [x] This change is protected by a config parameter `buffer.enable.single.buffer`.


src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java
```diff
@@ -237,18 +240,22 @@ public static ImmutableMap<String, String> validateStreamingSnowpipeConfig(
             BOOLEAN_VALIDATOR.ensureValid(
                 ERRORS_LOG_ENABLE_CONFIG, inputConfig.get(ERRORS_LOG_ENABLE_CONFIG));
           }
+          if (inputConfig.containsKey(ENABLE_CHANNEL_OFFSET_TOKEN_MIGRATION_CONFIG)) {
```
sfc-gh-wtrefon, line 243: see comment in https://github.com/snowflakedb/snowflake-kafka-connector/pull/822/files#r1568735719

sfc-gh-lshcharbaty (approved):
lgtm
It'd be easier to review if `ConnectorConfigTest` refactoring was in a dedicated pr.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
```diff
@@ -79,28 +85,52 @@ public StreamingClientProperties(Map<String, String> connectorConfig) {
...
     Optional<String> snowpipeStreamingMaxClientLag =
         Optional.ofNullable(connectorConfig.get(SNOWPIPE_STREAMING_MAX_CLIENT_LAG));
     snowpipeStreamingMaxClientLag.ifPresent(
-        overriddenValue -> {
-          parameterOverrides.put(MAX_CLIENT_LAG, String.format("%s second", overriddenValue));
-        });
+        overriddenValue ->
+            parameterOverrides.put(MAX_CLIENT_LAG, String.format("%s second", overriddenValue)));
+
+    if (isSingleBufferEnabled(connectorConfig)) {
```
sfc-gh-rcheng, line 96: question for my understanding - why are we only enabling the max channel and max memory size for the single buffer?

src/test/java/com/snowflake/kafka/connector/ConnectorConfigTest.java
```diff
@@ -53,133 +75,76 @@ public void testConfig_ConvertedInvalidAppName() {
...
+    NAME,
+    SNOWFLAKE_URL,
+    SNOWFLAKE_USER,
+    SNOWFLAKE_DATABASE,
+    SNOWFLAKE_SCHEMA,
+    SNOWFLAKE_PRIVATE_KEY,
+    BUFFER_SIZE_BYTES,
+    BUFFER_COUNT_RECORDS
+  })
+  public void shouldThrowExForEmptyProperty(String prop) {
```
sfc-gh-rcheng, line 116: beautiful, thanks!

src/test/java/com/snowflake/kafka/connector/ConnectorConfigTest.java
```diff
@@ -810,40 +726,32 @@ public void testValidSchematizationForStreamingSnowpipe() {
...
-      config.put(
-          SnowflakeSinkConnectorConfig.VALUE_CONVERTER_CONFIG_FIELD,
-          "org.apache.kafka.connect.storage.StringConverter");
-      config.put(Utils.SF_ROLE, "ACCOUNTADMIN");
-      Utils.validateConfig(config);
-    } catch (SnowflakeKafkaConnectorException exception) {
-      assert exception
-          .getMessage()
-          .contains(SnowflakeSinkConnectorConfig.ENABLE_SCHEMATIZATION_CONFIG);
-    }
```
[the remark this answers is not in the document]
sfc-gh-rcheng: yup, we should be able to support any converter (even though technically customer converters are not allowed) unless they are using schematization

sfc-gh-rcheng (approved):
overall lgtm, left some small notes 

for future PRs:
- Please include a description in the PR next time, you can simply copy the SNOW description over
- you're currently failing the snyk merge gates. This happens sometimes, you'll need to close and reopen the PR to get the gates running properly. it shouldn't ever fail unless you've edited the .pom

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -51,6 +53,9 @@ public class SnowflakeSinkConnectorConfig {
   public static final String BEHAVIOR_ON_NULL_VALUES_CONFIG = "behavior.on.null.values";
 
   // Buffer thresholds
+  public static final String BUFFER_ENABLE_SINGLE_BUFFER = "buffer.enable.single.buffer";
```
sfc-gh-tzhang, line 56: Is there a description on what is this used for?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -129,6 +134,13 @@ public class SnowflakeSinkConnectorConfig {
   public static final String SNOWPIPE_STREAMING_MAX_CLIENT_LAG =
       "snowflake.streaming.max.client.lag";
 
+  public static final String SNOWPIPE_STREAMING_MAX_CHANNEL_SIZE =
+      "snowflake.streaming.max.channel.size";
+  public static final long SNOWPIPE_STREAMING_MAX_CHANNEL_SIZE_DEFAULT = 128 * 1024 * 1024;
+
+  public static final String SNOWPIPE_STREAMING_MAX_MEMORY_LIMIT =
+      "snowflake.streaming.max.memory.limit";
+  public static final long SNOWPIPE_STREAMING_MAX_MEMORY_LIMIT_DEFAULT = -1L;
```
sfc-gh-tzhang, line 143: What's the reason of adding these parameters given that we already have SNOWPIPE_STREAMING_CLIENT_PROVIDER_OVERRIDE_MAP that can control these parameters?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-tzhang: and should we prefix it with `snowflake.streaming`?

sfc-gh-tzhang (changes requested):
I assume this is for removing the double buffer, could you have at least a short doc to describe the plan? It's not clear to me why we're adding a bunch of new parameters instead of using the existing `buffer.*` , thanks!

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -578,6 +595,19 @@ static ConfigDef newConfigDef() {
             6,
             ConfigDef.Width.NONE,
             SNOWPIPE_STREAMING_MAX_CLIENT_LAG)
+        .define(
```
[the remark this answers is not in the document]
sfc-gh-wtrefon: somehow i missed it, nice catch!

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java
```diff
@@ -237,18 +240,22 @@ public static ImmutableMap<String, String> validateStreamingSnowpipeConfig(
...
-                      inputConfig.get(SNOWPIPE_STREAMING_MAX_CLIENT_LAG)));
-            }
+            ensureValidLong(inputConfig, SNOWPIPE_STREAMING_MAX_CLIENT_LAG, invalidParams);
+          }
+
+          if (inputConfig.containsKey(SNOWPIPE_STREAMING_MAX_CHANNEL_SIZE)) {
+            ensureValidLong(inputConfig, SNOWPIPE_STREAMING_MAX_CHANNEL_SIZE, invalidParams);
+          }
+
+          if (inputConfig.containsKey(SNOWPIPE_STREAMING_MAX_MEMORY_LIMIT)) {
```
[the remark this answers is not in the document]
sfc-gh-wtrefon: yup

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
sfc-gh-wtrefon: First thing: isSingleBufferEnabled serves as parameter protection, we don't want to change any logic when the feature is disabled.

Second thing: these properties should only be available when using single buffer, the customer can still change the settings using overrideMap as it was before for multiple buffers but we don't want to explicitly expose them for (it should be somehow hidden)

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-wtrefon: We want to explicitly define them for better visibility and put emphasis on them so the customer is aware that they exist and can be modified. The override map was kinda hidden and only used in rare occasions as stated in the description "optional and recommended to use ONLY after consulting Snowflake Support"

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-wtrefon: added description and new prefix

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -443,6 +440,30 @@ static String validateConfig(Map<String, String> config) {
...
+            Utils.formatString(
+                "{} is only available with ingestion type: {}.",
+                SnowflakeSinkConnectorConfig.SNOWPIPE_STREAMING_MAX_MEMORY_LIMIT,
+                IngestionMethodConfig.SNOWPIPE_STREAMING.toString()));
+      }
+      if (config.containsKey(SnowflakeSinkConnectorConfig.BUFFER_ENABLE_SINGLE_BUFFER)) {
+        invalidConfigParams.put(
+            SnowflakeSinkConnectorConfig.BUFFER_ENABLE_SINGLE_BUFFER,
+            Utils.formatString(
+                "{} is only available with ingestion type: {}.",
```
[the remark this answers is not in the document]
sfc-gh-wtrefon: i left the validation as we need it in the future but added extra one that fails for both ingestion methods, this one should be removed in future when single buffer is implemented.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-tzhang: Yes, that's why I was asking for a design doc to describe the overall plan :) It seems to me that the decision is that we're exposing more parameters for customer to control which is not ideal in terms of ease of use, now they need to understand and use the new parameters instead of the old parameters on `buffer.*`. For example, I was imagine that the buffer size can still be controlled by `buffer.size.bytes` and we can use that value to set the max channel size so that there is no behavior change. 

Another concern is exposing the MAX_MEMORY_LIMIT, I'm not sure customer would be able to understand this easily and use it in a correct way since the number of clients and channels are controlled by Kafka Connector. I don't think we have heard from customers in the past that this needs to be updated.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -126,9 +128,20 @@ public class SnowflakeSinkConnectorConfig {
...
+  public static final String SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER =
+      "snowflake.streaming.enable.single.buffer";
+
+  public static final boolean SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER_DEFAULT = false;
   public static final String SNOWPIPE_STREAMING_MAX_CLIENT_LAG =
       "snowflake.streaming.max.client.lag";
 
+  public static final String SNOWPIPE_STREAMING_MAX_CHANNEL_SIZE =
+      "snowflake.streaming.max.channel.size";
+  public static final long SNOWPIPE_STREAMING_MAX_CHANNEL_SIZE_DEFAULT = 128 * 1024 * 1024;
```
sfc-gh-japatel, line 140: this is a bit confusing, max channel count size? or size of data in it?

what will this do?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-japatel: +1 with @sfc-gh-tzhang 
Please lets add new configs only if it is absolutely needed. 
A design doc describing what this will do would be very helpful.

sfc-gh-gjachimko (approved):

src/test/java/com/snowflake/kafka/connector/ConnectorConfigTest.java
```diff
@@ -1,10 +1,23 @@
...
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWFLAKE_PRIVATE_KEY;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWFLAKE_SCHEMA;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWFLAKE_URL;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWFLAKE_USER;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWPIPE_STREAMING_MAX_CLIENT_LAG;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWPIPE_STREAMING_MAX_MEMORY_LIMIT_IN_BYTES;
 import static com.snowflake.kafka.connector.Utils.HTTP_NON_PROXY_HOSTS;
 import static com.snowflake.kafka.connector.internal.TestUtils.getConfig;
+import static org.assertj.core.api.Assertions.*;
```
sfc-gh-lshcharbaty, line 20: Nit: wildcard import.

sfc-gh-lshcharbaty (approved):
lgtm

PR #824, opened 2024-04-23
SNOW-1269024 Installing multiple Java versions for test purposes

sfc-gh-krosinski:
# Overview

SNOW-1269024 - adding verification on different versions of JRE
Building and unit testing process remains the same (Java8) but e2e testing is run on three different versions of JAVA: 8, 11 and 17. This means that there are 3 times more testing jobs, but they all run in parallel, hence total run time should remain ~the same. 

## Pre-review checklist
- [ ] Snowpipe Changes
    - [x] `No` - This is test only change

*normal* 

Longer test runs

### DIFF
--- a/.github/workflows/End2EndTestApache.yml
+++ b/.github/workflows/End2EndTestApache.yml
@@ -18,9 +18,12 @@ jobs:
     - name: Checkout Code
       uses: actions/checkout@v2
     - name: "Install Java 8"
-      uses: actions/setup-java@v1
+      uses: actions/setup-java@v3
       with:
-        java-version: 1.8
+        java-version: |
+          1.8
+          11
+          17
     - name: "Cache local Maven repository"
       uses: actions/cache@v2
       with:
--- a/.github/workflows/End2EndTestApache.yml
+++ b/.github/workflows/End2EndTestApache.yml
@@ -80,7 +80,7 @@ jobs:
         distribution: 'zulu'
         java-version: ${{ matrix.java_test_version }}
 
-    - name: End to End Test of Apache Platform ${{ matrix.apache_kafka_version }} against Snowflake in ${{ matrix.snowflake_cloud }}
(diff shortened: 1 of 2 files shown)


sfc-gh-tzhang (approved):
LGTM, please see if @sfc-gh-japatel could take a look

.github/workflows/End2EndTestApache.yml
```diff
@@ -18,9 +18,12 @@ jobs:
     - name: Checkout Code
       uses: actions/checkout@v2
     - name: "Install Java 8"
```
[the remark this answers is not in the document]
sfc-gh-krosinski: Decided to go with different approach - here only Java 8 is installed, hence the name remains the same.

sfc-gh-japatel (commented):
any plans for confluent end to end test?

.github/workflows/End2EndTestApache.yml
```diff
@@ -14,13 +14,15 @@ jobs:
       matrix:
         apache_kafka_version: [ '2.5.1', '2.8.1', '3.2.1' ]
         snowflake_cloud: [ 'AWS', 'AZURE', 'GCS' ]
+        java_test_version: [ '8', '11', '17' ]
     steps:
     - name: Checkout Code
       uses: actions/checkout@v2
     - name: "Install Java 8"
```
sfc-gh-japatel, line 21: why do we need this step? as in we are installing java 8 here but using different versions below.

sfc-gh-krosinski: the whole point of my change is to use two different versions of java - java 8 for building and unit testing but then 8, 11 and 17 as jre to run the app and verify it against given runtime environment

sfc-gh-wtrefon (approved):
LGTM

sfc-gh-lshcharbaty (approved):

sfc-gh-dseweryn (approved):
Awesome 💪

PR #830, opened 2024-04-30
SNOW-1348656 Send records that could not be parsed to DLQ

sfc-gh-mbobowski:
# Overview

SNOW-1348656 Kafka Processing error not sent to DLQ correctly.

Prior to this change Snowpipe Streaming with schematization did not send records that could not be parsed to DLQ but throw exception and stop the ingestion instead.

Instead of catching all `SnowflakeKafkaConnectorException` (as suggested in the JIRA) I went with more conservative approach and to it for specific error code only.

## Pre-review checklist
- [ ] Snowpipe Changes


src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1228,7 +1230,7 @@ protected long getApproxSizeOfRecordInBytes(SinkRecord kafkaSinkRecord) {
           }
         }
       }
-    } catch (JsonProcessingException e) {
+    } catch (Exception e) {
       // We ignore any errors here because this is just calculating the record size
```
sfc-gh-wtrefon, line 1234: could we at least log sth here?

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskStreamingDLQTest.java
```diff
@@ -0,0 +1,100 @@
...
+import net.snowflake.ingest.utils.SFException;
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.SchemaAndValue;
+import org.apache.kafka.connect.sink.SinkRecord;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentMatchers;
+import org.mockito.Mockito;
+
+public class SnowflakeSinkTaskStreamingDLQTest {
```
sfc-gh-wtrefon, line 23: i would add one more test case:
when we throw exception that has different error code and is not sent to DLQ but rethrown

sfc-gh-wtrefon (approved):
just one test case missing, overall LGTM

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1320,6 +1322,16 @@ public Pair<List<Map<String, Object>>, List<Long>> getData() {
                 kafkaSinkRecord.kafkaOffset(),
                 kafkaSinkRecord.topic());
             kafkaRecordErrorReporter.reportError(kafkaSinkRecord, e);
```
sfc-gh-gjachimko, line 1324: shouldn't also records, which raised JsonProcessingException end up in the DLQ? 
i mean - we' ve consumed exception ,reported an error and the message is gone.

sfc-gh-wtrefon: but they end up, we call kafkaRecordErrorReporter.reportError(kafkaSinkRecord, e);
What do u mean?

sfc-gh-mbobowski: +1 they are in DLQ.

sfc-gh-gjachimko (approved):

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-lshcharbaty: Let's catch `SnowflakeKafkaConnectorException` with `ERROR_0010` code only. similar to what you do in `getData()` below.

sfc-gh-lshcharbaty (approved):

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-mbobowski: Good point. Done.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1228,8 +1230,16 @@ protected long getApproxSizeOfRecordInBytes(SinkRecord kafkaSinkRecord) {
...
-      // We ignore any errors here because this is just calculating the record size
+    } catch (Exception e) {
+      boolean isJsonProcessingEx = e instanceof JsonProcessingException;
+      boolean isSnowflakeParsingEx =
+          e instanceof SnowflakeKafkaConnectorException
+              && ((SnowflakeKafkaConnectorException) e).checkErrorCode(SnowflakeErrors.ERROR_0010);
+      if (!(isJsonProcessingEx || isSnowflakeParsingEx)) {
+        // We ignore any errors parsing exceptions here because this is just calculating the record
+        // size
+        throw new RuntimeException(e);
```
sfc-gh-tzhang, line 1241: Why we're not throwing the original exception?

sfc-gh-tzhang (approved):
lgtm, thanks!

sfc-gh-japatel (approved):
can you verify if data gets into DLQ in an end to end test scenario?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-mbobowski: `JsonProcessingException` is checked exception so the compiler requires to add throws clause to the signature of `getApproxSizeOfRecordInBytes`. The call stack leads to `SnowflakeSinkTask#put()` which is Kafka Connect method that we cannot modify. At some point we have to wrap checked exception into the unchecked one.

src/test/java/com/snowflake/kafka/connector/SnowflakeSinkTaskStreamingDLQTest.java
sfc-gh-mbobowski: There is no obvious code path to provide such test at task level (checked the one with null records but it turned out to be dead code).
It is possible to mock RecordService to achieve such behaviour but it requires a lot of additional refactoring. I think it is an overkill for that particular fix.

PR #831, opened 2024-05-02
SNOW-1056407 Cleanup metadata properties class. Migrate records tests to junit5

sfc-gh-lshcharbaty:
# Overview

SNOW-1056407
A preliminary PR for introducing a new connector timestamp property.

## Pre-review checklist
- [ ] Snowpipe Changes
    - [x] `No` - Simple refactoring and tests.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/records/RecordService.java
+++ b/src/main/java/com/snowflake/kafka/connector/records/RecordService.java
@@ -24,7 +24,6 @@ import com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig;
 import com.snowflake.kafka.connector.Utils;
 import com.snowflake.kafka.connector.internal.KCLogger;
 import com.snowflake.kafka.connector.internal.SnowflakeErrors;
-import com.snowflake.kafka.connector.internal.telemetry.SnowflakeTelemetryService;
 import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.text.SimpleDateFormat;
@@ -53,6 +52,10 @@ import org.apache.kafka.connect.header.Header;
 import org.apache.kafka.connect.header.Headers;
 import org.apache.kafka.connect.sink.SinkRecord;
 
+/**
+ * Process records output JSON format: <i>{ "meta": { "offset": 123, "topic": "topic name",
+ * "partition": 123, "key":"key name" } "content": "record content" }</i>
+ */
 public class RecordService {
   private final KCLogger LOGGER = new KCLogger(RecordService.class.getName());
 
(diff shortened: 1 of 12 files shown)


src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -93,26 +96,9 @@ public class RecordService {
...
-  public RecordService(SnowflakeTelemetryService telemetryService) {
-    this.telemetryService = telemetryService;
-  }
-
   /** Record service with null telemetry Service, only use it for testing. */
   @VisibleForTesting
-  public RecordService() {
-    this(null);
-  }
+  public RecordService() {}
```
sfc-gh-lshcharbaty, line 101: A new constructor accepting `Clock` is going to be introduced.
That's why this constructor is still marked as `@VisibleForTesting`

src/main/java/com/snowflake/kafka/connector/records/SnowflakeMetadataConfig.java
```diff
@@ -12,64 +19,37 @@ public class SnowflakeMetadataConfig {
...
+        Optional.ofNullable(config.get(property))
+            .orElse(SnowflakeSinkConnectorConfig.SNOWFLAKE_METADATA_DEFAULT);
 
-    createtimeFlag = createtime;
-    topicFlag = topic;
-    offsetAndPartitionFlag = offsetAndPartition;
-    allFlag = all;
+    // Cannot use Boolean.parseBoolean in order to ensure backward compatibility.
+    // The equality check must be case sensitive.
+    return "true".equals(value);
```
sfc-gh-gjachimko, line 44: can't we break that contract? i mean - using case insensitive comparison wouldn't break old customers but would be a a nice improvement for new installations.

sfc-gh-gjachimko (approved):

src/main/java/com/snowflake/kafka/connector/records/SnowflakeMetadataConfig.java
sfc-gh-lshcharbaty: I doubt anyone will use uppercase "TRUE" value to disable something.
I'd like this pr to introduce no changes, so I can change it in the following one.

sfc-gh-tzhang: Create a JIRA or add TODO to make sure this won't fall out of the cracks?

sfc-gh-tzhang (approved):
lgtm, thanks!

src/main/java/com/snowflake/kafka/connector/records/SnowflakeMetadataConfig.java
sfc-gh-lshcharbaty: [It's done already](https://github.com/snowflakedb/snowflake-kafka-connector/pull/833/files#diff-cf910e10942ebdab56b8d5ab0021e53cf6ee64f7105482fcdca09a0c2ce41797R45) in one of the PRs following.

PR #832, opened 2024-05-02
SNOW-1056407 Add tests MetaColumnTest for snowpipe streaming

sfc-gh-lshcharbaty:
# Overview

SNOW-1056407
Add metadata column tests for Snowpipe Streaming.
Depends on #831.

## Pre-review checklist
- [ ] Snowpipe Changes


src/test/java/com/snowflake/kafka/connector/builder/SinkRecordBuilder.java
```diff
@@ -50,4 +54,10 @@ public SinkRecordBuilder withOffset(long offset) {
     this.offset = offset;
     return this;
   }
+
+  public SinkRecordBuilder withTimestamp(long timestamp, TimestampType timestampType) {
```
sfc-gh-gjachimko, line 58: probably overkill, but should we assert on NO_TIMESTAMP_TYPE ? it seems like this value wouldn't be allowed for non-null timestamp value.

sfc-gh-gjachimko (approved):

PR #833, opened 2024-05-02
SNOW-1056407 Add SnowflakeConnectorPushTime to metadata

sfc-gh-lshcharbaty:
# Overview

SNOW-1056407
Adding a new `SnowflakeConnectorPushTime` property in metadata that represents a moment in time when the message was pushed further by the connector.

Depends on: #832.

## Pre-review checklist
- [x] This change should be part of a Behavior Change Release. See [go/behavior-change](http://go/behavior-change).
- [ ] Snowpipe Changes
- [x] This change is protected by a config parameter `snowflake.streaming.metadata.connectorPushTime`.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -98,13 +98,15 @@ public class SnowflakeSinkConnectorConfig {
   // Snowflake Metadata Flags
   private static final String SNOWFLAKE_METADATA_FLAGS = "Snowflake Metadata Flags";
   public static final String SNOWFLAKE_METADATA_CREATETIME = "snowflake.metadata.createtime";
-  public static final String SNOWFLAKE_METADATA_CONNECTOR_PUSH_TIME =
-      "snowflake.metadata.connectorPushTime";
   public static final String SNOWFLAKE_METADATA_TOPIC = "snowflake.metadata.topic";
   public static final String SNOWFLAKE_METADATA_OFFSET_AND_PARTITION =
       "snowflake.metadata.offset.and.partition";
   public static final String SNOWFLAKE_METADATA_ALL = "snowflake.metadata.all";
   public static final String SNOWFLAKE_METADATA_DEFAULT = "true";
+
+  public static final String SNOWFLAKE_STREAMING_METADATA_CONNECTOR_PUSH_TIME =
+      "snowflake.streaming.metadata.connectorPushTime";
+  public static final boolean SNOWFLAKE_STREAMING_METADATA_CONNECTOR_PUSH_TIME_DEFAULT = false;
 
   // Where is Kafka hosted? self, confluent or any other in future.
   // By default it will be None since this is not enforced and only used for monitoring
@@ -513,6 +515,17 @@ public class SnowflakeSinkConnectorConfig {
             3,
             ConfigDef.Width.NONE,
             SNOWFLAKE_METADATA_OFFSET_AND_PARTITION)
+        .define(
+            SNOWFLAKE_STREAMING_METADATA_CONNECTOR_PUSH_TIME,
+            Type.BOOLEAN,
+            SNOWFLAKE_STREAMING_METADATA_CONNECTOR_PUSH_TIME_DEFAULT,
+            Importance.LOW,
+            "Flag to control whether ConnectorPushTime is collected in snowflake metadata for"
+                + " Snowpipe Streaming",
+            SNOWFLAKE_METADATA_FLAGS,
+            4,
+            ConfigDef.Width.NONE,
+            SNOWFLAKE_STREAMING_METADATA_CONNECTOR_PUSH_TIME)
         .define(
             PROVIDER_CONFIG,
             Type.STRING,
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -98,7 +98,8 @@ public class SnowflakeSinkConnectorConfig {
   // Snowflake Metadata Flags
   private static final String SNOWFLAKE_METADATA_FLAGS = "Snowflake Metadata Flags";
   public static final String SNOWFLAKE_METADATA_CREATETIME = "snowflake.metadata.createtime";
-  public static final String SNOWFLAKE_METADATA_CONNECTOR_TIME = "snowflake.metadata.connectorTime";
+  public static final String SNOWFLAKE_METADATA_CONNECTOR_PUSH_TIME =
+      "snowflake.metadata.connectorPushTime";
   public static final String SNOWFLAKE_METADATA_TOPIC = "snowflake.metadata.topic";
   public static final String SNOWFLAKE_METADATA_OFFSET_AND_PARTITION =
       "snowflake.metadata.offset.and.partition";
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -106,7 +106,7 @@ public class SnowflakeSinkConnectorConfig {
 
   public static final String SNOWFLAKE_STREAMING_METADATA_CONNECTOR_PUSH_TIME =
       "snowflake.streaming.metadata.connectorPushTime";
-  public static final boolean SNOWFLAKE_STREAMING_METADATA_CONNECTOR_PUSH_TIME_DEFAULT = false;
+  public static final boolean SNOWFLAKE_STREAMING_METADATA_CONNECTOR_PUSH_TIME_DEFAULT = true;
 
   // Where is Kafka hosted? self, confluent or any other in future.
   // By default it will be None since this is not enforced and only used for monitoring
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -98,6 +98,7 @@ public class SnowflakeSinkConnectorConfig {
   // Snowflake Metadata Flags
   private static final String SNOWFLAKE_METADATA_FLAGS = "Snowflake Metadata Flags";
   public static final String SNOWFLAKE_METADATA_CREATETIME = "snowflake.metadata.createtime";
+  public static final String SNOWFLAKE_METADATA_CONNECTOR_TIME = "snowflake.metadata.connectorTime";
   public static final String SNOWFLAKE_METADATA_TOPIC = "snowflake.metadata.topic";
   public static final String SNOWFLAKE_METADATA_OFFSET_AND_PARTITION =
       "snowflake.metadata.offset.and.partition";
--- a/src/main/java/com/snowflake/kafka/connector/records/RecordService.java
+++ b/src/main/java/com/snowflake/kafka/connector/records/RecordService.java
@@ -27,6 +27,7 @@ import com.snowflake.kafka.connector.internal.SnowflakeErrors;
 import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.text.SimpleDateFormat;
+import java.time.Clock;
 import java.time.Instant;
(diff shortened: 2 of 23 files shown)


src/test/java/com/snowflake/kafka/connector/records/ProcessRecordTest.java
```diff
@@ -54,76 +61,79 @@ public static Iterable<Case> data() throws IOException {
             getAvro(),
             MAPPER.readTree(
                 "{\"content\":{\"int\":222},\"meta\":{\"offset\":0,\"topic\":\"test\",\"partition\":0,\"schema_id\":1,\"key\":\"string"
-                    + " value\"}}")),
+                    + " value\",\"SnowflakeConnectorTime\":1714655000000}}")),
```
sfc-gh-lshcharbaty, line 64: I'll refactor the json creation in the next pr.

sfc-gh-gjachimko (approved):

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -193,6 +201,10 @@ private SnowflakeTableRow processRecord(SinkRecord record) {
       meta.put(SCHEMA_ID, valueContent.getSchemaID());
     }
 
+    if (metadataConfig.connectorTimeFlag) {
+      meta.put(CONNECTOR_TIME, clock.instant().toEpochMilli());
```
sfc-gh-xhuang, line 205: could we get the timestamp of when insertRow(s) to the streaming SDK happened?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -98,6 +98,7 @@ public class SnowflakeSinkConnectorConfig {
   // Snowflake Metadata Flags
   private static final String SNOWFLAKE_METADATA_FLAGS = "Snowflake Metadata Flags";
   public static final String SNOWFLAKE_METADATA_CREATETIME = "snowflake.metadata.createtime";
+  public static final String SNOWFLAKE_METADATA_CONNECTOR_TIME = "snowflake.metadata.connectorTime";
```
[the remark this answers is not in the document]
sfc-gh-xhuang: approximate would be too long of a text to add in there. We can put that in documentation. 

I like the suggestion in the jira:

snowflake.metadata.connectorPushTime 

or snowflake.metadata.flushTime

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-lshcharbaty: This method is called right before pushing data to ingest-sdk. The metadata is written as json immediately, so if we want to call it in the exact place where there rows are being pushed further, refactoring of the record service is needed.

That's the call stack:
`TopicPartitionChannel.insertRowsWithFallback` -> `InsertRowsApiResponseSupplier` -> `StreamingBuffer.getData()` -> `RecordService.getProcessedRecordForStreamingIngest` -> `RecordService.processRecord`.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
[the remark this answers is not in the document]
sfc-gh-lshcharbaty: Changed to `connectorPushTime`

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -69,6 +71,7 @@ public class RecordService {
   static final String CONTENT = "content";
   static final String META = "meta";
   static final String SCHEMA_ID = "schema_id";
+  static final String CONNECTOR_PUSH_TIME = "ConnectorPushTime";
```
sfc-gh-japatel, line 74: we have used all lower case historically
to be consistent, may be use `connector_push_time`?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -98,6 +98,8 @@ public class SnowflakeSinkConnectorConfig {
   // Snowflake Metadata Flags
   private static final String SNOWFLAKE_METADATA_FLAGS = "Snowflake Metadata Flags";
   public static final String SNOWFLAKE_METADATA_CREATETIME = "snowflake.metadata.createtime";
+  public static final String SNOWFLAKE_METADATA_CONNECTOR_PUSH_TIME =
+      "snowflake.metadata.connectorPushTime";
```
sfc-gh-japatel, line 102: this field is applicable to streaming only I suppose, good to add comment. 
adding streaming specific name would mean we wont be able to add for snowpipe in future. SNowpipe wont need it since they have commit time in load history but just FYI.

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-japatel: nvm, I see `CreateTime`

test/test_suit/test_avro_avro.py
```diff
@@ -31,7 +31,7 @@ def verify(self, round):
         # validate content of line 1
         res = self.driver.snowflake_conn.cursor().execute(
             "Select * from {} limit 1".format(self.topic)).fetchone()
-        goldMeta = r'{"CreateTime":\d*,"key":[{"timestamp":\d*,"tweet":"Rock:Nerfpaper,scissorsisfine.",' \
+        goldMeta = r'{"CreateTime":\d*,"SnowflakeConnectorTime":\d*,"key":[{"timestamp":\d*,"tweet":"Rock:Nerfpaper,scissorsisfine.",' \
```
sfc-gh-japatel, line 34: is the default set to true? this seems like a breaking change?

sfc-gh-japatel (approved):
lgtm overall, I will let @sfc-gh-xhuang pitch in on the name of the field. 

my main concern is test changes, since default is false, why are null fields showing up in `goldMeta`? If they are just regex, please ignore. 

It would be good to add a test for enabling/disabling the parameter to verify expected behavior.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-tzhang: Agree, I think it's always good to be more clear in the property name even though more detail will be in the comment, I suggest we call it `snowflake.streaming.metadata.approximate.flushtime` (or pushtime). I would not concern about it being too long since it won't see part of the every row, plus we have pretty long ones already like `"value.converter.basic.auth.credentials.source"`

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-tzhang: should we just have `PushTime` (or flush) time to make it consistent with `CreateTime`?

test/test_suit/test_avro_avro.py
sfc-gh-tzhang: +1, default should be false. Also the name doesn't seem to much with above

sfc-gh-tzhang (approved):
- Left some suggestions about the naming, feel free to discuss it with the team and PMs and you can make the final call
- The default of this property should be false at the beginning, we can consider enable it some time later by default as a behavior change

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-lshcharbaty: Imo `PushTime` sounds quite ambiguous. It isn't clear where exactly the push happened. Ideally I'd have `SnowflakeConnectorPushTime`, but it's too long and I reckon in most cases it should be obvious for users what connector is implied. 

I'd like not to use "flush", as we still have an ingest-sdk that buffers data, so it isn't a time when a real flush to Snowflake happens. Also, after removing a buffering layer in the connector, nothing is going to be flushed as records will be passed to ingest-sdk immediately.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-lshcharbaty: Adding `streaming` in the property name makes perfect sense. I believe `snowflake.streaming.metadata.connectorPushTime` is quite descriptive.

test/test_suit/test_avro_avro.py
sfc-gh-lshcharbaty: Turned the parameter `false` by default. Yes, the property name in the test had not been updated.

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-japatel: I vote for `SnowflakeConnectorPushTime` too. 

TBH, it should be `SnowflakeConnectorPullTime` to be very specific. (Kafka connect polling so w.r.t snowflake it is pull time)

i am fine with either given we will and should be very descriptive in our docs.

src/main/java/com/snowflake/kafka/connector/records/SnowflakeMetadataConfig.java
```diff
@@ -1,9 +1,6 @@
 package com.snowflake.kafka.connector.records;
 
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWFLAKE_METADATA_ALL;
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWFLAKE_METADATA_CREATETIME;
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWFLAKE_METADATA_OFFSET_AND_PARTITION;
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWFLAKE_METADATA_TOPIC;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.*;
```
sfc-gh-japatel, line 3: all star import. probably an intellij senting

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-japatel: I would even go with adding `Ms` in suffix

sfc-gh-japatel (approved):
lgtm, left a comment. 
Thanks for reverting to false by default.

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-xhuang: I'm avoiding pull for pull-kafka reasons. 

I'm not sure if ms is necessary as the other createtime doesn;t have it and it's also millisecond value?

test/test_suit/test_avro_avro.py
sfc-gh-xhuang: We can have client breaking changes. It just needs to be documented. 
I don't think having it default true is a bad thing

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-lshcharbaty: 1. After reconsideration, naming the field as `SnowflakeConnectorPushTime` makes more sense, as customers can use some source connector, so it isn't clear which one's timestamp it is.
1. I wouldn't go with _ms_ suffix in order to align with the Kafka's naming for timestamp: `CreateTime`, `LogAppendTime`.

test/test_suit/test_avro_avro.py
sfc-gh-lshcharbaty: Shouldn't the connector follow BCR too?
If not then both options sound alright

sfc-gh-mbobowski (approved):
LGTM - I would consider adding TimeProvider abstraction

test/test_suit/test_avro_avro.py
sfc-gh-xhuang: Client BCRs are different than Snowflake server release BCRs. Yes, there is still approval and review but to customers, there is a notice and customers can choose to upgrade to the newest client version. 
I'll forward the internal doc.

sfc-gh-lshcharbaty:
Since `SnowflakeConnectorPushTime` is included in `record_metadata` by default, this change is a subject to BCR

PR #839, opened 2024-05-07
SNOW-1369271 Ignore SFException when closing a channel

sfc-gh-lshcharbaty:
# Overview

SNOW-1369271
[StreamingIngestChannel can throw an SFException](https://github.com/snowflakedb/snowflake-ingest-java/blob/master/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java#L285) when calling close. That causes connector tasks to be stopped. 

Ignoring `SFException` when channels are invalid remediates this issue. 

## Pre-review checklist
- [ ] Snowpipe Changes
    - [x] `No` - Straightforward change.

## Urgency
This review is **high** priority

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
@@ -1082,7 +1082,7 @@ public class TopicPartitionChannel {
       // telemetry and metrics
       this.telemetryServiceV2.reportKafkaPartitionUsage(this.snowflakeTelemetryChannelStatus, true);
       this.snowflakeTelemetryChannelStatus.tryUnregisterChannelJMXMetrics();
-    } catch (InterruptedException | ExecutionException e) {
+    } catch (InterruptedException | ExecutionException | SFException e) {
       final String errMsg =
           String.format(
               "Failure closing Streaming Channel name:%s msg:%s",
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
@@ -36,6 +36,7 @@ import dev.failsafe.function.CheckedSupplier;
 import java.io.ByteArrayOutputStream;
 import java.io.ObjectOutputStream;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
@@ -1088,7 +1089,12 @@ public class TopicPartitionChannel {
               "Failure closing Streaming Channel name:%s msg:%s",
               this.getChannelNameFormatV1(), e.getMessage());
       this.telemetryServiceV2.reportKafkaConnectFatalError(errMsg);
-      LOGGER.error(errMsg, e);
+      LOGGER.error(
+          "Closing Streaming Channel={} encountered an exception {}: {} {}",
+          this.getChannelNameFormatV1(),
+          e.getClass(),
+          e.getMessage(),
+          Arrays.toString(e.getStackTrace()));
     }
   }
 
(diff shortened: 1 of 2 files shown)


sfc-gh-wtrefon (approved):

sfc-gh-mbobowski (approved):

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelTest.java
```diff
@@ -251,6 +251,71 @@ public void testCloseChannelException() throws Exception {
...
+            mockKafkaRecordErrorReporter,
+            mockSinkTaskContext,
+            mockSnowflakeConnectionService,
+            new RecordService(),
+            mockTelemetryService,
+            false,
+            null);
+
+    // when
+    topicPartitionChannel.closeChannel();
```
sfc-gh-mbobowski, line 281: Adding `Assertions.assertDoesNotThrow` would make the reason of this test more explicit.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1082,7 +1082,7 @@ public void closeChannel() {
       // telemetry and metrics
       this.telemetryServiceV2.reportKafkaPartitionUsage(this.snowflakeTelemetryChannelStatus, true);
       this.snowflakeTelemetryChannelStatus.tryUnregisterChannelJMXMetrics();
-    } catch (InterruptedException | ExecutionException e) {
+    } catch (InterruptedException | ExecutionException | SFException e) {
       final String errMsg =
```
sfc-gh-japatel, line 1086: can you also add stacktrace here in logger.

sfc-gh-japatel (approved):
thank you! lgtm barring a minor comment

sfc-gh-japatel:
For future records, please add in description why catching the exception is fine.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1082,7 +1082,7 @@ public void closeChannel() {
       // telemetry and metrics
       this.telemetryServiceV2.reportKafkaPartitionUsage(this.snowflakeTelemetryChannelStatus, true);
       this.snowflakeTelemetryChannelStatus.tryUnregisterChannelJMXMetrics();
-    } catch (InterruptedException | ExecutionException e) {
+    } catch (InterruptedException | ExecutionException | SFException e) {
```
sfc-gh-tzhang, line 1085: Almost all the exceptions from the SDK will be a SFException, it's not ideal, but should we check the specific error code?

sfc-gh-tzhang: Ideally we should add a closeWithoutVerification function in the SDK and use that instead

sfc-gh-tzhang (approved):
lgtm, left a suggestion, thanks

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-lshcharbaty: After a discussion with @sfc-gh-japatel we've decided to catch all `SFException`s, as no matter what the problem is, we don't want to fail a connector task because of a channel.

sfc-gh-tzhang: In general, it's an anti-pattern to catch all exceptions. This would just masks other unexpected exceptions and it's always better to fail faster IMO.

sfc-gh-japatel: I agree this is catch all exception but this is catch all only for closeChannel right? I think idea is to not have any failure when there is a shutdown since the channels are being opened in task open API anyways.

PR #840, opened 2024-05-07
SNOW-988870 Update io.confluent packages to v7.6.0

sfc-gh-xhuang:
# Overview
SNOW-988870 Update io.confluent packages to v7.6.0

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
   ~ /*
-  ~  * Copyright (c) 2019 - 2020 Snowflake Computing Inc. All rights reserved.
+  ~  * Copyright (c) 2019 - 2024 Snowflake Computing Inc. All rights reserved.
   ~  */
   -->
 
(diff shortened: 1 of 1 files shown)


sfc-gh-japatel:
Tests are failing! 
also, I just realized, we should remove 5.5.x from our supported versions since confluent doesnt support that anymore. https://docs.confluent.io/platform/current/installation/versions-interoperability.html

sfc-gh-xhuang:
@sfc-gh-japatel Seem only the non-required tests are failing? 
I looked at our docs, it doesn't seem like we claim 5.5.x support 
https://docs.snowflake.com/en/user-guide/kafka-connector-install#installation-prerequisites

The only note we have for it is here: https://docs.snowflake.com/en/user-guide/kafka-connector-protobuf#configuring-the-confluent-version-of-the-protobuf-converter
which can either stay or be removed.

sfc-gh-japatel:
> @sfc-gh-japatel Seem only the non-required tests are failing? I looked at our docs, it doesn't seem like we claim 5.5.x support https://docs.snowflake.com/en/user-guide/kafka-connector-install#installation-prerequisites
> 
> The only note we have for it is here: https://docs.snowflake.com/en/user-guide/kafka-connector-protobuf#configuring-the-confluent-version-of-the-protobuf-converter which can either stay or be removed.

I suggest we remove them since it;s failing. 

I understand they are Non-required gates but we should investigate why they are failing with version bumps.

sfc-gh-mbobowski:
It duplicates https://github.com/snowflakedb/snowflake-kafka-connector/pull/854
Closing

PR #841, opened 2024-05-08
SNOW-1369280 Close channels asynchronously

sfc-gh-lshcharbaty:
# Overview

SNOW-1369280
By closing channels in parallel, speeds up task rebalancing. As in  #839, `SFException` is ignored and doesn't fail a connector task.

## Pre-review checklist
- [ ] Snowpipe Changes
- [x] This change is protected by a config parameter `snowflake.streaming.closeChannelsInParallel.enabled`.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -125,6 +125,11 @@ public class SnowflakeSinkConnectorConfig {
   public static final boolean SNOWPIPE_FILE_CLEANER_FIX_ENABLED_DEFAULT = false;
   public static final int SNOWPIPE_FILE_CLEANER_THREADS_DEFAULT = 1;
 
+  // Whether to close streaming channels in parallel.
+  public static final String SNOWPIPE_STREAMING_CLOSE_CHANNELS_IN_PARALLEL =
+      "snowflake.streaming.closeChannelsInParallel.enabled";
+  public static final boolean SNOWPIPE_STREAMING_CLOSE_CHANNELS_IN_PARALLEL_DEFAULT = false;
+
   // This is the streaming max client lag which can be defined in config
   public static final String SNOWPIPE_STREAMING_MAX_CLIENT_LAG =
       "snowflake.streaming.max.client.lag";
@@ -567,6 +572,13 @@ public class SnowflakeSinkConnectorConfig {
             Importance.LOW,
             "Defines number of worker threads to associate with the cleaner task. By default there"
                 + " is one cleaner per topic's partition and they all share one worker thread")
+        .define(
+            SNOWPIPE_STREAMING_CLOSE_CHANNELS_IN_PARALLEL,
+            Type.BOOLEAN,
+            SNOWPIPE_STREAMING_CLOSE_CHANNELS_IN_PARALLEL_DEFAULT,
+            Importance.MEDIUM,
+            "Whether to close Snowpipe Streaming channels in parallel during task shutdown or"
+                + " rebalancing")
         .define(
             SNOWPIPE_STREAMING_MAX_CLIENT_LAG,
             Type.LONG,
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
@@ -3,6 +3,8 @@ package com.snowflake.kafka.connector.internal.streaming;
 import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.BUFFER_SIZE_BYTES_DEFAULT;
 import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.ENABLE_STREAMING_CLIENT_OPTIMIZATION_DEFAULT;
 import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWFLAKE_ROLE;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWPIPE_STREAMING_CLOSE_CHANNELS_IN_PARALLEL;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWPIPE_STREAMING_CLOSE_CHANNELS_IN_PARALLEL_DEFAULT;
 import static com.snowflake.kafka.connector.internal.streaming.StreamingUtils.STREAMING_BUFFER_COUNT_RECORDS_DEFAULT;
 import static com.snowflake.kafka.connector.internal.streaming.StreamingUtils.STREAMING_BUFFER_FLUSH_TIME_DEFAULT_SEC;
 import static com.snowflake.kafka.connector.internal.streaming.TopicPartitionChannel.NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE;
@@ -27,6 +29,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import net.snowflake.ingest.streaming.SnowflakeStreamingIngestClient;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.connect.sink.SinkRecord;
@@ -92,7 +95,9 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
   // Config set in JSON
   private final Map<String, String> connectorConfig;
 
-  private boolean enableSchematization;
+  private final boolean enableSchematization;
+
+  private final boolean closeChannelsInParallel;
 
   /**
    * Key is formulated in {@link #partitionChannelKey(String, int)} }
@@ -130,6 +135,11 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
     this.enableSchematization =
         this.recordService.setAndGetEnableSchematizationFromConfig(this.connectorConfig);
 
+    this.closeChannelsInParallel =
+        Optional.ofNullable(connectorConfig.get(SNOWPIPE_STREAMING_CLOSE_CHANNELS_IN_PARALLEL))
+            .map(Boolean::parseBoolean)
+            .orElse(SNOWPIPE_STREAMING_CLOSE_CHANNELS_IN_PARALLEL_DEFAULT);
+
     this.streamingIngestClient =
         StreamingClientProvider.getStreamingClientProviderInstance()
             .getClient(this.connectorConfig);
@@ -162,6 +172,7 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
       SnowflakeStreamingIngestClient streamingIngestClient,
       Map<String, String> connectorConfig,
       boolean enableSchematization,
+      boolean closeChannelsInParallel,
       Map<String, TopicPartitionChannel> partitionsToChannel) {
     this.flushTimeSeconds = flushTimeSeconds;
     this.fileSizeBytes = fileSizeBytes;
@@ -180,6 +191,7 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
         StreamingClientProvider.getStreamingClientProviderInstance()
             .getClient(this.connectorConfig);
     this.enableSchematization = enableSchematization;
+    this.closeChannelsInParallel = closeChannelsInParallel;
     this.partitionsToChannel = partitionsToChannel;
 
     this.tableName2SchemaEvolutionPermission = new HashMap<>();
@@ -349,15 +361,40 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
 
   @Override
   public void closeAll() {
+    if (closeChannelsInParallel) {
+      closeAllInParallel();
+    } else {
+      closeAllSequentially();
+    }
+
+    partitionsToChannel.clear();
+
+    StreamingClientProvider.getStreamingClientProviderInstance()
+        .closeClient(this.connectorConfig, this.streamingIngestClient);
+  }
+
+  private void closeAllSequentially() {
     partitionsToChannel.forEach(
         (partitionChannelKey, topicPartitionChannel) -> {
           LOGGER.info("Closing partition channel:{}", partitionChannelKey);
           topicPartitionChannel.closeChannel();
         });
-    partitionsToChannel.clear();
+  }
 
-    StreamingClientProvider.getStreamingClientProviderInstance()
-        .closeClient(this.connectorConfig, this.streamingIngestClient);
+  private void closeAllInParallel() {
+    CompletableFuture<?>[] futures =
+        partitionsToChannel.entrySet().stream()
+            .map(
+                entry -> {
+                  String channelKey = entry.getKey();
+                  TopicPartitionChannel topicPartitionChannel = entry.getValue();
+
+                  LOGGER.info("Closing partition channel:{}", channelKey);
+                  return topicPartitionChannel.closeChannelAsync();
+                })
+            .toArray(CompletableFuture[]::new);
+
+    CompletableFuture.allOf(futures).join();
   }
 
   /**
@@ -374,6 +411,20 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
    */
   @Override
   public void close(Collection<TopicPartition> partitions) {
+    if (closeChannelsInParallel) {
+      closeInParallel(partitions);
+    } else {
+      closeSequentially(partitions);
+    }
+
+    LOGGER.info(
+        "Closing {} partitions and remaining partitions which are not closed are:{}, with size:{}",
+        partitions.size(),
+        partitionsToChannel.keySet().toString(),
+        partitionsToChannel.size());
+  }
+
+  private void closeSequentially(Collection<TopicPartition> partitions) {
     partitions.forEach(
         topicPartition -> {
           final String partitionChannelKey =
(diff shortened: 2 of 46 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -349,15 +361,40 @@ public void callAllGetOffset() {
...
-    partitionsToChannel.clear();
+  }
 
-    StreamingClientProvider.getStreamingClientProviderInstance()
-        .closeClient(this.connectorConfig, this.streamingIngestClient);
+  private void closeAllInParallel() {
+    CompletableFuture<?>[] futures =
+        partitionsToChannel.entrySet().stream()
+            .map(
+                entry -> {
```
sfc-gh-gjachimko, line 388: i think i'd combine the two variants of the close methods together,
just replace the return with
if  (closeChannelsInParallel) {
return topicPartitionChannel.closeChannelAsync() ;
} else {
   topicPartitionChannel.closeChannel();
 return CompletableFuture.completedFuture();
}

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -125,6 +125,11 @@ public class SnowflakeSinkConnectorConfig {
   public static final boolean SNOWPIPE_FILE_CLEANER_FIX_ENABLED_DEFAULT = false;
   public static final int SNOWPIPE_FILE_CLEANER_THREADS_DEFAULT = 1;
 
+  // Whether to close streaming channels in parallel.
+  public static final String SNOWPIPE_STREAMING_CLOSE_CHANNELS_IN_PARALLEL =
+      "snowflake.streaming.closeChannelsInParallel.enabled";
+  public static final boolean SNOWPIPE_STREAMING_CLOSE_CHANNELS_IN_PARALLEL_DEFAULT = false;
```
sfc-gh-tzhang, line 131: This doesn't feel like a configuration should be exposed to customer. As a customer, I don't know whether I should close the channels in parallel or not. In fact, I don't really care, I just want my system to run without any failure. WDYT?

sfc-gh-japatel: agreed but I think we want to try this out for customers who has a lot of partitions in their single task. so enabling by default will not benefit a lot of customers 
- also its good to have param protection. 
- We should stress test this and slow roll out in couple of releases.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -392,11 +443,33 @@ public void close(Collection<TopicPartition> partitions) {
...
-        partitionsToChannel.size());
+        "Closing partitionChannel:{}, partition:{}, topic:{}",
+        topicPartitionChannel == null ? null : topicPartitionChannel.getChannelNameFormatV1(),
+        topicPartition.topic(),
+        topicPartition.partition());
+
+    // It's possible that some partitions can be unassigned before their respective channels are
+    // even created.
+    return topicPartitionChannel == null
+        ? CompletableFuture.completedFuture(null)
```
sfc-gh-japatel, line 469: thinking out loud, does this need a NPE handling? or we are fine in `closeInParallel`?

test/rest_request_template/schema_evolution_nullable_values_after_smt.json
```diff
@@ -15,6 +15,7 @@
     "snowflake.schema.name": "SNOWFLAKE_SCHEMA",
     "snowflake.role.name": "SNOWFLAKE_ROLE",
     "snowflake.ingestion.method": "SNOWPIPE_STREAMING",
+    "snowflake.streaming.closeChannelsInParallel.enabled": true,
```
sfc-gh-japatel, line 18: can we add a test with false to make sure rollback is working like current behavior?

sfc-gh-japatel (commented):
added couple of comments but overall LGTM!

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-xhuang: What's the risk of having it default true and then getting rid of it in a future release?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-lshcharbaty: I want to keep the existing flow intact. Yes that introduces some duplication but I know for sure that nothing is changed when the parameter is disabled.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-lshcharbaty: @sfc-gh-tzhang 
This is not a publicly announced parameter. It's going to behave as a knob that's going to allow tho customer to easily roll the change back in case something goes wrong. Exactly what @sfc-gh-japatel wrote.

@sfc-gh-xhuang 
Imo it's better to be safe than sorry.  I'm not sure how the parallelization is going to work on the customer environments as channel closing is assigned to a shared `ForkJoinPool` that seems to exist as a single instance and shared across all tasks on the same host.

I suggest we approach the "change rollout" parameters the following way:
1. Protect the change with a parameter and disable it by default. If some customers experience problems that can be solved by the change, they can turn it on. 
1. Enable the change by default. If some customers experience problems that cause by the change, they can turn it off and communicate to us.
1. Get rid of the older implementation.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -349,15 +361,40 @@ public void callAllGetOffset() {
...
+                entry -> {
+                  String channelKey = entry.getKey();
+                  TopicPartitionChannel topicPartitionChannel = entry.getValue();
+
+                  LOGGER.info("Closing partition channel:{}", channelKey);
+                  return topicPartitionChannel.closeChannelAsync();
+                })
+            .toArray(CompletableFuture[]::new);
+
+    CompletableFuture.allOf(futures).join();
```
sfc-gh-wtrefon, line 397: can't we reuse here the closeInParallel(Collection<TopicPartition> partitions) and pass to it `partitionsToChannels`

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-lshcharbaty: No, null handling isn't needed. 
1. The returned future carries a `Void` value, so only `null` can be passed in there.
2. We don't do anything with the values returned by futures. The only thing we care about is their completion.

test/rest_request_template/schema_evolution_nullable_values_after_smt.json
sfc-gh-lshcharbaty: I'll try to test this in a unit test.
Each e2e test is run about 45 times, so I'd like to avoid writing e2e's for functionality that can be tested within a quicker set of tests.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-gjachimko: perhaps these futures could return "key" instead - then you could simply pass method pointer to the thenAccept(partitonsToChannel::remove) ?

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1098,6 +1103,59 @@ public void closeChannel() {
...
+        .exceptionally(this::tryRecoverFromCloseChannelError);
+  }
+
+  private CompletableFuture<Void> closeChannelWrapped() {
+    try {
+      return this.channel.close();
+    } catch (SFException e) {
+      // Calling channel.close() can throw an SFException instead of putting it into a completable
+      // future.
+      CompletableFuture<Void> future = new CompletableFuture<>();
```
sfc-gh-gjachimko, line 1122: this comment is a bit odd - it took me a few reads to understand your point. 
i'd either get rid of it completely - i think it is clear that the sync part of the channel.close can throw or rephrase it to something like "wrap SFException into CompletableFuture to keep consistent method chain"

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
```diff
@@ -1098,6 +1103,59 @@ public void closeChannel() {
...
+      future.completeExceptionally(e);
+      return future;
```
sfc-gh-dseweryn, line 1124: A bit shorter
```suggestion
      return CompletableFuture.failedFuture(e);
```

sfc-gh-lshcharbaty: `CompletableFuture.failedFuture` was introduced in Java 9 only :(

test/rest_request_template/schema_evolution_nullable_values_after_smt.json
sfc-gh-lshcharbaty: Added [CloseTopicPartitionChannelIT](https://github.com/snowflakedb/snowflake-kafka-connector/pull/841/files#diff-24f62287e02cb6967f8a0c0792c4bab2ef5fcec5460003c41b951daf399dc6f1R17)  that ensures all channels are closed no matter what the config value is.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-lshcharbaty: `TopicPartitionChannel` which returns a future from `closeChannelAsync` knows nothing about a "key"`. Therefore, it can't return it.

src/main/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannel.java
sfc-gh-dseweryn: Oh :C

sfc-gh-dseweryn (approved):
LGTM

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-xhuang: Understood but I was just thinking that there is not much difference/value between steps 1 and 2 but yes we can be a bit safer this way.

sfc-gh-tzhang: Would this still be a behavior change after we finish step 3? This configuration is gone and customer's KC won't be able to start, then they will need to get the logs to understand why. Assuming they know how to look into the log, then they realize that the configuration is being removed and they will probably need to reach out to Snowflake to understand why it's being removed?

It seems to me a better way is to do a beta release with the fix so customer can try it (or just provide a private jar)? Other customers won't use this version since it's a beta version.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -349,15 +361,40 @@ public void callAllGetOffset() {
...
         (partitionChannelKey, topicPartitionChannel) -> {
           LOGGER.info("Closing partition channel:{}", partitionChannelKey);
           topicPartitionChannel.closeChannel();
         });
-    partitionsToChannel.clear();
+  }
 
-    StreamingClientProvider.getStreamingClientProviderInstance()
-        .closeClient(this.connectorConfig, this.streamingIngestClient);
+  private void closeAllInParallel() {
```
sfc-gh-tzhang, line 384: Think about more I'm not sure this approach would help a lot. If you look into the close function in the SDK, it will flush everything in the client and then call get status on this particular channel. If we're doing it in parallel, the same number of flush will be called, every flush will still be run in serial and block other flushes. I thought most of time is spent on waiting for server side to commit, which will stay the same since everything will be flush at once. Have you verified locally that it reduces the close time dramatically with a large number of channels?

sfc-gh-akowalczyk: Hi @sfc-gh-tzhang, as Lex is OOO, I prepared a short test for that. I am not sure whether the IT is sufficient to give meaningful information, but I ran [such a test](https://github.com/snowflakedb/snowflake-kafka-connector/commit/7d6a84236a7ba7522e3a713e2fe5e61a151f20fb#diff-4005e4dba838ef1cd8e0fe2c4b34e061ed69ec1da322bd1ebf1503e8be051a1eR225-R269) closing 50 channels. It shows the parallelism speeded the execution up.

13513.7 ms mean of sequential execution time vs 1508.0 of parallel
image

But the scenario does not cover the case where we have rows present in `net.snowflake.ingest.streaming.internal.SnowflakeStreamingIngestChannelInternal` buffer (the scenario described by you) - I cannot reproduce it even though I played with the buffering parameters.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
sfc-gh-akowalczyk: @sfc-gh-tzhang I don't see any schema validation for configuration properties. We will just stop respecting the `snowflake.streaming.closeChannelsInParallel.enabled` config key, but no errors should be observed. So, I suppose there will be no behavioral change afterwards.

sfc-gh-tzhang (approved):
Approved to unblock the PR, but I think this change is a risky change since I'm not sure if something weird would happen if we close a few hundreds of channels in parallel, since every close will spin up a background flush thread. Also another concern is calling get status in parallel would potentially overload DS node. I think longer term, the better solution is to add a close function which doesn't call flush or get status so it will be very fast. Could we create a JIRA to track this? Thanks!

ebuildy:
As this is for better performance, is it a way to monitor it please?

This could help debugging crash/rebalance issues.

Thanks you, cant wait to test it!

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-akowalczyk: The log level was set to DEBUG so it added a few additional seconds to the measured time. I changed it to INFO and attached the profiler. It takes ~11 seconds and spends most of its time on `CompletableFuture.get()`. 
image

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -349,15 +361,40 @@ public void callAllGetOffset() {
...
+                  String channelKey = entry.getKey();
+                  TopicPartitionChannel topicPartitionChannel = entry.getValue();
+
+                  LOGGER.info("Closing partition channel:{}", channelKey);
+                  return topicPartitionChannel.closeChannelAsync();
+                })
+            .toArray(CompletableFuture[]::new);
+
+    CompletableFuture.allOf(futures).join();
   }
```
ebuildy, line 398: In my small comprehension,  `CompletableFuture` uses the `ForkJoinPool.commonPool()` as its default thread pool with `(Runtime.getRuntime().availableProcessors() - 1)` threads. We can cheat by setting JVM cpu count but I rather suggest to create a thread pools like this:

```
import java.util.concurrent.*;

private void closeAllInParallel() {
    ExecutorService executor = Executors.newFixedThreadPool(CONFIG_THREAD_POOL_SIZE);
    CompletableFuture<?>[] futures =
        partitionsToChannel.entrySet().stream()
            .map(
                entry -> {
                  String channelKey = entry.getKey();
                  TopicPartitionChannel topicPartitionChannel = entry.getValue();

                  LOGGER.info("Closing partition channel:{}", channelKey);
                  return CompletableFuture.runAsync(topicPartitionChannel::closeChannelAsync, executor);
                })
            .toArray(CompletableFuture[]::new);

    CompletableFuture.allOf(futures).join();
    executor.shutdown(); // Don't forget to shutdown the executor
}
```

ebuildy:
> Approved to unblock the PR, but I think this change is a risky change since I'm not sure if something weird would happen if we close a few hundreds of channels in parallel, since every close will spin up a background flush thread. Also another concern is calling get status in parallel would potentially overload DS node. I think longer term, the better solution is to add a close function in the SDK which doesn't call flush or get status so it will be very fast. Could we create a JIRA to track this? Thanks!

The default thread pool used by `CompletableFuture`  according doc at https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/CompletableFuture.html 

```
All async methods without an explicit Executor argument are performed using the [ForkJoinPool.commonPool()](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/ForkJoinPool.html#commonPool())
``` 

Would be better to create a static thread pool

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-akowalczyk: Thanks, @ebuildy, for providing an example. Let me merge the PR in its current form, I'll check with the SDK team on a possible built-in solution. If the timeline is not optimistic, I'll propose this change to the rest of the team.

sfc-gh-akowalczyk:
> Approved to unblock the PR, but I think this change is a risky change since I'm not sure if something weird would happen if we close a few hundreds of channels in parallel, since every close will spin up a background flush thread. Also another concern is calling get status in parallel would potentially overload DS node. I think longer term, the better solution is to add a close function in the SDK which doesn't call flush or get status so it will be very fast. Could we create a JIRA to track this? Thanks!

Thanks, created SNOW-1437461

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
ebuildy: fantastic! Any plan to do the same for open channel operations?

PR #845, opened 2024-05-17
SNOW-1417197 Check inherited roles when resolving schema evolution permissions

sfc-gh-mbobowski:
# Overview

SNOW-1417197 Check inherited roles when calculating schema evolution

A customer reported an issue that when schema evolution grants are inherited from the other role the table is created but all records are sent to DLQ.

We cannot test this change in e2e right now in a clean way, because test user does not have a `securityadmin` role and cannot create other roles - this is a gap that should be addressed at some point. I managed to test this change manually by running schematization tests with role mbobowski_testrole_kafka:
```
create role mbobowski_testrole_kafka;
grant role testrole_kafka to role mbobowski_testrole_kafka;
grant role mbobowski_testrole_kafka to user test_kafka;
```

I would like to extract permission checking logic into a separate class and write unit tests in a separate PR.

## Pre-review checklist
- [ ] Snowpipe Changes


sfc-gh-mbobowski:
Realised that it needs some improvements. Temporary closing.

PR #850, opened 2024-05-22
Change log level debug for fetched offset message

ebuildy:
# Overview

This log message is too verbose and not provide "info level" information. 

It is inside a loop from https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java#L340

Maybe here we should add info logs?

## Pre-review checklist
- [X] Snowpipe Changes
- [X] Snowpipe Streaming Changes

## Urgency

This review is *normal* priority

## Risks

No impact on operation, impact only if user relies on this log message for monitoring or alerting.


sfc-gh-achyzy (commented):
We understand that some default settings might not apply to all environments and scenarios. Still, we would like to keep the logging defaults a bit more verbose—or, in other words, on a level that allows us to troubleshoot most scenarios and customers. So, we'd like to ask you to change the logging configuration when the defaults are too verbose.

PR #851, opened 2024-05-23
SNOW-1061848: Introduce TopicPartitionChannel interface

sfc-gh-akowalczyk:
# Overview

SNOW-1061848

We want to remove double buffering from the project completely, which is quite a significant change, so we need to temporarily support running **with** and **without** the additional buffer.

This PR is a first attempt to extract an interface that both approaches (with and without buffer) will implement so we can easily switch between them.

## Pre-review checklist
- [ ] Snowpipe Changes


sfc-gh-dseweryn (commented):
Generally OK for me. I think that the ExposingInternalsTopicPartition interface should be either directly implemented on a particular TopicPartitionChannel implementation or available only in tests

src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/ExposingInternalsTopicPartitionChannel.java
```diff
@@ -0,0 +1,58 @@
+package com.snowflake.kafka.connector.internal.streaming.channel;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.snowflake.kafka.connector.internal.streaming.BufferedTopicPartitionChannel;
+import com.snowflake.kafka.connector.internal.streaming.telemetry.SnowflakeTelemetryChannelStatus;
+import com.snowflake.kafka.connector.internal.telemetry.SnowflakeTelemetryService;
+import dev.failsafe.Fallback;
+import net.snowflake.ingest.streaming.SnowflakeStreamingIngestChannel;
+
+public interface ExposingInternalsTopicPartitionChannel {
```
sfc-gh-akowalczyk, line 10: To avoid confusion, the intention was to create this interface as a midstep before investing time in further refactoring. These methods existed already in the previous implementation and were freely accessible by other classes (used to have `protected` modifier).

In the future, I'd like to get rid of the interface but for now, I'd like to focus on discussing whether having a single interface `TopicPartitionChannel` having 2 implementations (buffered and non-buffered) is the way to go.

Additional refactoring will be done afterward.

sfc-gh-mbobowski (approved):
I tried to imagine how I would approach this problem and I was thinking about:
1. Creating a builder for this class to have only one constructor used
2. Creating an abstract class instead of interface. Abstract class would contain `channel` and all methods that are not related to buffering (like getOffsetSafeToCommitToKafka, closeChannel, isChannelClose, getChannelNameFormat).

But I am also ok with the interface-based approach.

sfc-gh-akowalczyk:
> I tried to imagine how I would approach this problem and I was thinking about:
> 
> 1. Creating a builder for this class to have only one constructo
> 2. Creating an abstract class instead of interface. Abstract class would contain `channel` and all methods that are not related to buffering (like getOffsetSafeToCommitToKafka, closeChannel, isChannelClose, getChannelNameFormat).
> 
> But I am also ok with the interface-based approach.

Thanks, let me rethink this approach.

src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
```diff
@@ -1181,42 +1088,50 @@ public String toString() {
         .toString();
   }
 
+  @Override
   @VisibleForTesting
-  protected long getOffsetPersistedInSnowflake() {
+  public long getOffsetPersistedInSnowflake() {
```
sfc-gh-gjachimko, line 1093: do we really need to change visibility of these methods?

src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
```diff
@@ -0,0 +1,90 @@
...
+import com.snowflake.kafka.connector.internal.telemetry.SnowflakeTelemetryService;
+import net.snowflake.ingest.streaming.SnowflakeStreamingIngestChannel;
+import org.apache.kafka.connect.sink.SinkRecord;
+
+import java.util.concurrent.CompletableFuture;
+
+public class DirectTopicPartitionChannel implements TopicPartitionChannel {
+  @Override
+  public long getOffsetPersistedInSnowflake() {
+    return 0;
```
sfc-gh-gjachimko, line 14: this probably doesn't hurt, as it isn't used anywhere, but i tent do throw UnsupportedOperationException for such initial stub methods.

sfc-gh-gjachimko (approved):
lgtm. solid baseline for further work. i like the name refactor.

src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
sfc-gh-akowalczyk: let me leave it as it is now, it will be replaced in the next iterations

PR #855, opened 2024-06-03
SNOW-1061848: Fix unit tests

sfc-gh-akowalczyk:
# Overview

SNOW-1061848

This PR fixes the unit tests for non-buffered scenario - some of them used to still validate the buffered scenario even if it was not expected.

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
@@ -252,7 +252,7 @@ public class DirectTopicPartitionChannel implements TopicPartitionChannel {
             metricsJmxReporter,
             this.offsetPersistedInSnowflake,
             this.processedOffset,
-            new AtomicLong(-1));
+            new AtomicLong(0));
     this.telemetryServiceV2.reportKafkaPartitionStart(
         new SnowflakeTelemetryChannelCreation(this.tableName, this.channelNameFormatV1, startTime));
 
@@ -891,7 +891,7 @@ public class DirectTopicPartitionChannel implements TopicPartitionChannel {
   @VisibleForTesting
   @Deprecated
   public long getLatestConsumerOffset() {
-    return -1;
+    return 0;
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
```diff
@@ -252,7 +252,7 @@ public DirectTopicPartitionChannel(
             metricsJmxReporter,
             this.offsetPersistedInSnowflake,
             this.processedOffset,
-            new AtomicLong(-1));
+            new AtomicLong(0));
```
sfc-gh-akowalczyk, line 255: I need to change it back to conform with the expected behavior in tests.

src/test/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannelTest.java
```diff
@@ -0,0 +1,329 @@
...
+import net.snowflake.ingest.utils.ErrorCode;
+import net.snowflake.ingest.utils.SFException;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.connect.errors.DataException;
+import org.apache.kafka.connect.sink.SinkRecord;
+import org.apache.kafka.connect.sink.SinkTaskContext;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
```
sfc-gh-lshcharbaty, line 33: Let's use junit5 only in new classes

sfc-gh-lshcharbaty (approved):
lgtm

src/test/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannelTest.java
sfc-gh-akowalczyk: That are the tests just moved from the previous file, I would prefer avoid refactoring setup to the latest version at this moment. Eventually, we will, hopefully, move all of them to JUnit 5.

PR #856, opened 2024-06-07
NO-SNOW Add dependabot

sfc-gh-mbobowski:
# Overview

NO-SNOW Add dependabot

Adding dependabot to the project to track dependencies.

There is no way to handle multiple pom files located in the same directory so the PRs will update only the `pom.xml` file and it is developer responsibility to update `confluent_pom.xml` as well.

## Pre-review checklist
- [ ] Snowpipe Changes


sfc-gh-achyzy (approved):
LGTM. 
Thanks a lot for adding it.

PR #862, opened 2024-06-10
SNOW-1061848: Run TopicPartitionChannelIT against a non-buffered path

sfc-gh-akowalczyk:
# Overview

SNOW-1061848

Run `TopicPartitionChannelIT` for a non-buffered scenario.

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelIT.java
+++ b/src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelIT.java
@@ -14,8 +14,6 @@ import com.snowflake.kafka.connector.internal.TestUtils;
 import com.snowflake.kafka.connector.internal.streaming.channel.TopicPartitionChannel;
 import com.snowflake.kafka.connector.internal.streaming.telemetry.SnowflakeTelemetryServiceV2;
 import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelIT.java
```diff
@@ -535,10 +554,19 @@ public void testPartialBatchChannelInvalidationIngestion_schematization() throws
     service.closeAll();
   }
 
+  private Map<String, String> getConfForStreaming() {
+    Map<String, String> config = TestUtils.getConfForStreaming();
+    String enableSingleBufferValue = String.valueOf(!useDoubleBuffer);
```
sfc-gh-achyzy, line 559: It would be easier to use a `useSingleBuffer` for the name and not negate the value. There is an unneeded translation needed.

sfc-gh-achyzy (approved):
LTGM. 
Please consider reverting the property value to match the parameter.

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelIT.java
sfc-gh-akowalczyk: I preferred to have a _turning on_ parameter instead of _turning off_, as our desired behavior is to have a single buffer, but I've changed it as requested for simplicity.

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelIT.java
```diff
@@ -29,9 +31,14 @@
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
 
+@RunWith(Parameterized.class)
```
sfc-gh-mbobowski, line 37: Did you try to migrate to JUnit 5 before? It should be fairly easy and we could avoid this ugly syntax.

sfc-gh-lshcharbaty (approved):
lgtm except the usage of junit4 :)

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelIT.java
sfc-gh-akowalczyk: It's done in JUnit 5. I must admin the previous version was quite handy for parameterizing the whole class :)

sfc-gh-mbobowski (approved):

PR #796, opened 2024-03-01
SNOW-1161484: Use insertRows instead of insertRow for schematization

sfc-gh-tzhang:
# Overview

SNOW-1161484

- Create channel with SKIP_BATCH instead of CONTINUE, to avoid the case that KC can crash right after adding the good rows to table, and the bad rows will be missing in the DLQ
- Update the schematization code to use insertRows instead of insertRow
        - Better performance
        - If KC is configured with a longer buffer flush time, everything will be ingested as one batch so that the 1 second flush or 32MB channel size won't apply

## Pre-review checklist
- [ ] This change should be part of a Behavior Change Release. See [go/behavior-change](http://go/behavior-change).
- [ ] This change has passed Merge gate tests
- [ ] Snowpipe Changes
- [ ] Snowpipe Streaming Changes
- [ ] This change is TEST-ONLY
- [ ] This change is README/Javadocs only
- [ ] This change is protected by a config parameter <PARAMETER_NAME> eg `snowflake.ingestion.method`.
    - [ ] `Yes` - Added end to end and Unit Tests. 
    - [ ] `No` - Suggest why it is not param protected
- [ ] Is his change protected by parameter <PARAMETER_NAME> on the server side?
    - [ ] The parameter/feature is not yet active in production (partial rollout or PrPr, see [Changes for Unreleased Features and Fixes](http://go/ppp-prpr)).
    - [ ] If there is an issue, it can be safely mitigated by turning the parameter off. This is also verified by a test (See [go/ppp](http://go/ppp)).

PR #866, opened 2024-06-13
SNOW-1151484 Use insertRows for schema evolution

sfc-gh-wtrefon:
# Overview

SNOW-1161484 

Mostly copy of old PR https://github.com/snowflakedb/snowflake-kafka-connector/pull/796

Create channel with SKIP_BATCH instead of CONTINUE, to avoid the case that KC can crash right after adding the good rows to table, and the bad rows will be missing in the DLQ
Update the schematization code to use insertRows instead of insertRow
- Better performance
- If KC is configured with a longer buffer flush time, everything will be ingested as one batch so that the 1 second flush or 32MB channel size won't apply

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
@@ -507,8 +507,6 @@ public class BufferedTopicPartitionChannel implements TopicPartitionChannel {
     InsertRowsResponse response = null;
     try {
       response = insertRowsWithFallback(streamingBufferToInsert);
-      // Updates the flush time (last time we called insertRows API)
-      this.previousFlushTimeStampMs = System.currentTimeMillis();
 
       LOGGER.info(
           "Successfully called insertRows for channel:{}, buffer:{}, insertResponseHasErrors:{},"
@@ -517,10 +515,6 @@ public class BufferedTopicPartitionChannel implements TopicPartitionChannel {
           streamingBufferToInsert,
           response.hasErrors(),
           response.needToResetOffset());
-      if (response.hasErrors()) {
-        handleInsertRowsFailures(
-            response.getInsertErrors(), streamingBufferToInsert.getSinkRecords());
-      }
 
       // Due to schema evolution, we may need to reopen the channel and reset the offset in kafka
       // since it's possible that not all rows are ingested
@@ -528,6 +522,18 @@ public class BufferedTopicPartitionChannel implements TopicPartitionChannel {
         streamingApiFallbackSupplier(
             StreamingApiFallbackInvoker.INSERT_ROWS_SCHEMA_EVOLUTION_FALLBACK);
       }
+      // If there are errors other than schema mismatch, we need to handle them and reinsert the
+      // good rows
+      if (response.hasErrors()) {
+        handleInsertRowsFailures(
+            response.getInsertErrors(), streamingBufferToInsert.getSinkRecords());
+        insertRecords(
+            rebuildBufferWithoutErrorRows(streamingBufferToInsert, response.getInsertErrors()));
+      }
+
+      // Updates the flush time (last time we successfully insert some rows)
+      this.previousFlushTimeStampMs = System.currentTimeMillis();
+
       return response;
     } catch (TopicPartitionChannelInsertionException ex) {
       // Suppressing the exception because other channels might still continue to ingest
@@ -538,6 +544,22 @@ public class BufferedTopicPartitionChannel implements TopicPartitionChannel {
           ex);
     }
     return response;
+  }
+
+  /** Building a new buffer which contains only the good rows from the original buffer */
+  private StreamingBuffer rebuildBufferWithoutErrorRows(
+      StreamingBuffer streamingBufferToInsert,
+      List<InsertValidationResponse.InsertError> insertErrors) {
+    StreamingBuffer buffer = new StreamingBuffer();
+    int errorIdx = 0;
+    for (long rowIdx = 0; rowIdx < streamingBufferToInsert.getNumOfRecords(); rowIdx++) {
+      if (errorIdx < insertErrors.size() && rowIdx == insertErrors.get(errorIdx).getRowIndex()) {
+        errorIdx++;
+      } else {
+        buffer.insert(streamingBufferToInsert.getSinkRecord(rowIdx));
+      }
+    }
+    return buffer;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
```diff
@@ -517,17 +515,25 @@ public InsertRowsResponse insertRecords(StreamingBuffer streamingBufferToInsert)
...
       if (response.needToResetOffset()) {
         streamingApiFallbackSupplier(
             StreamingApiFallbackInvoker.INSERT_ROWS_SCHEMA_EVOLUTION_FALLBACK);
       }
+      // If there are errors other than schema mismatch, we need to handle them and reinsert the
+      // good rows
+      if (response.hasErrors()) {
+        handleInsertRowsFailures(
+            response.getInsertErrors(), streamingBufferToInsert.getSinkRecords());
+        insertRecords(
```
sfc-gh-wtrefon, line 530: do we need to reset the offset if we put here the schema evolution records as well?
@sfc-gh-tzhang

sfc-gh-tzhang: I don't think so, why we need to do that?

src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
```diff
@@ -540,6 +546,22 @@ public InsertRowsResponse insertRecords(StreamingBuffer streamingBufferToInsert)
     return response;
   }
 
+  /** Building a new buffer which contains only the good rows from the original buffer */
+  private StreamingBuffer rebuildBufferWithoutErrorRows(
```
sfc-gh-tzhang, line 550: let's add a unit test for this function

src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
```diff
@@ -620,65 +642,40 @@ public InsertRowsResponse get() throws Throwable {
...
+            SchematizationUtils.evolveSchemaIfNeeded(
+                this.conn,
+                this.channel.getTableName(),
+                new ArrayList<>(nonNullableColumns),
+                extraColNames,
+                this.insertRowsStreamingBuffer.getSinkRecord(insertError.getRowIndex()));
+            needToResetOffset = true;
+          } else {
+            // Simply added to the final response if it's not schema related errors
+            finalResponse.addError(insertError);
```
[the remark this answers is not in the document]
sfc-gh-tzhang: In my old PR, I don't have `finalResponse` anymore. I simply return `response` if there is no schema evolution related errors which should include all the errors already. Do you see any issue with that?

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelTest.java
```diff
@@ -806,10 +806,11 @@ public void testInsertRows_SuccessAfterReopenChannel() throws Exception {
...
 
   @Test
-  public void testInsertRowsWithSchemaEvolution() throws Exception {
+  public void testInsertRowsWithSchemaEvolution_onlySingleBuffer() throws Exception {
     if (this.sfConnectorConfig
-        .get(SnowflakeSinkConnectorConfig.ENABLE_SCHEMATIZATION_CONFIG)
-        .equals("true")) {
+            .get(SnowflakeSinkConnectorConfig.ENABLE_SCHEMATIZATION_CONFIG)
+            .equals("true")
+        && !useDoubleBuffer) {
```
sfc-gh-tzhang, line 813: I'm not sure what needs to be updated without double buffer now

src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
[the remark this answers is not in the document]
sfc-gh-wtrefon: without the `finalResponse`, the `response` will contain the schema evolution errors as well. This means in handleInsertRowsFailures we will put the schema evolutions erros in DLQ, this is a behaviour change.

Additionally, the rebuildBufferWithoutErrorRows will filter out the schema evolution errors and such rows wont be inserted ever, right? We don't wont such behaviour

src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
sfc-gh-wtrefon: we are setting this flag needToResetOffset to be true in case of schema evolution error. I'm not sure what it exactly does and if we need this if we now reinsert the rows by rebuild the buffer with rebuildBufferWithoutErrorRows

sfc-gh-akowalczyk:
> @sfc-gh-tzhang another question, do we need this change? With removal of the extra buffer we will get rid of this code anyway, do we want to invest time here?

@sfc-gh-tzhang  friendly reminder about the question above ☝️

src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
sfc-gh-tzhang: if you see how needToResetOffset will be used, when it's true, we will reopen the channel and restart from the latest committed offset token

src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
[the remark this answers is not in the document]
sfc-gh-tzhang: > This means in handleInsertRowsFailures we will put the schema evolutions erros in DLQ, this is a behaviour change.

In my old PR, we will return the response before calling handleInsertRowsFailures, right?

> Additionally, the rebuildBufferWithoutErrorRows will filter out the schema evolution errors and such rows wont be inserted ever, right? We don't wont such behaviour

Reset the offset token will take care of that, Kafka will send us the same batch again.

sfc-gh-tzhang:
> @sfc-gh-tzhang another question, do we need this change? With removal of the extra buffer we will get rid of this code anyway, do we want to invest time here?

How would you do schema evolution with the removal of the extra buffer? I don't see a design doc so I don't know the answer.

sfc-gh-mbobowski:
I doubt if it makes sense to keep it open after 6 months of inactivity.

PR #867, opened 2024-06-14
SNOW-1061854 Run e2e tests with single buffer

sfc-gh-lshcharbaty:
# Overview

SNOW-1061854

Running e2e tests with single and double buffers. In every snowpipe_streaming test we'll need to specify `$SNOWFLAKE_STREAMING_ENABLE_SINGLE_BUFFER` property that's later substituted with `true` and `false` values. To achieve isolation between different parameter sets, a parameter set index is added to test salt.

I wouldn't use github test matrix, as this solution seems to be temporary and removed shortly.

**Note that currently the tests are run sequentially that will take about 2x time!** File generation logic needs to be rewritten to allow running a test suite with different parameters in parallel.

All related changes are in [test_verfiy.py](https://github.com/snowflakedb/snowflake-kafka-connector/pull/867/files#diff-0eb8c275e2e077d9fe6b7d9a9e45cd7887d93c752cb0013ccf646fb41e90de59). 
Also removed a check for enabled `SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER` in [Utils.java](https://github.com/snowflakedb/snowflake-kafka-connector/pull/867/files#diff-371eabc61cd3788bddba04e3204eaa5ad54d315769e30def9b88095307e5a4b1).
The rest are json request and test verification adjustments.

A few e2e tests don't pass with a single buffer. See SNOW-1061855 and SNOW-1475664 for the details.

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/test/rest_request_template/travis_correct_snowpipe_streaming_string_json.json
+++ b/test/rest_request_template/travis_correct_snowpipe_streaming_string_json.json
@@ -14,6 +14,7 @@
     "snowflake.schema.name": "SNOWFLAKE_SCHEMA",
     "snowflake.role.name": "SNOWFLAKE_ROLE",
     "snowflake.ingestion.method": "SNOWPIPE_STREAMING",
+    "snowflake.streaming.enable.single.buffer": "$SNOWFLAKE_STREAMING_ENABLE_SINGLE_BUFFER",
     "snowflake.streaming.closeChannelsInParallel.enabled": true,
     "snowflake.enable.schematization": "false",
     "key.converter": "org.apache.kafka.connect.storage.StringConverter",
--- a/test/run_test_apache.sh
+++ b/test/run_test_apache.sh
@@ -175,7 +175,6 @@ if [ $testError -ne 0 ]; then
     RED='\033[0;31m'
     NC='\033[0m' # No Color
     echo -e "${RED} There is error above this line ${NC}"
-    #    Ignore for e2e test draft. Todo: turn back again?
-    #    cat $APACHE_LOG_PATH/kc.log
+    cat $APACHE_LOG_PATH/kc.log
     error_exit "=== test_verify.py failed ==="
 fi
--- a/test/run_test_apache.sh
+++ b/test/run_test_apache.sh
@@ -175,6 +175,7 @@ if [ $testError -ne 0 ]; then
     RED='\033[0;31m'
     NC='\033[0m' # No Color
     echo -e "${RED} There is error above this line ${NC}"
-    cat $APACHE_LOG_PATH/kc.log
+    #    Ignore for e2e test draft. Todo: turn back again?
+    #    cat $APACHE_LOG_PATH/kc.log
     error_exit "=== test_verify.py failed ==="
 fi
--- a/test/test_verify.py
+++ b/test/test_verify.py
@@ -29,11 +29,11 @@ class ConnectorParametersList:
     def __init__(self, connectorParametersList: list[ConnectorParameters]):
         self.connectorParametersList = connectorParametersList
 
-    def for_each(self, function: Callable[[ConnectorParameters], None]) -> None:
-        for conenctor_parameters in self.connectorParametersList:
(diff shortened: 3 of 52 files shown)


test/rest_request_template/travis_correct_snowpipe_streaming_string_json.json
```diff
@@ -14,6 +14,7 @@
     "snowflake.schema.name": "SNOWFLAKE_SCHEMA",
     "snowflake.role.name": "SNOWFLAKE_ROLE",
     "snowflake.ingestion.method": "SNOWPIPE_STREAMING",
+    "snowflake.streaming.enable.single.buffer": "$SNOWFLAKE_STREAMING_ENABLE_SINGLE_BUFFER",
```
sfc-gh-lshcharbaty, line 17: This line is going to be added for each `SNOWPIPE_STREAMING` test.

sfc-gh-krosinski (approved):

test/rest_request_template/travis_correct_snowpipe_streaming_string_json.json
sfc-gh-achyzy: Why introduce a new way of introducing parameters with '$'?

test/test_verify.py
```diff
@@ -275,6 +295,9 @@ def cleanTableStagePipe(self, connectorName, topicName="", partitionNumber=1):
             print(datetime.now().strftime("%H:%M:%S "), "=== Drop pipe {} ===".format(pipeName))
             self.snowflake_conn.cursor().execute("DROP pipe IF EXISTS {}".format(pipeName))
 
+        print(datetime.now().strftime("%H:%M:%S "), f"=== Delete Kafka topic {topicName} ===")
+        self.deleteTopic(topicName)
```
sfc-gh-achyzy, line 299: Why exactly do we need to delete the topic here? Is this a fix, as it was missing, or did you add some logic related to that?

test/rest_request_template/travis_correct_snowpipe_streaming_string_json.json
sfc-gh-lshcharbaty: Using caps only doesn't make the template parameters stand out. 
In this example `SNOWFLAKE_SCHEMA` and `SNOWFLAKE_ROLE` are parameter names while `SNOWPIPE_STREAMING` is a value.

I'm also inclined to adjust the naming for all template parameters.

test/run_test_apache.sh
```diff
@@ -175,6 +175,7 @@ if [ $testError -ne 0 ]; then
     RED='\033[0;31m'
     NC='\033[0m' # No Color
     echo -e "${RED} There is error above this line ${NC}"
-    cat $APACHE_LOG_PATH/kc.log
+    #    Ignore for e2e test draft. Todo: turn back again?
+    #    cat $APACHE_LOG_PATH/kc.log
```
sfc-gh-lshcharbaty, line 179: It's horrible to check what has failed when printing **all** Kafka connector debug logs. Once all tests have passed, this change will be reverted back.

sfc-gh-achyzy (approved):
LGTM. I hope you'll be able to get to the state where all merge gates pass.

test/test_verify.py
```diff
@@ -18,14 +20,31 @@
...
+class ConnectorParameters:
+    snowflake_streaming_enable_single_buffer: str
+
+
+class ConnectorParametersList:
+    def __init__(self, connectorParametersList: list[ConnectorParameters]):
+        self.connectorParametersList = connectorParametersList
+
+    def for_each(self, function: Callable[[int, ConnectorParameters], None]) -> None:
+        for idx, conenctor_parameters in enumerate(self.connectorParametersList):
```
sfc-gh-akowalczyk, line 33: typo, `connector_parameters`,

shouldn't it be `connector_parameter`, BTW?

sfc-gh-akowalczyk (approved):
Good idea to set the parameter at the test suite top level

PR #875, opened 2024-07-09
SNOW-1514185: Do assign the reopened channel unless kafka offsets are fully reset

sfc-gh-akowalczyk:
# Overview

SNOW-1514185

The change applies to channel reopen logic. The process used to consists of 3 steps:
 - reopen the channel
 - fetch the offset from the channel
 - reset internal buffer
 - rest offset in Kafka

If any of the steps 2-4 fails during `getOffsetSafeToCommitToKafka` the error will be ignored in [SnowflakeSinkTask::preCommit](https://github.com/snowflakedb/snowflake-kafka-connector/blob/15df7115dc26a2ad015f8486d6dd3f48a460e525/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java#L347-L349)

After the change the channel is assigned to the TopicPartitionChannel after all the steps are done.

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
@@ -861,7 +861,8 @@ public class BufferedTopicPartitionChannel implements TopicPartitionChannel {
         this.getChannelNameFormatV1());
     long offsetRecoveredFromSnowflake = fetchLatestOffsetFromChannel(newChannel);
 
-    resetChannelMetadataAfterRecovery(streamingApiFallbackInvoker, offsetRecoveredFromSnowflake, newChannel);
+    resetChannelMetadataAfterRecovery(
+        streamingApiFallbackInvoker, offsetRecoveredFromSnowflake, newChannel);
 
     return offsetRecoveredFromSnowflake;
   }
@@ -874,15 +875,16 @@ public class BufferedTopicPartitionChannel implements TopicPartitionChannel {
    * <p>Idea behind resetting offset (1 more than what we found in snowflake) is that Kafka should
    * send offsets from this offset number so as to not miss any data.
    *
-   * @param streamingApiFallbackInvoker  Streaming API which is using this fallback function. Used
-   *                                     for logging mainly.
+   * @param streamingApiFallbackInvoker Streaming API which is using this fallback function. Used
+   *     for logging mainly.
    * @param offsetRecoveredFromSnowflake offset number found in snowflake for this
-   *                                     channel(partition)
-   * @param newChannel                   a channel to assign to the current instance
+   *     channel(partition)
+   * @param newChannel a channel to assign to the current instance
    */
   private void resetChannelMetadataAfterRecovery(
-          final StreamingApiFallbackInvoker streamingApiFallbackInvoker,
-          final long offsetRecoveredFromSnowflake, SnowflakeStreamingIngestChannel newChannel) {
+      final StreamingApiFallbackInvoker streamingApiFallbackInvoker,
+      final long offsetRecoveredFromSnowflake,
+      SnowflakeStreamingIngestChannel newChannel) {
     if (offsetRecoveredFromSnowflake == NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE) {
       LOGGER.info(
           "{} Channel:{}, offset token is NULL, will use the consumer offset managed by the"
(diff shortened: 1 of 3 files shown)


sfc-gh-mbobowski (approved):

src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
```diff
@@ -853,9 +853,18 @@ public long fetchOffsetTokenWithRetry() {
...
+    SnowflakeStreamingIngestChannel newChannel = reopenChannel(streamingApiFallbackInvoker);
+
+    LOGGER.warn(
+        "{} Fetching offsetToken after re-opening the channel:{}",
+        streamingApiFallbackInvoker,
+        this.getChannelNameFormatV1());
+    long offsetRecoveredFromSnowflake = fetchLatestOffsetFromChannel(newChannel);
+
     resetChannelMetadataAfterRecovery(streamingApiFallbackInvoker, offsetRecoveredFromSnowflake);
+    ;
```
sfc-gh-gjachimko, line 865: why the ';' here?

src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
```diff
@@ -958,11 +962,16 @@ private long getRecoveredOffsetFromSnowflake(
    *     snowflake.
    */
   private long fetchLatestCommittedOffsetFromSnowflake() {
+    SnowflakeStreamingIngestChannel channelToGetOffset = this.channel;
+    return fetchLatestOffsetFromChannel(channelToGetOffset);
+  }
+
+  private long fetchLatestOffsetFromChannel(SnowflakeStreamingIngestChannel channelToGetOffset) {
```
sfc-gh-gjachimko, line 969: channelToGetOffset -> rename to just the "channel" ?

src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
```diff
@@ -640,9 +640,17 @@ public long fetchOffsetTokenWithRetry() {
    */
   private long streamingApiFallbackSupplier(
```
sfc-gh-gjachimko, line 641: i'm wondering if there is a risk of having race condition here - i.e. two callers invoking this method would invoke resetChannelMetadataAfterRecovery - this method mutates underlying state, so - i'm wondering if the entire method content should be synchronized?

src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
sfc-gh-akowalczyk: just a leftover, removed

src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
sfc-gh-akowalczyk: 👍

src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
sfc-gh-akowalczyk: that's true, I put the channel assignment into the synchronized sections, so all properties are being modified there

src/test/java/com/snowflake/kafka/connector/internal/streaming/TopicPartitionChannelTest.java
```diff
@@ -1245,4 +1247,68 @@ public void testOffsetTokenVerificationFunction() {
     Assert.assertFalse(StreamingUtils.offsetTokenVerificationFunction.verify("1", "3", "4", 3));
     Assert.assertFalse(StreamingUtils.offsetTokenVerificationFunction.verify("2", "1", "4", 3));
   }
+
+  @Test
+  public void assignANewChannelAfterTheSetupIsFullyDone() throws Exception {
+    // given
+
+    SnowflakeStreamingIngestChannel channel1 = Mockito.mock(SnowflakeStreamingIngestChannel.class);
```
sfc-gh-achyzy, line 1255: Nit: I would like to use `FakeSnowflakeStreamingIngestChannel` or extend that implementation so we don't have to do so much mocking.

sfc-gh-akowalczyk: Either way, I must modify `FakeSnowflakeStreamingIngestChannel` for a specific edge case. This means it won't be significantly different from mocking an object.

src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
```diff
@@ -915,6 +925,7 @@ private void resetChannelMetadataAfterRecovery(
       // Set the flag so that any leftover rows in the buffer should be skipped, it will be
       // re-ingested since the offset in kafka was reset
       needToSkipCurrentBatch = true;
+      this.channel = newChannel;
```
sfc-gh-tzhang, line 928: Is this correct? At line 904 above, we simply return if the channel offset is null, this means that we will never use the new channel if the offset is null?

PR #882, opened 2024-07-15
SNOW-1061855 Fix the E2E test for SchemaEvolutionDropTable with a single buffer

sfc-gh-gjachimko:
SNOW-1061855

Fix e2e tests for schema migration with non-buffered channel as underlying transport. The functionality is in main, this PR enables the tests and sets the channel backwards compatible with buffered channel. 

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
@@ -83,7 +83,7 @@ public class DirectTopicPartitionChannel implements TopicPartitionChannel {
   // This offset is would not be required for buffer-less channel, but we add it to keep buffered
   // and non-buffered
   // channel versions compatible.
-  private final AtomicLong latestConsumerOffset =
+  private final AtomicLong currentConsumerGroupOffset =
       new AtomicLong(NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE);
 
   // Indicates whether we need to skip and discard any leftover rows in the current batch, this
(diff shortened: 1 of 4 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
```diff
@@ -80,6 +80,12 @@ public class DirectTopicPartitionChannel implements TopicPartitionChannel {
   private final AtomicLong processedOffset =
       new AtomicLong(NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE);
 
+  // This offset is would not be required for buffer-less channel, but we add it to keep buffered
+  // and non-buffered
+  // channel versions compatible.
+  private final AtomicLong latestConsumerOffset =
```
sfc-gh-lshcharbaty, line 86: Nit: since this offset represents a current consumer group offset let's name it `currentConsumerGroupOffset`. I was always confused by `latestConsumerOffset`

src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
```diff
@@ -689,19 +700,17 @@ private void resetChannelMetadataAfterRecovery(
...
-      LOGGER.info(
-          "There is no offset registered for {} channel in Snowflake. Stop recovering the channel"
-              + " metadata.",
-          this.getChannelNameFormatV1());
-      return;
-    }
+    final long offsetToResetInKafka =
+        offsetRecoveredFromSnowflake == NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE
+            ? latestConsumerOffset.get()
+            : offsetRecoveredFromSnowflake + 1L;
```
sfc-gh-lshcharbaty, line 713: I expect we can fall into this method even when `latestConsumerOffset` has not been initialized yet. Even though there is a consumer group offset.
In these cases we don't want to reset consumer group offset to 0.
In the Buffered channel there is an additional check for this scenario:
```java
    final long offsetToResetInKafka =
        offsetRecoveredFromSnowflake == NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE
            ? latestConsumerOffset.get()
            : offsetRecoveredFromSnowflake + 1L;
    if (offsetToResetInKafka == NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE) {
      return;
    }
```

sfc-gh-lshcharbaty (approved):
lgtm

PR #883, opened 2024-07-16
NOSNOW: improve troubleshooting experience for scenario with skipped offsets

sfc-gh-gjachimko:
NOSNOW

Change the verbosity (of first log message occurrence) and message content to improve troubleshooting expience as requested by support team. 

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
@@ -402,13 +402,13 @@ public class BufferedTopicPartitionChannel implements TopicPartitionChannel {
         insertRecords(copiedStreamingBuffer);
       }
     } else {
-      LOGGER.debug(
-          "Skip adding offset:{} to buffer for channel:{} because"
-              + " offsetPersistedInSnowflake:{}, processedOffset:{}",
-          kafkaSinkRecord.kafkaOffset(),
+      LOGGER.warn(
+          "Channel {} - skipping current record - expected offset {} but received {}. The current"
+              + " offset stored in Snowflake: {}",
           this.getChannelNameFormatV1(),
-          currentOffsetPersistedInSnowflake,
-          currentProcessedOffset);
+          currentProcessedOffset,
+          kafkaSinkRecord.kafkaOffset(),
+          currentOffsetPersistedInSnowflake);
     }
   }
 
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
@@ -324,13 +324,13 @@ public class DirectTopicPartitionChannel implements TopicPartitionChannel {
         || kafkaSinkRecord.kafkaOffset() >= currentProcessedOffset + 1) {
       transformAndSend(kafkaSinkRecord);
     } else {
-      LOGGER.debug(
-          "Skip adding offset:{} to buffer for channel:{} because"
-              + " offsetPersistedInSnowflake:{}, processedOffset:{}",
-          kafkaSinkRecord.kafkaOffset(),
+      LOGGER.warn(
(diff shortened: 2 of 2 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
```diff
@@ -323,14 +328,20 @@ public void insertRecord(SinkRecord kafkaSinkRecord, boolean isFirstRowPerPartit
     if (currentProcessedOffset == NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE
         || kafkaSinkRecord.kafkaOffset() >= currentProcessedOffset + 1) {
       transformAndSend(kafkaSinkRecord);
+      isFirstMismatchedOffset.set(true);
```
sfc-gh-achyzy, line 331: I'm not a fan of logging something on the first occurrence and later skipping. 
I propose to keep it consistent.

src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
```diff
@@ -323,14 +328,20 @@ public void insertRecord(SinkRecord kafkaSinkRecord, boolean isFirstRowPerPartit
...
-              + " offsetPersistedInSnowflake:{}, processedOffset:{}",
-          kafkaSinkRecord.kafkaOffset(),
-          this.getChannelNameFormatV1(),
-          currentOffsetPersistedInSnowflake,
-          currentProcessedOffset);
+      final BiConsumer<String, Object[]> logMsgDelegate =
+          isFirstMismatchedOffset.getAndSet(false) ? LOGGER::warn : LOGGER::debug;
+
+      logMsgDelegate.accept(
+          "Connector's channel {} expected record at offset {} but received {}. Offset stored in"
```
sfc-gh-achyzy, line 337:
```suggestion
          "Channel {} expected record at offset {} but received {}. The current offset stored in"
```

sfc-gh-achyzy (commented):
I'm not sure I like the usage of different levels depending on the number of occurrences.

src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
sfc-gh-gjachimko: ok, lets keep it as info then. warn seems to me too high.

sfc-gh-hachouraria: Will it be OK that we print this infinite times, especially in the case of expected offset being earlier to current offset?

Personally I think its alright since an action is certainly required but I was just wondering from a performance/cost impact to customers (and iff any of our previous logging decisions were led by that).

src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
```diff
@@ -402,13 +402,13 @@ public void insertRecord(SinkRecord kafkaSinkRecord, boolean isFirstRowPerPartit
         insertRecords(copiedStreamingBuffer);
       }
     } else {
-      LOGGER.debug(
-          "Skip adding offset:{} to buffer for channel:{} because"
-              + " offsetPersistedInSnowflake:{}, processedOffset:{}",
-          kafkaSinkRecord.kafkaOffset(),
+      LOGGER.warn(
+          "Channel {} expected record at offset {} but received {}. The current offset stored in"
```
sfc-gh-hachouraria, line 406: Thank you very much for addressing this!

Could we keep the skip messaging too (my suggested change in Slack was meant to be additive, sorry for the confusion!)

The new warn message is great, but it doesn't let the reader know what the connector's action was (skip) so they can mentally match it to their observed lag/loss situation.

sfc-gh-gjachimko: ok, extended the message a bit.

sfc-gh-lshcharbaty (approved):

Issue #878, opened 2024-07-12
schematization=true,  error  Insert Row Error , Columns not present in the table shouldn't be specified


marystory:
I am testing the schematization feature for the Snowpipe Streaming ingestion method locally. I have tried it with JSON and Avro value serialization. In both cases, the destination table does not exist and is supposed to be created by the Kafka connector.

I get the error below:

**ERROR [snowflake-sink-connector|task-0] [SF_KAFKA_CONNECTOR] Insert Row Error message:The given row cannot be converted to the internal format: Extra columns: ["EMAIL", "AGE", "NAME"]. Columns not present in the table shouldn't be specified, rowIndex:1 (com.snowflake.kafka.connector.internal.streaming.TopicPartitionChannel:94)**

Setup: 
kafka_2.13-3.2.3 : for kafka connect
snowflake-kafka-connector-2.2.2
```
name=snowflake-sink-connector
connector.class=com.snowflake.kafka.connector.SnowflakeSinkConnector
snowflake.ingestion.method=SNOWPIPE_STREAMING
tasks.max=1
topics=user
snowflake.topic2table.map=user:user
buffer.flush.time=20
buffer.count.records=10000
buffer.size.bytes=5000000

errors.tolerance=all
errors.log.enable=true

key.converter.schemas.enable=false
value.converter.schemas.enable=true
key.converter=org.apache.kafka.connect.storage.StringConverter
value.converter=io.confluent.connect.avro.AvroConverter
value.converter.schema.registry.url=http://localhost:8081

# Snowflake configurations
# The URL for accessing your Snowflake account
snowflake.url.name=<>
snowflake.database.name=<>
snowflake.schema.name=INGESTION
snowflake.user.name=<>
snowflake.role.name=<>
snowflake.private.key=<>
snowflake.private.key.passphrase=<>

snowflake.enable.schematization=true
```

Schematization should be able to create the table with my schema. So I am confused about the error I got. I get the same error even when I create the destination table manually and set its ENABLE_SCHEMA_EVOLUTION to true.

Any guidance on whether I am missing a setup or using wrong versions? I am using redpanda cluster and kafka connect.

Thanks

sfc-gh-mbobowski:
Hello @marystory, thank you for trying out schema evolution feature!

I don't see any error within your config and v2.2.2 should definitely support schematization. Please provide information on:
- the connector logs (at least INFO level)
- data that is being ingested to Snowflake (at least a couple of rows)
[...]

PR #884, opened 2024-07-18
Adding logs to track schematization

sfc-gh-mbobowski:
In the recent issue (https://github.com/snowflakedb/snowflake-kafka-connector/issues/878) I discovered that sadly we do not have any logs that can be used to track schematization settings and code execution. The reason for this PR is to close this gap.

The logs can be added at INFO level because they appear only at connector startup and when schema change is detected.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
@@ -669,9 +669,10 @@ public class BufferedTopicPartitionChannel implements TopicPartitionChannel {
               // Simply added to the final response if it's not schema related errors
               finalResponse.addError(insertError);
             } else {
-              LOGGER.info("Triggering schema evolution. NonNullableColumns={}, extraColumns={}",
-                      String.join("", nonNullableColumns),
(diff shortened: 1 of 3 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
```diff
@@ -669,6 +669,10 @@ public InsertRowsResponse get() throws Throwable {
               // Simply added to the final response if it's not schema related errors
               finalResponse.addError(insertError);
             } else {
+              LOGGER.info(
+                  "Triggering schema evolution. NonNullableColumns={}, extraColumns={}",
+                  String.join("", nonNullableColumns),
```
sfc-gh-bzabek, line 674: Maybe we want a spacebar or other separator for `nonNullableColumns`? (I didnt check how it looks now)

sfc-gh-bzabek (approved):

src/main/java/com/snowflake/kafka/connector/internal/streaming/BufferedTopicPartitionChannel.java
sfc-gh-mbobowski: Good catch I missed that.

sfc-gh-achyzy (approved):
LGTM

PR #885, opened 2024-07-18
SNOW-1541942 Extend Snowpipe initialization logging at INFO level

sfc-gh-mbobowski:
# Overview

SNOW-1541942

Customer using Snowpipe reported random errors thrown at stage and table validations. I believe that we can log details of initialization at INFO level because it happens only on task start. One comment states that initialization happens once for 5 second with up to 120 tries but it is still not a big amount of logs. 

## Pre-review checklist
- [x] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
@@ -351,7 +351,7 @@ public class SnowflakeConnectionServiceV1 implements SnowflakeConnectionService 
       }
       compatible = hasMeta && hasContent && allNullable;
     } catch (SQLException e) {
-      LOGGER.debug("table {} doesn't exist", tableName);
+      LOGGER.info("Table {} doesn't exist. Exception {}", tableName, e.getStackTrace());
       compatible = false;
     } finally {
       try {
@@ -370,6 +370,7 @@ public class SnowflakeConnectionServiceV1 implements SnowflakeConnectionService 
         e.printStackTrace();
       }
     }
+    LOGGER.info("Table {} compatibility is {}", tableName, compatible);
     return compatible;
   }
 
@@ -577,16 +578,17 @@ public class SnowflakeConnectionServiceV1 implements SnowflakeConnectionService 
     checkConnection();
     InternalUtils.assertNotEmpty("stageName", stageName);
     if (!stageExist(stageName)) {
-      LOGGER.debug("stage {} doesn't exists", stageName);
+      LOGGER.info("stage {} doesn't exists", stageName);
       return false;
     }
     List<String> files = listStage(stageName, "");
     for (String name : files) {
       if (!FileNameUtils.verifyFileName(name)) {
-        LOGGER.debug("file name {} in stage {} is not valid", name, stageName);
+        LOGGER.info("file name {} in stage {} is not valid", name, stageName);
         return false;
       }
     }
+    LOGGER.info("Stage {} is valid", stageName);
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -577,16 +578,17 @@ public boolean isStageCompatible(final String stageName) {
     checkConnection();
     InternalUtils.assertNotEmpty("stageName", stageName);
     if (!stageExist(stageName)) {
-      LOGGER.debug("stage {} doesn't exists", stageName);
+      LOGGER.info("stage {} doesn't exists", stageName);
```
sfc-gh-bzabek, line 581: I'm thinking how will you find this and above log? Imagine you want to search for it in a file. When you search by "stage" or "doesn't exists" you will find many of words like this. 
Maybe we should use some convention, adding some kind of a signature for easier search? Is my concern valid, WDYT?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -351,7 +351,7 @@ public boolean isTableCompatible(final String tableName) {
       }
       compatible = hasMeta && hasContent && allNullable;
     } catch (SQLException e) {
-      LOGGER.debug("table {} doesn't exist", tableName);
+      LOGGER.info("Table {} doesn't exist. Exception {}", tableName, e.getStackTrace());
```
sfc-gh-achyzy, line 354: That message in logs can be interpreted as a serious situation without context. I would keep it on debug. The next one is much better for INFO.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-achyzy: I would keep it on debug, similar to table one, and use the next one for INFO.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -577,16 +578,17 @@ public boolean isStageCompatible(final String stageName) {
...
     if (!stageExist(stageName)) {
-      LOGGER.debug("stage {} doesn't exists", stageName);
+      LOGGER.info("stage {} doesn't exists", stageName);
       return false;
     }
     List<String> files = listStage(stageName, "");
     for (String name : files) {
       if (!FileNameUtils.verifyFileName(name)) {
-        LOGGER.debug("file name {} in stage {} is not valid", name, stageName);
+        LOGGER.info("file name {} in stage {} is not valid", name, stageName);
```
sfc-gh-achyzy, line 587: I think we can leave it on debug.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -577,16 +578,17 @@ public boolean isStageCompatible(final String stageName) {
...
         return false;
       }
     }
+    LOGGER.info("Stage {} is valid", stageName);
```
sfc-gh-achyzy, line 591: How about keeping it consistent with the table one? like
```suggestion
    LOGGER.info("Stage {} compatibility is {}", stageName, compatible);
```

sfc-gh-mbobowski: It required some refactor, but done

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-mbobowski: I would like to give it a try on the INFO level. 

Customer reported an error that occasionally some files are left on a stage and it causes connector to restart. Getting file names could greatly improve our debugging capabilities and DEBUG level is unfortunately not available.

sfc-gh-achyzy (approved):
LGTM

sfc-gh-mbobowski:
Close / open to restart snyk check

PR #895, opened 2024-07-31
SNOW-1229108 Refactor config validation

sfc-gh-mbobowski:
SNOW-1229108

Initial refactor:
- extract config validation from Utils into dedicated class
- create interface


src/main/java/com/snowflake/kafka/connector/IConnectorConfigValidator.java
```diff
@@ -0,0 +1,14 @@
+package com.snowflake.kafka.connector;
+
+import java.util.Map;
+
+public interface IConnectorConfigValidator {
```
sfc-gh-wtrefon, line 5: do we need to follow here the "I" interface naming convention? I would just do as usual with the implementation having postfix Impl or prefix Main/Default

sfc-gh-mbobowski: Changed to Default prefix. We should align to one convention as right now we have no interfaces = no convention ;)

sfc-gh-mbobowski:
Closing - will post another PR with complete change

PR #896, opened 2024-08-01
SNOW-1229108 Add warning about buffer config when single buffer enabled

sfc-gh-mbobowski:
# Overview

SNOW-1229108

I usually avoid mixing refactoring with adding logic but I didn't expect that it would be that much code.
I added comments to make your review easier but what I tried to achieve is:
- move config validation logic outside of Utils / StreamingUtils classes
- create separate config validation path for single and double buffer
- add warning logs about some parameters being skipped for single buffer

## Pre-review checklist
- [ ] Snowpipe Changes


src/main/java/com/snowflake/kafka/connector/internal/streaming/DefaultStreamingConfigValidator.java
```diff
@@ -0,0 +1,222 @@
...
+import com.snowflake.kafka.connector.internal.KCLogger;
+import com.snowflake.kafka.connector.internal.parameters.InternalBufferParameters;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import org.apache.kafka.common.config.ConfigException;
+
+public class DefaultStreamingConfigValidator implements StreamingConfigValidator {
```
sfc-gh-mbobowski, line 21: This class contains everything from the `StreamingUtils.java` validation. The only difference is adding `SingleBufferConfigValidator` that logs warnings and `DoubleBufferConfigValidator` that performs validations as before.

It would be great to unit test this class separately from `ConnectorConfigValidatorTest` but I won't do it in this PR.

src/test/java/com/snowflake/kafka/connector/ConnectorConfigValidatorLogsTest.java
```diff
@@ -0,0 +1,86 @@
...
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.Map;
+import org.bouncycastle.operator.OperatorCreationException;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+public class ConnectorConfigValidatorLogsTest {
```
sfc-gh-mbobowski, line 22: This is refactored `SecurityTest` with one more test added.

sfc-gh-mbobowski:
Close / reopen to test snyk

src/main/java/com/snowflake/kafka/connector/internal/streaming/DefaultStreamingConfigValidator.java
```diff
@@ -0,0 +1,222 @@
...
+
+    private static final KCLogger LOGGER =
+        new KCLogger(SingleBufferConfigValidator.class.getName());
+
+    private void logDoubleBufferingParametersWarning(Map<String, String> config) {
+      if (InternalBufferParameters.isSingleBufferEnabled(config)) {
+        Set<String> ignoredParameters =
+            new HashSet<>(
+                Arrays.asList(BUFFER_FLUSH_TIME_SEC, BUFFER_SIZE_BYTES, BUFFER_COUNT_RECORDS));
+        ignoredParameters.forEach(
```
sfc-gh-wtrefon, line 194: nit: could be done in more streaming way
ignoredParameters.stream().filter(param -> config.containsKey(param)).forEach(param -> LOGGER...)

src/main/java/com/snowflake/kafka/connector/internal/streaming/DefaultStreamingConfigValidator.java
```diff
@@ -0,0 +1,222 @@
...
+
+  /** Config validations specific to single buffer architecture */
+  private static class SingleBufferConfigValidator {
+
+    private static final KCLogger LOGGER =
+        new KCLogger(SingleBufferConfigValidator.class.getName());
+
+    private void logDoubleBufferingParametersWarning(Map<String, String> config) {
+      if (InternalBufferParameters.isSingleBufferEnabled(config)) {
+        Set<String> ignoredParameters =
```
sfc-gh-wtrefon, line 191: a list is enough, it does not need to be a set, we just iterate over it

sfc-gh-wtrefon (approved):

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingConfigValidator.java
```diff
@@ -0,0 +1,14 @@
+package com.snowflake.kafka.connector.internal.streaming;
+
+import com.google.common.collect.ImmutableMap;
+import java.util.Map;
+
+/** Validates connector config for Snowpipe Streaming */
+public interface StreamingConfigValidator {
```
sfc-gh-bzabek, line 7: Do you plan to create more validators? As for now It looks like these new interfaces are not really usefull.

src/main/java/com/snowflake/kafka/connector/internal/streaming/DefaultStreamingConfigValidator.java
sfc-gh-mbobowski: Great idea thx!

src/main/java/com/snowflake/kafka/connector/internal/streaming/DefaultStreamingConfigValidator.java
sfc-gh-mbobowski: Yeah, maybe it's too much to create a set here

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingConfigValidator.java
sfc-gh-mbobowski: I would say they might be useful for testing purpose.

PR #906, opened 2024-08-06
NO-SNOW Ignore OAuth test

sfc-gh-mbobowski:
# Overview

NO-SNOW Ignore UTs and ITs that connects to SF in order to get OAuth token.

Due to recent changes in security for test users on sfctest0 (including `test_kafka`) Kafka Connector tests started to fail. The reason for this failure is still under investigation.

One of the one client tests ITs is flaky - I disabled it.

The root level logging for tests was changed from `ALL` to `INFO`. The volume of logs produced on `ALL` level was so big that it was impossible to download them and check the reason of failure...

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/test/java/com/snowflake/kafka/connector/UtilsTest.java
+++ b/src/test/java/com/snowflake/kafka/connector/UtilsTest.java
@@ -6,6 +6,7 @@ import com.snowflake.kafka.connector.internal.TestUtils;
 import java.util.HashMap;
 import java.util.Map;
 import org.junit.Assert;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.contrib.java.lang.system.EnvironmentVariables;
@@ -283,6 +284,8 @@ public class UtilsTest {
   }
 
   @Test
+  @Ignore
+  // OAuth tests are temporary disabled
   public void testGetSnowflakeOAuthAccessToken() {
     Map<String, String> config = TestUtils.getConfForStreamingWithOAuth();
     if (config != null) {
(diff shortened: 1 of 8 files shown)


src/test/java/com/snowflake/kafka/connector/internal/ConnectionServiceIT.java
```diff
@@ -56,6 +57,8 @@ public void testEncryptedKey() {
   }
 
   @Test
+  @Disabled
+  // OAuth tests are temporary disabled
```
sfc-gh-lshcharbaty, line 61: Junit's disabled accepts reason as a parameter. Then this reason can be displayed in IDE and other tools
```java
@Disabled("OAuth tests are temporary disabled")
```

src/test/java/com/snowflake/kafka/connector/UtilsTest.java
```diff
@@ -283,6 +284,8 @@ public void testGetExceptionMessage() throws Exception {
   }
 
   @Test
+  @Ignore
+  // OAuth tests are temporary disabled
```
sfc-gh-lshcharbaty, line 288: Junit's ignore accepts the reason as a parameter. Then this reason can be displayed in IDE and other tools
```java
@ Ignore("OAuth tests are temporary disabled")
```

src/test/java/com/snowflake/kafka/connector/internal/ConnectionServiceIT.java
sfc-gh-mbobowski: Thx, done

PR #909, opened 2024-08-09
SNOW-1571459 parse jbdc properties

sfc-gh-bzabek:
# Overview

[SNOW-1571459 parse properties](https://snowflakecomputing.atlassian.net/browse/SNOW-1571459)
[SNOW-1571515 detect overrides](https://snowflakecomputing.atlassian.net/browse/SNOW-1571515)

Add new property: `"snowflake.jdbc.map": "isInsecureMode : true, notYetExistingProp : true"`

This will allow user to set custom snowflake jdbc properties and to handle properties that will be created for JDBC Driver in the future.
The map is not allowed to override other jdbc properties like: `snowflake.database.name, snowflake.schema,name, snowflake.private.key` etc. nor `jvm.proxy.xxx`. In such a case the error is thrown.
It is also not designed to replace them.

Properties passed to a map are not validated, for example `"foo: bar"` is allowed. In such a case we rely on Jdbc Driver behaviour.

## Pre-review checklist
- [ ] Snowpipe Changes

[SNOW-1571459]: https://snowflakecomputing.atlassian.net/browse/SNOW-1571459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/com/snowflake/kafka/connector/internal/JdbcProperties.java
```diff
@@ -0,0 +1,89 @@
...
+
+import java.util.Properties;
+
+/** Wrapper class for all snowflake jdbc properties */
+public class JdbcProperties {
+
+  /** All jdbc properties including proxyProperties */
+  private final Properties properties;
+  /** Proxy related properties */
+  private final Properties proxyProperties;
```
sfc-gh-bzabek, line 11: There is a special use case for it

sfc-gh-bzabek:
```
    this.internalStage =
        new SnowflakeInternalStage(
            (SnowflakeConnectionV1) this.conn, credentialExpireTimeMillis, proxyProperties);
```

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -60,30 +58,27 @@ public class SnowflakeConnectionServiceV1 implements SnowflakeConnectionService
...
     this.stageType = null;
-    this.proxyProperties = proxyProperties;
     this.kafkaProvider = kafkaProvider;
+    Properties proxyProperties = jdbcProperties.getProxyProperties();
+    Properties combinedProperties = jdbcProperties.getProperties();
     try {
-      if (proxyProperties != null && !proxyProperties.isEmpty()) {
-        Properties combinedProperties =
-            mergeProxyAndConnectionProperties(this.prop, this.proxyProperties);
+      if (!proxyProperties.isEmpty()) {
```
sfc-gh-bzabek, line 76: nullability is handled in JdbcProperties create().

src/main/java/com/snowflake/kafka/connector/internal/JdbcProperties.java
```diff
@@ -0,0 +1,89 @@
...
+
+  private JdbcProperties(Properties combinedProperties, Properties proxyProperties) {
+    this.properties = combinedProperties;
+    this.proxyProperties = proxyProperties;
+  }
+
+  public Properties getProperties() {
+    Properties copy = new Properties();
```
sfc-gh-mbobowski, line 19: Is there a reason for making a copy here?

sfc-gh-bzabek: Super anti bug safety? I can remove it. I would not insist on keeping that.

src/test/java/com/snowflake/kafka/connector/internal/JdbcPropertiesTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThatThrownBy;
+import static org.junit.Assert.*;
+
+import java.util.Properties;
+import org.junit.Test;
+
+public class JdbcPropertiesTest {
+
+  @Test
```
sfc-gh-mbobowski, line 11: Please add new tests using JUnit5 (import from `junit.jupiter` package)

sfc-gh-bzabek: Done in both tests.

src/test/java/com/snowflake/kafka/connector/internal/InternalUtilsTest.java
```diff
@@ -134,4 +136,18 @@ public void testResultSize() throws SQLException {
     resultSet = new MockResultSetForSizeTest(100);
     assert InternalUtils.resultSize(resultSet) == 100;
   }
+
+  @Test
```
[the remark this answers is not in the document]
sfc-gh-bzabek: Done, only `@Test` needed migration.

src/test/java/com/snowflake/kafka/connector/internal/JdbcPropertiesTest.java
sfc-gh-bzabek: Done in both classes.

src/main/java/com/snowflake/kafka/connector/internal/JdbcProperties.java
sfc-gh-bzabek: I reconsidered and removed the copy.

src/test/java/com/snowflake/kafka/connector/internal/JdbcPropertiesTest.java
sfc-gh-mbobowski: Thanks!

PR #915, opened 2024-08-19
NO-SNOW Ignore major updates on Kafka dependencies

sfc-gh-mbobowski:
# Overview

NO-SNOW Ignore major updates on Kafka dependencies

Dependabot gets confused with Confluent Kafka versioning and wants to update to community edition like [here](https://github.com/snowflakedb/snowflake-kafka-connector/pull/902/files). Preventing major updates should stop this behavior because ce / cs updates are not present in 3.x.x ([source](https://mvnrepository.com/artifact/org.apache.kafka/kafka-clients?repo=confluent-packages))

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -7,3 +7,6 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "weekly"
+    ignore:
+      - dependency-name: "org.apache.kafka:*"
+        update-types: ["version-update:semver-major"]


sfc-gh-dseweryn (approved):
I was thinking that there may be a better way to make dependabot what we want but apparently there is no easy way. We should probably spend the time elsewhere. Question though — how will we keep the kafka updated? There are already versions 3.7.1 and 3.8.0 available.

sfc-gh-mbobowski:
> I was thinking that there may be a better way to make dependabot what we want but apparently there is no easy way. We should probably spend the time elsewhere. Question though — how will we keep the kafka updated? There are already versions 3.7.1 and 3.8.0 available.

@sfc-gh-dseweryn not sure if I understand your question. The dependabot will keep track of the new Kafka releases that are not major so basically anything 3.x.y is valid.

sfc-gh-dseweryn:
> There are already versions 3.7.1 and 3.8.0 available.

This was only a side-note that we're not on top of the available versions.
I meant that if 4.0.0 will be released we won't notice

sfc-gh-mbobowski:
> > There are already versions 3.7.1 and 3.8.0 available.
> 
> This was only a side-note that we're not on top of the available versions. I meant that if 4.0.0 will be released we won't notice

That's true we have to track it on our own.

PR #916, opened 2024-08-21
SNOW-1623269 Fail sink task on authorization exception from Snowflake

sfc-gh-mbobowski:
# Overview

SNOW-1623269

Details of this error can be found in JIRA and comments left in the code. The fix is not very complicated but decided to hide it behind a parameter for safety reasons.

I tested this change end-to-end on sfctest0 cause it's tricky to test it automatically.

## Pre-review checklist
- [ ] Snowpipe Changes
- [x] This change is protected by a config parameter `enable.task.fail.on.authorization.errors`.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
@@ -16,10 +16,10 @@ import java.util.Map;
  */
 public class SnowflakeSinkTaskAuthorizationExceptionTracker {
 
-  public static final String AUTHORIZATION_EXCEPTION_MESSAGE = "Authorization failed after retry";
+  private static final String AUTHORIZATION_EXCEPTION_MESSAGE = "Authorization failed after retry";
 
-  boolean authorizationTaskFailureEnabled;
-  boolean authorizationErrorReported;
+  private boolean authorizationTaskFailureEnabled;
+  private boolean authorizationErrorReported;
 
   public SnowflakeSinkTaskAuthorizationExceptionTracker() {
     this.authorizationTaskFailureEnabled = true;
@@ -40,8 +40,7 @@ public class SnowflakeSinkTaskAuthorizationExceptionTracker {
    * @param ex - any exception that occurred during preCommit
    */
   public void reportPrecommitException(Exception ex) {
-    if (authorizationTaskFailureEnabled
-        && ex.getMessage().contains(AUTHORIZATION_EXCEPTION_MESSAGE)) {
+    if (ex.getMessage().contains(AUTHORIZATION_EXCEPTION_MESSAGE)) {
       authorizationErrorReported = true;
     }
   }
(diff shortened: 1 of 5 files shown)


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * <p>Note that exceptions thrown during preCommit() are swallowed by Kafka Connect and will not
+ * cause task failure.
+ */
+public class SnowflakeSinkTaskAuthorizationExceptionTracker {
+
+  public static final String AUTHORIZATION_EXCEPTION_MESSAGE = "Authorization failed after retry";
+
+  boolean authorizationTaskFailureEnabled;
+  boolean authorizationErrorReported;
```
sfc-gh-dseweryn, line 22: should these be private?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
```diff
@@ -0,0 +1,55 @@
...
+      authorizationErrorReported = true;
+    }
+  }
+
+  /** Throw exception if authorization has failed before */
+  public void throwExceptionIfAuthorizationFailed() {
+    if (authorizationTaskFailureEnabled && authorizationErrorReported) {
+      throw ERROR_1005.getException();
+    }
+  }
```
sfc-gh-dseweryn, line 54: I think that in `reportPrecommitException()` we do not need a feature check. toggling a field is not dangerous and gives us another possibility...

...in `throwExceptionIfAuthorizationFailed()` if the feature is disabled by default (can we check if it is default vs explicitly disabled?), we could add a warning log — something like:
> Authorization problem detected. You may want to configure "enable.task.fail.on.authorization.errors = true" to make the connector automatically pickup rotated credentials

It could help users with discoverability and fixing problem on their own — WDYT?

sfc-gh-dseweryn (approved):
Generally LGTM, two minors

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
sfc-gh-mbobowski: I think it might not be the best idea because of the following reasons:
- failing task doesn't mean that it's going to be restarted. This behaviour is specific to Confluent platform.
- authorization failure might happend from different reasons than key rotation
- there would be a lot of this log as they would be produced on every put() call

Overall it's a good idea, but I have doubts about this specific scenario.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
sfc-gh-mbobowski: Sure, done.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
sfc-gh-dseweryn: What you have just wrote could be a very fitting warning message then. We could create a latch to print it only once. Food for thought. Anyway — good job :)

sfc-gh-dseweryn (approved):

PR #920, opened 2024-09-04
SNOW-1649161 Fix NaN value handling in schematization

sfc-gh-mbobowski:
# Overview

SNOW-1649161

Special floating point value `NaN` was causing exception with schematization enabled. The reason for this failure was parsing data to an intermediate JsonNode format and then calling `writeValueAsString` which returned `""NaN""`.

This PR contains:
- bugfix itself
- integration test that was rewritten and moved to a separated file
- e2e tests to validate `NaN` working in Snowpipe / Snowpipe Streaming ingestion

## Pre-review checklist
- [ ] Snowpipe Changes


src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2AvroSchematizationIT.java
```diff
@@ -0,0 +1,243 @@
...
+import org.apache.kafka.connect.data.Struct;
+import org.apache.kafka.connect.sink.SinkRecord;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public class SnowflakeSinkServiceV2AvroSchematizationIT {
```
sfc-gh-mbobowski, line 36: The logic of this test is the same as before with added `NaN` field. A couple of things I fixed here:
1. Use String constants to avoid typos
2. Use awaitility instead of custom code
3. Use `StringUtils.deleteWhitespace` instead of custom code
4. Use JUnit assertions instead of assert
5. Use Junit @BeforeEach / @AfterEach
6. Restructure the code to have a given / when / then
7.  Move test data to test class
8. Make assertion explicit (if you have doubts take a look at what happens in `checkTableContentOneRow()`)

So while it still can be better please keep your review bar at reasonable level here :)

test/run_test_confluent.sh
```diff
@@ -204,6 +204,5 @@ if [ $testError -ne 0 ]; then
     RED='\033[0;31m'
     NC='\033[0m' # No Color
     echo -e "${RED} There is error above this line ${NC}"
-    cat $APACHE_LOG_PATH/kc.log
```
sfc-gh-mbobowski, line 207: Not sure if we need it at all. Printing 10k log lines might be useful in CI but for local development it makes things worse.

sfc-gh-gjachimko (approved):

sfc-gh-achyzy (approved):
LGTM

sfc-gh-wtrefon (approved):

PR #925, opened 2024-09-17
SNOW-1658905 Limit SnowflakeSinkConnectorConfig responsibilities

sfc-gh-mbobowski:
# Overview

SNOW-1658905 Initial config refactor

`SnowflakeSinkConnectorConfig` class has a lot of responsibilities:
- holding config keys and default values
- applying default parameters values to the given config
- storing config validators
- serving config definition in a format required by KC framework

This PR extracts some parts of this class to dedicated files. There are still many other things to be fixed around config but this is a 1 hour of work that gives some effect.

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/config/KafkaProviderValidator.java
+++ b/src/main/java/com/snowflake/kafka/connector/config/KafkaProviderValidator.java
@@ -1,0 +1,33 @@
+package com.snowflake.kafka.connector.config;
+
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.PROVIDER_CONFIG;
+
+import com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig;
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.common.config.ConfigException;
+
+/** Validator to validate Kafka Provider values which says where kafka is hosted */
+class KafkaProviderValidator implements ConfigDef.Validator {
+  public KafkaProviderValidator() {}
+
+  // This API is called by framework to ensure the validity when connector is started or when a
+  // validate REST API is called
+  @Override
+  public void ensureValid(String name, Object value) {
+    assert value instanceof String;
+    final String strValue = (String) value;
+    // The value can be null or empty.
+    try {
+      SnowflakeSinkConnectorConfig.KafkaProvider kafkaProvider =
+          SnowflakeSinkConnectorConfig.KafkaProvider.of(strValue);
+    } catch (final IllegalArgumentException e) {
+      throw new ConfigException(PROVIDER_CONFIG, value, e.getMessage());
+    }
+  }
+
+  public String toString() {
+    return "Whether kafka is running on Confluent code, self hosted or other managed service."
+        + " Allowed values are:"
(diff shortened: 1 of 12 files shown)


src/main/java/com/snowflake/kafka/connector/config/KafkaProviderValidator.java
```diff
@@ -0,0 +1,33 @@
...
+/** Validator to validate Kafka Provider values which says where kafka is hosted */
+class KafkaProviderValidator implements ConfigDef.Validator {
+  public KafkaProviderValidator() {}
+
+  // This API is called by framework to ensure the validity when connector is started or when a
+  // validate REST API is called
+  @Override
+  public void ensureValid(String name, Object value) {
+    assert value instanceof String;
+    final String strValue = (String) value;
```
sfc-gh-gjachimko, line 18: should we assert on instance of string?

src/main/java/com/snowflake/kafka/connector/config/TopicToTableValidator.java
```diff
@@ -0,0 +1,25 @@
...
+
+import com.snowflake.kafka.connector.Utils;
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.common.config.ConfigException;
+
+class TopicToTableValidator implements ConfigDef.Validator {
+  public TopicToTableValidator() {}
+
+  public void ensureValid(String name, Object value) {
+    String s = (String) value;
```
sfc-gh-gjachimko, line 11: same as above

src/main/java/com/snowflake/kafka/connector/config/KafkaProviderValidator.java
```diff
@@ -0,0 +1,33 @@
...
+
+/** Validator to validate Kafka Provider values which says where kafka is hosted */
+class KafkaProviderValidator implements ConfigDef.Validator {
+  public KafkaProviderValidator() {}
+
+  // This API is called by framework to ensure the validity when connector is started or when a
+  // validate REST API is called
+  @Override
+  public void ensureValid(String name, Object value) {
+    assert value instanceof String;
```
sfc-gh-achyzy, line 17: I think the use of `assert` is a legacy part that we should get rid of. Since we cannot enforce that they are enabled at runtime, we assume they can be disabled.

src/main/java/com/snowflake/kafka/connector/config/KafkaProviderValidator.java
sfc-gh-mbobowski: It's copied. Even if it's wrong I don't want to change anything within this PR.

src/main/java/com/snowflake/kafka/connector/config/TopicToTableValidator.java
sfc-gh-mbobowski: Same as above :)

src/main/java/com/snowflake/kafka/connector/config/KafkaProviderValidator.java
sfc-gh-mbobowski: Yes, but not in the same PR as refactor.

PR #931, opened 2024-09-20
SNOW-1642799: added exception handler when moving invalid file to table stage; file cleaner will preserve 'dirty' files a bit longer

sfc-gh-gjachimko:
# Overview

SNOW-1642799

Based on the logs analysis - when failed file is handled, it should happen that file is moved to table stage. However, it may happen that the file was removed by external process (i.e. manually) - in such case - exception would interrupt file cleaner process, which in turn - could potentially delete freshly uploaded files upon trying to resume normal operation. 

This PR does 2 things:
- in case of error when moving file to table stage - the cleaning process is not interrupted - file could be removed manually, so we just log the error and continue operation. Worst case scenario - the file would be moved in subsequent cleanup cycles - it may occupy stage for some time. 
- in case of forced restart, the cleaner will not delete "fresh" files - if cleaner process is restarting, the file may be processed by pipe logic, so the file may end up being ingested - worst case scenario - file will stay on stage for 1 hour and eventually be deleted. 

## Pre-review checklist
- [ ] Snowpipe Changes
- [x] This change is protected by a config parameter `snowflake.snowpipe.stageFileNameExtensionEnabled`.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/config/TopicToTableModeExtractor.java
+++ b/src/main/java/com/snowflake/kafka/connector/config/TopicToTableModeExtractor.java
@@ -1,38 +1,37 @@
 package com.snowflake.kafka.connector.config;
 
 import com.snowflake.kafka.connector.Utils;
-
 import java.util.Map;
 
 public class TopicToTableModeExtractor {
 
-    /**
-     * Defines whether single target table is fed by one or many source topics.
-     */
-    public enum Topic2TableMode {
-        // Single topic = single table
-        SINGLE_TOPIC_SINGLE_TABLE,
-        // Multiple topics = single table
-        MANY_TOPICS_SINGLE_TABLE,
+  /** Defines whether single target table is fed by one or many source topics. */
+  public enum Topic2TableMode {
+    // Single topic = single table
+    SINGLE_TOPIC_SINGLE_TABLE,
+    // Multiple topics = single table
+    MANY_TOPICS_SINGLE_TABLE,
+  }
 
-    }
-    private TopicToTableModeExtractor(){}
+  private TopicToTableModeExtractor() {}
 
-    /**
-     * Util method - checks if given topic is defined in topic2Table map - if it is more than once, it means
-     * multiple topics will store data in single table - in such case, for SNOWPIPE ingestion we need to
-     * uniquely identify stage files so different instances of file cleaner won't handle other's channel files.
-     * @param topic
-     * @return
-     */
-    public static Topic2TableMode determineTopic2TableMode(Map<String, String> topic2TableMap, String topic) {
-        String tableName = Utils.tableName(topic, topic2TableMap);
-        return topic2TableMap
-                .values()
-                .stream()
+  /**
+   * Util method - checks if given topic is defined in topic2Table map - if it is more than once, it
+   * means multiple topics will store data in single table - in such case, for SNOWPIPE ingestion we
+   * need to uniquely identify stage files so different instances of file cleaner won't handle
+   * other's channel files.
+   *
+   * @param topic
+   * @return
+   */
+  public static Topic2TableMode determineTopic2TableMode(
+      Map<String, String> topic2TableMap, String topic) {
+    String tableName = Utils.tableName(topic, topic2TableMap);
+    return topic2TableMap.values().stream()
                 .filter(table -> table.equalsIgnoreCase(tableName))
-                .count() > 1
-                ? Topic2TableMode.MANY_TOPICS_SINGLE_TABLE
-                : Topic2TableMode.SINGLE_TOPIC_SINGLE_TABLE;
-    }
+                .count()
+            > 1
+        ? Topic2TableMode.MANY_TOPICS_SINGLE_TABLE
+        : Topic2TableMode.SINGLE_TOPIC_SINGLE_TABLE;
+  }
 }
--- a/src/main/java/com/snowflake/kafka/connector/config/TopicToTableModeExtractor.java
+++ b/src/main/java/com/snowflake/kafka/connector/config/TopicToTableModeExtractor.java
@@ -1,0 +1,38 @@
+package com.snowflake.kafka.connector.config;
+
+import com.snowflake.kafka.connector.Utils;
+
+import java.util.Map;
+
+public class TopicToTableModeExtractor {
+
+    /**
+     * Defines whether single target table is fed by one or many source topics.
+     */
+    public enum Topic2TableMode {
+        // Single topic = single table
+        SINGLE_TOPIC_SINGLE_TABLE,
+        // Multiple topics = single table
+        MANY_TOPICS_SINGLE_TABLE,
+
+    }
+    private TopicToTableModeExtractor(){}
+
+    /**
+     * Util method - checks if given topic is defined in topic2Table map - if it is more than once, it means
+     * multiple topics will store data in single table - in such case, for SNOWPIPE ingestion we need to
+     * uniquely identify stage files so different instances of file cleaner won't handle other's channel files.
+     * @param topic
+     * @return
+     */
+    public static Topic2TableMode determineTopic2TableMode(Map<String, String> topic2TableMap, String topic) {
+        String tableName = Utils.tableName(topic, topic2TableMap);
+        return topic2TableMap
+                .values()
+                .stream()
+                .filter(table -> table.equalsIgnoreCase(tableName))
+                .count() > 1
+                ? Topic2TableMode.MANY_TOPICS_SINGLE_TABLE
+                : Topic2TableMode.SINGLE_TOPIC_SINGLE_TABLE;
+    }
+}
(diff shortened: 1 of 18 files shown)


src/main/java/com/snowflake/kafka/connector/config/TopicToTableModeExtractor.java
```diff
@@ -0,0 +1,38 @@
...
+     * @param topic
+     * @return
+     */
+    public static Topic2TableMode determineTopic2TableMode(Map<String, String> topic2TableMap, String topic) {
+        String tableName = Utils.tableName(topic, topic2TableMap);
+        return topic2TableMap
+                .values()
+                .stream()
+                .filter(table -> table.equalsIgnoreCase(tableName))
+                .count() > 1
```
sfc-gh-gjachimko, line 34: in basic scenario - the topic2TableMap would be empty - thus the condition > 1

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -430,7 +435,12 @@ private ServiceContext(
       this.cleanerFileNames = new LinkedList<>();
       this.buffer = new SnowpipeBuffer();
       this.ingestionService = conn.buildIngestService(stageName, pipeName);
-      this.prefix = FileNameUtils.filePrefix(conn.getConnectorName(), tableName, partition);
+      // SNOW-1642799 = if multiple topics load data into single table, we need to ensure prefix is unique per table - otherwise, file cleaners for different channels may run into race condition
+      if (determineTopic2TableMode(topic2TableMap, topicName) == TopicToTableModeExtractor.Topic2TableMode.MANY_TOPICS_SINGLE_TABLE) {
+        this.prefix = FileNameUtils.filePrefix(conn.getConnectorName(), tableName, topicName, partition);
+      } else {
+        this.prefix = FileNameUtils.filePrefix(conn.getConnectorName(), tableName, "", partition);
```
sfc-gh-gjachimko, line 442: we could enable this for all connectors, but this would mean breaking backward compatibility - for scenarios, where there is just 1:1 mapping between topic and table - this fix doesn't change logic, but the changed prefix would mean there could be garbage files left on stage after upgrading/downgrading. 
for scenarios when there is  n:1 mapping - there would be garbage on stage, requiring manual cleanup, but customer would get the fixed connector instead.

src/main/java/com/snowflake/kafka/connector/internal/FileNameUtils.java
```diff
@@ -78,11 +89,33 @@ static String brokenRecordFileName(String prefix, long offset, boolean isKey) {
...
+      // 4. add partition id (which should in production use cases never reach a value above 5.000
+      // partitions pers topic).
+      // In theory - we would support 32767 partitions, which is more than
+      byte[] bytes = topic.toUpperCase().getBytes(StandardCharsets.UTF_8);
+      BigInteger hash = BigInteger.valueOf(Crc32C.compute(bytes, 0, bytes.length));
+      partitionPart =
+          hash.abs()
+              .multiply(BigInteger.valueOf(0x10000))
+              .add(BigInteger.valueOf(0x8000))
+              .add(partitionPart);
```
sfc-gh-lshcharbaty, line 116: Why are we calculating the hash? The change is breaking for some of the customers anyway. Why not write a topic name normally?

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-lshcharbaty: We're breaking backward compatibility for some of the customers anyway. Isn't it simpler to make a breaking change for everyone and simplify the filename generation logic?

src/main/java/com/snowflake/kafka/connector/internal/FileNameUtils.java
sfc-gh-gjachimko: because of length limitation - AFAIK - the name of the file on stage is limited  - i did come upon a number of around 200 character (can't find it now though). the topic name could be anything up to 250 characters, so adding it as is would create huge file names.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-gjachimko: taking into account, that only 3 customers reported this specific error after migration (based on the spreadsheet Xin provided a few days ago, we've got over 200 customers using KC with affected version - (for simplicity >= 2.3.0) using SNOWPIPE ingestion, I think it would be better to keep the generation backwards compatible for them. It seems the configuration, where multiple topics feed single table is rather exception than the rule - otherwise we would probably see much more raised incidents.

src/main/java/com/snowflake/kafka/connector/internal/FileNameUtils.java
```diff
@@ -78,11 +89,34 @@ static String brokenRecordFileName(String prefix, long offset, boolean isKey) {
...
    * @return file prefix
    */
-  static String filePrefix(String appName, String table, int partition) {
-    return appName + "/" + table + "/" + partition + "/";
+  static String filePrefix(String appName, String table, String topic, int partition) {
+    if (partition >= 0x8000) {
+      throw new IllegalArgumentException(
+          String.format("partition id=%d is too large (max=%d)", partition, 0x8000));
+    }
+    BigInteger partitionPart = BigInteger.valueOf(partition);
```
sfc-gh-achyzy, line 101: Nit: We could extract it to a separate class or method

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -370,6 +406,35 @@ protected static String getNameIndex(String topic, int partition) {
     return topic + "_" + partition;
   }
 
+  public void configureSingleTableLoadFromMultipleTopcis(boolean fixEnabled) {
+    enableStageFilePrefixExtension = fixEnabled;
+  }
+
+  /**
+   * util method, checks if there are stage files present matching "old" file name format, if they
```
sfc-gh-achyzy, line 414: Nit: can we be more specific than just saying "old"? After some time, it would be hard to understand.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -430,7 +496,28 @@ private ServiceContext(
...
+      // condition
+      TopicToTableModeExtractor.Topic2TableMode mode =
+          determineTopic2TableMode(topic2TableMap, topicName);
+      if (mode == TopicToTableModeExtractor.Topic2TableMode.MANY_TOPICS_SINGLE_TABLE
+          && !enableStageFilePrefixExtension) {
+        LOGGER.warn(
+            "The table {} is used as ingestion target by multiple topics - including this one"
+                + " '{}'.\n"
+                + "To prevent potential data loss consider setting"
+                + " 'snowflake.snowpipe.stageFileNameExtensionEnabled' to true",
```
sfc-gh-achyzy, line 510: Nit: Maybe we can reference `SNOWPIPE_SINGLE_TABLE_MULTIPLE_TOPICS_FIX_ENABLED`?

sfc-gh-achyzy (approved):
LGTM. 
I left some nit comments to consider.

src/main/java/com/snowflake/kafka/connector/internal/FileNameUtils.java
```diff
@@ -11,20 +15,27 @@ public class FileNameUtils {
    * Note: all file names should using the this format
    *
    * @param appName connector name
-   * @param table table name
+   * @param table table name @Param topic name
```
sfc-gh-wtrefon, line 18: this looks strange
```suggestion
   * @param table table name 
      @param topic name
```

src/main/java/com/snowflake/kafka/connector/internal/FileNameUtils.java
sfc-gh-wtrefon: yes, and also move the methods used only in tests from production code to test code

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -370,6 +406,35 @@ protected static String getNameIndex(String topic, int partition) {
     return topic + "_" + partition;
   }
 
+  public void configureSingleTableLoadFromMultipleTopcis(boolean fixEnabled) {
```
sfc-gh-wtrefon, line 409: typo
```suggestion
  public void configureSingleTableLoadFromMultipleTopics(boolean fixEnabled) {
```

sfc-gh-wtrefon (approved):

PR #933, opened 2024-09-23
SNOW-1658905 Improve Iceberg table validations

sfc-gh-mbobowski:
# Overview

SNOW-1658905

Solving most of the TODOs and improving tests a little bit. It's still not a complete feature as it's not plugged into the connector startup.

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/test/java/com/snowflake/kafka/connector/streaming/iceberg/BaseIcebergIT.java
+++ b/src/test/java/com/snowflake/kafka/connector/streaming/iceberg/BaseIcebergIT.java
@@ -21,9 +21,15 @@ public class BaseIcebergIT {
     conn.close();
   }
 
-  protected static void createIcebergTable(String tableName) throws Exception {
+  protected static void createIcebergTable(String tableName) {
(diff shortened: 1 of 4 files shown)


src/test/java/com/snowflake/kafka/connector/streaming/iceberg/BaseIcebergIT.java
```diff
@@ -21,9 +21,15 @@ public static void teardown() {
     conn.close();
   }
 
-  protected static void createIcebergTable(String tableName) throws Exception {
+  protected static void createIcebergTable(String tableName) {
+    createIcebergTableWithColumnClause(tableName, "record_metadata object()");
+  }
+
+  protected static void createIcebergTableWithColumnClause(String tableName, String columnClause) {
```
sfc-gh-wtrefon, line 28: nice, i was just thinking about same thing for my case

sfc-gh-lshcharbaty (approved):

PR #934, opened 2024-09-23
SNOW-1658914: init record_meta column type

sfc-gh-wtrefon:
# Overview

SNOW-1658914

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
@@ -162,14 +162,15 @@ public class SnowflakeConnectionServiceV1 implements SnowflakeConnectionService 
   public void initializeMetadataColumnTypeForIceberg(String tableName) {
     checkConnection();
     InternalUtils.assertNotEmpty("tableName", tableName);
-    String query = "ALTER ICEBERG TABLE identifier(?) ALTER COLUMN RECORD_METADATA SET DATA TYPE OBJECT AS (" +
-            "offset NUMBER," +
+    String query = "ALTER ICEBERG TABLE identifier(?) ALTER COLUMN RECORD_METADATA SET DATA TYPE OBJECT(" +
+            "offset INTEGER," +
             "topic STRING," +
-            "partition NUMBER," +
+            "partition INTEGER," +
             "key STRING," +
-            "schema_id NUMBER," +
-            "key_schema_id NUMBER," +
-            "CreateTime NUMBER," +
+            "schema_id INTEGER," +
+            "key_schema_id INTEGER," +
+            "CreateTime BIGINT," +
+            "SnowflakeConnectorPushTime BIGINT," +
             "headers MAP(VARCHAR, VARCHAR)" +
             ")";
     try {
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
@@ -175,6 +175,7 @@ public class SnowflakeConnectionServiceV1 implements SnowflakeConnectionService 
             + "schema_id INTEGER,"
             + "key_schema_id INTEGER,"
             + "CreateTime BIGINT,"
+            + "LogAppendTime BIGINT,"
             + "SnowflakeConnectorPushTime BIGINT,"
             + "headers MAP(VARCHAR, VARCHAR)"
             + ")";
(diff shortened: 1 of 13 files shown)


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -158,6 +162,37 @@ public void createTableWithOnlyMetadataColumn(final String tableName) {
     LOGGER.info("Created table {} with only RECORD_METADATA column", tableName);
   }
 
+  @Override
+  public void initializeMetadataColumnTypeForIceberg(String tableName) {
+    checkConnection();
+    InternalUtils.assertNotEmpty("tableName", tableName);
+    String query =
+        "ALTER ICEBERG TABLE identifier(?) ALTER COLUMN RECORD_METADATA SET DATA TYPE OBJECT("
+            + "offset INTEGER,"
```
sfc-gh-wtrefon, line 171: double-check if the list of properties is correct and the data types

sfc-gh-mbobowski: Looks good, but I wonder if we should add `LogAppendTime BIGINT` as well. I found the following code
```
// ignore if no timestamp
    if (record.timestampType() != TimestampType.NO_TIMESTAMP_TYPE
        && metadataConfig.createtimeFlag) {
      meta.put(record.timestampType().name, record.timestamp());
    }
```

and there are also tests like `public void testTimeStamp() throws IOException {`

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergInitServiceIT.java
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  @Test
+  void shouldThrowExceptionWhenTableDoesNotExist() {
+    assertThatThrownBy(() -> icebergInitService.initializeIcebergTableProperties(tableName))
+        .isInstanceOf(SnowflakeKafkaConnectorException.class);
+  }
+
+  @Test
+  void shouldThrowExceptionWhenRecordMetadataDoesNotExist() {
```
sfc-gh-bzabek, line 66: I feel dumb. What is tested here?
We `ALTER ICEBERG TABLE identifier(?) ALTER COLUMN RECORD_METADATA ` in IcebergInitService... and also throw an error when there is no record_metadata ??

sfc-gh-bzabek: By the way you can also verify exception code. Will be more readable.
```
    assertThatThrownBy(() -> JdbcProperties.create(connection, proxy, jdbcMap))
        .isInstanceOfSatisfying(
            SnowflakeKafkaConnectorException.class,
            ex -> {
              // property key is printed not value
              assertTrue(ex.getMessage().contains("user"));
              assertEquals("0031", ex.getCode());
            });
```

sfc-gh-bzabek:
```
net.snowflake.client.jdbc.SnowflakeSQLException: SQL compilation error: error line 1 at position 47
invalid identifier 'RECORD_METADATA'
```
I don't get it.

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergInitServiceIT.java
```diff
@@ -0,0 +1,74 @@
...
+    tableName = TestUtils.randomTableName();
+  }
+
+  @AfterEach
+  public void tearDown() {
+    dropIcebergTable(tableName);
+  }
+
+  @Test
+  void shouldInitializeMetadataType() {
```
sfc-gh-bzabek, line 37: And here we create iceberg table with "record_metadata" column and then run a query to alter it with "record_metadata". How doeas it work?

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergInitServiceIT.java
sfc-gh-bzabek: aa I see we throw an error that there is no record_metadata column, when sql query throws an error...

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-wtrefon: done

sfc-gh-mbobowski (approved):

PR #935, opened 2024-09-23
SNOW-1658905 Do not run Iceberg tests outside of AWS

sfc-gh-mbobowski:
# Overview

SNOW-1658905

Iceberg Integration tests so far are configured only with AWS external volume and breaks in AZURE and GCS.

CI configuration is always taken from the master branch so this chabne can be validated only after merging. Works locally.

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/test/build_runtime_jar.sh
+++ b/test/build_runtime_jar.sh
@@ -85,9 +85,8 @@ case $BUILD_METHOD in
 	  # mvn clean should clean the target directory, hence using default pom.xml
 	  mvn -f $POM_FILE_NAME clean
 
-	  # mvn verify runs the integration test
+	  # skip Iceberg tests outside of AWS
 	  if [BUILD_FOR_CLOUD == "AWS"]; then
-	    # skip Iceberg tests outside of AWS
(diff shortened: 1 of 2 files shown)


test/build_runtime_jar.sh
```diff
@@ -80,8 +84,14 @@ case $BUILD_METHOD in
 	verify)
 	  # mvn clean should clean the target directory, hence using default pom.xml
 	  mvn -f $POM_FILE_NAME clean
+
 	  # mvn verify runs the integration test
-    mvn -f $POM_FILE_NAME verify -Dgpg.skip=true -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
+	  if [BUILD_FOR_CLOUD == "AWS"]; then
+	    # skip Iceberg tests outside of AWS
```
sfc-gh-wtrefon, line 90: i would move the comment to the `else` as here it is misleading a bit

sfc-gh-mbobowski: Done

PR #939, opened 2024-09-26
SNOW-1692725 Profiles to handle AWS specific tests

sfc-gh-mbobowski:
# Overview

SNOW-XXXXX

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/pom_confluent.xml
+++ b/pom_confluent.xml
@@ -325,29 +325,6 @@
             </plugin>
 
             <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-failsafe-plugin</artifactId>
-                <version>2.22.2</version>
-                <configuration>
-                    <includes>
-                        <include>**/*IT.java</include>
-                    </includes>
-                    <excludes>
-                        <exclude>none</exclude>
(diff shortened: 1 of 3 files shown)


pom_confluent.xml
```diff
@@ -324,29 +324,6 @@
                 </configuration>
             </plugin>
 
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-failsafe-plugin</artifactId>
-                <version>2.22.2</version>
-                <configuration>
-                    <includes>
-                        <include>**/*IT.java</include>
```
sfc-gh-wtrefon, line 333: might it be that case that your `-Dtest=!*Iceberg*IT'` param was ignored due to this setting?

sfc-gh-wtrefon (approved):

pom_confluent.xml
sfc-gh-mbobowski: We have two plugins in our project: failsafe plugin (for integration tests) and surefire plugin (for unit tests). The previous version seems to be the flag for surefire but not for failsafe.

PR #942, opened 2024-10-01
SNOW-1694636: Schematization utils refactor (PART 1)

sfc-gh-wtrefon:
# Overview

SNOW-1694636

This is the first part of the refactoring of SchematizationUtils class. It mostly splits the current code into 3 classes/layers: SchemaEvolutionService, ColumnTypeMapper and TableSchemaResolver. I tried to alter the core logic code as little as possible in this PR to be sure we don't change any current logic.

More changes to the core logic code will come in next PRs when there will be more tests covering existing logic.

Plan for future PRs:
* refactor TableSchemaResolver.resolveTableSchemaFromRecord and split it in two independent methods/classes so the Schema mapping and json mapping is not tied so tight as it is now
* write more unit tests for each of the 3 layers SchemaEvolutionService, ColumnTypeMapper and TableSchemaResolver
* extract logic for parsing InsertValidationResponse.InsertError into separate class
* most likely there will be needed more refactoring to ColumnTypeMapper classes when complex schema evolution for Iceberg comes

## Pre-review checklist
- [ ] Snowpipe Changes


src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/TableSchemaResolver.java
```diff
@@ -0,0 +1,107 @@
...
+          if (columnInfos == null) {
+            // only when the type of the value is unrecognizable for JAVA
+            throw SnowflakeErrors.ERROR_5022.getException(
+                "column: " + field.getKey() + " schemaMap: " + schemaMap);
+          }
+        }
+        columnToType.put(colName, columnInfos);
+      }
+    }
+    return new TableSchema(columnToType);
```
sfc-gh-wtrefon, line 66: this still needs refactor but it will come with next PRs, see this PR description

sfc-gh-gjachimko (approved):
that type mapping code always looks scary... lets give it a try!

PR #944, opened 2024-10-03
SNOW-1694636: Schematization utils refactor (PART 3) - Refactor schema resolver

sfc-gh-wtrefon:
# Overview

SNOW-1694636

## Pre-review checklist
- [ ] Snowpipe Changes


src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/TableSchemaResolver.java
```diff
@@ -30,40 +32,71 @@ protected TableSchemaResolver(ColumnTypeMapper columnTypeMapper) {
...
+            .map(ColumnValuePair::of)
+            .filter(pair -> columnNamesSet.contains(pair.getQuotedColumnName()))
+            .map(
+                pair ->
+                    Maps.immutableEntry(
+                        pair.getQuotedColumnName(),
+                        new ColumnInfos(inferDataTypeFromJsonObject(pair.getJsonNode()))))
+            .collect(
+                Collectors.toMap(
+                    Map.Entry::getKey, Map.Entry::getValue, (oldValue, newValue) -> newValue));
```
sfc-gh-wtrefon, line 98: open to suggestions how to make the refactoring nicer, java 8 does not help, kotlin would be perfect here ;(

sfc-gh-wtrefon: in theory we could split the schemaFromJson and schemaFromRecordSchema into two separate classes but i wouldn't invest here much more as this class is likely to be fully changed for the complex iceberg types

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/ColumnInfos.java
```diff
@@ -12,6 +12,11 @@ public ColumnInfos(String columnType, String comments) {
     this.comments = comments;
   }
 
+  public ColumnInfos(String columnType) {
```
sfc-gh-akowalczyk, line 15: Why plural form here?

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/TableSchemaResolver.java
```diff
@@ -104,4 +138,32 @@ private String inferDataTypeFromJsonObject(JsonNode value) {
     // Passing null to schemaName when there is no schema information
     return columnTypeMapper.mapToColumnType(schemaType);
   }
+
+  private static class ColumnValuePair {
+    private final String columnName;
+    private final String quotedColumnName;
+    private final JsonNode jsonNode;
+
+    public static ColumnValuePair of(Map.Entry<String, JsonNode> field) {
```
sfc-gh-akowalczyk, line 147: `of` suggests we construct the object with a given value. However, here, we apply some conversion - extracting fields, so `from` would be more suitable.

sfc-gh-akowalczyk (approved):
Only a few comments related to the naming convention.
It looks much better when the given refactoring is applied.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/TableSchemaResolver.java
```diff
@@ -30,40 +32,72 @@ protected TableSchemaResolver(ColumnTypeMapper columnTypeMapper) {
...
+
+  private TableSchema getTableSchemaFromRecordSchema(
+      SinkRecord record, Set<String> columnNamesSet) {
+    JsonNode recordNode = RecordService.convertToJson(record.valueSchema(), record.value(), true);
+    Map<String, ColumnInfos> schemaMap = getFullSchemaMapFromRecord(record);
+    Map<String, ColumnInfos> columnsInferredFromSchema =
+        Streams.stream(recordNode.fields())
+            .map(ColumnValuePair::of)
+            .filter(pair -> columnNamesSet.contains(pair.getQuotedColumnName()))
+            .peek(
```
sfc-gh-achyzy, line 67: I'm not a fan of `peek` as the docs explicitly stated: `This method exists mainly to support debugging, where you want to see the elements as they flow past a certain point in a pipeline:`
Maybe it would be better to partition/group it into separate streams?

sfc-gh-achyzy (approved):
Looks good although think about replacing `peek` operation on stream

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/ColumnInfos.java
sfc-gh-wtrefon: you are the author of this class :D

sfc-gh-wtrefon: and this is constructor so it has to be class name

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/TableSchemaResolver.java
sfc-gh-wtrefon: partitionBy looks ok

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/ColumnInfos.java
sfc-gh-akowalczyk: so that's my mistake, sorry about that 😄

PR #958, opened 2024-10-11
SNOW-0000:  extended telemetry with partitionId, fixed the event type sen…

sfc-gh-gjachimko:
…t on cleaner progress, extended telemetry logging;

# Overview

SNOW-0000:

Various telemetry improvements:
- sending kafka_pipe_usage instead of kafka_pipe_start on progress report,
- added partitionId to kafka_pipe_usage (important when troubleshooting issues for topics with many partitions)
- added telemetry logging for failed stage access

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
@@ -598,6 +598,7 @@ class SnowflakeSinkServiceV1 implements SnowflakeSinkService {
         } catch (Exception e) {
           LOGGER.warn("Cleaner and Flusher threads shut down before initialization");
         }
+        // with v2 cleaner enabled, this event is raised by the cleaner itself
         telemetryService.reportKafkaPartitionStart(pipeCreation);
       }
     }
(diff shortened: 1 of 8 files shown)


src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -593,8 +598,8 @@ private void init(long recordOffset) {
         } catch (Exception e) {
           LOGGER.warn("Cleaner and Flusher threads shut down before initialization");
         }
+        telemetryService.reportKafkaPartitionStart(pipeCreation);
       }
-      telemetryService.reportKafkaPartitionStart(pipeCreation);
```
sfc-gh-gjachimko, line 597: when v2 cleaner is used - this event is sent in file cleaner, no need to send it twice!

src/test/java/com/snowflake/kafka/connector/internal/StageFilesProcessorTest.java
```diff
@@ -457,9 +491,28 @@ void fileProcessor_WillCleanHistoryEntries_OlderThanOneHour() {
     // check if these old not matched files have been removed from history and do not result in
     // memory leak
     assertThat(register.currentProcessorContext.ingestHistory).isEmpty();
+    initAssert.run();
+  }
+
+  private Runnable configureInitialState() {
```
sfc-gh-mbobowski, line 497: It's a little bit confusing to me as the test setup and result verification is mixed. Wdyt about splitting it into two separate methods?

sfc-gh-mbobowski: Then it wouldn't have to return Runnable which suggests running it from a different thread.

sfc-gh-mbobowski (approved):

PR #963, opened 2024-10-17
SNOW-1728000 TestExecutor and TestSelector for e2e tests

sfc-gh-mbobowski:
# Overview

SNOW-1728000

`test_verify.py` was so complicated that adding any new logic there seemed impossible to me. I decided to introduce the following abstractions:
- TestSelector - which one and only role is to filter tests according to given parameters
- TestExecutor - which one and only role is to run tests

Overall there is still huge space for improvement but this refactor makes cloud vendor specific filtering much easier.

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/test/test_selector.py
+++ b/test/test_selector.py
@@ -1,0 +1,18 @@
+from test_suites import create_end_to_end_test_suites
+import test_suit
+
+# TestSelector is responsible for selecting a subset of tests to be run
+# It is meant to filter tests by platform, cloud vendor or any other predicate needed
+class TestSelector:
+
+    def selectTestsToBeRun(self, driver, nameSalt, schemaRegistryAddress, testPlatform, allowedTestsCsv):
+        test_suites = create_end_to_end_test_suites(driver, nameSalt, schemaRegistryAddress, testPlatform, allowedTestsCsv)
+
+        if testPlatform == "apache":
+            return [single_end_to_end_test.test_instance for single_end_to_end_test in test_suites.values() if single_end_to_end_test.run_in_apache == True]
+        elif testPlatform == "confluent":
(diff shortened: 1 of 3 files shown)


test/test_selector.py
```diff
@@ -0,0 +1,18 @@
+from test_suites import create_end_to_end_test_suites
+import test_suit
+
+# TestSelector is responsible for selecting a subset of tests to be run
+# It is meant to filter tests by platform, cloud vendor or any other predicate needed
+class TestSelector:
+
+    def selectTestsToBeRun(self, driver, nameSalt, schemaRegistryAddress, testPlatform, allowedTestsCsv):
+        test_suites = create_end_to_end_test_suites(driver, nameSalt, schemaRegistryAddress, testPlatform, allowedTestsCsv)
```
sfc-gh-mbobowski, line 9: `allowedTestCsv` filtering will be moved here as well.

test/test_selector.py
```diff
@@ -0,0 +1,18 @@
...
+
+        if testPlatform == "apache":
+            return [single_end_to_end_test.test_instance for single_end_to_end_test in test_suites.values() if single_end_to_end_test.run_in_apache == True]
+        elif testPlatform == "confluent":
+            return [single_end_to_end_test.test_instance for single_end_to_end_test in test_suites.values() if single_end_to_end_test.run_in_confluent == True]
+        elif testPlatform == "clean":
```
sfc-gh-mbobowski, line 15: This `clean` stuff is very wrong but separating it requires a lot of work.

sfc-gh-achyzy (approved):
LGTM

PR #969, opened 2024-10-22
SNOW-1737840: Adapt record mapping in RecordService

sfc-gh-wtrefon:
# Overview

SNOW-1737840

## Pre-review checklist
- [ ] Snowpipe Changes


src/main/java/com/snowflake/kafka/connector/records/SnowflakeTableStreamingRecordMapper.java
```diff
@@ -0,0 +1,83 @@
...
+    for (JsonNode node : row.getContent().getData()) {
+      if (schematizationEnabled) {
+        streamingIngestRow.putAll(getMapFromJsonNodeForStreamingIngest(node));
+      } else {
+        streamingIngestRow.put(TABLE_COLUMN_CONTENT, mapper.writeValueAsString(node));
+      }
+    }
+    if (includeAllMetadata) {
+      streamingIngestRow.put(TABLE_COLUMN_METADATA, mapper.writeValueAsString(row.getMetadata()));
+    }
```
sfc-gh-wtrefon, line 39: not sure if there was some trick here but i moved it from the forEach loop above to here. I don't see any sense in parsing and putting it all over again for every jsonNode. This should improve overall processing speed as we do not serialise it for every json node we have

src/test/java/com/snowflake/kafka/connector/records/IcebergTableStreamingRecordMapperTest.java
```diff
@@ -0,0 +1,392 @@
...
+import java.util.Map;
+import java.util.stream.Stream;
+import net.snowflake.client.jdbc.internal.fasterxml.jackson.core.JsonProcessingException;
+import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class IcebergTableStreamingRecordMapperTest {
```
sfc-gh-wtrefon, line 20: please think of more test cases if possible with example :)

src/main/java/com/snowflake/kafka/connector/records/IcebergTableStreamingRecordMapper.java
```diff
@@ -0,0 +1,104 @@
...
+  private static final TypeReference<Map<String, Object>> OBJECTS_MAP_TYPE_REFERENCE =
+      new TypeReference<Map<String, Object>>() {};
+
+  public IcebergTableStreamingRecordMapper(ObjectMapper objectMapper) {
+    this.mapper = objectMapper;
+  }
+
+  @Override
+  public Map<String, Object> processSnowflakeRecord(
+      SnowflakeTableRow row, boolean schematizationEnabled, boolean includeAllMetadata)
```
sfc-gh-mbobowski, line 31: nit: maybe just `boolean includeMetadata`?

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
```diff
@@ -103,54 +95,29 @@ public class RecordService {
...
   private SnowflakeMetadataConfig metadataConfig = new SnowflakeMetadataConfig();
 
-  RecordService(Clock clock) {
+  RecordService(
+      Clock clock,
+      boolean enableSchematization,
+      StreamingRecordMapper streamingRecordMapper,
+      ObjectMapper mapper) {
     this.clock = clock;
+    this.enableSchematization = enableSchematization;
```
sfc-gh-mbobowski, line 104: I wonder if `RecordService` needs to be aware of `enableSchematization` value after creating `StreamingRecordMapper` abstraction. In fact we could pass this flag to a mapper in a factory instead of `processSnowflakeRecord` argument.

src/test/java/com/snowflake/kafka/connector/records/IcebergTableStreamingRecordMapperTest.java
sfc-gh-mbobowski: Looks exhaustive to me, let's wait for the bugs to happen :)

sfc-gh-mbobowski (approved):
Left some minor comments but great change overall!

src/main/java/com/snowflake/kafka/connector/records/RecordService.java
sfc-gh-wtrefon: good idea

PR #924, opened 2024-09-12
NO-SNOW Enable single buffering in Snowpipe Streaming by default

sfc-gh-mbobowski:
# Overview

NO-SNOW Enable single buffering in Snowpipe Streaming by default

## Pre-review checklist
- [ ] This change should be part of a Behavior Change Release. See [go/behavior-change](http://go/behavior-change).
- [ ] This change has passed Merge gate tests
- [ ] Snowpipe Changes
- [x] Snowpipe Streaming Changes
- [ ] This change is TEST-ONLY
- [ ] This change is README/Javadocs only
- [ ] This change is protected by a config parameter <PARAMETER_NAME> eg `snowflake.ingestion.method`.
    - [ ] `Yes` - Added end to end and Unit Tests. 
    - [ ] `No` - Suggest why it is not param protected
- [ ] Is his change protected by parameter <PARAMETER_NAME> on the server side?
    - [ ] The parameter/feature is not yet active in production (partial rollout or PrPr, see [Changes for Unreleased Features and Fixes](http://go/ppp-prpr)).
    - [ ] If there is an issue, it can be safely mitigated by turning the parameter off. This is also verified by a test (See [go/ppp](http://go/ppp)).

PR #981, opened 2024-10-29
Revert "NO-SNOW Enable single buffering in Snowpipe Streaming by default (#924)"

sfc-gh-mbobowski:
This reverts commit f0e5f230ff710e809671f5f0a27b23302c8cb3ba.

# Overview

We decided to not enable this setting by default in 2.5.0 release because we are not sure what is the impact on the first read performance.

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -133,7 +133,7 @@ public class SnowflakeSinkConnectorConfig {
   public static final String SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER =
       "snowflake.streaming.enable.single.buffer";
 
-  public static final boolean SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER_DEFAULT = true;
+  public static final boolean SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER_DEFAULT = false;
   public static final String SNOWPIPE_STREAMING_MAX_CLIENT_LAG =
       "snowflake.streaming.max.client.lag";
 
--- a/src/test/java/com/snowflake/kafka/connector/ConnectorConfigValidatorTest.java
+++ b/src/test/java/com/snowflake/kafka/connector/ConnectorConfigValidatorTest.java
@@ -1,6 +1,19 @@
 package com.snowflake.kafka.connector;
 
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.*;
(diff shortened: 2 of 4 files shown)


sfc-gh-rsawicki (approved):

src/test/java/com/snowflake/kafka/connector/ConnectorConfigValidatorTest.java
```diff
@@ -495,7 +495,6 @@ public void testStreamingEmptyFlushTime() {
         IngestionMethodConfig.SNOWPIPE_STREAMING.toString());
     config.put(Utils.SF_ROLE, "ACCOUNTADMIN");
     config.remove(SnowflakeSinkConnectorConfig.BUFFER_FLUSH_TIME_SEC);
-    config.put(SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER, "false");
```
sfc-gh-wtrefon, line 498: we could keep that, it will be later easier to set the default back to true

sfc-gh-wtrefon (approved):
lgtm, just consider keeping `config.put(SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER, "false"); ` in the tests

src/test/java/com/snowflake/kafka/connector/ConnectorConfigValidatorTest.java
sfc-gh-akowalczyk: But leaving the line may suggest we put a meaningful configuration value in this test. Deleting this makes it clear we use the default one.

sfc-gh-akowalczyk (approved):

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -133,7 +133,7 @@ public class SnowflakeSinkConnectorConfig {
   public static final String SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER =
       "snowflake.streaming.enable.single.buffer";
 
-  public static final boolean SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER_DEFAULT = true;
+  public static final boolean SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER_DEFAULT = false;
```
sfc-gh-dseweryn, line 136: So we decided not to introduce / change default values of the SDK flush timeout / flush size?

sfc-gh-mbobowski: Not in 2.5.0. We'll try to figure this out for the next release.

sfc-gh-dseweryn (approved):
Oh, it's a revert 😓

PR #994, opened 2024-11-12
No snow use com.fasterxml json

sfc-gh-bzabek:
# Overview

NO-SNOW

I need to have standard `com.fasterxml` to be compatible with IcebergUtils when parsing a KafkaRecord into Iceberg Type. Entire connector uses `net.snowflake.client.jdbc.internal.fasterxml`.

It's because snowflake JDBC driver uses "shading" in its pom.xml. Even though the driver uses standard `com.fasterxml` after compiling it into a jar the package is `net.snowflake.client.jdbc.internal.fasterxml`. 
JDBC driver exposed telemetry method that requires `net.snowflake.client.jdbc.internal.fasterxml` as an argument - `TelemetryUtil.buildJobData(ObjectNode obj)`. As a result the whole connector is bounded to the type inherited from the driver.

There are to approaches two solve it:
- change imports in the entire connector and map `com.fasterxml` -> `net.snowflake.client.jdbc.internal.fasterxml` in `SnowflakeTelemetryService.send` method which I made in a first commit (see: [link](https://github.com/snowflakedb/snowflake-kafka-connector/pull/994/commits/5a4f0cb8002e3a1083305f9b297f3456498668dd#diff-906ff03bb77ba2299218cd6967f8fba65b7dd5452e01959e2def8bca9e49913eR206))
- leave `net.snowflake.client.jdbc.internal.fasterxml` as it is only in telemetry related services (second commit)

Or I can do hacks before parsing a json to Iceberg.

## Pre-review checklist
- [ ] Snowpipe Changes


sfc-gh-mbobowski (approved):
Kafka Connector and snowflake-jdbc use the same `2.17.2` version of Jackson so it shouldn't be a problem to replace it.

Could you ask JDBC team what is the reason for using shaded library in the public interface? Maybe they can provide more convenient api for us?

sfc-gh-bzabek:
> Kafka Connector and snowflake-jdbc use the same `2.17.2` version of Jackson so it shouldn't be a problem to replace it.
> 
> Could you ask JDBC team what is the reason for using shaded library in the public interface? Maybe they can provide more convenient api for us?

Yes I will ask them.

PR #995, opened 2024-11-12
SNOW-1728002 get rid of reflection when enabling iceberg streaming

sfc-gh-bzabek:
# Overview

related to SNOW-1728002
It's temporary approach. Finally we will set iceberg mode by invoking SnowflakeStreamingIngestClientFactory.builder.setIceberg(true) but the method is not yet public.

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
@@ -91,7 +91,7 @@ public class StreamingClientProperties {
     // Override only if the streaming client properties are explicitly set in config
     this.parameterOverrides = new HashMap<>();
     if (isIcebergEnabled) {
-      parameterOverrides.put("ENABLE_ICEBERG_STREAMING", "true");
+      parameterOverrides.put("enable_iceberg_streaming", "true");
     }
     Optional<String> snowpipeStreamingMaxClientLag =
         Optional.ofNullable(connectorConfig.get(SNOWPIPE_STREAMING_MAX_CLIENT_LAG));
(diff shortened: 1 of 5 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
```diff
@@ -90,6 +90,11 @@ public StreamingClientProperties(Map<String, String> connectorConfig) {
 
     // Override only if the streaming client properties are explicitly set in config
     this.parameterOverrides = new HashMap<>();
+    if (isIcebergEnabled) {
+      // TODO should be replaced by SnowflakeStreamingIngestClientFactory.builder.setIceberg(true)
+      // call in SNOW-1728002
+      this.parameterOverrides.put("enable_iceberg_streaming", "true");
+    }
```
sfc-gh-bzabek, line 97: By the way the param is set in IcebergIngestionIT setUp.

sfc-gh-bzabek:
I see my commits name are bad, i'll change them while squashing.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
sfc-gh-bzabek: There are no enum for "enable_iceberg_streaming" in ingest-sdk.

sfc-gh-mbobowski: I found one. Please try: `ParameterProvider.ENABLE_ICEBERG_STREAMING`

sfc-gh-mbobowski: Additionally please remove the TODO comment and add `SNOW-1728002` to the PR name. This is everything that needs to be done in this scope.

sfc-gh-bzabek: > I found one. Please try: ParameterProvider.ENABLE_ICEBERG_STREAMING

I had to update ingest-sdk for it.

> Additionally please remove the TODO comment and add SNOW-1728002 to the PR name. This is everything that needs to be done in scope of this JIRA.

Done

sfc-gh-bzabek: When I used variable from ParameterProvider it fails to compile, I think ingest-sdk 3.0.0 is not yet available

sfc-gh-mbobowski (approved):

PR #999, opened 2024-11-14
SNOW-1805174 Block iceberg mode with double buffer

sfc-gh-mbobowski:
# Overview

SNOW-1805174

Iceberg ingestion is failing for double buffer and it is not worth fixing it as single buffer will become a new default soon.

To make it user friendly I propose to make single buffer default for iceberg and return error when explicitly set to false.

## Pre-review checklist
- [ ] Snowpipe Changes


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -253,6 +254,18 @@ public static void setDefaultValues(Map<String, String> config) {
         config, BUFFER_FLUSH_TIME_SEC, BUFFER_FLUSH_TIME_SEC_DEFAULT, "seconds");
 
     if (isSnowpipeStreamingIngestion(config)) {
```
sfc-gh-mbobowski, line 256: Note that this code is executed before config validators are run.

PR #1005, opened 2024-11-20
SNOW-1805174 Enable single buffer by default

sfc-gh-mbobowski:
# Overview

SNOW-1805174

This PR contains:
- enabling single buffer by default + fixes for some tests
- validation that blocks running iceberg ingestion with double buffer

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/test/test_suit/test_schema_evolution_multi_topic_drop_table.py
+++ b/test/test_suit/test_schema_evolution_multi_topic_drop_table.py
@@ -62,7 +62,7 @@ class TestSchemaEvolutionMultiTopicDropTable(BaseE2eTest):
             self.driver.sendBytesData(topic, value, key)
 
         # Sleep for some time and then verify the rows are ingested
-        sleep(120)
+        sleep(100)
         self.verify("0")
 
         # Recreate the table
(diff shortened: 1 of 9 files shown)


test/test_suit/test_schema_evolution_multi_topic_drop_table.py
```diff
@@ -62,7 +62,7 @@ def send(self):
             self.driver.sendBytesData(topic, value, key)
 
         # Sleep for some time and then verify the rows are ingested
-        sleep(120)
+        sleep(100)
```
sfc-gh-mbobowski, line 65: Investigated this test a little bit as it was the only one e2e that started to fail after my changes.

The test lasts for a significant amount of time not only because of ingest sdk lag, but also because of retry lags in KC framework itself (tried to tune it with "errors.retry.delay.max.ms" but didn't see any improvement). I also validated that the columns appear in the table with `sleep(240)` and default lag settings - anyway I decided to make it faster with lower lag.

sfc-gh-rsawicki (approved):

PR #1007, opened 2024-11-21
SNOW-1729292 modify iceberg tree based on record data

sfc-gh-bzabek:
# Overview

SNOW-1729292

There is still a lot work to be done. Tell me if you find this approach good.
1. Ingest-sdk returns list of columnsToEvolve inside an InsertError
2. get schemas from channel for these columns
3. get column schemas from record
4. resolve which column is new 
- if column from record is not present in channels schema than it's new
- add this column
5. resolve which column is modified
- if the column is present in a channel it means it's not new - hence it's to be modified
- merge column from channel with a column from record
- set data type 

The whole logic responsible for iceberg schema evolution is package private. Only IcebergSchemaEvolutionService is public. 

It doesn't work with ingest-sdk 3.0.0, local jar needs to be build from ingest-sdk master.

**TODO:**
- I didn't inspect if columns nullability evolution is handled
- column comments are not handled
- tests for potential SQL injection (I think we need another Jira for it)

## Pre-review checklist
- [ ] Snowpipe Changes


src/test/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/ParseIcebergColumnTreeTest.java
```diff
@@ -0,0 +1,205 @@
...
+        arguments(
+            "{ \"testColumnName\": {" + "\"k1\" : 1," + "\"k2\" : 2" + "} " + "}",
+            "OBJECT(k1 LONG, k2 LONG)"),
+        arguments(
+            "{ \"testColumnName\": {"
+                + "\"k1\" : { \"nested_key1\" : 1},"
+                + "\"k2\" : { \"nested_key2\" : 2}"
+                + "}}",
+            "OBJECT(k1 OBJECT(nested_key1 LONG), k2 OBJECT(nested_key2 LONG))"),
+        arguments(
```
sfc-gh-bzabek, line 113: The order of vehicle1 and vehicle2 has changed after the parsing it's worrisome. However k1 and k2 do not change order.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergColumnTree.java
```diff
@@ -1,15 +1,39 @@
...
+class IcebergColumnTree {
 
   private final IcebergFieldNode rootNode;
 
-  public IcebergColumnTree(ApacheIcebergColumnSchema columnSchema) {
-    this.rootNode = new IcebergFieldNode(columnSchema.getColumnName(), columnSchema.getSchema());
+  String getColumnName() {
+    return rootNode.name;
+  }
+
```
sfc-gh-bzabek, line 13: Below two contructor may be refactored.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -514,13 +514,36 @@ public void appendColumnsToTable(String tableName, Map<String, ColumnInfos> colu
...
    *
    * @param tableName the name of the table
-   * @param columnInfosMap the mapping from the columnNames to their infos
+   * @param addColumnsQuery
+   * @param alterSetDataTypeQuery
    */
   @Override
   public void appendColumnsToIcebergTable(
-      String tableName, Map<String, ColumnInfos> columnInfosMap) {
+      String tableName, String addColumnsQuery, String alterSetDataTypeQuery) {
```
sfc-gh-mbobowski, line 522: Wdyt about using `Optional<String>`? I think it's more explicit than passing an empty String.

sfc-gh-bzabek: Yes this is not good... I will change it.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -144,10 +145,8 @@ public SnowflakeSinkServiceV2(
             Utils.isIcebergEnabled(connectorConfig), schematizationEnabled);
     this.icebergTableSchemaValidator = new IcebergTableSchemaValidator(conn);
     this.icebergInitService = new IcebergInitService(conn);
-    this.schemaEvolutionService =
-        Utils.isIcebergEnabled(connectorConfig)
-            ? new IcebergSchemaEvolutionService(conn)
-            : new SnowflakeSchemaEvolutionService(conn);
+    this.schemaEvolutionService = new SnowflakeSchemaEvolutionService(conn);
```
sfc-gh-mbobowski, line 148: I liked the idea of strategy pattern for `IcebergSchemaEvolutionService` and `SnowflakeSchemaEvolutionService` because we could hide some context from 'DirectTopicPartitionChannel`. Can we just pass `channel.getTableSchema()` even when it's not used for non-iceberg?

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergColumnTypeMapper.java
```diff
@@ -19,23 +18,25 @@
...
-  public static String mapToSnowflakeDataType(Type apacheIcebergType) {
+  String mapToSnowflakeDataType(Type apacheIcebergType) {
     switch (apacheIcebergType.typeId()) {
       case BOOLEAN:
         return "BOOLEAN";
       case INTEGER:
         return "NUMBER(10,0)";
       case LONG:
-        return "NUMBER(19,0)";
+        return "LONG";
```
sfc-gh-mbobowski, line 36: Did it fail with the previous types?

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergColumnTypeMapper.java
```diff
@@ -101,9 +112,15 @@ public String mapToColumnType(Schema.Type kafkaType, String schemaName) {
...
+        return "ARRAY";
+      case STRUCT:
+        return "OBJECT";
+      case MAP:
+        return "MAP";
       default:
-        // MAP and STRUCT will go here
-        throw new IllegalArgumentException("Arrays, struct and map not supported!");
+        // todo try to throw info about a whole record - this is pure
+        throw new IllegalArgumentException(
```
sfc-gh-mbobowski, line 122: nit: I think we could introduce a new 5xxx error code (like ERROR_5001).

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergColumnTypeMapper.java
```diff
@@ -19,23 +18,25 @@
 import org.apache.kafka.connect.data.Time;
 import org.apache.kafka.connect.data.Timestamp;
 
-public class IcebergColumnTypeMapper extends ColumnTypeMapper {
+class IcebergColumnTypeMapper {
+
+  static IcebergColumnTypeMapper INSTANCE = new IcebergColumnTypeMapper();
```
sfc-gh-mbobowski, line 23: While I understand why it was introduced, I think we can do better. If the mapping is moved from the node class to some kind of factory service this singleton won't be needed.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergFieldNode.java
```diff
@@ -1,24 +1,83 @@
...
+    modifiedNode.children.forEach(
+        (key, value) -> {
+          IcebergFieldNode thisChild = this.children.get(key);
+          if (thisChild == null) {
+            this.children.put(key, value);
+          } else {
+            thisChild.merge(value);
+          }
+        });
+    addNewChildren(modifiedNode);
```
sfc-gh-mbobowski, line 48: I would say this is more straightforward: `addNewChildren(modifiedNode.children)`

sfc-gh-mbobowski: or even `addChildrenIfAbsent(modifiedNode.children)`

sfc-gh-mbobowski: As discussed - this line is most probably not needed because new nodes were already added in the loop above.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergFieldNode.java
```diff
@@ -1,24 +1,83 @@
...
+    this.name = name;
+    this.snowflakeIcebergType = mapper.mapToColumnDataTypeFromJson(jsonNode);
+    this.children = produceChildren(jsonNode);
+  }
+
+  /**
+   * Method does not modify, delete any existing nodes and its types, names. It is meant only to add
+   * new children.
+   */
+  void merge(IcebergFieldNode modifiedNode) {
```
sfc-gh-mbobowski, line 38: nit: find a better name for modifiedNode. Maybe just `nodeToMerge`? Not an easy one to name.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergFieldNode.java
```diff
@@ -61,26 +120,23 @@ StringBuilder buildQuery(StringBuilder sb, String parentType) {
     return sb;
   }
 
-  private StringBuilder appendNameAndType(StringBuilder sb) {
```
sfc-gh-mbobowski, line 64: Let's separate SQL generation to a separate class. Node class is having a lot of responsibilities right now.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergSchemaEvolutionService.java
```diff
@@ -3,53 +3,98 @@
...
-  private final TableSchemaResolver tableSchemaResolver;
+  private final IcebergTableSchemaResolver icebergTableSchemaResolver;
 
   public IcebergSchemaEvolutionService(SnowflakeConnectionService conn) {
     this.conn = conn;
-    this.tableSchemaResolver = new IcebergTableSchemaResolver();
+    this.icebergTableSchemaResolver = new IcebergTableSchemaResolver();
   }
 
   @VisibleForTesting
```
sfc-gh-mbobowski, line 28: Actually it is not used anymore. You can safely delete it.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergTableSchemaResolver.java
```diff
@@ -1,16 +1,101 @@
...
+class IcebergTableSchemaResolver {
 
-  @VisibleForTesting
-  IcebergTableSchemaResolver(IcebergColumnTypeMapper columnTypeMapper) {
-    super(columnTypeMapper);
+  /**
+   * Retrieve IcebergSchema stored in a channel, then parse it into a tree. Filter out columns that
+   * do not need to be modified.
+   */
+  public List<IcebergColumnTree> resolveIcebergSchemaFromChannel(
```
sfc-gh-mbobowski, line 22: nit: I am not 100% sure but perhaps it would be more elegant to return a single tree here rather than list of trees.

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergIngestionSchemaEvolutionIT.java
```diff
@@ -109,20 +186,21 @@ private void assertRecordsInTable() {
...
-              new DescribeTableRow("ID_INT16", "NUMBER(10,0)"),
-              new DescribeTableRow("ID_INT32", "NUMBER(10,0)"),
-              new DescribeTableRow("ID_INT64", "NUMBER(19,0)"),
-              new DescribeTableRow("DESCRIPTION", "VARCHAR(16777216)"),
-              new DescribeTableRow("RATING_FLOAT32", "FLOAT"),
-              new DescribeTableRow("RATING_FLOAT64", "FLOAT"),
-              new DescribeTableRow("APPROVAL", "BOOLEAN")
-            },
-            true),
+        // READING SCHEMA FROM A RECORD IS NOT YET SUPPORTED.
```
sfc-gh-mbobowski, line 189: unfinished test?

src/main/java/com/snowflake/kafka/connector/records/IcebergTableStreamingRecordMapper.java
```diff
@@ -54,7 +53,7 @@ private Map<String, Object> getMapForSchematization(JsonNode node) {
             entry ->
                 new AbstractMap.SimpleEntry<>(
                     Utils.quoteNameIfNeeded(entry.getKey()), entry.getValue()))
-        .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+        .collect(HashMap::new, (m, v) -> m.put(v.getKey(), v.getValue()), HashMap::putAll);
```
sfc-gh-bzabek, line 56: This lin threw NPE when entry value was null. This is known JDK bug https://stackoverflow.com/questions/24630963/nullpointerexception-in-collectors-tomap-with-null-entry-values.
Record that contained `"array" : null` threw that NPE.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-bzabek: Having method with unused field is a drawback hence I tried to go the new way, I thought it'a a lesser evil. 
Anyway I can go back to previous state.

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergIngestionSchemaEvolutionIT.java
sfc-gh-bzabek: I temporary broke schema evolution based on schema. Now it works again.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/ApacheIcebergColumnSchema.java
```diff
@@ -3,22 +3,22 @@
 import org.apache.iceberg.types.Type;
 
 /** Wrapper class for Iceberg schema retrieved from channel. */
-public class ApacheIcebergColumnSchema {
+class ApacheIcebergColumnSchema {
```
sfc-gh-wtrefon, line 6: I dont think we need org name, we never used it any other places
```suggestion
class IcebergColumnSchema {
```

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -499,39 +499,54 @@ public boolean hasSchemaEvolutionPermission(String tableName, String role) {
...
+   * @param addColumnsQuery ADD COLUMN query
+   * @param alterSetDataTypeQuery SET DATA TYPE query
    */
   @Override
-  public void appendColumnsToTable(String tableName, Map<String, ColumnInfos> columnInfosMap) {
-    LOGGER.debug("Appending columns to snowflake table");
-    appendColumnsToTable(tableName, columnInfosMap, false);
+  public void appendColumnsToIcebergTable(
+      String tableName, Optional<String> addColumnsQuery, Optional<String> alterSetDataTypeQuery) {
+    LOGGER.debug("Appending columns to iceberg table");
```
sfc-gh-wtrefon, line 509: move the log to the if statements and now it's not entirely true.

sfc-gh-wtrefon: and also add log for the case when altering datatype

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergTableSchemaResolver.java
sfc-gh-bzabek: Single tree represent one column. I wanted to get all columns for modification/addition.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergFieldNode.java
sfc-gh-bzabek: Good catch.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergColumnTree.java
```diff
@@ -1,15 +1,39 @@
...
+  IcebergColumnTree(ApacheIcebergColumnSchema columnSchema) {
```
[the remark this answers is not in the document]
sfc-gh-bzabek: Isn't it a scope of: https://snowflakecomputing.atlassian.net/browse/SNOW-1731255 ?

The PR is very huge without it :/

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergSchemaEvolutionService.java
```diff
@@ -61,4 +110,56 @@ public void evolveSchemaIfNeeded(SchemaEvolutionTargetItems targetItems, SinkRec
...
+      String dataType = column.buildType();
+
+      setDataTypeQuery.append(columnName).append(" set data type ").append(dataType).append(", ");
+    }
+    // remove last comma and whitespace
+    setDataTypeQuery.deleteCharAt(setDataTypeQuery.length() - 1);
+    setDataTypeQuery.deleteCharAt(setDataTypeQuery.length() - 1);
+
+    return Optional.of(setDataTypeQuery.toString());
+  }
```
sfc-gh-wtrefon, line 164: I don't like this. We mix abstraction layers here. We shouldn't create any sql here. We should stick to the existing abstraction and create the sql inside the SnowflakeConnectionService

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -499,39 +499,46 @@ public boolean hasSchemaEvolutionPermission(String tableName, String role) {
...
    *
    * @param tableName the name of the table
-   * @param columnInfosMap the mapping from the columnNames to their infos
+   * @param query ADD COLUMN or ALTER SET DATA TYPE query
    */
   @Override
-  public void appendColumnsToTable(String tableName, Map<String, ColumnInfos> columnInfosMap) {
-    LOGGER.debug("Appending columns to snowflake table");
-    appendColumnsToTable(tableName, columnInfosMap, false);
+  public void evolveIcebergColumns(String tableName, String query) {
```
sfc-gh-wtrefon, line 508: see my comment in IcebergSchemaEvolutionService. The current implementation is just a pure proxy and executes arbitrary queries, it isn't in any way related to the iceberg. IMO we should generate the sql code here

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/SchemaEvolutionService.java
```diff
@@ -12,5 +14,8 @@ public interface SchemaEvolutionService {
    *     nullability, and columns to add
    * @param record the sink record that contains the schema and actual data
    */
-  void evolveSchemaIfNeeded(SchemaEvolutionTargetItems targetItems, SinkRecord record);
+  void evolveSchemaIfNeeded(
+      SchemaEvolutionTargetItems targetItems,
+      SinkRecord record,
+      Map<String, ColumnProperties> schemaAlreadyInUse);
```
sfc-gh-wtrefon, line 20: can't we put it inside SchemaEvolutionTargetItems so we don't change the api here? I think it fits queite well

sfc-gh-wtrefon, line 20: wouldn't it be better name?
```suggestion
      Map<String, ColumnProperties> existingSchema);
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergFieldNode.java
sfc-gh-wtrefon: agree

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergFieldNode.java
```diff
@@ -61,26 +68,145 @@ StringBuilder buildQuery(StringBuilder sb, String parentType) {
...
+  /**
+   * Method does not modify, delete any existing nodes and its types, names. It is meant only to add
```
sfc-gh-wtrefon, line 72: this description is misleading, it modifies the node by adding children

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergFieldNode.java
```diff
@@ -61,26 +68,145 @@ StringBuilder buildQuery(StringBuilder sb, String parentType) {
...
+  void merge(IcebergFieldNode modifiedNode) {
+    modifiedNode.children.forEach(
+        (key, node) -> {
+          IcebergFieldNode thisChild = this.children.get(key);
+          if (thisChild == null) {
+            this.children.put(key, node);
+          } else {
+            thisChild.merge(node);
+          }
+        });
```
sfc-gh-wtrefon, line 84: what if we have nodes on different lvl with same name, how we know where we should add it?

sfc-gh-wtrefon: and as this is called recursive, we might end up with adding children to many nodes on different levels

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergFieldNode.java
```diff
@@ -61,26 +68,145 @@ StringBuilder buildQuery(StringBuilder sb, String parentType) {
...
+      return produceChildrenFromStruct(connectSchema);
+    }
+    if (connectSchema.type() == Schema.Type.MAP) {
+      return produceChildrenFromMap(connectSchema);
+    }
+    if (connectSchema.type() == Schema.Type.ARRAY) {
+      return produceChildrenForArray(connectSchema);
+    } else { // isPrimitive == true
+      return new LinkedHashMap<>();
+    }
```
sfc-gh-wtrefon, line 141: i would skip the else statement for readability
```suggestion
    } 
    // primitive types don't have children
    return new LinkedHashMap<>();
    
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergSchemaEvolutionService.java
sfc-gh-mbobowski: I mean the constructor below marked with `@VisibleForTesting` (not affected by your changes).

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergFieldNode.java
```diff
@@ -61,26 +68,145 @@ StringBuilder buildQuery(StringBuilder sb, String parentType) {
...
+                    new IcebergFieldNode(
+                        stringJsonNodeEntry.getKey(), stringJsonNodeEntry.getValue()),
+                (v1, v2) -> {
+                  throw new IllegalArgumentException("Two same keys: " + v1);
+                },
+                LinkedHashMap::new));
+  }
+
+  private LinkedHashMap<String, IcebergFieldNode> produceChildren(Schema connectSchema) {
+    Schema.Type type = connectSchema.type();
```
sfc-gh-wtrefon, line 130: unused

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergFieldNode.java
```diff
@@ -1,54 +1,61 @@
...
 import java.util.stream.Collectors;
 import org.apache.iceberg.types.Type;
 import org.apache.iceberg.types.Types;
+import org.apache.kafka.connect.data.Field;
+import org.apache.kafka.connect.data.Schema;
 
 class IcebergFieldNode {
 
-  public final String name;
+  // todo consider refactoring into some more classes
```
sfc-gh-wtrefon, line 16: i would split it into two classes at least. 
1. IcebergFieldNode - simple DTO class containing name, type and list of children
2. Some service/bean class responsible for creating and merging IcebergFieldNode

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergSchemaEvolutionService.java
```diff
@@ -17,39 +20,85 @@ public class IcebergSchemaEvolutionService implements SchemaEvolutionService {
...
+                    .filter(c -> c.getColumnName().equals(existingColumn.getColumnName()))
+                    .collect(Collectors.toList())
+                    .get(0);
+            existingColumn.merge(mewVersion);
+          });
+      Optional<String> alterSetDataTypeQuery = alterSetDataTypeQuery(alreadyExistingColumns);
       try {
-        conn.appendColumnsToIcebergTable(tableName, tableSchema.getColumnInfos());
+        addColumnsQuery.ifPresent(query -> conn.evolveIcebergColumns(tableName, query));
+        alterSetDataTypeQuery.ifPresent(query -> conn.evolveIcebergColumns(tableName, query));
```
sfc-gh-wtrefon, line 101: I would split this big wall of text to at least two methods: addingColumns, alteringColumnTypes

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergSchemaEvolutionService.java
```diff
@@ -61,4 +110,56 @@ public void evolveSchemaIfNeeded(SchemaEvolutionTargetItems targetItems, SinkRec
...
+
+  /**
+   * Get only column names, ignore nested field names. Remove double quotes.
+   *
+   * <p>example: TEST_STRUCT.field1 -> TEST_STRUCT
+   */
+  private Set<String> getColumnsToEvolve(SchemaEvolutionTargetItems targetItems) {
+    return targetItems.getColumnsToAdd().stream()
+        // remove double quotes
+        .map(targetItem -> targetItem.split("\\.")[0].replaceAll("\"", ""))
```
sfc-gh-wtrefon, line 122: instead of commenting, you could just extract it to method removeDoubleQouetes :P

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-bzabek: [refactor the way we execute query, plus format.sh](https://github.com/snowflakedb/snowflake-kafka-connector/pull/1007/commits/da4bbf93dc186ad278eeb12b16695bf20e0de26b)

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergTableSchemaResolver.java
```diff
@@ -1,16 +1,133 @@
...
+      ArrayList<Field> foundColumns = new ArrayList<>();
+      ArrayList<Field> notFoundColumns = new ArrayList<>();
+
+      for (Field field : schema.fields()) {
+        if (columnsToEvolve.contains(field.name().toUpperCase())) {
+          foundColumns.add(field);
+        } else {
+          notFoundColumns.add(field);
+        }
+      }
```
sfc-gh-wtrefon: we could shorten it like this
```suggestion
List<Field> foundColumns = schema.fields().stream()
    .filter(field -> columnsToEvolve.contains(field.name().toUpperCase()))
    .collect(Collectors.toList());

List<Field> notFoundColumns = schema.fields().stream()
    .filter(field -> !columnsToEvolve.contains(field.name().toUpperCase()))
    .collect(Collectors.toList());
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergTableSchemaResolver.java
```diff
@@ -1,16 +1,133 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-public class IcebergTableSchemaResolver extends TableSchemaResolver {
+class IcebergTableSchemaResolver {
 
-  @VisibleForTesting
-  IcebergTableSchemaResolver(IcebergColumnTypeMapper columnTypeMapper) {
-    super(columnTypeMapper);
+  private final IcebergColumnTypeMapper mapper = IcebergColumnTypeMapper.INSTANCE;
```
sfc-gh-wtrefon, line 20: unused

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergTableSchemaResolver.java
```diff
@@ -1,16 +1,133 @@
...
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(IcebergTableSchemaResolver.class);
```
sfc-gh-wtrefon, line 22: logger is unfortunately unused ;/

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-bzabek: I wanted to pass a `String` query here so I don't expose List<IcebergoColumnTree> outside of the module.

"connectionService" is a better place to generate queries than a SchemaEvolutionService generates a schema etc...? Let the connection service just give me a connection and execute a statement that I want.

> it isn't in any way related to the iceberg
So maybe I'll just invoke 'executeStatement' method in Evolution service and generate log in evolution service.

src/test/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/ParseIcebergColumnTreeTest.java
```diff
@@ -0,0 +1,236 @@
...
+            "{ \"testColumnName\": {"
+                + "\"vehicle1\" : { \"car\" : { \"brand\" : \"vw\" } },"
+                + "\"vehicle2\" : { \"car\" : { \"brand\" : \"toyota\" } }"
+                + "}}",
+            "OBJECT(vehicle2 OBJECT(car OBJECT(brand VARCHAR)), "
+                + "vehicle1 OBJECT(car OBJECT(brand VARCHAR)))"),
+        // <- todo lol with k1, k2 the order is natural, however it changes an order when I used
```
sfc-gh-wtrefon, line 123: ?

src/test/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/ParseIcebergColumnTreeTest.java
```diff
@@ -0,0 +1,236 @@
...
+        arguments(
+            "{ \"testColumnName\": {"
+                + "\"k1\" : { \"car\" : { \"brand\" : \"vw\" } },"
+                + "\"k2\" : { \"car\" : { \"brand\" : \"toyota\" } }"
+                + "}}",
+            "OBJECT(k1 OBJECT(car OBJECT(brand VARCHAR)), k2"
+                + " OBJECT(car"
+                + " OBJECT(brand"
+                + " VARCHAR)))"),
+        arguments(
```
sfc-gh-wtrefon, line 134: we should add the descripton to the argument as it;s getting quite complex

src/test/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/ParseIcebergColumnTreeTest.java
```diff
@@ -0,0 +1,236 @@
...
+                + "    \"nested_object\": { "
+                + "      \"nested_key1\" : \"string\", "
+                + "      \"nested_key2\" : \"blah\", "
+                + "      \"nested_object2\" : { "
+                + "        \"nested_key2\" : 23.5 "
+                + "    }}"
+                + "}}",
+            "OBJECT(k1 INT, k2 INT, nested_object OBJECT(nested_key1"
+                + " VARCHAR(16777216), nested_key2 VARCHAR(16777216), nested_object2"
+                + " OBJECT(nested_key2 DOUBLE)))"),
```
sfc-gh-wtrefon, line 203: we should consider keeping the schema in separate txt files and the messages in json files as this is becoming really difficult to read

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergIngestionSchemaEvolutionIT.java
```diff
@@ -138,4 +110,185 @@ private static Stream<Arguments> prepareData() {
...
+        columns.get(1).getType(),
+        "OBJECT(k1 NUMBER(19,0), k2 NUMBER(19,0), k3 VARCHAR(16777216), k4 FLOAT)");
+
+    // k2, k3, k4
+    String testStruct4 = "{ \"testStruct\": { \"k2\" : 2, \"k3\" : 3, \"k4\" : 4.34 } }";
+    insertWithRetry(testStruct4, 3, false);
+    waitForOffset(4);
+
+    columns = describeTable(tableName);
+    assertEquals(
```
sfc-gh-wtrefon, line 144: let's stick to the older tests and use assertJ with assertThat, imho is much nicer than this junit sh*t :D

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergIngestionSchemaEvolutionIT.java
```diff
@@ -138,4 +110,185 @@ private static Stream<Arguments> prepareData() {
...
+        columns.get(1).getType(),
+        "OBJECT(k1 NUMBER(19,0), k2 NUMBER(19,0), k3 VARCHAR(16777216), k4 FLOAT, k5 NUMBER(19,0),"
+            + " k6 NUMBER(19,0))");
+    assertEquals(columns.size(), 2);
+  }
+
+  @Test
+  public void alterAlreadyExistingStructure_timestamp() throws Exception {
```
sfc-gh-wtrefon: i have no idea what this test does, could we find better name?

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergIngestionSchemaEvolutionIT.java
```diff
@@ -138,4 +110,185 @@ private static Stream<Arguments> prepareData() {
...
+            + "      }"
+            + "    ],"
+            + "    \"greeting\": \"Hello, Davis Heath! You have 4 unread messages.\","
+            + "    \"favoriteFruit\": \"strawberry\""
+            + "  }";
+    insertWithRetry(testStruct1, 0, false);
+    waitForOffset(1);
+
+    List<DescribeTableRow> columns = describeTable(tableName);
+    assertEquals(columns.size(), 23);
```
sfc-gh-wtrefon, line 218: what we are checking here? Some magic numbers to me. Do we check some field was added or all of the fields? We should run also describeTable before schema evolution to compare

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/TestJsons.java
```diff
@@ -0,0 +1,144 @@
+package com.snowflake.kafka.connector.streaming.iceberg;
+
+/**
+ * Class to provide Iceberg schema evolution tests with schemas and payload. It decreases the size
+ * of the test file .
+ */
+class TestJsons {
```
sfc-gh-wtrefon, line 7: maybe it would be worth moving the schemas to separate json files and in this class only read from the files?

sfc-gh-wtrefon (changes requested):
Really great change, good job. You did some really difficult things here. 
We need some changes here, mostly refactoring and splitting into smaller classes therefore for now I'm blocking it but if we decide to do the refactoring in a separate PR I will approve this one.

I didn't check if the merge/parsing logic is correct though, waiting for the refactoring so it's easier to read and reason

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/SchemaEvolutionService.java
sfc-gh-bzabek: done

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergSchemaEvolutionService.java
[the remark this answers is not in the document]
sfc-gh-bzabek: Done

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergColumnTypeMapper.java
sfc-gh-bzabek: - At first run it was resolved fromRecord as "LONG" - add column ... LONG
- After second run it was resolved from icebergSchema as NUMBER(19,0)
- then it failed when alter set data type NUMBER(19,0) was invoked. 

So I made it consise with int64 cause it returns long.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-bzabek: I generate query in connction service now.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergColumnTypeMapper.java
sfc-gh-bzabek: * It is a case for sturctured types

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergColumnTypeMapper.java
sfc-gh-bzabek: Done.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergFieldNode.java
sfc-gh-bzabek: done

src/test/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/ParseIcebergColumnTreeTest.java
sfc-gh-bzabek: I checked it and it shouldn't affect anything.

src/test/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/ParseIcebergColumnTreeTest.java
sfc-gh-bzabek: I don't know. Should I verbally describe Jsons that I parse here...

src/test/java/com/snowflake/kafka/connector/streaming/schemaevolution/iceberg/ParseIcebergColumnTreeTest.java
```diff
@@ -1,71 +0,0 @@
...
-import com.snowflake.kafka.connector.internal.streaming.schemaevolution.iceberg.IcebergColumnTree;
-import com.snowflake.kafka.connector.internal.streaming.schemaevolution.iceberg.IcebergDataTypeParser;
-import java.util.stream.Stream;
-import org.apache.iceberg.types.Type;
-import org.junit.jupiter.api.Assertions;
-import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.Arguments;
-import org.junit.jupiter.params.provider.MethodSource;
-
-public class ParseIcebergColumnTreeTest {
```
sfc-gh-bzabek, line 15: This class was moved to a correct module.

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergIngestionSchemaEvolutionIT.java
sfc-gh-bzabek: I removed it.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -498,6 +498,36 @@ public boolean hasSchemaEvolutionPermission(String tableName, String role) {
...
+    StringBuilder setDataTypeQuery = new StringBuilder("alter iceberg ");
+    setDataTypeQuery.append("table identifier(?) alter column ");
+    for (Map.Entry<String, ColumnInfos> column : columnsToModify.entrySet()) {
+      String columnName = column.getKey();
+      String dataType = column.getValue().getColumnType();
+
+      setDataTypeQuery.append(columnName).append(" set data type ").append(dataType).append(", ");
+    }
+    // remove last comma and whitespace
+    setDataTypeQuery.deleteCharAt(setDataTypeQuery.length() - 1);
```
sfc-gh-mbobowski, line 525: I think you can even try to avoid deleting chars by utilizing Collectors.joining:
```
    String columnsPart = columnsToModify.entrySet().stream().map(column -> {
      String columnName = column.getKey();
      String dataType = column.getValue().getColumnType();
      return columnName + " set data type " + dataType;
    }).collect(Collectors.joining(", "));
    
    setDataTypeQuery.append(columnsPart);
```

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
```diff
@@ -552,18 +582,22 @@ private void appendColumnsToTable(
       logColumn.append(columnName).append(" (").append(columnInfosMap.get(columnName)).append(")");
     }
 
+    executeStatement(tableName, appendColumnQuery.toString());
+
+    logColumn.insert(0, "Following columns created for table {}:\n").append("]");
```
sfc-gh-mbobowski, line 587: I know that you tried to fit into the code that already exists, but I doubt if it makes sense to construct log message in a such complicated way at all. We log ALTER TABLE statement in `executeQuery` and that's all we need.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/SchemaEvolutionService.java
```diff
@@ -12,5 +14,8 @@ public interface SchemaEvolutionService {
    *     nullability, and columns to add
    * @param record the sink record that contains the schema and actual data
    */
-  void evolveSchemaIfNeeded(SchemaEvolutionTargetItems targetItems, SinkRecord record);
+  void evolveSchemaIfNeeded(
+      SchemaEvolutionTargetItems targetItems,
+      SinkRecord record,
+      Map<String, ColumnProperties> existingSchema);
```
sfc-gh-mbobowski, line 20: nit: javadoc for new parameter

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergColumnTreeFactory.java
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  IcebergColumnTree fromJson(IcebergColumnJsonValuePair pair) {
+    LOGGER.debug(
+        "Attempting to parse schema from records payload for column: " + pair.getColumnName());
+    IcebergFieldNode rootNode = createNode(pair.getColumnName().toUpperCase(), pair.getJsonNode());
+    return new IcebergColumnTree(rootNode);
+  }
+
+  IcebergColumnTree fromConnectSchema(Field kafkaConnectField) {
```
sfc-gh-mbobowski, line 41: We have two separated flows: JSON without schema and AVRO/Protobuf.

I would extract `fromConnectSchema` to a separate class `SchematizedIcebergColumnTreeFactory` and rename this class to `NoSchemaIcebergColumnTreeFactory`. WDYT?

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergColumnTreeTypeBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+        });
+    removeLastSeparator(sb);
+  }
+
+  private void removeLastSeparator(StringBuilder sb) {
+    sb.deleteCharAt(sb.length() - 1);
+    sb.deleteCharAt(sb.length() - 1);
+  }
+
+  private static final String ROOT_NODE_TYPE = "ROOT_NODE";
```
sfc-gh-mbobowski, line 57: nit: let's place it at the beginning of class for better visibility

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergSchemaEvolutionService.java
```diff
@@ -17,48 +20,163 @@ public class IcebergSchemaEvolutionService implements SchemaEvolutionService {
...
+      LOGGER.debug("Adding columns to iceberg table: {} columns: {}", tableName, columnsToEvolve);
+      // some of the column might already exist, and we will modify them, not create
+      List<IcebergColumnTree> alreadyExistingColumns =
+          icebergTableSchemaResolver.resolveIcebergSchemaFromChannel(
+              existingSchema, columnsToEvolve);
+
+      List<IcebergColumnTree> modifiedOrAddedColumns =
+          icebergTableSchemaResolver.resolveIcebergSchemaFromRecord(record, columnsToEvolve);
+
+      List<IcebergColumnTree> columnsToAdd =
```
sfc-gh-mbobowski, line 59: There is no need to iterate twice over the list. The columns from `modifiedOrAddedColumns` are either modified or added, right?

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergSchemaEvolutionService.java
```diff
@@ -17,48 +20,163 @@ public class IcebergSchemaEvolutionService implements SchemaEvolutionService {
...
+
+  /** Columns that are not present in a current schema are to be added */
+  private List<IcebergColumnTree> distinguishColumnsToAdd(
+      List<IcebergColumnTree> alreadyExistingColumns,
+      List<IcebergColumnTree> modifiedOrAddedColumns) {
+    return modifiedOrAddedColumns.stream()
+        .filter(
+            modifiedOrAddedColumn ->
+                alreadyExistingColumns.stream()
+                    .noneMatch(
```
sfc-gh-mbobowski, line 99: It would be easier to convert `alreadyExistingColumns` from List to Set.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergSchemaEvolutionService.java
```diff
@@ -17,48 +20,163 @@ public class IcebergSchemaEvolutionService implements SchemaEvolutionService {
...
+
+  private void mergeChangesIntoExistingColumns(
+      List<IcebergColumnTree> alreadyExistingColumns, List<IcebergColumnTree> modifiedColumns) {
+    alreadyExistingColumns.forEach(
+        existingColumn -> {
+          IcebergColumnTree mewVersion =
+              modifiedColumns.stream()
+                  .filter(c -> c.getColumnName().equals(existingColumn.getColumnName()))
+                  .collect(Collectors.toList())
+                  .get(0);
```
sfc-gh-mbobowski, line 157: I am not so sure that we don't end up with `ArrayIndexOutOfBoundsException` here. Please add some meaningful logging for that case.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergTableSchemaResolver.java
```diff
@@ -1,16 +1,134 @@
...
+  // todo remove in 1820155 when getIcebergSchema() method is made public
+  private static String getIcebergSchema(ColumnProperties columnProperties) {
+    try {
+      java.lang.reflect.Field field =
+          columnProperties.getClass().getDeclaredField("icebergColumnSchema");
+      field.setAccessible(true);
+      return (String) field.get(columnProperties);
+    } catch (IllegalAccessException | NoSuchFieldException e) {
+      throw new IllegalStateException(
+          "Couldn't set iceberg by accessing private field: " + "isIceberg", e);
```
sfc-gh-mbobowski, line 80: No need to concatenate strings here.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergTableSchemaResolver.java
```diff
@@ -1,16 +1,134 @@
...
+        .filter(pair -> columnsToEvolve.contains(pair.getColumnName().toUpperCase()))
+        .map(treeFactory::fromJson)
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * Given a SinkRecord, get the schema information from it
+   *
+   * @param record the sink record that contains the schema and actual data
+   * @return list of column represantation in a form of tree
```
sfc-gh-mbobowski, line 105: typo

src/test/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/ParseIcebergColumnTreeTest.java
```diff
@@ -0,0 +1,243 @@
...
+                + "\"type\":\"struct\",\"fields\":["
+                + "{\"id\":1,\"name\":\"primitive\",\"required\":true,\"type\":\"boolean\"}"
+                + "]},"
+                + "\"element-required\":true}",
+            "{\"TESTSTRUCT\": [ { \"primitive\" : true, \"new_field\" : 25878749237493287429348 }]"
+                + " }",
+            "ARRAY(OBJECT(primitive BOOLEAN, new_field LONG))"));
+  }
+
+  protected SinkRecord createKafkaRecord(String jsonString, boolean withSchema) {
```
sfc-gh-mbobowski, line 223: `withSchema` is always false. Do we need this argument?

src/test/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/ParseIcebergColumnTreeTest.java
```diff
@@ -0,0 +1,243 @@
...
+    IcebergColumnTree alreadyExistingTree = treeFactory.fromIcebergSchema(apacheSchema);
+
+    // tree parsed from a record
+    SinkRecord record = createKafkaRecord(recordJson, false);
+    JsonNode recordNode = RecordService.convertToJson(record.valueSchema(), record.value(), true);
+    IcebergColumnJsonValuePair columnValuePair =
+        IcebergColumnJsonValuePair.from(recordNode.fields().next());
+
+    IcebergColumnTree modifiedTree = treeFactory.fromJson(columnValuePair);
+    // then
```
sfc-gh-mbobowski, line 170: it's rather `when`

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/TestJsons.java
```diff
@@ -0,0 +1,509 @@
...
+        + "  ],"
+        + "  \"optional\": true,"
+        + "  \"name\": \"sf.kc.test\""
+        + " },"
+        + " \"payload\": "
+        + nestedObjectsPayload
+        + "}";
+  }
+
+  static String nestedObjectsPayload =
```
sfc-gh-mbobowski, line 39: it can be private

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/TestJsons.java
```diff
@@ -0,0 +1,509 @@
...
+  static final String INT8_PAYLOAD = "\"test_int8\" : 2137 ";
+  static final String FLOAT_PAYLOAD = "\"test_float\" : 1939.30 ";
+  static final String DOUBLE_PAYLOAD = "\"test_double\" : 123.45793247859723 ";
+  static final String STRING_PAYLOAD = "\"test_string\" : \"very long string\" ";
+
+  private static final String SCHEMA_BEGINNING =
+      "{ \"schema\": { \"type\": \"struct\", \"fields\": [";
+  private static final String SCHEMA_END = "]},";
+
+  private static final String OBJECT_SCHEMA_BEGINNING =
```
sfc-gh-mbobowski, line 505: unused?

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergIngestionSchemaEvolutionIT.java
```diff
@@ -107,35 +193,294 @@ private void assertRecordsInTable() {
...
+            true),
+        Arguments.of(
+            objectVarcharPayload,
+            objectWithNestedObjectPayload(),
+            twoObjectsWithSchemaPayload(),
+            twoObjectsExtendedWithMapAndArrayPayload(),
             false));
   }
+
+  private static final String RECORD_METADATA_TYPE =
```
sfc-gh-mbobowski, line 480: Let's move it at the beginning of the class.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/SchemaEvolutionService.java
sfc-gh-bzabek: Done

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceV1.java
sfc-gh-bzabek: Yea it's more elegant. I apply it.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergColumnTreeFactory.java
sfc-gh-bzabek: Hmm, so what we have:
* from Iceberg (channel) schema - it's used before **both** workflows
* from json payload - workflow without schema
* from record schema - now there is only fromConnectSchema method. Do we have to also write seperate logic to parse avro and protobuf schema? I think not because Converter will parse it into connect schema when we ancounter AVRO or Protobuf. (For sure we ust test it)

When I wrote the factory, I thought it will be a bit over engineering to split it. However if we are going to need more methods then sure. Logically we have 3 parts.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergSchemaEvolutionService.java
sfc-gh-bzabek: Righ, however I found my approach simpler and didn't care very much about performance of that.

Create 2 lists and write an if to put an element into one or another list.

```
  private Pair<List<IcebergColumnTree>, List<IcebergColumnTree>> distinguish(
          List<IcebergColumnTree> alreadyExistingColumns,
          List<IcebergColumnTree> modifiedOrAddedColumns) {

    ArrayList<IcebergColumnTree> columnsToModify = new ArrayList<>();
    ArrayList<IcebergColumnTree> columnsToAdd = new ArrayList<>();
    
    for (IcebergColumnTree tree : modifiedOrAddedColumns) {
      if (alreadyExistingColumns.stream()
              .anyMatch(alreadyExisting -> alreadyExisting.getColumnName()
                      .equalsIgnoreCase(tree.getColumnName()))) {
        
        columnsToModify.add(tree);
      } else {
        columnsToAdd.add(tree);
      }
    }
    return Pair<List<IcebergColumnTree>, List<IcebergColumnTree>>(columnsToAdd, columnsToModify);
  }
```
But there is not "Pair" in JDK8. 
This approach is a bit more messy. Slightly more performant.


~(It's messy ->) The second alternative I see is to distinguish columnsToAdd, and then having columnsToAdd iterate over columns to evolve and again match them with modifiedOrAddedColumns list.~

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergSchemaEvolutionService.java
sfc-gh-bzabek: I can use Set everywhere instead of a list.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergSchemaEvolutionService.java
sfc-gh-bzabek: Yea I agree. It's too optimistic I'll change it.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergTableSchemaResolver.java
```diff
@@ -1,16 +1,134 @@
...
+        .collect(Collectors.toList());
+  }
+
+  public List<IcebergColumnTree> resolveIcebergSchemaFromRecord(
+      SinkRecord record, Set<String> columnsToEvolve) {
+    if (columnsToEvolve == null || columnsToEvolve.isEmpty()) {
+      return ImmutableList.of();
+    }
+    if (hasSchema(record)) {
+      LOGGER.debug(
```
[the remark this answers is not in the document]
sfc-gh-bzabek: Sure, done

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergTableSchemaResolver.java
```diff
@@ -1,16 +1,134 @@
...
+          "Schema found. Evolve columns basing on a record schema, column: " + columnsToEvolve);
+      return getTableSchemaFromRecordSchema(record, columnsToEvolve);
+    } else {
+      LOGGER.debug(
+          "Schema NOT found. Evolve columns basing on a records payload, columns: "
```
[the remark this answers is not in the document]
sfc-gh-bzabek: Yes. `records -> record's` I meant that this payload belongs to a record.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergTableSchemaResolver.java
sfc-gh-bzabek: Done

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergTableSchemaResolver.java
sfc-gh-bzabek: represEntation - done

src/test/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/ParseIcebergColumnTreeTest.java
sfc-gh-bzabek: No it's not. I changed it.

src/test/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/ParseIcebergColumnTreeTest.java
sfc-gh-bzabek: yep, plus "then" two lines below

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/TestJsons.java
sfc-gh-bzabek: Let's keep it consise with others. I don't want to restrict it's usage.

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/TestJsons.java
sfc-gh-bzabek: deleted.

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergIngestionSchemaEvolutionIT.java
sfc-gh-bzabek: Done

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergSchemaEvolutionService.java
sfc-gh-bzabek: None is perfect.

sfc-gh-mbobowski: For me it's a matter of avoiding code duplication rather than optimizing the code execution time. I don't see any problem with creating a wrapper class for two lists.
Anyway it's not a blocker to me.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergSchemaEvolutionService.java
sfc-gh-mbobowski: Good point. Is there any problem with implementing these methods or the final code doesn't look cleaner?

sfc-gh-mbobowski (approved):
I still see some space for minor improvements, but I am ok with merging it at this point.

src/main/java/com/snowflake/kafka/connector/internal/streaming/schemaevolution/iceberg/IcebergSchemaEvolutionService.java
sfc-gh-bzabek: If we implement only for columnName - there is not a problem. Using it for both doesn't make sense. 
I don't think using Set will change anything in a logic. It may (shouldn't) sneakily replace a column somewhere... I can change it, never say never.

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergIngestionIT.java
```diff
@@ -51,6 +49,8 @@ public void setUp() {
     config.put(ICEBERG_ENABLED, "TRUE");
     config.put(ENABLE_SCHEMATIZATION_CONFIG, isSchemaEvolutionEnabled().toString());
     config.put(SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER, "true");
+    // "snowflake.streaming.max.client.lag" = 1 second, for faster tests
+    config.put(SNOWPIPE_STREAMING_MAX_CLIENT_LAG, "1");
```
sfc-gh-wtrefon, line 53: nice

PR #1008, opened 2024-11-22
Update ConnectorConfigDefinition.java

sudeshwasnik:
# Overview
Configurations that may contain sensitive information (password, token) should be defined as PASSWORD type, there are two reasons : 
1. Raw value of these configurations must not be visible to users on UI/ API responses and should be masked. Values of PASSWORD type configurations are masked in responses. 
2. Connector configurations (and values) are logged at INFO level at various places (raw values for non-secret configurations) in Kafka Connect while it creates tasks/ connector.

This change is backward-compatible, since configuration values are never fetched from`SnowflakeSinkConnectorConfig` object in connector (yet). Raw values passed to [start()](https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java#L87) (`parsedConfig`) are directly used.

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/config/ConnectorConfigDefinition.java
+++ b/src/main/java/com/snowflake/kafka/connector/config/ConnectorConfigDefinition.java
@@ -123,7 +123,7 @@ public class ConnectorConfigDefinition {
             OAUTH_CLIENT_ID)
         .define(
             OAUTH_CLIENT_SECRET,
-            ConfigDef.Type.STRING,
+            ConfigDef.Type.PASSWORD,
             "",
             ConfigDef.Importance.HIGH,
             "Client secret of target OAuth integration",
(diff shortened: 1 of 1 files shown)


sudeshwasnik:
@sfc-gh-mbobowski Could you please take a look at this small change / PR? 
We want to make sure that raw-values of sensitive configurations do not get logged anywhere.

sfc-gh-achyzy:
@sudeshwasnik Have you signed CLA before? If not, you'd have to sign it first before contributing to the public repo.

sfc-gh-mbobowski (approved):

sfc-gh-mbobowski:
Already merged in https://github.com/snowflakedb/snowflake-kafka-connector/pull/1009

PR #1011, opened 2024-11-29
NO-SNOW Fix missing ubuntu packages error in CI

sfc-gh-mbobowski:
# Overview

NO-SNOW

CI recently started to fail because of some ubuntu packages disappearance.

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/.github/workflows/End2EndTestApache.yml
+++ b/.github/workflows/End2EndTestApache.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build_and_test:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-22.04
     strategy:
       fail-fast: false # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast
       matrix:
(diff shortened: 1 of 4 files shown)


sfc-gh-rsawicki (approved):

.github/workflows/End2EndTestApache.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build_and_test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
sfc-gh-achyzy, line 11: That is super risky... I don't think we should do that.

sfc-gh-achyzy: Let's use the current latest instead of the latest tag

sfc-gh-achyzy: Besides, I think I tried to bump it some time ago, and there were issues with some of the packages.

sfc-gh-mbobowski: Ok, let's try with 24.04 then.

sfc-gh-mbobowski: 24.04 works for non-Python workflows. For e2e tests there is some package version collision in 24.04, but not in 22.04.

If CI passes I will go on with combination of 22 and 24 to unblock development.

PR #1014, opened 2024-12-03
SNOW-1842220 Add comments to columns created by schema evolution

sfc-gh-bzabek:
# Overview

SNOW-1842220

Get comment from schema and append it to a column. 
It is supported only for records with schema, otherwise the default comment is used.
If comment does not exist use default comment. 

Comments are not modified when we modify column (alter set data type), Sql syntax does not support it. You can play around with it [HERE](https://preprod.app.snowflake.com/sfengineering/streaming_connectors/w1JaFHWtYX9x#query) in line 107

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergIngestionSchemaEvolutionIT.java
+++ b/src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergIngestionSchemaEvolutionIT.java
@@ -483,4 +483,15 @@ public class IcebergIngestionSchemaEvolutionIT extends IcebergIngestionIT {
             twoObjectsExtendedWithMapAndArrayPayload(),
             false));
   }
+
+  @Test
+  // @Disabled
+  void shouldAppendedCommentTest() throws Exception {
+    // when
+    insertWithRetry(schemaAndPayloadWithComment(), 0, true);
+    waitForOffset(1);
+    // then
+    List<DescribeTableRow> columns = describeTable(tableName);
+    assertEquals("Test comment", columns.get(1).getComment());
+  }
 }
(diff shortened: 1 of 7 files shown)


src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergIngestionSchemaEvolutionIT.java
```diff
@@ -483,4 +483,15 @@ private static Stream<Arguments> testEvolutionOfComplexTypes_dataSource() {
             twoObjectsExtendedWithMapAndArrayPayload(),
             false));
   }
+
+  @Test
+  @Disabled
+  void shouldAppendedCommentTest() throws Exception {
```
sfc-gh-mbobowski, line 489: Could you test adding the default comment in any of the tests?

sfc-gh-bzabek: I can.

sfc-gh-wtrefon, line 489:
```suggestion
  void shouldAppendCommentTest() throws Exception {
```

sfc-gh-bzabek: +1

sfc-gh-bzabek: done

sfc-gh-bzabek: done

sfc-gh-mbobowski (approved):

PR #1017, opened 2024-12-04
SNOW-1731255: Put null or empty json nodes to DLQ for schema evolution

sfc-gh-wtrefon:
# Overview

SNOW-1731255

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeErrors.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeErrors.java
@@ -340,7 +340,11 @@ public enum SnowflakeErrors {
       "Could not allocate thread for file cleaner to start processing in given time. If problem"
           + " persists, please try setting snowflake.snowpipe.use_new_cleaner to false"),
   ERROR_5025(
-      "5025", "Unexpected data type", "Unexpected data type encountered during schema evolution.");
+      "5025", "Unexpected data type", "Unexpected data type encountered during schema evolution."),
+  ERROR_5026(
+      "5026",
+      "Invalid SinkRecord received",
+      "Cannot infer type from null or empty object/list during schema evolution.");
 
   // properties
 
--- a/src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergIngestionNoSchemaEvolutionIT.java
+++ b/src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergIngestionNoSchemaEvolutionIT.java
@@ -1,9 +1,12 @@
 package com.snowflake.kafka.connector.streaming.iceberg;
 
+import static com.snowflake.kafka.connector.streaming.iceberg.sql.ComplexJsonRecord.complexJsonPayloadExample;
+import static com.snowflake.kafka.connector.streaming.iceberg.sql.ComplexJsonRecord.complexJsonPayloadWithWrongValueTypeExample;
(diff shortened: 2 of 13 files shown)


src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergIngestionNoSchemaEvolutionIT.java
```diff
@@ -91,10 +96,34 @@ void shouldInsertRecords(String description, String message, boolean withSchema)
...
+            wrongValueRecord2,
+            createKafkaRecord(complexJsonPayloadExample, 3, false),
+            createKafkaRecord(complexJsonPayloadExample, 4, false)));
+    waitForOffset(5);
+
+    assertRecordsInTable(1L, 3L, 4L);
+    List<InMemoryKafkaRecordErrorReporter.ReportedRecord> reportedRecords =
+        kafkaRecordErrorReporter.getReportedRecords();
+    assertThat(reportedRecords).hasSize(2);
+    assertThat(reportedRecords.stream().map(it -> it.getRecord()).collect(Collectors.toList()))
```
sfc-gh-mbobowski, line 122: If only it would be Kotlin...

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeErrors.java
```diff
@@ -340,7 +340,11 @@ public enum SnowflakeErrors {
       "Could not allocate thread for file cleaner to start processing in given time. If problem"
           + " persists, please try setting snowflake.snowpipe.use_new_cleaner to false"),
   ERROR_5025(
-      "5025", "Unexpected data type", "Unexpected data type encountered during schema evolution.");
+      "5025", "Unexpected data type", "Unexpected data type encountered during schema evolution."),
+  ERROR_5026(
```
sfc-gh-mbobowski, line 344: My understanding of SnowflakeErrors is that they represent errors exposed to the customer. In this particular situation we use exception to handle one of the expected scenarios (let's hope uncle Bob is not around) so I would choose an ordinary package-private exception. Wdyt?

src/test/java/com/snowflake/kafka/connector/streaming/iceberg/IcebergIngestionNoSchemaEvolutionIT.java
```diff
@@ -91,10 +96,34 @@ void shouldInsertRecords(String description, String message, boolean withSchema)
     service.insert(Collections.singletonList(createKafkaRecord(message, 2, withSchema)));
     waitForOffset(3);
 
-    assertRecordsInTable();
+    assertRecordsInTable(0L, 1L, 2L);
   }
 
-  private void assertRecordsInTable() {
+  @Test
+  @Disabled
```
sfc-gh-mbobowski, line 103: sdk 3.0.1 has just been merged to master. You can rebase and enable all tests.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeErrors.java
sfc-gh-wtrefon: well, the exception ends up in dlq so it's kinda public,  i would leave it as it is so the customer can find the root cause faster by the error code?

sfc-gh-mbobowski: Didn't know that it's visible. Let's leave it as it is.

sfc-gh-mbobowski (approved):

PR #1035, opened 2024-12-20
SNOW-1859651 Salt prefix for files mapped from different topics

sfc-gh-dseweryn:
# Overview

SNOW-1859651

Given two topics that are setup to be ingested to a single table using Snowpipe ingestion, each of the topics has an assigned cleaner for deleting already ingested files. This change prevents these cleaners from incorrectly cleaning files from the topic that should be handled by the other cleaner.
Any topic that is mapped by the `topic2table.map` should have filename prefix salted with the topic as it is otherwise complicated (and unnecessary) to determine which other topic can have a clash on cleaning filenames.

## Pre-review checklist
- [x] Snowpipe Changes
- [x] This change is protected by a config parameter `snowflake.snowpipe.stageFileNameExtensionEnabled`.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/Utils.java
+++ b/src/main/java/com/snowflake/kafka/connector/Utils.java
@@ -431,6 +431,29 @@ public class Utils {
   }
 
   /**
+   * Class for returned GeneratedName. isNameFromMap equal to True indicates that the name was
+   * resolved by using the map passed to appropriate function. {@link
+   * Utils#generateTableName(String, Map)}
+   */
+  public static class GeneratedName {
+    public final String name;
+    public final boolean isNameFromMap;
+
+    private GeneratedName(String name, boolean isNameFromMap) {
+      this.name = name;
+      this.isNameFromMap = isNameFromMap;
+    }
+
+    private static GeneratedName fromMap(String name) {
+      return new GeneratedName(name, true);
+    }
+
+    private static GeneratedName generated(String name) {
+      return new GeneratedName(name, false);
+    }
+  }
+
+  /**
    * modify invalid application name in config and return the generated application name
    *
    * @param config input config object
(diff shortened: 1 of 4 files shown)


src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -430,6 +430,29 @@ public static boolean isSnowpipeStreamingIngestion(Map<String, String> config) {
...
   }
 
+  /**
+   * Class for returned GeneratedName. isNameFromMap equal to True indicates that the name was
+   * resolved by using the map passed to appropriate function. {@link
+   * Utils#generateTableName(String, Map)}
+   */
+  public static class GeneratedName {
+    public final String name;
+    public final boolean isNameFromMap;
```
sfc-gh-wtrefon, line 440: nit: getters instead of public members

sfc-gh-wtrefon (approved):

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
```diff
@@ -486,44 +496,38 @@ private class ServiceContext {
...
-      if (mode == TopicToTableModeExtractor.Topic2TableMode.MANY_TOPICS_SINGLE_TABLE
-          && enableStageFilePrefixExtension) {
+      {
+        final String topicForPrefix =
+            generatedTableName.isNameFromMap() && enableStageFilePrefixExtension ? topicName : "";
         this.prefix =
-            FileNameUtils.filePrefix(conn.getConnectorName(), tableName, topicName, partition);
-      } else {
-        this.prefix = FileNameUtils.filePrefix(conn.getConnectorName(), tableName, "", partition);
+            FileNameUtils.filePrefix(conn.getConnectorName(), tableName, topicForPrefix, partition);
```
sfc-gh-gjachimko, line 530: so, if i read this correctly - we will not generate the salted prefix for all cases but only for a subset of them, right?
why not enable this by default for all?

sfc-gh-gjachimko (approved):
lgtm - left one question though.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkServiceV1.java
sfc-gh-dseweryn: Correct. There are three benefits I can think of when `topic2table.map` is not used for particular topic:
- no chances of regression
- no leftover files just because the connector has been updated
- easier to identify partition numbers

PR #1039, opened 2025-01-07
Add support for config provider for any property

VitoMakarevich:
# Overview

SNOW-XXXXX

Currently, config providers are allowed for connection-related properties only - for any other property, validation runs and fails, as it's run before the config provider resolves values.
This PR adds a check of validation error and if it's caused by configuration provider - removes the error.
Unfortunately - the way it's done is hacky(uses some bad encapsulation from KafkaConnect authors).
I can find a better way to do so(such as rewriting code to extend each validator from a config-aware one, so that the value is checked only if it's not supposed to be resolved by the config provider).
Please let me know if this improvement can be accepted on your side, and I'll adjust the code as you'd like.

The motivation is that likely user should be able to provide any config parameter using the config provider for whatever reason, and there should be no hidden assumptions about the secrecy of the parameters.
P.S. The true motivation in my case is that we are planning to consume from ~100 topics in one setup, and the connector's config is just too long for AWS MSK to handle(not sure whether this is MSK or KafkaConnect itself). The issue is exaggerated by an even bigger topic2table map(compared to a list of topics).
P.P.S. I was also considering why validation happens before config resolution, but cannot find meaningful results. Also I was considering to resolve these values by writing some code(it's possible 100%) - but didn't find existing KafkaConnect Connectors to do so(checked JDBC/S3 sinks).

## Pre-review checklist
- [ ] Snowpipe Changes

Likely no risks, as existing code will remain the same, while for some portion of users it may be useful.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnector.java
@@ -29,7 +29,6 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.regex.Pattern;
 import org.apache.kafka.common.config.Config;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.connect.connector.Task;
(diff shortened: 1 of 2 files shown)


VitoMakarevich:
Hi @sfc-gh-mbobowski @sfc-gh-achyzy could you please review this

PR #1040, opened 2025-01-09
SNOW-1875173: Support external OAuth

sfc-gh-akowalczyk:
# Overview

SNOW-1875173

The change brings support for the external OAuth feature in the connector.

The configuration is now extended with new properties:

- snowflake.oauth.client.id
- snowflake.oauth.client.secret
- snowflake.oauth.refresh.token
- snowflake.oauth.token.endpoint

## Pre-review checklist
- [ ] Snowpipe Changes


src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -793,8 +795,7 @@ public static String getSnowflakeOAuthToken(
                   // Trim surrounding quotation marks
                   return respBody.get(tokenType).toString().replaceAll("^\"|\"$", "");
                 } catch (Exception e) {
-                  throw SnowflakeErrors.ERROR_1004.getException(
-                      "Failed to get Oauth access token after retries");
+                  throw SnowflakeErrors.ERROR_1004.getException(e);
```
sfc-gh-dseweryn, line 798: Maybe we could add message to the exception?

src/main/java/com/snowflake/kafka/connector/internal/OAuthURL.java
```diff
@@ -0,0 +1,82 @@
...
+import static org.apache.commons.lang3.StringUtils.EMPTY;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.apache.commons.lang3.StringUtils;
+
+public class OAuthURL implements URL {
+  private static final KCLogger LOGGER = new KCLogger(OAuthURL.class.getName());
+
+  private static final String SNOWFLAKE_URL_REGEX_PATTERN =
```
sfc-gh-dseweryn, line 12: Why it is "SNOWFLAKE" url regex pattern? Does not seem to be related to Snowflake in any way?

src/main/java/com/snowflake/kafka/connector/internal/URL.java
```diff
@@ -0,0 +1,11 @@
+package com.snowflake.kafka.connector.internal;
+
+public interface URL {
+  String hostWithPort();
+
+  String getScheme();
+
+  String path();
+
+  boolean sslEnabled();
```
sfc-gh-dseweryn, line 10: `isSslEnabled`?

sfc-gh-dseweryn (approved):
Generally — nice

src/main/java/com/snowflake/kafka/connector/internal/URL.java
sfc-gh-akowalczyk: I only extracted an interface here. It indicates whether it's an `HTTP` vs `HTTPS` connection.

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-akowalczyk: The message is taken from the underlying exception

src/main/java/com/snowflake/kafka/connector/internal/OAuthURL.java
sfc-gh-akowalczyk: yup, fixed

src/main/java/com/snowflake/kafka/connector/internal/URL.java
sfc-gh-akowalczyk: It's just one of the conventions. Both the classic and the current one are commonly used.

PR #1041, opened 2025-01-13
Add configuration to disable reprocess files cleanup by v1Cleaner

nirutgupta:
# Overview

Issue: Missing records in Snowflake when streaming data using the Snowflake Sink Connector. Customers who have set up multiple topics to map to a single table might see data loss.

Cause: Cleaning up reprocessFiles may remove files that haven’t been fully loaded yet, leading to data loss.

Solution: To avoid this, enable the snowflake.snowpipe.v1Cleaner.disable.reprocessFiles.cleanup flag. This will stop the cleanup of reprocessFiles and help prevent any data loss during the process.

## Pre-review checklist
- [ ] Snowpipe Changes


sfc-gh-mbobowski:
Hi @nirutgupta, thanks for your contribution!

The bug causing cleaner to delete files from other topics was fixed ([link to PR)](https://github.com/snowflakedb/snowflake-kafka-connector/pull/1035) last week and will be released in 3.0.1. From now on the filename is prefixed with topic name when topicsToTableMap is used so we are sure that no more data loss will happen.

Having that in mind there is no need to put additional toggle flag for the cleaner itself.

sfc-gh-achyzy:
Is there a chance that #1035 already fixed it @nirutgupta ?

sfc-gh-mbobowski:
Merged with some changes in https://github.com/snowflakedb/snowflake-kafka-connector/pull/1045

PR #1047, opened 2025-01-27
NO-SNOW Set default max_client_lag to 30s

sfc-gh-mbobowski:
# Overview

SNOW-XXXXX

The current default value of 120 seconds was copied from the double buffer settings. However the documentation states 10 seconds for this value... anyway it was decided to set the default for 30 seconds to significantly reduce default latency.

## Pre-review checklist
- [ ] Snowpipe Changes


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -141,7 +141,7 @@ public class SnowflakeSinkConnectorConfig {
   public static final boolean SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER_DEFAULT = true;
   public static final String SNOWPIPE_STREAMING_MAX_CLIENT_LAG =
       "snowflake.streaming.max.client.lag";
-  public static final int SNOWPIPE_STREAMING_MAX_CLIENT_LAG_SECONDS_DEFAULT = 120;
+  public static final int SNOWPIPE_STREAMING_MAX_CLIENT_LAG_SECONDS_DEFAULT = 30;
```
sfc-gh-tjones, line 144: We should document this change in the release notes or README

PR #1049, opened 2025-01-31
SNOW-1899609 Bump bpckix-fips

sfc-gh-rsawicki:
# Overview

[SNOW-1899609]

## Pre-review checklist
- [ ] Snowpipe Changes

[SNOW-1899609]: https://snowflakecomputing.atlassian.net/browse/SNOW-1899609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -276,7 +276,7 @@
         <dependency>
             <groupId>org.bouncycastle</groupId>
             <artifactId>bcpkix-fips</artifactId>
-            <version>1.0.3</version>
+            <version>2.1.8</version>
             <scope>provided</scope>
         </dependency>
 
--- a/pom_confluent.xml
+++ b/pom_confluent.xml
@@ -427,7 +427,7 @@
             <!-- This dependency won't be present in the kafka connect runtime, hence we are packaging this in an uber jar -->
             <groupId>org.bouncycastle</groupId>
             <artifactId>bcpkix-fips</artifactId>
-            <version>1.0.3</version>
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -276,7 +276,7 @@
         <dependency>
             <groupId>org.bouncycastle</groupId>
             <artifactId>bcpkix-fips</artifactId>
-            <version>1.0.3</version>
+            <version>2.1.8</version>
```
sfc-gh-mbobowski, line 279: 1. I think we should also upgrade bc-fips
2. And downkloaded version in `run_test_apache.sh` so we can actually test it

sfc-gh-mbobowski: And we need to update the docs: https://docs.snowflake.com/en/user-guide/kafka-connector-install#download-the-kafka-connector-files

sfc-gh-mbobowski: Confluent still recommends using 1.0.2, so we have to ask them first https://docs.snowflake.com/en/user-guide/kafka-connector-install#download-the-kafka-connector-files

sfc-gh-mbobowski:
I have manually tested that 2.1.8 works for both apache and confluent. Tested also that it does not break if older jar is present on the classpath.

sfc-gh-mbobowski (approved):

PR #1055, opened 2025-02-11
SNOW-1904571 Add option to configure v2 cleaner interval

sfc-gh-mbobowski:
# Overview

SNOW-1904571

For Snowpipe based connector when having a topic with multiple partitions with no data produced the cleaner is never started because pipe doe not exists. 
```
                // cleaner starts along with the partition task, but until table, stage and pipe
                // aren't created - there is no point in querying the stage.
                if (isFirstRun.get()
                    && checkPreRequisites() != CleanerPrerequisites.PIPE_COMPATIBLE) {
                  LOGGER.debug(
                      "neither table {} nor stage {} nor pipe {} have been initialized yet,"
                          + " skipping cycle...",
                      tableName,
                      stageName,
                      pipeName);
                  return;
                }
```

This generates additional `describe table / stage / pipe` calls that some users find costly and noisy. 

This PR adds `snowflake.snowpipe.v2CleanerIntervalSeconds` configuration options to run cleaner with a lower frequency.

### DIFF
--- a/test/test_verify.py
+++ b/test/test_verify.py
@@ -84,7 +84,8 @@ class KafkaTest:
             }
         else:
             self.client_config = {
-                "bootstrap.servers": kafkaAddress
+                "bootstrap.servers": kafkaAddress,
+                "broker.address.family": "v4",
             }
 
         self.adminClient = AdminClient(self.client_config)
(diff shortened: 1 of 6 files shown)


test/test_verify.py
```diff
@@ -84,7 +84,8 @@ def __init__(self, kafkaAddress, schemaRegistryAddress, kafkaConnectAddress, cre
             }
         else:
             self.client_config = {
-                "bootstrap.servers": kafkaAddress
+                "bootstrap.servers": kafkaAddress,
+                "broker.address.family": "v4",
```
sfc-gh-akowalczyk, line 88: is it related to the change?

PR #1058, opened 2025-02-19
NO-SNOW Use Iceberg String instead of Snowflake VARCHAR in schema evolution

sfc-gh-mbobowski:
Fixes the following error that started to appear in tests

```
For Iceberg tables, only max length (16777216) is supported for 'VARCHAR(L)/STRING(L)' or use STRING directly.
```


sfc-gh-wtrefon (approved):

sfc-gh-wtrefon:
@sfc-gh-mbobowski u need to fix IcebergIngestionNoSchemaEvolutionIT as well

sfc-gh-mbobowski:
The error disappeared...

sfc-gh-mbobowski:
...and it's back again

PR #1087, opened 2025-04-15
CONN-10468 More cleanup after double buffer removal

sfc-gh-mbobowski:
# Overview

CONN-10468

A follow up to the https://github.com/snowflakedb/snowflake-kafka-connector/pull/1086

There is a couple of small cleanups in this PR:
- `BufferThreshold` - delete most of buffer logic, left validations as they are needed for Snowpipe
- `SnowflakeSinkService` - remove 3 buffer related methods from interface
- `TopicPartitionChannel` - remove `insertBufferedRecordsIfFlushTimeThresholdReached` method from interface
- `SnowflakeSinkServiceV2` - remove `flushTimeSeconds / fileSizeBytes / recordNum` fields and getters. Remove logic from setters (need another round of refactor to get rid of them completely)
- `StreamingBufferThreshold` - remove this class and it's usages completely


src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/TopicPartitionChannel.java
```diff
@@ -81,19 +80,5 @@ static String generateChannelNameFormatV2(String channelNameFormatV1, String con
...
-   *
-   * <p>Note: We acquire buffer lock since we copy the buffer.
-   *
-   * <p>Threshold is config parameter: {@link SnowflakeSinkConnectorConfig#BUFFER_FLUSH_TIME_SEC}
-   *
-   * <p>Previous flush time here means last time we called insertRows API with rows present in
-   */
-  void insertBufferedRecordsIfFlushTimeThresholdReached();
-
-  // todo it should belong to a buffered channel
```
sfc-gh-mbobowski, line 97: This comment was misleading. We decided to leave this behavior in single-buffer to be consistent with double-buffer.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -563,13 +534,7 @@ public boolean isClosed() {
...
   @Override
   public void setRecordNumber(long num) {
-    if (num < 0) {
-      LOGGER.error("number of record in each file is {}, it is negative, reset to 0", num);
-      this.recordNum = STREAMING_BUFFER_COUNT_RECORDS_DEFAULT;
-    } else {
-      this.recordNum = num;
-      LOGGER.info("Set number of records for buffer threshold to {}", num);
-    }
+    // TODO - remove from this class
```
sfc-gh-rzukowski, line 537: is this in use? 
maybe throw some kind of NotImplementedExeption will be more secure?

sfc-gh-mbobowski: No, it is not in use anymore. I am going to remove it in the next round of refactor.

sfc-gh-rzukowski: Ok, maybe it is not needed, but if the code can be released I would consider https://docs.oracle.com/javase/8/docs/api/java/lang/UnsupportedOperationException.html

sfc-gh-rzukowski (approved):

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-mbobowski: I will have to revert throwing UnsupportedOperationException and leave methods empty.
E2E tests showed that this code can be accidentally run by `SnowflakeSinkServiceFactory` when customer has old properties set in config. We don't want to break the whole task in that case.

PR #1092, opened 2025-04-24
SNOW-2034182 [Data loss fix] Change OnErrorOption from CONTINUE to SKIP_BATCH

sfc-gh-mbobowski:
# Overview

SNOW-2034182

When schematization is enabled it is possible for the connector to lose a single message that triggers changing a table schema due to a race condition. Assuming that `offset=n` triggers the schema evolution:
1. `this.channel.insertRow(transformedRecord, Long.toString(offset))` is called for offset=n-1 (no schema evolution)
2. `this.channel.insertRow(transformedRecord, Long.toString(offset))` is called for offset=n
3. [Race condition] The background ingest sdk flush thread sends and commits the data with offsets up to n (inclusive) into Snowflake. Record n is considered as error and is skipped due to `OnErrorOption.CONTINUE` client setting.
4. The channel is invalidated and reopened. Connector reports offset n+1 as the next starting point for Kafka Connect

Note that data won't be loss if step 4 happens before step 3.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
@@ -255,7 +255,7 @@ public class DirectTopicPartitionChannel implements TopicPartitionChannel {
     }
 
     // Open channel and reset the offset in kafka
-    this.channel = Preconditions.checkNotNull(openChannelForTable());
+    this.channel = Preconditions.checkNotNull(openChannelForTable(this.enableSchemaEvolution));
     final long lastCommittedOffsetToken = fetchOffsetTokenWithRetry();
     this.offsetPersistedInSnowflake.set(lastCommittedOffsetToken);
     this.processedOffset.set(lastCommittedOffsetToken);
@@ -755,7 +755,7 @@ public class DirectTopicPartitionChannel implements TopicPartitionChannel {
       final StreamingApiFallbackInvoker streamingApiFallbackInvoker) {
     LOGGER.warn(
         "{} Re-opening channel:{}", streamingApiFallbackInvoker, this.getChannelNameFormatV1());
-    return Preconditions.checkNotNull(openChannelForTable());
+    return Preconditions.checkNotNull(openChannelForTable(this.enableSchemaEvolution));
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
```diff
@@ -814,7 +814,7 @@ private SnowflakeStreamingIngestChannel openChannelForTable() {
             .setDBName(this.sfConnectorConfig.get(Utils.SF_DATABASE))
             .setSchemaName(this.sfConnectorConfig.get(Utils.SF_SCHEMA))
             .setTableName(this.tableName)
-            .setOnErrorOption(OpenChannelRequest.OnErrorOption.CONTINUE)
+            .setOnErrorOption(OpenChannelRequest.OnErrorOption.SKIP_BATCH)
```
sfc-gh-dseweryn, line 817: shouldn't this be only for the schematized flow?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -68,7 +68,7 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
   private final IcebergTableSchemaValidator icebergTableSchemaValidator;
   private final IcebergInitService icebergInitService;
 
-  private final SchemaEvolutionService schemaEvolutionService;
+  private SchemaEvolutionService schemaEvolutionService;
```
sfc-gh-dseweryn, line 71: Why not final anymore?

sfc-gh-mbobowski: Because of the two constructors that exists for the `SnowflakeSinkServiceV2`. I started to clean up this mess recently and once I have single constructor I will change it back to final.

src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
sfc-gh-mbobowski: This is a good question actually... error flag affects the non-schematized flow only for Iceberg imo, as otherwise KC packs everything into a VARIANT column so the data always match the schema. For Iceberg `SKIP_BATCH` can lower performance if e.g. only 1 message in the batch doesn't match the schema.

Having that in mind I will leave `CONTINUE` for non-schematized, thanks.

sfc-gh-dseweryn (approved):
LGTM!

PR #1097, opened 2025-04-29
CONN-10462 Remove single buffer parameter

sfc-gh-mbobowski:
# Overview

CONN-10462

## Pre-review checklist
- [ ] Snowpipe Changes


src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryService.java
```diff
@@ -285,13 +285,6 @@ protected void addUserConnectorPropertiesToDataNode(
           SNOWPIPE_STREAMING_CLIENT_PROVIDER_OVERRIDE_MAP,
           userProvidedConfig.get(SNOWPIPE_STREAMING_CLIENT_PROVIDER_OVERRIDE_MAP));
     }
-    if (userProvidedConfig.containsKey(SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER)) {
-      dataObjectNode.put(
-          SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER,
-          userProvidedConfig.getOrDefault(
-              SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER,
-              String.valueOf(SNOWPIPE_STREAMING_ENABLE_SINGLE_BUFFER_DEFAULT)));
-    }
```
sfc-gh-dseweryn, line 294: Should we keep it with value "false" for ease-of-use? On one hand we can infer this from the version, on the other we will have to infer in some cases

sfc-gh-dseweryn:
Will the removed parameter be ignored if customers will not remove it?

sfc-gh-mbobowski:
> Will the removed parameter be ignored if customers will not remove it?

Yes

sfc-gh-dseweryn (approved):

PR #1101, opened 2025-05-09
CONN-10476 Extract common part of SSv1 and SSv2

sfc-gh-mbobowski:
CONN-10476

During the SSv2 PoC implementation I discovered that a significant part of `DirectTopicPartitionChannel` code will be used in SSv2 as well. This PR makes the code available to be reused with the composition.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingRecordService.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingRecordService.java
@@ -1,0 +1,155 @@
+package com.snowflake.kafka.connector.internal.streaming;
+
+import static org.apache.kafka.common.record.TimestampType.NO_TIMESTAMP_TYPE;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.google.common.collect.ImmutableMap;
+import com.snowflake.kafka.connector.dlq.KafkaRecordErrorReporter;
+import com.snowflake.kafka.connector.internal.KCLogger;
+import com.snowflake.kafka.connector.internal.SnowflakeErrors;
+import com.snowflake.kafka.connector.internal.SnowflakeKafkaConnectorException;
+import com.snowflake.kafka.connector.records.RecordService;
+import com.snowflake.kafka.connector.records.SnowflakeJsonSchema;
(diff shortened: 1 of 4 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingRecordService.java
```diff
@@ -0,0 +1,155 @@
...
+import com.snowflake.kafka.connector.records.SnowflakeJsonSchema;
+import com.snowflake.kafka.connector.records.SnowflakeRecordContent;
+import java.io.ByteArrayOutputStream;
+import java.io.ObjectOutputStream;
+import java.util.Map;
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.errors.DataException;
+import org.apache.kafka.connect.sink.SinkRecord;
+
+public class StreamingRecordService {
```
sfc-gh-dseweryn, line 21: We could use a Javadoc here

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingRecordService.java
```diff
@@ -0,0 +1,155 @@
...
+  private final RecordService recordService;
+  private final KafkaRecordErrorReporter kafkaRecordErrorReporter;
+
+  public StreamingRecordService(
+      RecordService recordService, KafkaRecordErrorReporter kafkaRecordErrorReporter) {
+    this.recordService = recordService;
+    this.kafkaRecordErrorReporter = kafkaRecordErrorReporter;
+  }
+
+  public Map<String, Object> transformDataBeforeSending(SinkRecord kafkaSinkRecord) {
```
sfc-gh-dseweryn, line 33: Also Javadoc for a public function is encouraged

sfc-gh-dseweryn (approved):
Generally looks OK. Please add a little bit of Javadoc

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingRecordService.java
sfc-gh-mbobowski: Good point. Actually method name is also not very descriptive.

PR #1102, opened 2025-05-12
CONN-10486 Failsafe logic separation

sfc-gh-mbobowski:
Failsafe logic is shared across Snowpipe Streaming v1 and v2 so it's worth extracting to a separate class.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java
@@ -11,38 +11,11 @@ import java.util.Map;
 import java.util.Properties;
 import net.snowflake.ingest.streaming.OffsetTokenVerificationFunction;
 import net.snowflake.ingest.utils.Constants;
-import org.apache.kafka.common.record.DefaultRecord;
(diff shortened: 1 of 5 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingUtils.java
```diff
@@ -11,39 +11,12 @@
 import java.util.Properties;
 import net.snowflake.ingest.streaming.OffsetTokenVerificationFunction;
 import net.snowflake.ingest.utils.Constants;
-import org.apache.kafka.common.record.DefaultRecord;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /* Utility class/Helper methods for streaming related ingestion. */
 public class StreamingUtils {
-  private static final Logger LOGGER = LoggerFactory.getLogger(StreamingUtils.class);
```
sfc-gh-mbobowski, line 20: Deleted fields were unused

sfc-gh-dseweryn (approved):
Looks legit. I see some room for improvement in DirectTopicPartitionChannel but it is not related to this PR.

PR #1103, opened 2025-05-13
CONN-10486 SnowflakeSinkServiceV2IT refactor

sfc-gh-mbobowski:
In order to ensure that SSv2 works I would like to test the same ITs as we have for SSv1. This PR fixes a couple of problems with the existing ITs:
- All tests were run twice for password and oAuth authentication. This is fundamentally wrong because authentication method doesn't change anything in the service logic. I extracted a single oAuth test and left rest of them as password based.
- Configuration validation tests were inappropiate for IT level testing so I decided to delete it. It's just to heavy.
- Some of the tests were strongly connected with double buffering logic that was removed recently
- Schematization tests were moved to a a separate file

Of course it is faaar from being perfect but it is good enough to continue development of SSv2.

### DIFF
--- a/src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
+++ b/src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
@@ -1,6 +1,5 @@
 package com.snowflake.kafka.connector.internal.streaming;
 
-import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.ENABLE_SCHEMATIZATION_CONFIG;
 import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWPIPE_STREAMING_MAX_CLIENT_LAG;
 import static com.snowflake.kafka.connector.internal.streaming.SnowflakeSinkServiceV2.partitionChannelKey;
 import static com.snowflake.kafka.connector.internal.streaming.channel.TopicPartitionChannel.NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE;
@@ -8,7 +7,6 @@ import static com.snowflake.kafka.connector.internal.streaming.channel.TopicPart
 import com.codahale.metrics.Gauge;
 import com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig;
 import com.snowflake.kafka.connector.Utils;
-import com.snowflake.kafka.connector.builder.SinkRecordBuilder;
 import com.snowflake.kafka.connector.dlq.InMemoryKafkaRecordErrorReporter;
 import com.snowflake.kafka.connector.internal.SnowflakeConnectionService;
 import com.snowflake.kafka.connector.internal.SnowflakeConnectionServiceFactory;
@@ -39,7 +37,6 @@ import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.json.JsonConverter;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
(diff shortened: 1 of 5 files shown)


src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -1061,393 +901,31 @@ public void testStreamingIngestionWithExactlyOnceSemanticsOverlappingOffsets() t
...
     TestUtils.assertWithRetry(() -> TestUtils.tableSize(table) == totalRecordsExpected, 30, 20);
 
-    assert service2.getOffset(new TopicPartition(topic, partition)) == totalRecordsExpected;
+    assert service2.getOffset(topicPartition) == totalRecordsExpected;
 
     service2.closeAll();
   }
 
-  @Test
-  public void testSchematizationWithTableCreationAndJsonInput() throws Exception {
```
sfc-gh-mbobowski, line 1070: Moved to `SnowflakeSinkServiceV2SchematizationIT`

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -1061,393 +901,31 @@ public void testStreamingIngestionWithExactlyOnceSemanticsOverlappingOffsets() t
...
-        () -> service.getOffset(new TopicPartition(topic, partition)) == startOffset + 1, 20, 5);
-
-    TestUtils.checkTableContentOneRow(
-        table, SchematizationTestUtils.CONTENT_FOR_JSON_TABLE_CREATION);
-
-    service.closeAll();
-  }
-
-  @Test
-  public void testSchematizationSchemaEvolutionWithNonNullableColumn() throws Exception {
```
sfc-gh-mbobowski, line 1157: Moved to `SnowflakeSinkServiceV2SchematizationIT`

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -1061,393 +901,31 @@ public void testStreamingIngestionWithExactlyOnceSemanticsOverlappingOffsets() t
...
-      TestUtils.assertWithRetry(
-          () -> service.getOffset(new TopicPartition(topic, partition)) == noOfRecords, 5, 4);
-    } catch (Exception e) {
-      // do nothing
-    }
-    service.closeAll();
-  }
-
-  @Test
-  public void snowflakeSinkTask_put_whenJsonRecordCannotBeSchematized_sendRecordToDLQ() {
```
sfc-gh-mbobowski, line 1373: Moved to `SnowflakeSinkServiceV2SchematizationIT`

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -1538,92 +1016,4 @@ public void testStreamingIngest_multipleChannel_distinctClients() throws Excepti
         .getRegisteredClients()
         .containsKey(new StreamingClientProperties(fishConfig));
   }
-
-  @Test
-  void testSkippingOffsetsInSchemaEvolution() throws Exception {
```
sfc-gh-mbobowski, line 1543: Moved to `SnowflakeSinkServiceV2SchematizationIT`

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -1061,393 +901,31 @@ public void testStreamingIngestionWithExactlyOnceSemanticsOverlappingOffsets() t
...
-              .build();
-      service.startPartition(table, new TopicPartition(topic, partition));
-    } catch (IllegalArgumentException ex) {
-      Assertions.assertEquals(NumberFormatException.class, ex.getCause().getClass());
-    }
-  }
-
-  @ParameterizedTest(name = "useOAuth: {0}")
-  @ValueSource(booleans = {true, false})
-  public void testStreamingIngestionValidClientPropertiesOverride(boolean useOAuth)
```
sfc-gh-mbobowski, line 1304: Moved to SnowflakeSinkServiceV2LegacyIT

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -1061,393 +901,31 @@ public void testStreamingIngestionWithExactlyOnceSemanticsOverlappingOffsets() t
...
-    service.closeAll();
-  }
-
-  /**
-   * Even if override key is invalid, we will still create a client since we dont verify key and
-   * values, only format.
-   */
-  @ParameterizedTest(name = "useOAuth: {0}")
-  @ValueSource(booleans = {true, false})
-  public void testStreamingIngestion_invalidClientPropertiesOverride(boolean useOAuth)
```
sfc-gh-mbobowski, line 1342: Removing. This test is not even doing what it claims it does
```
// Wait 20 seconds here and no flush should happen since the max client lag is 30 seconds
```
max client lag is 1s not 30s. It passes because of the flawed assertion.

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -1061,393 +901,31 @@ public void testStreamingIngestionWithExactlyOnceSemanticsOverlappingOffsets() t
...
-    // Wait for enough time, the rows should be flushed
-    TestUtils.assertWithRetry(
-        () -> service.getOffset(new TopicPartition(topic, partition)) == noOfRecords, 30, 30);
-
-    service.closeAll();
-  }
-
-  @Test
-  public void testStreamingIngestionInvalidClientLag() {
-    conn = getConn(false);
```
sfc-gh-mbobowski, line 1280: Removing test - the code does not throw an exception that test asserts. It passes only becuase of the flawed assertion logic.

sfc-gh-dseweryn (commented):
Thank you for addressing most of the comments I made offline. I would argue that we should keep the OAuth tests intact to have certainty that SSV1 will not stop working in the future.

sfc-gh-mbobowski:
> Thank you for addressing most of the comments I made offline. I would argue that we should keep the OAuth tests intact (not reduced to a smoke test) to have certainty that SSV1 will not stop working in the future.

OAuth is an authentication mechanism. I see a value in having a smoke test to validate if it works. I don't see any value in running twice test that e.g. validates what happens when channel was closed before ingestion as it is an orthogonal feature to the authentication itself.

I am not against running more tests on auth but let's be specific about which of them brings value.

sfc-gh-dseweryn (approved):

PR #1104, opened 2025-05-14
CONN-10478 Java 11 migration

sfc-gh-mbobowski:
Migration to JDK 11 including removal of JDK 8 testing.

The main reasons for the migration are:
- enable support for Snowpipe Streaming v2
- long-lasting deprecation and upcoming removal of JDK 8 from Kafka / Kafka Connect environment

### DIFF
--- a/src/test/java/com/snowflake/kafka/connector/ConnectorConfigValidatorTest.java
+++ b/src/test/java/com/snowflake/kafka/connector/ConnectorConfigValidatorTest.java
@@ -1,7 +1,6 @@
 package com.snowflake.kafka.connector;
 
 import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.*;
-import static com.snowflake.kafka.connector.Utils.HTTP_NON_PROXY_HOSTS;
 import static com.snowflake.kafka.connector.internal.TestUtils.getConfig;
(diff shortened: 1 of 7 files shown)


src/test/java/com/snowflake/kafka/connector/ConnectorConfigValidatorTest.java
```diff
@@ -143,33 +141,6 @@ public void testEmptyHost() {
         .hasMessageContaining(JVM_PROXY_PORT);
   }
 
-  @Test
```
sfc-gh-mbobowski, line 146: This test caused other ITs to fail after JDK migration. The reason for that is that call
```
Utils.enableJVMProxy(config);
```
Sets a bunch of environment properties that are not clear in a proper way:
```
System.clearProperty(HTTP_USE_PROXY);
System.clearProperty(HTTP_PROXY_HOST);
System.clearProperty(HTTP_PROXY_PORT);
```
However, I have a strong feeling that testing if method sets environment variable on a unit test level is just wrong. This functionality is tested on IT and e2e level.

sfc-gh-mbobowski:
> I would rather fix this test and leave it be. It takes little time, would give quicker feedback than IT/E2E and is unlikely to change in the future

Test restored. Although I still hate it.

PR #1108, opened 2025-05-23
CONN-10463 Record mapping logic for Iceberg ssv2

sfc-gh-mbobowski:
CONN-10463

 This part of SSv2 logic can be independently review and merged so the noise in the upcoming PRs can be limited.

As discovered during the PoC it's much better to ingest POJO instead of Map<String, Object> in the `RECORD_METADATA` field for Iceberg ssv2. With Map<> the ingestion silently skips record if any field is missing which is very often in the metadata. Fixing it on the pipe definition level is possible but much more complicated.


src/main/java/com/snowflake/kafka/connector/records/RecordMetadata.java
```diff
@@ -0,0 +1,120 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * POJO for writing metadata with SSv2. Passing this class to sdk instead of String containing JSON
+ * makes pipe definition simpler.
+ */
+class RecordMetadata {
```
sfc-gh-wtrefon, line 11: wanted to propose record class here as we upgraded the java version but we are not still there :D

src/main/java/com/snowflake/kafka/connector/records/RecordMetadata.java
```diff
@@ -0,0 +1,120 @@
...
+  private final Long offset;
+  private final String topic;
+  private final Integer partition;
+  private final String key;
```
sfc-gh-wtrefon, line 15: is the key always a String?

src/main/java/com/snowflake/kafka/connector/records/RecordMetadata.java
sfc-gh-wtrefon, line 11: let's override toString(), can be useful if we log it somewhere

src/main/java/com/snowflake/kafka/connector/records/IcebergTableStreamingRecordMapper.java
```diff
@@ -35,11 +35,35 @@ public Map<String, Object> processSnowflakeRecord(SnowflakeTableRow row, boolean
...
     if (includeMetadata) {
-      streamingIngestRow.put(TABLE_COLUMN_METADATA, getMapForMetadata(row.getMetadata()));
+      Map<String, Object> mapForMetadata = getMapForMetadata(row.getMetadata());
+      if (ssv2Enabled) {
+        // ssv2 requires explicit type casting in pipe definition. For Map<> it does not ingest data
+        // when any of the values is missing.
+        // Passing POJO solves this problem and makes pipe definition easier.
+        RecordMetadata metadata =
+            new RecordMetadata(
+                getNullSafeLong(mapForMetadata, OFFSET),
```
sfc-gh-wtrefon, line 45: double checked the Long type for offset :D "experience"

src/main/java/com/snowflake/kafka/connector/records/RecordMetadata.java
```diff
@@ -0,0 +1,120 @@
...
+  private final Integer schema_id;
+  private final Integer key_schema_id;
+  private final Long CreateTime;
+  private final Long LogAppendTime;
+  private final Long SnowflakeConnectorPushTime;
```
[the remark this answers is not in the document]
sfc-gh-mbobowski: Good point. Actually I found that you have already created exactly the same class for Iceberg tests so I will reuse it instead of creating a new one :) https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/src/test/java/com/snowflake/kafka/connector/streaming/iceberg/sql/MetadataRecord.java

src/main/java/com/snowflake/kafka/connector/records/RecordMetadata.java
```diff
@@ -0,0 +1,120 @@
...
+    return key;
+  }
+
+  public Integer getSchema_id() {
+    return schema_id;
+  }
+
+  public Integer getKey_schema_id() {
+    return key_schema_id;
+  }
```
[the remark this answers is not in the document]
sfc-gh-mbobowski: Solved by https://github.com/snowflakedb/snowflake-kafka-connector/pull/1108#discussion_r2106869954

PR #1114, opened 2025-06-04
CONN-10496 Record data mapping for SSv2

sfc-gh-mbobowski:
# Overview

CONN-10496 Record data mapping for SSv2

In SSv1 Objects and Arrays were converted into Strings containing JSON.

In SSv2 it would be problematic because it would require placing additional logic within the pipe e.g. `PARSE_JSON($1) AS RECORD_CONTENT`. To keep pipe simple and avoid its recreation when schema change the data is mapped into `Map<String, Object>`.

### DIFF
--- a/src/test/java/com/snowflake/kafka/connector/records/StreamingRecordMapperTest.java
+++ b/src/test/java/com/snowflake/kafka/connector/records/StreamingRecordMapperTest.java
@@ -4,7 +4,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.Map;
 
-public class StreamingRecordMapperTest {
+abstract class StreamingRecordMapperTest {
 
   protected static final ObjectMapper objectMapper = new ObjectMapper();
 
@@ -26,9 +26,6 @@ public class StreamingRecordMapperTest {
           + "\"header2\": \"testheaderstring\","
           + "\"header3\": 3.5}"
           + "}";
-
-  protected static final String fullMetadataJsonExampleWithoutWhitespace =
-      fullMetadataJsonExample.replaceAll("\\s+", "");
 
   protected static final String fullMetadataWithoutWhitespace =
       fullMetadataJsonExample.replaceAll("\\s+", "");
@@ -52,7 +49,7 @@ public class StreamingRecordMapperTest {
               "objectAsJsonStringHeader",
               "{\"key1\":\"value1\",\"key2\":\"value2\"}"));
 
-  protected static RecordService.SnowflakeTableRow buildRow(String content)
+  protected static RecordService.SnowflakeTableRow buildRowWithDefaultMetadata(String content)
       throws JsonProcessingException {
     return buildRow(content, fullMetadataJsonExample);
   }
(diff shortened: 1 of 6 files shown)


src/test/java/com/snowflake/kafka/connector/records/StreamingRecordMapperTest.java
```diff
@@ -0,0 +1,66 @@
+package com.snowflake.kafka.connector.records;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.Map;
+
+public class StreamingRecordMapperTest {
```
sfc-gh-dseweryn: If it is not meant to be used directly, mark as abstract. Also does it need to be public?
```suggestion
abstract class StreamingRecordMapperTest {
```

src/test/java/com/snowflake/kafka/connector/records/StreamingRecordMapperTest.java
```diff
@@ -0,0 +1,66 @@
...
+          + "},"
+          + "\"header2\": \"testheaderstring\","
+          + "\"header3\": 3.5}"
+          + "}";
+
+  protected static final String fullMetadataJsonExampleWithoutWhitespace =
+      fullMetadataJsonExample.replaceAll("\\s+", "");
+
+  protected static final String fullMetadataWithoutWhitespace =
+      fullMetadataJsonExample.replaceAll("\\s+", "");
```
sfc-gh-dseweryn: Both values are equal. Copy-paste error?

src/test/java/com/snowflake/kafka/connector/records/StreamingRecordMapperTest.java
```diff
@@ -0,0 +1,66 @@
...
+          5L,
+          Map.of(
+              "header3",
+              "3.5",
+              "header2",
+              "testheaderstring",
+              "objectAsJsonStringHeader",
+              "{\"key1\":\"value1\",\"key2\":\"value2\"}"));
+
+  protected static RecordService.SnowflakeTableRow buildRow(String content)
```
sfc-gh-dseweryn: How about? The function name gives the information on what metadata is available
```suggestion
  protected static RecordService.SnowflakeTableRow buildRowWithDefaultMetadata(String content)
```

src/test/java/com/snowflake/kafka/connector/records/StreamingRecordMapperTest.java
sfc-gh-mbobowski: Yes, it can be abstract. Thx.

src/test/java/com/snowflake/kafka/connector/records/StreamingRecordMapperTest.java
sfc-gh-mbobowski: Good catch, fixed.

src/test/java/com/snowflake/kafka/connector/records/StreamingRecordMapperTest.java
sfc-gh-mbobowski: It was moved from another file, but I agree. Renamed.

PR #1115, opened 2025-06-04
CONN-10496 Creating pipe for SSv2

sfc-gh-mbobowski:
# Overview

CONN-10496 Creating pipe for SSv2

Initially this interface was more complicated but after [recent changes](https://github.com/snowflakedb/snowflake-kafka-connector/pull/1114) that's all left. `create if not exists` is called at channel/partition startup and `create or replace` when schema evolution happens.


sfc-gh-dseweryn (approved):
Generally OK, though it is an interesting philosophical question whether the same exact string should be repeated in code and tests

PR #1117, opened 2025-06-05
Mbobowski ssv2 poc

sfc-gh-mbobowski:
# Overview

SNOW-XXXXX

## Pre-review checklist
- [ ] Snowpipe Changes


src/main/java/com/snowflake/kafka/connector/config/IcebergConfigValidator.java
```diff
@@ -14,6 +15,9 @@ public class IcebergConfigValidator implements StreamingConfigValidator {
   private static final String INCOMPATIBLE_INGESTION_METHOD =
       "Ingestion to Iceberg table is supported only for Snowpipe Streaming";
 
+  private static final String ICEBERG_NOT_SUPPORT_IN_SSV2 =
```
sfc-gh-wtrefon:
```suggestion
  private static final String ICEBERG_NOT_SUPPORTED_IN_SSV2 =
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/DefaultStreamingIngestClientV2Provider.java
```diff
@@ -0,0 +1,86 @@
...
+public class DefaultStreamingIngestClientV2Provider implements StreamingIngestClientV2Provider {
+
+  private static final String STREAMING_CLIENT_V2_PREFIX_NAME = "KC_CLIENT_V2_";
+  private static final String DEFAULT_CLIENT_NAME = "DEFAULT_CLIENT";
+  private static final AtomicInteger createdClientId = new AtomicInteger(0);
+
+  private final Map<String, SnowflakeStreamingIngestClient> pipeToClientMap = new HashMap<>();
+
+  @Override
+  public SnowflakeStreamingIngestClient getClient(
```
sfc-gh-wtrefon: just to be sure, it cannot be called from multiple threads?

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
```diff
@@ -0,0 +1,763 @@
...
+
+  private void validateRecordAndEvolveTableSchema(
+      SinkRecord kafkaSinkRecord, Map<String, Object> transformedRecord) {
+    Map<String, Object> fieldsToValidate = new HashMap<>(transformedRecord);
+    // skip RECORD_METADATA cause SSv1 validations don't accept POJOs
+    fieldsToValidate.remove("RECORD_METADATA");
+
+    // validation return only first encountered error type so it might be needed to call it more
+    // than once
+    for (int i = 0; i < 5; i++) {
```
sfc-gh-wtrefon: what is this magic number 5? Could we name it?

sfc-gh-wtrefon: and why exactly 5? I guess because we can have 5 different types of schema ev errors? But let's make it clear with proper variable name and comment

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
```diff
@@ -0,0 +1,763 @@
...
+  @Override
+  public void reopenChannelAfterSchemaEvolved() {
+    rowSchema = rowSchemaProvider.getRowSchema(tableName, connectorConfig);
+    channel = openChannelForTable(channelName);
+  }
+
+  @Override
+  public void waitForLastProcessedRecordCommitted() {
+    long intervalMs = 1000;
+    int maxRetries = 20;
```
sfc-gh-wtrefon: maybe we could have an exponential backoff instead of 20 retries?

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
```diff
@@ -0,0 +1,763 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Consumer;
+import net.snowflake.ingest.streaming.InsertValidationResponse;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.connect.errors.ConnectException;
+import org.apache.kafka.connect.sink.SinkRecord;
+import org.apache.kafka.connect.sink.SinkTask;
+import org.apache.kafka.connect.sink.SinkTaskContext;
+
+public class SnowpipeStreamingV2PartitionChannel implements TopicPartitionChannel {
```
sfc-gh-wtrefon, line 42: by how much is this inspired by V1? Would like to know on which parts to focus mostly

sfc-gh-mbobowski: I would say around 50% was inspired by V1. The part with transforming records, calling sdk and schema evolution is the most important.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/DefaultStreamingIngestClientV2Provider.java
sfc-gh-mbobowski: Yes, a task in Kafka world is a single thread so only one `put()` is processed at a time. Records in the connector are processed one-by-one. I don't see any race condition that could happen here.

src/main/java/com/snowflake/kafka/connector/config/IcebergConfigValidator.java
sfc-gh-mbobowski: Done

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-mbobowski: Actually there are 3 distinct types of errors so 5 was a very defensive approach :P 
Changed magic number into named variable and added comment.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-mbobowski: Good idea, added.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
```diff
@@ -349,4 +349,12 @@ ChannelMigrateOffsetTokenResponseDTO migrateStreamingChannelOffsetToken(
...
    */
   Optional<List<DescribeTableRow>> describeTable(String tableName);
+
+  /**
+   * execute sql query
+   *
+   * @param query sql query string
+   * @param parameter parameter to be inserted at index 1
+   */
+  void executeQueryWithParameter(String query, String parameter);
```
sfc-gh-dseweryn: nit: for a more versatile function this could accept varargs
```suggestion
  void executeQueryWithParameter(String query, String... parameters);
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
```diff
@@ -0,0 +1,764 @@
...
+  private void validateRecordAndEvolveTableSchema(
+      SinkRecord kafkaSinkRecord, Map<String, Object> transformedRecord) {
+    Map<String, Object> fieldsToValidate = new HashMap<>(transformedRecord);
+    // skip RECORD_METADATA cause SSv1 validations don't accept POJOs
+    fieldsToValidate.remove("RECORD_METADATA");
+
+    // validation return only first encountered error type so it might be needed to call it more
+    // than once. Possible validation errors are: extra column, null value for non-null column or
+    // missing value for non-null column
+    int typesOfValidationErrors = 3;
```
sfc-gh-dseweryn: Can we somehow bind this number with the actual verification code?
So if there will be more types of errors added, this number would automatically update

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -239,26 +252,68 @@ private TopicPartitionChannel createTopicPartitionChannel(
...
+    List<TopicPartitionChannel> channelsForTable =
+        partitionsToChannel.values().stream()
+            .filter(topicPartitionChannel -> topicPartitionChannel.tableName().equals(table))
+            .collect(Collectors.toList());
+    LOGGER.info("Closing {} channels for table {}", channelsForTable.size(), table);
+    channelsForTable.forEach(TopicPartitionChannel::closeChannel);
+    streamingIngestClientV2Provider.close(
+        PipeNameProvider.pipeName(connectorConfig.get(Utils.NAME), table));
+    channelsForTable.forEach(TopicPartitionChannel::reopenChannelAfterSchemaEvolved);
+  }
```
sfc-gh-dseweryn: It seems to me that this function can be called from multiple workers concurrently. Should it be somehow synchronized?

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SSv2PipeCreator.java
```diff
@@ -0,0 +1,6 @@
+package com.snowflake.kafka.connector.internal.streaming.v2;
+
+/** (Re)creates the pipe */
+interface SSv2PipeCreator {
+  void createPipe(boolean recreate);
+}
```
sfc-gh-dseweryn: This interface could be more explicit, there would be no need to check the implementation
```
enum Mode { CREATE_IF_NOT_EXISTS, CREATE_OR_REPLACE }
...
void createPipe(Mode)
```

Personal opinion: I dislike the idea of having an interface that is a dedicated side-effect. Ideally it should return a Pipe object that would be used somewhere else — maybe I will find a right place continuing the review.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SSv2PipeCreator.java
```diff
@@ -0,0 +1,6 @@
+package com.snowflake.kafka.connector.internal.streaming.v2;
+
+/** (Re)creates the pipe */
+interface SSv2PipeCreator {
```
sfc-gh-dseweryn: Also, why interface for a single implementation?

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
```diff
@@ -0,0 +1,764 @@
...
+
+  // This offset represents the data buffered in KC. More specifically it is the KC offset to ensure
+  // exactly once functionality. On the creation it is set to the latest committed token in
+  // Snowflake (see offsetPersistedInSnowflake) and updated on each new row from KC.
+  private final AtomicLong processedOffset =
+      new AtomicLong(NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE);
+
+  // This offset is would not be required for buffer-less channel, but we add it to keep buffered
+  // and non-buffered
+  // channel versions compatible.
```
sfc-gh-dseweryn: Perhaps this comment can be improved? Also, could you add more on what this parameter is used for? As I understand even for non-buffered channels we still have a state within the API/SSV2 client that is asynchronous, therefore it is a kind of buffer
```suggestion
  // This offset is not required for non-buffered channel, but we add it to keep buffered
  // and non-buffered channel versions compatible.
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
```diff
@@ -0,0 +1,764 @@
...
+          currentOffsetPersistedInSnowflake);
+    }
+  }
+
+  private void transformAndSend(SinkRecord kafkaSinkRecord) {
+    try {
+      Map<String, Object> transformedRecord = streamingRecordService.transformData(kafkaSinkRecord);
+      if (schemaEvolutionEnabled) {
+        validateRecordAndEvolveTableSchema(kafkaSinkRecord, transformedRecord);
+      }
```
sfc-gh-dseweryn: Again, a matter of taste, but I would rather see two implementations of SchemaEvolutionService (one normal and one noop) rather than a service + a boolean

src/main/java/com/snowflake/kafka/connector/records/MetadataRecord.java
```diff
@@ -62,22 +62,27 @@ public String getKey() {
     return key;
   }
 
+  @JsonProperty("schema_id")
```
sfc-gh-dseweryn, line 65: why these do not follow SnakeCase?

sfc-gh-dseweryn (commented):
Nothing blocking, some thoughts though

src/main/java/com/snowflake/kafka/connector/records/MetadataRecord.java
sfc-gh-mbobowski: Matching legacy format :(

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SSv2PipeCreator.java
sfc-gh-mbobowski: Deleted interface

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionService.java
sfc-gh-mbobowski: Changed to handle varargs

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SSv2PipeCreator.java
sfc-gh-mbobowski: Added enum.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-mbobowski: I changed this comment to `// Used for telemetry and logging only` because this is the only use case for this value. I was also thinking about deleting it completely but then I would have to send some magic number in the telemetry which is not perfect either.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-mbobowski: To be precise a "worker" in Kafka Connect terminology is a process that can handle multiple tasks https://docs.confluent.io/platform/current/connect/index.html#workers

My understanding of synchronization in Kafka Connect is as follows:
- each task use a single thread for put / precommit calls
- in `SnowflakeSinkTask` we simply call `getSink().insert(records);` - still single thread
- in `SnowflakeSinkServiceV2` we iterate over records and call sdk - still single thread
- `closeClientAndReopenChannelsForTable()` is also a thread-safe code as we close and open channel synchronously

This is super important so I would be very happy if you find any mistake in my logic. I even believe that there is too much synchronization in `DirectTopicPartitionChannel` but I don't want to deal with it right now.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-mbobowski: Thanks for pointing that out, I think that in general a new level of abstraction was needed here to separate schema evolution code.
I created an interface for SSv2 that may be also integrated with SSv1 (but it is beyond the scope of this PR).

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-mbobowski: I know this code is not great. Any improvements would require changing the validation logic that I would like to keep consistent with the old sdk.

The root cause of this problem is a poor design of validations that could return all errors at once. This is not a simple thing to change though because the logic relies heavily on exceptions handling.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-mbobowski: Needed to revert this one... the problem with extracting schema evolution logic is that maintaining state in `rowSchema` becomes complex as it changes with every validation iteration. The current version is ugly but it's simple and stupid.

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2SchematizationIT.java
```diff
@@ -114,53 +86,38 @@ public void testSchematizationWithTableCreationAndJsonInput() throws Exception {
...
-        "org.apache.kafka.connect.json.JsonConverter");
-    config.put(SnowflakeSinkConnectorConfig.VALUE_SCHEMA_REGISTRY_CONFIG_FIELD, "http://fake-url");
-    config.put("schemas.enable", "false");
-    // get rid of these at the end
-
-    Schema schema =
-        SchemaBuilder.struct()
-            .field("id_int8", Schema.INT8_SCHEMA)
-            .field("id_int8_non_nullable_null_value", Schema.OPTIONAL_INT8_SCHEMA)
+  public void testSchematizationWithTableCreationAndJsonInput_ssv2() throws Exception {
```
sfc-gh-wtrefon: we don't need those _ssv2 postfixes in names as we moved it to separated class, right?

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2SchematizationIT.java
```diff
@@ -185,9 +142,26 @@ public void testSchematizationSchemaEvolutionWithNonNullableColumn() throws Exce
...
     service.insert(Collections.singletonList(jsonRecordValue));
-    TestUtils.assertWithRetry(() -> service.getOffset(topicPartition) == startOffset + 1, 20, 5);
+    TestUtils.assertWithRetry(() -> service.getOffset(topicPartition) == 1, 20, 5);
+  }
 
-    service.closeAll();
+  @Test
+  public void testSchematizationSchemaEvolutionWithNonNullableColumn_ssv2() throws Exception {
+    // given
+    config.put(SNOWPIPE_STREAMING_V2_ENABLED, "true");
```
sfc-gh-wtrefon: this can be moved to @BeforeEach method

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2SchematizationIT.java
```diff
@@ -231,19 +200,53 @@ void testSkippingOffsetsInSchemaEvolution() throws Exception {
...
+        .until(() -> service.getOffset(topicPartition) == 4);
+  }
+
+  @ParameterizedTest
+  @ValueSource(booleans = {true, false})
+  public void snowflakeSinkTask_put_whenJsonRecordCannotBeSchematized_sendRecordToDLQ(
+      boolean ssv2Enabled) {
     // given
     config.put(ENABLE_SCHEMATIZATION_CONFIG, "true");
+    config.put(SNOWPIPE_STREAMING_V2_ENABLED, String.valueOf(ssv2Enabled));
```
sfc-gh-wtrefon, line 194: why?

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-wtrefon: still, it's risky, the validation is on our side now, can we do anything to change that and validate everyithing?

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
```diff
@@ -0,0 +1,762 @@
...
+          LOGGER.info(
+              "Record doesn't match table schema - triggering schema evolution. topic={},"
+                  + " partition={}, offset={}",
+              kafkaSinkRecord.topic(),
+              kafkaSinkRecord.kafkaOffset(),
+              kafkaSinkRecord.kafkaPartition());
+          waitForAllPartitionsToCommitData.run();
+          evolveSchemaIfNeeded(kafkaSinkRecord, error);
+          ssv2PipeCreator.createPipe(CreatePipeMode.CREATE_OR_REPLACE);
+          closeClientAndReopenChannelsForTable.accept(tableName);
```
sfc-gh-wtrefon: how often do you think we will have more that one type of error? We could call some of the operations only once, right? But don't want to complicate the code more if it's rare situation

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
```diff
@@ -0,0 +1,762 @@
...
+      return;
+    }
+
+    streamingIngestClientV2Provider
+        .getClient(connectorConfig, pipeName, streamingClientProperties)
+        .initiateFlush();
+
+    WaitForLastOffsetCommittedPolicy.getPolicy(
+        () -> {
+          long offsetCommitedToBackend = fetchLatestCommittedOffsetFromSnowflake();
```
sfc-gh-wtrefon:
```suggestion
          long offsetCommittedToBackend = fetchLatestCommittedOffsetFromSnowflake();
```
:)

sfc-gh-wtrefon (commented):
Generally ok, some minor things to improve

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2SchematizationIT.java
sfc-gh-mbobowski: Nope, both tests are in the same class.

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2SchematizationIT.java
sfc-gh-mbobowski: Not really, we have v1 tests in the same file.

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2SchematizationIT.java
sfc-gh-mbobowski: Why what? :D 
I parameterized tests to validate that v1 and v2 behavior is the same. Can you add more context?

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-mbobowski: It is very unlikely but we have to support it. The good part is that this code will be gone in the future because sdk folks work on the solution to eliminate the need  of pipe recreation when schema is changed (it's in the GA scope).

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-mbobowski: Done

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-dseweryn: As we discussed offline, `rowSchema` being updated as a side effect in the verification loop is very surprising. Please add a manager or at least a comment

sfc-gh-mbobowski: Thanks for this suggestion, after adding `RowSchemaManager` it looks better indeed.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-dseweryn: Yeah, my bad — this looks like could be called by many tasks / threads concurrently. I don't see any synchronization to this method in the call-stack, so apparently it may be that multiple threads are calling it

sfc-gh-mbobowski: Thanks for pointing that out. Due to recent changes that removed pipe recreation I added synchronization only to `DefaultStreamingIngestClientV2Provider` methods.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-mbobowski: Code was deleted due to schema evolution removal.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
```diff
@@ -0,0 +1,675 @@
...
+    this.sinkTaskContext = sinkTaskContext;
+    this.streamingIngestClientV2Provider = streamingIngestClientV2Provider;
+    this.rowSchemaManager = rowSchemaManager;
+    this.streamingErrorHandler = streamingErrorHandler;
+
+    this.telemetryServiceV2 = conn.getTelemetryClient();
+    this.pipeName = PipeNameProvider.pipeName(connectorConfig.get(Utils.NAME), tableName);
+    this.streamingClientProperties = new StreamingClientProperties(connectorConfig);
+
+    ssv2PipeCreator.createPipeIfNotExists();
```
sfc-gh-lshcharbaty: Would it be possible to create a pipe before the instantiation of this class? There is a lot of dependencies already, and it doesn't seem `ssv2PipeCreator` is used anywhere else.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
```diff
@@ -0,0 +1,675 @@
...
+      LOGGER.info(
+          "Ignore inserting offset:{} for channel:{} because we recently reset offset in"
+              + " Kafka. currentProcessedOffset:{}",
+          kafkaSinkRecord.kafkaOffset(),
+          this.getChannelNameFormatV1(),
+          currentProcessedOffset);
+      return;
+    }
+    // Accept the incoming record only if we don't have a valid offset token at server side, or the
+    // incoming record offset is 1 + the processed offset
```
sfc-gh-lshcharbaty: This isn't entirely true. It can be paraphrased as `the incoming record offset is larger the processed offset`. Or the comment can be removed entirely.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
```diff
@@ -0,0 +1,675 @@
...
+          currentProcessedOffset,
+          kafkaSinkRecord.kafkaOffset(),
+          currentOffsetPersistedInSnowflake);
+    }
+  }
+
+  private void transformAndSend(SinkRecord kafkaSinkRecord) {
+    try {
+      Map<String, Object> transformedRecord = streamingRecordService.transformData(kafkaSinkRecord);
+      if (schemaEvolutionEnabled) {
```
sfc-gh-lshcharbaty: Where does schema evolution take place in this code? In `streamingErrorHandler.handleError`?

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
```diff
@@ -0,0 +1,675 @@
...
+      LOGGER.warn(
+          "Failed to insert row for channel:{}. Will be retried by Kafka. Exception: {}",
+          this.getChannelNameFormatV1(),
+          ex);
+    }
+  }
+
+  private Optional<RowSchema.Error> validateRecord(Map<String, Object> transformedRecord) {
+    Map<String, Object> fieldsToValidate = new HashMap<>(transformedRecord);
+    // skip RECORD_METADATA cause SSv1 validations don't accept POJOs
```
sfc-gh-lshcharbaty, line 259: Why do we care about SSv1 here?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -243,22 +255,53 @@ private TopicPartitionChannel createTopicPartitionChannel(
...
+          this.enableCustomJMXMonitoring,
+          this.metricsJmxReporter,
+          this.schemaEvolutionService,
+          new InsertErrorMapper(),
+          streamingErrorHandler);
+    }
+  }
+
+  private void waitForAllChannelsToCommitData() {
+    partitionsToChannel.forEach((key, value) -> value.waitForLastProcessedRecordCommitted());
```
sfc-gh-lshcharbaty: With this approach it seems the channels flushes happen sequentially. Can we make `waitForLastProcessedRecordCommitted` return a `CompletableFuture` and trigger the flush for all channels at once?

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
```diff
@@ -0,0 +1,675 @@
...
+                event ->
+                    LOGGER.error(
+                        "{} Failed to open Channel or fetching offsetToken for channel:{}."
+                            + " Exception: {}",
+                        StreamingApiFallbackInvoker.APPEND_ROW_FALLBACK,
+                        this.getChannelNameFormatV1(),
+                        event.getException()))
+            .build();
+
+    return Failsafe.with(reopenChannelFallbackExecutorForInsertRows)
```
sfc-gh-lshcharbaty: We created `WaitForLastOffsetCommittedPolicy` to compose the `Fallbacks`. Shall we move this configuration to a separate policy as well?

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
```diff
@@ -0,0 +1,675 @@
...
+              "Failure closing Streaming Channel name:%s msg:%s",
+              this.getChannelNameFormatV1(), e.getMessage());
+      this.telemetryServiceV2.reportKafkaConnectFatalError(errMsg);
+      LOGGER.error(
+          "Closing Streaming Channel={} encountered an exception {}: {} {}",
+          this.getChannelNameFormatV1(),
+          e.getClass(),
+          e.getMessage(),
+          Arrays.toString(e.getStackTrace()));
+    }
```
sfc-gh-lshcharbaty: Shall we replace it with `closeChannelAsync().get()`?

src/main/java/com/snowflake/kafka/connector/internal/streaming/validation/JWTManagerProvider.java
```diff
@@ -5,23 +5,31 @@
...
+  public static JWTManager fromConfig(Map<String, String> config) {
     SnowflakeURL snowflakeURL = new SnowflakeURL(config.get(Utils.SF_URL));
-    PrivateKey privateKey = parsePrivateKey(config.get(Utils.SF_PRIVATE_KEY));
+    String rawPrivateKey = config.get(Utils.SF_PRIVATE_KEY);
+    Optional<String> privateKeyPassphrase =
+        Optional.ofNullable(config.get(Utils.PRIVATE_KEY_PASSPHRASE));
+    PrivateKey privateKey =
+        privateKeyPassphrase
+            .map(passphrase -> EncryptionUtils.parseEncryptedPrivateKey(rawPrivateKey, passphrase))
+            .orElse(parsePrivateKey(rawPrivateKey));
```
sfc-gh-lshcharbaty: `orElse` always invokes the function, even when optional is not empty. For lazy computation we can use `orElseGet`. I assume that's exactly what's desired here.

src/main/java/com/snowflake/kafka/connector/records/MetadataRecord.java
```diff
@@ -62,22 +62,27 @@ public String getKey() {
...
   public Long getCreateTime() {
     return createTime;
   }
 
+  @JsonProperty("LogAppendTime")
   public Long getLogAppendTime() {
     return logAppendTime;
   }
 
+  @JsonProperty("SnowflakeConnectorPushTime")
```
sfc-gh-lshcharbaty, line 85: Are these changes still relevant taking into account #1141?

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
```diff
@@ -38,27 +40,33 @@
...
   public void afterEach() {
     TestUtils.dropTable(table);
+    TestUtils.dropPipe(pipe);
   }
 
-  @Test
-  public void testChannelCloseIngestion() throws Exception {
+  @ParameterizedTest
+  @ValueSource(booleans = {true, false})
+  public void testChannelCloseIngestion(boolean ssv2Enabled) throws Exception {
```
sfc-gh-lshcharbaty, line 68: Nit: perhaps it would be easier to create 2 subclasses, one that does no additional configuration (e.g. uses ssv1) and another one that does `config.put(SNOWPIPE_STREAMING_V2_ENABLED, "true");` in before each.
Anyway, the current organization of the test is fine for me

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2SchematizationIT.java
```diff
@@ -116,6 +147,28 @@ public void testSchematizationSchemaEvolutionWithNonNullableColumn() throws Exce
...
+            .withSinkTaskContext(new InMemorySinkTaskContext(Collections.singleton(topicPartition)))
+            .withErrorReporter(errorReporter)
+            .build();
+    service.startPartition(table, topicPartition);
+
+    // when
+    service.insert(Collections.singletonList(jsonRecordValue));
+
+    // then
+    // schema evolution not available for ssv2
```
sfc-gh-lshcharbaty: Nit: does it make sense to create multiple tests ensuring that schema evolution does not work?

test/test_suites.py
```diff
@@ -274,7 +274,7 @@ def create_end_to_end_test_suites(driver, nameSalt, schemaRegistryAddress, testS
                     ),
                     run_in_confluent=True,
                     run_in_apache=True,
-                    cloud_platform=CloudPlatform.ALL,
```
sfc-gh-lshcharbaty, line 277: What's the intention behind running these tests on AWS?
If that's for ssv2 purposes only, we still should run them with ssv1 on all clouds, I reckon

sfc-gh-lshcharbaty (commented):
A bunch of comments. Looks good from functional perspective

test/test_suites.py
sfc-gh-mbobowski: SSv2 currently works only on AWS so we cannot run any shared v1/v2 tests on other providers :( Azure is close to being unlocked but it's not there yet.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-mbobowski: Yes, it is a leftover after the schema evolution implementation. Dependency removed.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-mbobowski: True, it was copied from the v1. Deleted.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-mbobowski: Changed the parameter name from `schemaEvolutionEnabled` to `schematizationEnabled` and added a comment 🙈 There is no schema evolution in SSv2. In SSv1 schema evolution and schematization were bound together and that caused the naming confusion.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-mbobowski: Because the validations were copied from SSv1 SDK. Actually it won't be true after https://github.com/snowflakedb/snowflake-kafka-connector/pull/1141 is merged. I will remove this comment and code in a separate PR.

src/main/java/com/snowflake/kafka/connector/internal/streaming/validation/JWTManagerProvider.java
sfc-gh-mbobowski: Wow, this API is ridiculous to me... thanks.

src/main/java/com/snowflake/kafka/connector/records/MetadataRecord.java
sfc-gh-mbobowski: I will remove it in a separate PR because I have to merge https://github.com/snowflakedb/snowflake-kafka-connector/pull/1141 first.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-mbobowski: Done, although I couldn't extract all of the Failsafe boilerplate.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingV2PartitionChannel.java
sfc-gh-mbobowski: Yes it makes sense. However I still have to catch exceptions from `get()` so it's not that much shorter.

test/test_suites.py
sfc-gh-mbobowski: Correct, it's because of SSv2. It's difficult to split the execution in order to run SSv1 on all platforms and SSv2 on AWS only. Didn't want to waste time on that.

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2SchematizationIT.java
sfc-gh-mbobowski: Initially I implemented all of them for schema evo but I admin that once it got out it doesn't bring much value. Deleted.

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2IT.java
sfc-gh-mbobowski: I will leave it as-is for now. When schema evolution in SSv2 arrives it will be a better moment to rethink tests organization.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-mbobowski: Good idea, done.

sfc-gh-lshcharbaty (approved):
lgtm

PR #1122, opened 2025-06-13
CONN-10458 Disable schema evolution for ssv2

sfc-gh-mbobowski:
Schema evolution in the current SSv2 architecture will lose data. The reason is that multiple nodes share the same pipe object and when it is recreated from one worker node, other nodes cannot synchronize.

This PR block the ability to evolve schema. If schema doesn't match, the connector will either send record to DLQ or stop (depending on the user configuration).


sfc-gh-mbobowski:
Closing - cherry-picked

PR #1126, opened 2025-06-26
CONN-10463 SSv2 config and mini-refactors

sfc-gh-mbobowski:
To make review of the main SSv2 PR easier I extracted the part related to the parameter management:
- Adding parameter `snowflake.streaming.v2.enabled`
- Iceberg not supported in SSv2 validation
- SSv2 in PuPr supports only the default role - `snowflake.role.name` cannot be provided in SSv2
- a couple of one-liners

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/Utils.java
+++ b/src/main/java/com/snowflake/kafka/connector/Utils.java
@@ -18,6 +18,7 @@ package com.snowflake.kafka.connector;
 
 import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.ICEBERG_ENABLED;
 import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.INGESTION_METHOD_OPT;
+import static com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig.SNOWPIPE_STREAMING_V2_ENABLED_DEFAULT_VALUE;
 
 import com.google.common.collect.ImmutableMap;
 import com.snowflake.kafka.connector.internal.InternalUtils;
@@ -43,7 +44,7 @@ import java.util.Base64;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Random;
+import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -115,8 +116,6 @@ public class Utils {
 
   // jdbc log dir
   public static final String JAVA_IO_TMPDIR = "java.io.tmpdir";
-
-  private static final Random random = new Random();
 
   // mvn repo
   private static final String MVN_REPO =
@@ -412,8 +411,10 @@ public class Utils {
   }
 
   public static boolean isSnowpipeStreamingV2Enabled(Map<String, String> config) {
-    return Boolean.parseBoolean(
-        config.get(SnowflakeSinkConnectorConfig.SNOWPIPE_STREAMING_V2_ENABLED));
+    return Optional.ofNullable(
+            config.get(SnowflakeSinkConnectorConfig.SNOWPIPE_STREAMING_V2_ENABLED))
+        .map(Boolean::parseBoolean)
+        .orElse(SNOWPIPE_STREAMING_V2_ENABLED_DEFAULT_VALUE);
   }
 
   /**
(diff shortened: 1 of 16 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -248,7 +251,7 @@ private TopicPartitionChannel createTopicPartitionChannel(
         topicPartition,
         partitionChannelKey, // Streaming channel name
         tableName,
-        hasSchemaEvolutionPermission,
+        schemaEvolutionEnabled,
```
sfc-gh-wtrefon, line 251: are we sure this does not change something? Just to be sure, the logic is slightly different now

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -420,6 +425,14 @@ public static String role(Map<String, String> config) {
     return config.get(SF_ROLE);
   }
 
+  public static String database(Map<String, String> config) {
+    return config.get(SF_DATABASE);
+  }
+
+  public static String schema(Map<String, String> config) {
```
sfc-gh-wtrefon: nit: maybe to stick to convention: getDatabase, getSchema

src/main/java/com/snowflake/kafka/connector/internal/streaming/DefaultStreamingConfigValidator.java
```diff
@@ -106,6 +100,39 @@ public ImmutableMap<String, String> validate(Map<String, String> inputConfig) {
...
+    }
+    return Optional.empty();
+  }
+
+  private static Optional<Map.Entry<String, String>> validateRoleForSSv2(
+      Map<String, String> inputConfig) {
+    if (inputConfig.containsKey(Utils.SF_ROLE)) {
+      String rolePresentForSSv2 =
+          String.format(
+              "'%s' must not be present when '%s' is enabled",
```
sfc-gh-lshcharbaty: Nit: shall we mention that a default user role is always used instead?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-mbobowski: Thanks, I double checked on that. This value is taken from the cache that is populated like:
```
if (!tableName2SchemaEvolutionPermission.containsKey(tableName)) {
      if (Utils.isSchematizationEnabled(connectorConfig)) {
...
          tableName2SchemaEvolutionPermission.put(tableName, hasSchemaEvolutionPermission);
```
So my change shouldn't affect the e2e behavior but makes configuration be checked twice. I will simply rename the variable.

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-mbobowski: The conventions are actually mixed in this file, but I am ok with adding get prefix.

src/main/java/com/snowflake/kafka/connector/internal/streaming/DefaultStreamingConfigValidator.java
sfc-gh-mbobowski: Good point. Changed to be more precise.

src/main/java/com/snowflake/kafka/connector/Utils.java
```diff
@@ -420,6 +425,14 @@ public static String role(Map<String, String> config) {
...
   }
 
+  public static String getDatabase(Map<String, String> config) {
+    return config.get(SF_DATABASE);
+  }
+
+  public static String getSchema(Map<String, String> config) {
+    return config.get(SF_SCHEMA);
+  }
+
```
sfc-gh-dseweryn, line 435: Could we extend the mini-refactor to all getters so they would be consistent?

sfc-gh-dseweryn (approved):
Generally OK. I would love to have consistent function naming though

src/main/java/com/snowflake/kafka/connector/Utils.java
sfc-gh-mbobowski: Unified to get...

PR #1133, opened 2025-07-04
CONN-10463 Revert role validation + telemetry improvements + Java 11 fixes

sfc-gh-mbobowski:
This combo PR contains 3 changes:
1. Revert of role validation for SSv2. Testing on MSK showed that connector throws NPE in `boolean hasSchemaEvolutionPermission(String tableName, String role);`. Service ITs missed that one, because config validations are called on a different level.
2. Adding usage of SSv2 and Iceberg to telemetry.
3. Change guava Immutable collection to Java 11 collections

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -19,7 +19,6 @@ package com.snowflake.kafka.connector;
 import static com.snowflake.kafka.connector.Utils.isSnowpipeStreamingIngestion;
 
 import com.google.common.base.Strings;
-import com.google.common.collect.ImmutableSet;
 import com.snowflake.kafka.connector.internal.KCLogger;
 import com.snowflake.kafka.connector.internal.streaming.IngestionMethodConfig;
 import java.util.Arrays;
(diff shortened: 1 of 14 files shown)


sfc-gh-mbobowski:
> LGTM, Two questions:
> 
> 1. Have you considered a broader push to use other Java std lib, eg. `Map.of()`?
> 2. IIUC role checking is OK for both SSV1 and SSV2?

1. Not yet, just following a scout rule.
2. Yes, we will require user to provide a role for either v1 and v2. In v2 PuPr it will be the default user role.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
```diff
@@ -251,7 +250,7 @@ public class SnowflakeSinkConnectorConfig {
       "value.converter.schema.registry.url";
 
   public static final Set<String> CUSTOM_SNOWFLAKE_CONVERTERS =
-      ImmutableSet.of(
+      Set.of(
```
sfc-gh-wtrefon, line 253: 👍

src/main/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryService.java
```diff
@@ -263,6 +263,17 @@ protected void addUserConnectorPropertiesToDataNode(
         userProvidedConfig.getOrDefault(
             ENABLE_SCHEMATIZATION_CONFIG, ENABLE_SCHEMATIZATION_DEFAULT));
 
+    dataObjectNode.put(
+        SNOWPIPE_STREAMING_V2_ENABLED,
+        userProvidedConfig.getOrDefault(
+            SNOWPIPE_STREAMING_V2_ENABLED,
+            String.valueOf(SNOWPIPE_STREAMING_V2_ENABLED_DEFAULT_VALUE)));
+
+    dataObjectNode.put(
```
sfc-gh-wtrefon, line 272: nice, will be useful in future.

PR #1141, opened 2025-07-15
CONN-10463 Get rid of Jackson based serialization for SSv2

sfc-gh-mbobowski:
Initially I implemented passing metadata to SSv2 SDK as a POJO (mostly due to problems with Iceberg serialization). During performance tests though I noticed that Jackson annotations are not working due to some regression that happened after SDK 0.3.0. As a result there was a mismatch between `SnowflakeConnectorPushTime` and `snowflakeConnectorPushTime` which would be very inconvenient for the end user.

It is not clear if the POJO serialization with custom Jackson annotations is supported by the SDK so I decided to switch back to Map.


src/main/java/com/snowflake/kafka/connector/records/SnowflakeTableStreamingRecordMapper.java
```diff
@@ -126,8 +126,7 @@ private Map<String, Object> getMapFromJsonNodeForStreamingIngest(
 
   private Object getMetadata(RecordService.SnowflakeTableRow row) throws JsonProcessingException {
     if (ssv2Enabled) {
-      Map<String, Object> mapForMetadata = getMapForMetadata(row.getMetadata());
-      return metadataFromMap(mapForMetadata);
```
sfc-gh-lshcharbaty, line 130: I see `MetadataRecord` pojo is still used with iceberg. Is that intended?

sfc-gh-mbobowski: I was hesitant but after giving it a second thought I deleted the serialization code for Iceberg and move MetadataRecord class back to test package.

Iceberg is currently not supported in SSv2 and while this code might be useful it's better to not keep redundant codebase in the repository.

src/test/java/com/snowflake/kafka/connector/records/SnowflakeTableStreamingRecordMapperTest.java
```diff
@@ -117,38 +137,30 @@ public static Stream<Arguments> ssv2NoSchematizationData() throws JsonProcessing
...
             Map.of(
-                TABLE_COLUMN_METADATA,
-                new MetadataRecord(null, null, null, null, null, null, null, null, null, Map.of()),
-                TABLE_COLUMN_CONTENT,
-                new HashMap<>())),
+                TABLE_COLUMN_METADATA, EMPTY_SSV2_METADATA, TABLE_COLUMN_CONTENT, new HashMap<>())),
         Arguments.of(
             buildRowWithDefaultMetadata("{}"),
             Map.of(
-                TABLE_COLUMN_METADATA, fullRecordMetadata, TABLE_COLUMN_CONTENT, new HashMap<>())),
```
sfc-gh-lshcharbaty, line 127: `fullRecordMetadata` in the parent class is not used anymore.

sfc-gh-mbobowski: Thx, deleted

PR #1145, opened 2025-07-28
SNOW-2227742 Add exponential backoff to handle openChannel errors

sfc-gh-mbobowski:
SNOW-2227742

When a huge number of channels is being opened (~7k) ingest sdk may throw SFException caused by 429 HTTP status. Currently even a single channel failure causes the whole connector to break and retry to open all channels again...

There are some retries configured in the underlying sdk code, but it's neither perfect nor configurable. Adding retries in the connector is a cheap solution that can greatly benefit in some scenarios.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/OpenChannelRetryPolicy.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/OpenChannelRetryPolicy.java
@@ -24,12 +24,12 @@ class OpenChannelRetryPolicy {
 
   // Retry policy constants
   /** Initial delay before the first retry attempt. */
-  private static final Duration INITIAL_DELAY = Duration.ofSeconds(3);
+  private static final Duration INITIAL_DELAY = Duration.ofSeconds(2);
 
   /** Maximum delay between retry attempts. */
   private static final Duration MAX_DELAY = Duration.ofMinutes(1);
 
-  /** Exponential backoff multiplier (retry delays: 3s, 6s, 12s, 24s, 48s, 60s max). */
+  /** Exponential backoff multiplier (retry delays: 4s, 8s, 16s, 32s, 60s max). */
   private static final double BACKOFF_MULTIPLIER = 2.0;
 
   /** Random jitter added to retry delays to prevent thundering herd. */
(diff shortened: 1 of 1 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/OpenChannelRetryPolicy.java
```diff
@@ -0,0 +1,87 @@
...
+            });
+
+    RetryPolicy<SnowflakeStreamingIngestChannel> retryPolicy =
+        RetryPolicy.<SnowflakeStreamingIngestChannel>builder()
+            .handle(SFException.class)
+            .withDelay(Duration.ofSeconds(3)) // initial delay
+            .withBackoff(
+                Duration.ofSeconds(3),
+                Duration.ofMinutes(1),
+                2.0) // retry after 3,6,12,... seconds up to 1 minute
```
sfc-gh-wtrefon: i would extract the values as constants at the top of the class for better visibility

sfc-gh-mbobowski: Good point, extracted

src/main/java/com/snowflake/kafka/connector/internal/streaming/OpenChannelRetryPolicy.java
```diff
@@ -0,0 +1,87 @@
...
+                  "Open channel {} - max retry attempts reached. Last exception: {}",
+                  channelName,
+                  e.getLastException().getMessage(),
+                  e.getLastException());
+              throw e.getLastException();
+            });
+
+    RetryPolicy<SnowflakeStreamingIngestChannel> retryPolicy =
+        RetryPolicy.<SnowflakeStreamingIngestChannel>builder()
+            .handle(SFException.class)
```
sfc-gh-wtrefon, line 75: just wondering if we want to retry for every possible SF exception?

sfc-gh-mbobowski: I considered being more specific here, but what made me choose this solution was that `SFException` does not provide a contract for status code. We can only base on `private String vendorCode;` and I was hestitant about parsing some random string like 82 without guarantee that it won't change in the future.

There is even an API in failsafe that we could use `handleIf()` https://failsafe.dev/policies/#failure-handling

PR #1149, opened 2025-07-29
SNOW-2227742 Retry indefinitely on 429 and 500

sfc-gh-mbobowski:
# Overview

SNOW-XXXXX

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/OpenChannelRetryPolicy.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/OpenChannelRetryPolicy.java
@@ -1,13 +1,10 @@
 package com.snowflake.kafka.connector.internal.streaming;
 
-import com.snowflake.kafka.connector.SnowflakeSinkConnectorConfig;
 import com.snowflake.kafka.connector.internal.KCLogger;
 import dev.failsafe.Failsafe;
-import dev.failsafe.Fallback;
 import dev.failsafe.RetryPolicy;
 import dev.failsafe.function.CheckedSupplier;
 import java.time.Duration;
-import java.util.Map;
 import net.snowflake.ingest.streaming.SnowflakeStreamingIngestChannel;
 import net.snowflake.ingest.utils.SFException;
 
@@ -16,20 +13,23 @@ import net.snowflake.ingest.utils.SFException;
  * backoff and jitter.
  *
  * <p>This class provides a clean interface to execute channel opening operations with automatic
- * retry on various streaming-related exceptions.
+ * retry on HTTP 429 (rate limiting) errors from Snowflake streaming service.
  */
 class OpenChannelRetryPolicy {
 
   private static final KCLogger LOGGER = new KCLogger(OpenChannelRetryPolicy.class.getName());
+
+  private static final String RATE_LIMIT_MESSAGE_PART = "429";
+  private static final String INTERNAL_ERROR_MESSAGE_PART = "500";
 
   // Retry policy constants
   /** Initial delay before the first retry attempt. */
   private static final Duration INITIAL_DELAY = Duration.ofSeconds(2);
 
   /** Maximum delay between retry attempts. */
-  private static final Duration MAX_DELAY = Duration.ofMinutes(1);
+  private static final Duration MAX_DELAY = Duration.ofSeconds(8);
 
-  /** Exponential backoff multiplier (retry delays: 4s, 8s, 16s, 32s, 60s max). */
+  /** Exponential backoff multiplier (retry delays: 2s, 4s, 8s max). */
   private static final double BACKOFF_MULTIPLIER = 2.0;
 
   /** Random jitter added to retry delays to prevent thundering herd. */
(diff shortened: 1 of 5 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/OpenChannelRetryPolicy.java
```diff
@@ -38,59 +38,36 @@ class OpenChannelRetryPolicy {
...
-                        "Open channel {} retries exceeded. Last exception: {}",
-                        channelName,
-                        event.getException().getMessage()))
             .build();
 
-    return Failsafe.with(fallback).compose(retryPolicy).get(channelOpenAction);
+    return Failsafe.with(retryPolicy).get(channelOpenAction);
+  }
+
+  private static boolean shouldBeRetried(Throwable e) {
```
sfc-gh-wtrefon:
```suggestion
  private static boolean isRetryableError(Throwable e) {
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/OpenChannelRetryPolicy.java
```diff
@@ -16,20 +13,23 @@
...
  *
  * <p>This class provides a clean interface to execute channel opening operations with automatic
- * retry on various streaming-related exceptions.
+ * retry on HTTP 429 (rate limiting) errors from Snowflake streaming service.
  */
 class OpenChannelRetryPolicy {
 
   private static final KCLogger LOGGER = new KCLogger(OpenChannelRetryPolicy.class.getName());
 
+  private static final String RATE_LIMIT_MESSAGE_PART = "429";
```
sfc-gh-lshcharbaty: Let's use at least `HTTP Status: 429`. So that there is smaller likelihood for false positives.

src/main/java/com/snowflake/kafka/connector/internal/streaming/OpenChannelRetryPolicy.java
```diff
@@ -38,59 +38,36 @@ class OpenChannelRetryPolicy {
...
-                  e.getLastException().getMessage(),
-                  e.getLastException());
-              throw e.getLastException();
-            });
+      CheckedSupplier<SnowflakeStreamingIngestChannel> channelOpenAction, String channelName) {
 
     RetryPolicy<SnowflakeStreamingIngestChannel> retryPolicy =
         RetryPolicy.<SnowflakeStreamingIngestChannel>builder()
-            .handle(SFException.class)
+            .handleIf(e -> e instanceof SFException && shouldBeRetried(e))
```
sfc-gh-wtrefon: i would extract whole statement (e instanceof SFException && shouldBeRetried(e)) as shouldBeRetried
```suggestion
            .handleIf(OpenChannelRetryPolicy::shouldBeRetried)
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/OpenChannelRetryPolicy.java
```diff
@@ -16,20 +13,23 @@
...
  * <p>This class provides a clean interface to execute channel opening operations with automatic
- * retry on various streaming-related exceptions.
+ * retry on HTTP 429 (rate limiting) errors from Snowflake streaming service.
  */
 class OpenChannelRetryPolicy {
 
   private static final KCLogger LOGGER = new KCLogger(OpenChannelRetryPolicy.class.getName());
 
+  private static final String RATE_LIMIT_MESSAGE_PART = "429";
+  private static final String INTERNAL_ERROR_MESSAGE_PART = "500";
```
sfc-gh-wtrefon: don't we have anything more that we could add to the string besides just simple numbers? So we don't retry for some other numbers like 4295, 5001 etc. Maybe at least spaces on both sides?

sfc-gh-wtrefon: from the test example I see that it could be "HTTP Status: 429 " but not sure if it comes from real errors

sfc-gh-lshcharbaty: While I agree with infinite retries for 429 errors, I'm not so sure about 500.
429 means all is fine, but try later, so there is no problem in continuous retrying, since the service instructs us to do that.
500 on the other hand means something is very wrong. It's fine to retry a couple (dozens) times, to see if the issue is intermittent. But infinite retries mean we're constantly burning cpu, without any guarantees the service will be operational in reasonable time. Also a customer has no signal, besides the warning logs, that something goes wrong.

src/test/java/com/snowflake/kafka/connector/internal/streaming/OpenChannelRetryPolicyTest.java
```diff
@@ -37,77 +39,73 @@ void shouldReturnChannelOnFirstAttemptSuccess() {
...
+            SFException.class,
+            () -> OpenChannelRetryPolicy.executeWithRetry(supplier, channelName));
 
     assertSame(nonRetryableException, thrownException);
+    assertEquals(1, attemptCount.get()); // Should only attempt once
   }
 
   @Test
-  void shouldEventuallyFailAfterMaxRetries() {
+  void shouldRetryMultipleTimesOn429Exception() {
```
sfc-gh-lshcharbaty, line 89: Shall we test 500 too?

src/main/java/com/snowflake/kafka/connector/internal/streaming/OpenChannelRetryPolicy.java
sfc-gh-mbobowski: Agree

src/main/java/com/snowflake/kafka/connector/internal/streaming/OpenChannelRetryPolicy.java
sfc-gh-mbobowski: Done

src/test/java/com/snowflake/kafka/connector/internal/streaming/OpenChannelRetryPolicyTest.java
sfc-gh-mbobowski: Deleted retries for 500. Let SDK handle that.

src/main/java/com/snowflake/kafka/connector/internal/streaming/OpenChannelRetryPolicy.java
sfc-gh-mbobowski: Done

src/main/java/com/snowflake/kafka/connector/internal/streaming/OpenChannelRetryPolicy.java
sfc-gh-mbobowski: Done

sfc-gh-lshcharbaty (approved):

PR #1152, opened 2025-08-04
CONN-10517 Add retry on memory backpressure when inserting rows

sfc-gh-dseweryn:
# Overview

CONN-10517
When the SDK is not quick enough with sending out data it can throw an error informing that memory threshold is exceeded. This can be used as a back-pressure mechanism for the connector to slow down. Added an unlimited retry in that case.

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithRetryAndFallbackPolicy.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithRetryAndFallbackPolicy.java
@@ -23,10 +23,7 @@ class AppendRowWithRetryAndFallbackPolicy {
 
   // Retry policy constants
   /** Delay before next retry attempt. */
-  private static final Duration DELAY = Duration.ofSeconds(5);
-
-  /** Maximum duration for retries before giving up. */
-  private static final Duration MAX_RETRIES_DURATION = Duration.ofMinutes(5);
+  private static final Duration RETRY_DELAY = Duration.ofSeconds(5);
 
   /** Random jitter added to retry delays to prevent potential partition starving. */
   private static final Duration JITTER_DURATION = Duration.ofMillis(200);
@@ -72,9 +69,8 @@ class AppendRowWithRetryAndFallbackPolicy {
     RetryPolicy<AppendResult> memoryBackpressureRetryPolicy =
         RetryPolicy.<AppendResult>builder()
             .handleIf(AppendRowWithRetryAndFallbackPolicy::isMemoryBackpressure)
-            .withDelay(DELAY)
+            .withDelay(RETRY_DELAY)
             .withJitter(JITTER_DURATION)
-            .withMaxDuration(MAX_RETRIES_DURATION)
             .withMaxAttempts(-1)
             .onRetry(
                 event ->
--- a/src/test/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithRetryAndFallbackPolicyTest.java
+++ b/src/test/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithRetryAndFallbackPolicyTest.java
@@ -58,7 +58,7 @@ public class AppendRowWithRetryAndFallbackPolicyTest {
             supplier, failingFallback(), channelName);
 
     // Then
-    assertEquals(3, attemptCounter.get()); // Should retry once
+    assertEquals(3, attemptCounter.get()); // Should retry thrice (1 initial + 2 retries)
     assertSame(appendResult, result);
   }
 
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java
@@ -41,7 +41,7 @@ public class SnowflakeSinkConnectorConfig {
   // For tombstone records
   public static final String BEHAVIOR_ON_NULL_VALUES_CONFIG = "behavior.on.null.values";
 
-  // Buffer thresholds
+  // Buffer thresholds for Snowpipe
   public static final String BUFFER_FLUSH_TIME_SEC = "buffer.flush.time";
   public static final long BUFFER_FLUSH_TIME_SEC_DEFAULT = 120;
   public static final long BUFFER_FLUSH_TIME_SEC_MIN = 10;


src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithRetryAndFallbackPolicy.java
```diff
@@ -0,0 +1,118 @@
...
...bdkEPwoHgoC4c1F8hLXi4LD2oSiUJZqneL30K5+GCWWotEmDlxtkunjfL6ZpVbC9kCxAfqtuaiRD+q/jJTqfVPWOqRFKOYm6khmaQgmcWt4RYy+FOR2RTjFvxdjcyUGjmyNlvUKVqRwFrPBnGV4+v+Y9wDeXeqmveoGRGNL3GSk00c+jLotrirIomGNBAlldM1CPwXSYujrfId7Kg3oMwjEDul3l2HAMFWHX1zUHRRoFAsQGULvx0YcroFs4szHLcRYXCE4aXuNsha1Jhj4+NJpVBhp8Dda0UsRuNTpb6n/VwfF1OHSMQYEaXAg0ryRpmQKBgQD8uOXfLs7aiPgQVpOfU26PqK9teQxqy/lydiQfH6lnltCO2nNpsXyOAQmvr4rl4BgKBQGTcsUNvBaCqahhdQ56qSKdEtE9NqU0DR4jnocLG73fUXCsyATx51YqPhGL7Vzuygb9eSl//xXU2Xf9ri30VFYyoOhHYslbQARUv22++QKBgQDF2dveKKwaiRMyi9+M6KtkJjC/DfA5wQpLY3pK5mej7CuadXb4EolflfcXJtRUWmnqHK47V8nTHyzt11QgPdQw8/3NwnmG02mW7CxPOUF5BJhfMi1A4DPsaauAm2/3Ba7/z3CNGeyhRfjIh+lAJPhpbr5U24sSG5K+G9K01L/1kwKBgFogZnotaz+sUSsidQsA6hD5i3ZzEOaVT6RuVXt0AyoaMs5/QWGUFmNeI2IwLEzGmkwYdPNsnrHDPhPjBVWcNUmxP9LcfxphfpENjI4PXwGrJw+ZL6Guw3Sxl50A/aj1QYs3KT8EErS+974swAoML/9OhF8R20R6Cm0sa1791dmhAoGAPuYy6x7dEBaR/tjO+joh+tXVW3fv7jx7rFn00AkN2+RCPHmpl9Y0zCmeA0M575NK5r21enI/XJolq3twaEufwOxgvOBdyEo/4FY3A/y83t9iHamTD3P3reouNtWzs8Gfz/2/12vteZwIKqBivBzEyKgHZGzjep8/eKbZxtl4KXcCgYBgQdtqdagUWCsfJEPByVuciEjNO+oOjkJm3jzePWBOGC67uOeHSaWZR7tbX9IhaAZh5fFQu7urS4pS+nCsffUc1/xUIW3llShCshdiTcqT3/Y2srUgkq/5cQXL2kNLHr770VlPn0s2SKDzVhHIVOflsYRBTQJf69PZwx3OSlkKuQ==
```
sfc-gh-lkucharski, line 114: Why PEM keys in comments?

sfc-gh-lkucharski (changes requested):

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithRetryAndFallbackPolicy.java
```diff
@@ -14,9 +16,17 @@
...
-  private static final KCLogger LOGGER = new KCLogger(AppendRowWithFallbackPolicy.class.getName());
+  private static final KCLogger LOGGER =
+      new KCLogger(AppendRowWithRetryAndFallbackPolicy.class.getName());
+
+  // Retry policy constants
+  /** Delay before next retry attempt. */
+  private static final Duration DELAY = Duration.ofSeconds(5);
+
+  /** Random jitter added to retry delays to prevent potential partition starving. */
+  private static final Duration JITTER_DURATION = Duration.ofMillis(200);
```
sfc-gh-lkucharski, line 29: Did you mean to generate a random duration here? Looks like it's fixed

sfc-gh-lkucharski (changes requested):

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithRetryAndFallbackPolicy.java
sfc-gh-dseweryn: It is random, see [documentation](https://failsafe.dev/javadoc/core/dev/failsafe/RetryPolicyBuilder.html#withJitter-double-).

sfc-gh-lkucharski: got it

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithRetryAndFallbackPolicy.java
```diff
@@ -56,7 +66,29 @@ static AppendResult executeWithFallback(
                         event.getException()))
             .build();
 
-    return Failsafe.with(reopenChannelFallbackExecutor).get(appendRowAction);
+    RetryPolicy<AppendResult> memoryBackpressureRetryPolicy =
+        RetryPolicy.<AppendResult>builder()
+            .handleIf(AppendRowWithRetryAndFallbackPolicy::isMemoryBackpressure)
+            .withDelay(DELAY)
+            .withJitter(JITTER_DURATION)
+            .withMaxAttempts(-1)
```
lkuchars, line 74: Is it ok to retry indefinitely on this error? Perhaps it should be better to set this number to sth high but let the reopenChannelFallbackExecutor reopen a new channel?

sfc-gh-dseweryn: I was about to say that from what we know this error is recoverable on its own eventually. On the other hand if there is some unknown bug in the SDK manual close/reopen can help. Adding 5 minutes max retry.

lkuchars: I was thinking along the same lines. However, I had a bigger timeout in mind, like 30 minutes.

sfc-gh-lkucharski (approved):
LGTM

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithRetryAndFallbackPolicy.java
```diff
@@ -14,9 +16,20 @@
...
-class AppendRowWithFallbackPolicy {
+class AppendRowWithRetryAndFallbackPolicy {
 
-  private static final KCLogger LOGGER = new KCLogger(AppendRowWithFallbackPolicy.class.getName());
+  private static final KCLogger LOGGER =
+      new KCLogger(AppendRowWithRetryAndFallbackPolicy.class.getName());
+
+  // Retry policy constants
+  /** Delay before next retry attempt. */
+  private static final Duration DELAY = Duration.ofSeconds(5);
```
sfc-gh-wtrefon:
```suggestion
  private static final Duration RETRY_DELAY = Duration.ofSeconds(5);
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithRetryAndFallbackPolicy.java
```diff
@@ -14,9 +16,20 @@
...
+
+  /** Maximum duration for retries before giving up. */
+  private static final Duration MAX_RETRIES_DURATION = Duration.ofMinutes(5);
+
+  /** Random jitter added to retry delays to prevent potential partition starving. */
```
sfc-gh-wtrefon: i dont think we need comments for such basic things

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithRetryAndFallbackPolicy.java
sfc-gh-wtrefon: i think we should retry here indefinitely, no max duration. We cannot just assume that everywhere is a bug. If there is, we will report it and use fixed sdk version. **And most importantly, if we do not commit in time, rebalancing will just kill the task and create a new one. With MAX_RETRIES_DURATION we will throw an exception which will kill the task and require manual restart**

src/test/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithRetryAndFallbackPolicyTest.java
```diff
@@ -0,0 +1,125 @@
...
+          return appendResult;
+        };
+
+    // When
+    AppendResult result =
+        AppendRowWithRetryAndFallbackPolicy.executeWithRetryAndFallback(
+            supplier, failingFallback(), channelName);
+
+    // Then
+    assertEquals(3, attemptCounter.get()); // Should retry once
```
sfc-gh-wtrefon:
```suggestion
    assertEquals(3, attemptCounter.get()); // Should retry twice
```

sfc-gh-wtrefon (changes requested):
I think we should not have retry timeout

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithRetryAndFallbackPolicy.java
sfc-gh-mbobowski: I agree with @sfc-gh-wtrefon comment. The connector won't get stuck in this place indefinitely so timeout can be safely skipped.

I also wonder if the existing sdk throttling strategy is safe when considering OOM prevention. Is it always activated at 80% of the available memory? Should we add a parameter to limit the total memory consumption?

sfc-gh-dseweryn: Removing then.

There are several parameters available for handling backpressure — I would need to contact the development team to pick correct ones

sfc-gh-dseweryn: I checked the defaults and for now I think there is no need to add new parameters because:
- they are not straightforward to understand
- the SDK is not yet stable so they may change
- defaults look pretty Ok for my taste

sfc-gh-wtrefon: can they be provided by the snowflake.streaming.client.provider.override.map ?

sfc-gh-mbobowski (approved):

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithRetryAndFallbackPolicy.java
sfc-gh-dseweryn: it seems so

PR #1157, opened 2025-08-22
SNOW-2277590 new maven central release plugin

sfc-gh-dseweryn:
# Overview

SNOW-2277590

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,12 +88,6 @@
         </repository>
     </repositories>
 
-    <distributionManagement>
-        <snapshotRepository>
-            <id>ossrh</id>
-            <url>https://ossrh-staging-api.central.sonatype.com/content/repositories/snapshots</url>
-        </snapshotRepository>
-    </distributionManagement>
     <build>
         <plugins>
             <plugin>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -12,7 +12,7 @@
 
     <groupId>com.snowflake</groupId>
     <artifactId>snowflake-kafka-connector</artifactId>
-    <version>3.2.4</version>
+    <version>3.2.8</version>
```
sfc-gh-mbobowski, line 15: I would like to avoid bumping version by 5 releases only because we tested a new release process.

sfc-gh-mbobowski: ok, let's release next version as 3.3.0

sfc-gh-dseweryn: While 4 non-production releases are less than ideal, they will not break anything. We can release 3.2.9 or 3.3.0 without problems.

Test releases are marked as `preview` so should not spark any confusion.

sfc-gh-mbobowski: As @sfc-gh-lshcharbaty pointed out when some people see 3.2.4 followed by 3.2.9 they assume that we broken missing versions. I agree it's not a technical problem.

PR #1165, opened 2025-09-15
SNOW-2335144 Add test for large blobs and snowpipe

sfc-gh-lshcharbaty:
# Overview

SNOW-2335144

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/test/test_suit/test_large_blob_snowpipe.py
+++ b/test/test_suit/test_large_blob_snowpipe.py
@@ -1,0 +1,77 @@
+from test_suit.test_utils import RetryableError, NonRetryableError
+import json
+import random
+import string
+from test_suit.base_e2e import BaseE2eTest
+
+
+class TestLargeBlobSnowpipe(BaseE2eTest):
+    def __init__(self, driver, nameSalt):
+        self.driver = driver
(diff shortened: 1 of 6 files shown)


test/test_suit/test_large_blob_snowpipe.py
```diff
@@ -0,0 +1,77 @@
...
+from test_suit.base_e2e import BaseE2eTest
+
+
+class TestLargeBlobSnowpipe(BaseE2eTest):
+    def __init__(self, driver, nameSalt):
+        self.driver = driver
+        self.fileName = "travis_correct_large_blob_snowpipe"
+        self.topic = self.fileName + nameSalt
+        self.recordCount = 5
+        self.blobSize = 20 * 1024 * 1024  # 20 MiB
```
sfc-gh-lkucharski, line 14: Did you mean to test 30MB blob size here?

sfc-gh-lshcharbaty: I wanted to test 20 MiB. The default supported blob size is 16 MiB, so any value larger than it works.
Any value larger tan 20 MiB in the server setting works. I chose 30 MiB, but this can be less.

PR #1175, opened 2025-10-06
CONN-10559 Do not allow schematization config option when using SF Streaming SSv2

sfc-gh-lkucharski:
…

# Overview

CONN-10559 Do not allow schematization config option when using SF Streaming SSv2

## Pre-review checklist
- [ ] Snowpipe Changes

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/DefaultStreamingConfigValidator.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/DefaultStreamingConfigValidator.java
@@ -170,9 +170,9 @@ public class DefaultStreamingConfigValidator implements StreamingConfigValidator
       if (isSchematizationEnabled && Utils.isSnowpipeStreamingV2Enabled(inputConfig)) {
         invalidParams.put(
             SnowflakeSinkConnectorConfig.ENABLE_SCHEMATIZATION_CONFIG,
-            "Schematization is not supported with Snowpipe Streaming V2. Please set"
-                + " snowflake.enable.schematization to false when"
-                + " snowflake.streaming.v2.enabled is true.");
+            "Schematization is not yet supported with Snowpipe Streaming: High-Performance"
+                + " Architecture. Please set snowflake.enable.schematization to false when"
+                + " snowflake.streaming.v2.enabled is set to true.");
       }
 
       if (isSchematizationEnabled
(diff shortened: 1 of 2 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/DefaultStreamingConfigValidator.java
```diff
@@ -162,8 +162,20 @@ private static Map<String, String> validateSchematizationConfig(Map<String, Stri
...
-              inputConfig.get(SnowflakeSinkConnectorConfig.ENABLE_SCHEMATIZATION_CONFIG))
+      boolean isSchematizationEnabled =
+          Boolean.parseBoolean(
+              inputConfig.get(SnowflakeSinkConnectorConfig.ENABLE_SCHEMATIZATION_CONFIG));
+
+      // Validate that schematization and streaming V2 are mutually exclusive
+      if (isSchematizationEnabled && Utils.isSnowpipeStreamingV2Enabled(inputConfig)) {
+        invalidParams.put(
+            SnowflakeSinkConnectorConfig.ENABLE_SCHEMATIZATION_CONFIG,
+            "Schematization is not supported with Snowpipe Streaming V2. Please set"
```
sfc-gh-nsienkiewicz-1:
```suggestion
            "Schematization is not yet supported with Snowpipe Streaming: High-Performance Architecture. Please set"
```

SSv1 / SSv2 is our internal numenclature, not something customers are aware of. In documentation, we should use official naming

sfc-gh-mbobowski: Then we should also reconsider changing the user visible parameter name from `snowflake.streaming.v2.enabled` to `snowflake.streaming.high.performance.enabled`. Perhaps in a separate PR.

src/test/java/com/snowflake/kafka/connector/ConnectorConfigValidatorTest.java
```diff
@@ -983,4 +983,52 @@ public void testENABLE_REPROCESS_FILES_CLEANUP_invalid_value() {
...
+            .withSchematizationEnabled(false)
+            .withSnowpipeStreamingV2Enabled()
+            .build();
+
+    assertThatCode(() -> connectorConfigValidator.validateConfig(config))
+        .doesNotThrowAnyException();
+  }
+
+  @Test
+  public void testSchematizationEnabledWithoutStreamingV2_shouldPass() {
```
sfc-gh-mbobowski: nit: I don't think that these two tests bring much value. Configurations are already tested in other scenarios.

sfc-gh-mbobowski (approved):
LGTM, left minor comments

src/test/java/com/snowflake/kafka/connector/ConnectorConfigValidatorTest.java
sfc-gh-lkucharski: Removed

PR #1185, opened 2025-10-15
SNOW-2193898 Recreate streaming ingest client on CLIENT_CLOSED error

sfc-gh-aminyaylov:
# Overview

SNOW-2193898 Support transparent failover with [Client Redirect](https://docs.snowflake.com/en/user-guide/client-redirect) in streaming mode

## Problem

When Streaming Ingest SDK returns a `CLOSED_CLIENT` error, Kafka Connector will attempt to reopen channels with the same closed client, leading to repeated failures.

Customers are currently required to manually restart their KC after changing deployment: https://docs.snowflake.com/en/user-guide/snowpipe-streaming/snowpipe-streaming-classic-kafka#failover-limitations

## Solution

Implement automatic client recreation when a `CLOSED_CLIENT` error is detected. Streaming Ingest SDK will separately be updated to close the client when it detects a change to the primary deployment.

### SSv1

- `DirectTopicPartitionChannel`
    - Make `streamingIngestClient` field non-final to allow updates
    - Update `insertRowFallbackSupplier()` to call `StreamingClientProvider.recreateClient()` when client is closed

### SSv2

- `SnowpipeStreamingV2PartitionChannel`
    - Add `CLIENT_CLOSED_ERROR_CODE` constant with TODO (SSv2 SDK does not yet return structured error codes)
    - Update `insertRowFallbackSupplier()` to call `StreamingIngestClientV2Provider.recreateClient()`

### Client optimization

Be mindful to support both optimized and unoptimized path. To summarize:

- If optimization is **enabled**, all client with the same client properties will share a single client, managed by the `StreamingClientProvider`​. In this case, the `SinkService`​ **does not** cache the client.
- If optimization is **disabled**, every `SinkService`​ manages its own client and the shared cache managed by StreamingClientProvider is not used. In this case, `SinkService`​ **must** maintain its own reference to the client object and handle closing (which in the optimized path is done by the provider cache eviction policy).

## Testing

- [x] Added IT `testClientRecreationOn_ClosedClientError()` that validates:
    1. Client is closed (simulating CLOSED_CLIENT error)
    2. New client is created on next operation
    3. Data continues to be inserted successfully
    4. Works for both SSv1 and SSv2

## Urgency

This review is **high** priority. Customer requesting KC streaming ingest support for Client Redirect.

## Risks

Client recreate is an expensive operation. However:

1. The current response of reopening channel without fixing the underlying issue is wasteful.
2. A well-behaved customer following documentation will already be performing this action automatically when changing primary deployments.

## Backward/forward compatible

[x] This change does not introduce any breaking API or data format changes.

## Pre-review checklist

- [ ] Snowpipe Changes
- [ ] This change is protected by a config parameter \<PARAMETER_NAME> eg `snowflake.ingestion.method`.
    - [x] `No` - This change should be enabled on all KC instances.
- [ ] Is this change protected by parameter \<PARAMETER_NAME> on the server side?

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
@@ -414,9 +414,8 @@ public class DirectTopicPartitionChannel implements TopicPartitionChannel {
   private void insertRowFallbackSupplier(Throwable ex)
       throws TopicPartitionChannelInsertionException {
     if (isClientInvalidError(ex)) {
-      this.streamingIngestClient =
-          StreamingClientProvider.getStreamingClientProviderInstance()
-              .recreateClient(sfConnectorConfig);
+      LOGGER.warn("Client is invalid, recreating client for channel: {}", getChannelNameFormatV1());
+      this.streamingIngestClient = StreamingClientProvider.getStreamingClientProviderInstance().getClient(sfConnectorConfig);
     }
     final long offsetRecoveredFromSnowflake =
         streamingApiFallbackSupplier(StreamingApiFallbackInvoker.INSERT_ROWS_FALLBACK);
@@ -676,11 +675,9 @@ public class DirectTopicPartitionChannel implements TopicPartitionChannel {
     try {
       return OpenChannelRetryPolicy.executeWithRetry(
           () -> {
-            // Recreate client if invalid before each attempt
             if (!StreamingClientHandler.isClientValid(streamingIngestClient)) {
-              this.streamingIngestClient =
-                  StreamingClientProvider.getStreamingClientProviderInstance()
-                      .recreateClient(sfConnectorConfig);
+              LOGGER.warn("Client is invalid, recreating client for channel: {}", getChannelNameFormatV1());
+              this.streamingIngestClient = StreamingClientProvider.getStreamingClientProviderInstance().getClient(sfConnectorConfig);
             }
             return streamingIngestClient.openChannel(channelRequest);
           },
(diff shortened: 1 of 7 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/DirectTopicPartitionChannel.java
```diff
@@ -478,6 +484,19 @@ public long fetchOffsetTokenWithRetry() {
     return offsetTokenExecutor.get(this::fetchLatestCommittedOffsetFromSnowflake);
   }
 
+  /**
+   * Checks if the exception indicates a client invalidation error.
+   *
+   * @param e the exception to check
+   * @return true if the exception is a CLOSED_CLIENT error
+   */
+  private boolean isClientInvalidError(Throwable e) {
```
sfc-gh-mbobowski, line 493: I think we also need to handle this in `streamingIngestClient.openChannel()`

sfc-gh-aminyaylov: Good point. I added a `client.isClosed()` check to the `openChannelForTable` retry policy.

Furthermore, I'm thinking about possible race conditions when reopening a client that's shared by multiple channels, for example:

```
Channel A: insertRows() → recreateClient()
Channel B: openChannel() → recreateClient()
```

The following implicit locking should handle this safely. I also added a validity check to avoid double recreation:

```
// ConcurrentHashMap.compute() ensures that only one thread recreates a client for the specified key.
return this.registeredClients.asMap().compute(clientProperties, (key, client) -> {
  if (StreamingClientHandler.isClientValid(client)) {
    LOGGER.info("Client was already recreated by another thread: {}", client.getName());
    return client;
  }
  LOGGER.info("Recreating client: {}", clientProperties.toString());
  return this.streamingClientHandler.createClient(clientProperties);
});
```

PR #1317, opened 2026-02-28
SNOW-3136086 create tables and streaming clients in parallel

sfc-gh-skurella:
# Overview

SNOW-3136086

Checking for table and pipe existence is embarrassingly parallel so in this PR we're dispatching it to the IO thread pool.

Also, even though streaming client creation is already executing on the IO thread pool, `startPartitions` created them sequentially due to the blocking nature of `getClient`. We're parallelizing that here also. I.e. previously parallelism was between tasks, now it's also leveraged within a task.


src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPool.java
```diff
@@ -136,24 +140,40 @@ SnowflakeStreamingIngestClient getClient(
...
+        });
+  }
 
-    LOGGER.info(
-        "Task {} now using pipe {} for connector {}, total tasks on this pipe: {}",
-        taskId,
-        pipeName,
-        connectorName,
-        entry.taskCount());
+  SnowflakeStreamingIngestClient getClient(
```
sfc-gh-skurella: TODO:
- [ ] this method is only used in tests now, remove it - we handle the synchronous `getClient` in `StreamingClientPools`

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -173,23 +175,38 @@ public void startPartition(TopicPartition topicPartition) {
...
+                topic ->
+                    CompletableFuture.runAsync(
+                        () -> {
+                          final String tableName =
+                              getTableName(topic, this.topicToTableMap, this.enableSanitization);
+                          createTableIfNotExists(tableName);
+
+                          // Look for a pipe with the same name as the table. Otherwise, use the
+                          // default pipe.
+                          final boolean pipeExists = this.conn.pipeExist(tableName);
```
sfc-gh-mbobowski: Previously the connection object was used by a single thread and now it is not. Under the hood it's Snowflake JDBC Connection object that might not be thread-safe according to docs: https://docs.snowflake.com/en/developer-guide/driver-connections

Please, verify this behavior.

sfc-gh-skurella: Good point, thanks! I went through those limitations and some definitely apply. I'll create a connection pool for this logic.

### Session state

> Sessions keep track of the current database, schema, and role. If one thread changes these values (like USE DATABASE), the other thread might be affected. This impact is particularly important because changing to another schema with the same tables might cause a thread to accidentally modify the wrong table. Additionally, changing connection or configuration parameters in one session can affect all threads that use that session.

AFAIK we don't mutate the session state beyond connector creation. We don't use multiple databases, schemas, etc.

### Transaction state

> A transaction might start in one session. If multiple threads have access to that session, each one can potentially modify data in the same transaction, which might cause the data to be accidentally persisted or lost if a transaction is committed or rolled back.

We don't use transactions here.

### Sequence counter

> Drivers use a sequence counter to retry requests. Because sequence counters are global for a session, reusing a session in different threads might also inadvertently alter the global sequence counter that can result in unpredictable behavior for retrying requests.

**This could be an issue.**

### Query context cache

> For improved performance, sessions keep track of some internal information in a driver-specific or internal cache. The cache updates after every query, so running multiple queries concurrently in a session could result in data corruption.

**That too.**

### Last query ID

> Connections keep track of the last query ID, which can then be retrieved and used. If two queries run in parallel in different threads, a race condition can affect which one sets the last query ID.

We don't use that here.

sfc-gh-skurella:
Closing for now - likely not a useful optimization at this time.

PR #1400, opened 2026-04-01
Fix v4 channel recovery after CREATE OR REPLACE TABLE

sfc-gh-ppopov:
## Summary
After `CREATE OR REPLACE TABLE`, the SSv2 pipe is destroyed. The server returns `ERR_CHANNEL_DOES_NOT_EXIST_OR_NOT_AUTHORIZED` as a normal `ChannelStatus` response (not an `SFException`). `BatchOffsetFetcher` was only catching `SFException`, so this status went undetected — channels stayed connected to the dead pipe, silently losing all subsequent data.

## Fix
Added `PipeInvalidatedException` to detect non-SUCCESS status codes in `BatchOffsetFetcher.getCommittedOffsetsForPipe()`. When a channel returns a non-SUCCESS status:

1. Throws `PipeInvalidatedException` (caught alongside `SFException`)
2. Invalidates the cached SDK client via `StreamingClientPools.invalidateClient()` (evicts and closes)
3. Closes affected channels via `closeChannelAsync()` (sets `cancelled` flag)

On the next `put()` call, `isChannelClosed()` returns `true`, triggering `startPartition()` which creates a fresh channel with a new SDK client against the replaced table.

## Changes
- `BatchOffsetFetcher` — add `PipeInvalidatedException` inner class, check `statusCode` before `processChannelStatus`, widen catch to include both `SFException` and `PipeInvalidatedException`
- `StreamingClientPool` / `StreamingClientPools` — add `invalidateClient(pipeName)` to evict broken clients
- `SnowpipeStreamingPartitionChannel.isChannelClosed()` — check `cancelled` flag so closed channels are detected immediately
- E2E tests — enable replace-table tests for v4, pre-create tables with full schema to avoid ALTER TABLE DDL pipe invalidation during Wave 1

## Test plan
- [x] Unit tests: `BatchOffsetFetcherTest` — 9 tests pass (2 new: `channelErrorStatusInvalidatesClientAndClosesChannels`, `channelErrorStatusForOnePipeDoesNotAffectOthers`)
- [x] Unit tests: `StreamingClientPoolTest` — 3 new tests for `invalidateClient()`
- [x] All existing unit tests pass
- [ ] E2E `test_se_replace_table[v4]` — blocked by pre-existing SDK v1.2.0 Rust FFI bug (background channel status cache refresh fails for pipes with hyphens, causing self-invalidation before data materializes)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sfc-gh-bmikaili:
What run and what exact logs is this based on? can we trace the log lines that prove this is what occured?

PR #1419, opened 2026-04-10
Fix data type normalization in RowValidator for v3/v4-compat parity

sfc-gh-ppopov:
## Background

In KC v3, `StreamingRecordMapper.getTextualValue()` serialized **every value to a String** before the SSv1 SDK saw it — Integer `42` became `"42"`, Boolean `true` became `"true"`, Map `{a:1}` became `"{\"a\":1}"`. The SSv1 SDK then re-parsed these strings using type-specific validators (`convertStringToBoolean`, `parseInstantGuessScale`, etc.).

In KC v4, commit `6be86dd2` (FLOW-7464) replaced `StreamingRecordMapper` with `KafkaRecordConverter`, which passes **native Java types** directly (Boolean stays `Boolean`, Integer stays `Integer`, Map stays `Map`). This was correct for SSv2-native mode (v4-ht) — SSv2 serializes the `Map<String, Object>` to NDJSON via Jackson, and pre-stringifying would cause double-encoding.

However, in v4-compat mode, `RowValidator` (vendored SSv1 `DataValidationUtil`) sits between `KafkaRecordConverter` and SSv2 SDK. It receives native types but SSv1's validators were designed to receive strings. This type convention mismatch caused all the divergences fixed here.

## Summary

Fixes data type handling divergences between KC v3 (SSv1 SDK) and v4-compat (RowValidator + SSv2 SDK):

- **BOOLEAN**: Normalize Integer 0/1 and String tokens to `Boolean` before the SSv2 SDK call. Pre-reject non-0/1 numeric values to match end-to-end KC v3 behavior (KC v3's record mapper converts integers to strings before SSv1, and SSv1's `convertStringToBoolean` only accepts "0"/"1"/"true"/"false"/"yes"/"no"/"on"/"off" — so "42" is rejected).
- **VARIANT (String input)**: Parse validated JSON to native Java object (Map/List/primitive) via `validateAndParseVariantAsObject` so the SSv2 SDK stores the parsed object, not a double-quoted JSON literal.
- **ARRAY (String input)**: Parse validated JSON array to native `List` via `validateAndParseArrayAsList` so the SSv2 SDK stores a proper array instead of a single-element wrapper.
- **VARCHAR (Map/Collection input)**: Serialize Map/Collection values to JSON strings before validation, replicating KC v3's record mapper serialization (`StreamingRecordMapper.getTextualValue → writeValueAsStringOrNanOrInfinity`). The SSv1 SDK itself rejects Map/Collection for VARCHAR — the serialization happens at the KC pipeline level.
- **TIMESTAMP (Integer/Long epoch)**: Normalize Integer/Long epoch values to ISO timestamp strings via `validateAndFormatTimestamp`. The SSv2 SDK passes raw integers to the Snowflake backend which interprets them using the channel's default timezone (America/Los_Angeles) instead of UTC. SSv1 SDK converted epochs to UTC client-side.

## Test plan

- [x] Unit tests: 104 cases pass (72 RowValidatorTest + 32 DataValidationUtilTest)
- [x] E2E tests: `test_type_compatibility.py` — 96 passed, 7 xfailed, 2 skipped across 3 modes (v3, v4-compat, v4-ht)
  - v4-compat `tsntz_int_epoch` now matches v3 (10:00:00 UTC) — was 02:00:00 before the fix
  - v4-ht still diverges on integer epochs (expected — no client-side validation)
- [x] Negative test cases cover: invalid JSON → OBJECT/ARRAY, bad strings → BOOLEAN, non-0/1 integers → BOOLEAN, cross-type Map/List → VARCHAR(10), invalid timestamps

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
@@ -274,6 +274,42 @@ class DataValidationUtil {
   }
 
   /**
+   * Validates and parses input for VARIANT columns, returning a native Java object (Map, List, or
+   * primitive) instead of a JSON string. For String inputs this avoids the serialize→re-parse
+   * roundtrip of {@link #validateAndParseVariant}.
+   *
+   * @param input Object to validate
+   * @param insertRowIndex
+   * @return Native Java object (Map, List, String, Number, Boolean, or null for missing nodes)
+   */
+  static Object validateAndParseVariantAsObject(
+      String columnName, Object input, long insertRowIndex) {
+    JsonNode node =
+        validateAndParseSemiStructuredAsJsonTree(columnName, input, "VARIANT", insertRowIndex);
+
+    if (node.isMissingNode()) {
+      return null;
+    }
+
+    String output = node.toString();
+    int stringLength = output.getBytes(StandardCharsets.UTF_8).length;
+    if (stringLength > MAX_SEMI_STRUCTURED_LENGTH) {
+      throw valueFormatNotAllowedException(
+          columnName,
+          "VARIANT",
+          String.format(
+              "Variant too long: length=%d maxLength=%d", stringLength, MAX_SEMI_STRUCTURED_LENGTH),
+          insertRowIndex);
+    }
+    try {
+      return objectMapper.treeToValue(node, Object.class);
+    } catch (JsonProcessingException e) {
+      // Should never happen: node was already validated by validateAndParseSemiStructuredAsJsonTree
+      throw new IllegalStateException("Failed to convert validated JsonNode to Object", e);
+    }
+  }
+
+  /**
    * Validates and parses input as JSON. All types in the object tree must be valid variant types,
    * see {@link DataValidationUtil#isAllowedSemiStructuredType}.
    *
@@ -433,6 +469,43 @@ class DataValidationUtil {
           insertRowIndex);
     }
     return output;
+  }
+
+  /**
+   * Validates and parses input for ARRAY columns, returning a native Java List instead of a JSON
+   * string. For String inputs this avoids the serialize→re-parse roundtrip of {@link
+   * #validateAndParseArray}.
+   *
+   * @param input Object to validate
+   * @param insertRowIndex
+   * @return Native Java List
+   */
+  @SuppressWarnings("unchecked")
+  static List<Object> validateAndParseArrayAsList(
+      String columnName, Object input, long insertRowIndex) {
+    JsonNode jsonNode =
+        validateAndParseSemiStructuredAsJsonTree(columnName, input, "ARRAY", insertRowIndex);
+
+    if (!jsonNode.isArray()) {
+      jsonNode = objectMapper.createArrayNode().add(jsonNode);
+    }
+
+    String output = jsonNode.toString();
+    int stringLength = output.getBytes(StandardCharsets.UTF_8).length;
+    if (stringLength > MAX_SEMI_STRUCTURED_LENGTH) {
+      throw valueFormatNotAllowedException(
+          columnName,
+          "ARRAY",
+          String.format(
+              "Array too large. length=%d maxLength=%d", stringLength, MAX_SEMI_STRUCTURED_LENGTH),
+          insertRowIndex);
+    }
+    try {
+      return objectMapper.treeToValue(jsonNode, List.class);
+    } catch (JsonProcessingException e) {
+      // Should never happen: node was already validated by validateAndParseSemiStructuredAsJsonTree
+      throw new IllegalStateException("Failed to convert validated JsonNode to List", e);
+    }
   }
 
   /**
@@ -682,6 +755,52 @@ class DataValidationUtil {
               insertRowIndex, columnName, offsetDateTime));
     }
     return new TimestampWrapper(offsetDateTime, scale);
+  }
+
+  /**
+   * Validates a timestamp value and returns an ISO-formatted string. Unlike {@link
+   * #validateAndParseTimestamp} (which returns a {@link TimestampWrapper} for Parquet
+   * serialization), this method returns a human-readable ISO string suitable for passing to the
+   * SSv2 SDK.
+   *
+   * <p>This is used by RowValidator to normalize Integer/Long epoch values into unambiguous ISO
+   * strings, so the Snowflake backend interprets them correctly regardless of channel timezone.
+   *
+   * <p>Note: Unlike {@link #validateAndParseTimestamp}, this method omits the {@code scale}
+   * parameter because it only handles Integer/Long epoch inputs which have no fractional seconds.
+   *
+   * @param columnName Column name, used in error messages
+   * @param input Timestamp value (Integer, Long, String, or java.time.* object)
+   * @param defaultTimezone Timezone for inputs without timezone info
+   * @param trimTimezone true for TIMESTAMP_NTZ (strip timezone), false for LTZ/TZ
+   * @param insertRowIndex Row index for error messages
+   * @return ISO timestamp string (e.g., "2024-01-15T10:00" for NTZ, "2024-01-15T10:00Z" for LTZ)
+   */
+  static String validateAndFormatTimestamp(
+      String columnName,
+      Object input,
+      ZoneId defaultTimezone,
+      boolean trimTimezone,
+      long insertRowIndex) {
+    if (input instanceof Integer || input instanceof Long) {
+      input = input.toString();
+    }
+
+    OffsetDateTime offsetDateTime =
+        inputToOffsetDateTime(columnName, "TIMESTAMP", input, defaultTimezone, insertRowIndex);
+
+    if (trimTimezone) {
+      offsetDateTime = offsetDateTime.withOffsetSameLocal(ZoneOffset.UTC);
+    }
+    if (offsetDateTime.getYear() < 1 || offsetDateTime.getYear() > 9999) {
+      throw new SFExceptionValidation(
+          ErrorCode.INVALID_VALUE_ROW,
+          String.format(
+              "Timestamp out of representable inclusive range of years between 1 and 9999,"
+                  + " rowIndex:%d, column:%s, value:%s",
+              insertRowIndex, columnName, offsetDateTime));
+    }
+    return trimTimezone ? offsetDateTime.toLocalDateTime().toString() : offsetDateTime.toString();
   }
 
   /**
@@ -1120,7 +1239,8 @@ class DataValidationUtil {
    * @param rowIndex Index of the Input row primarily for debugging purposes.
    * @return SFExceptionValidation is thrown
    */
-  private static SFExceptionValidation valueFormatNotAllowedException(
+  // Package-private: used by RowValidator for consistent error formatting
+  static SFExceptionValidation valueFormatNotAllowedException(
       String columnName, String snowflakeType, String reason, final long rowIndex) {
     return new SFExceptionValidation(
         ErrorCode.INVALID_VALUE_ROW,
--- a/src/main/java/com/snowflake/kafka/connector/internal/validation/RowValidator.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/validation/RowValidator.java
@@ -6,13 +6,18 @@
 
 package com.snowflake.kafka.connector.internal.validation;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.snowflake.kafka.connector.Utils;
+import java.math.BigDecimal;
 import java.time.ZoneId;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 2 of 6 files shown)


sfc-gh-ppopov:
* **#1433** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1433?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1419** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1419?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1419?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

src/main/java/com/snowflake/kafka/connector/internal/validation/RowValidator.java
```diff
@@ -138,8 +145,27 @@ private Object validateAndNormalizeColumnValue(ColumnSchema col, Object value)
...
     switch (col.getLogicalType()) {
       case BOOLEAN:
-        DataValidationUtil.validateAndParseBoolean(col.getName(), value, insertRowIndex);
-        break;
+        // Normalize to Boolean so the SSv2 SDK receives an unambiguous type.
+        // The SDK only accepts Boolean for BOOLEAN columns; Integer/Long/String inputs
+        // would be silently dropped without this normalization (same pattern as BINARY).
+        //
+        // DataValidationUtil.validateAndParseBoolean accepts any Number (non-zero → 1),
+        // but SSv1 SDK only accepts 0/1 at the row buffer level — other integers (42, -1, 999)
```
sfc-gh-ppopov: **This claim is incorrect.** SSv1 SDK does NOT reject non-0/1 numeric values for BOOLEAN.

I verified against the SSv1 SDK source (`DataValidationUtil.java:1088-1092`):
```java
} else if (input instanceof Number) {
  return new BigDecimal(input.toString()).compareTo(BigDecimal.ZERO) == 0 ? 0 : 1;
}
```
Then in `SnowflakeParquetValueParser.java:49-53`:
```java
int intValue = DataValidationUtil.validateAndParseBoolean(...);
value = intValue > 0;  // 1 → true
```

SSv1 accepts **any** numeric value: non-zero → 1 (TRUE), zero → 0 (FALSE). The SSv1 test suite (`DataValidationUtilTest.java:1267-1304`) explicitly tests `1.1`, `-1.1`, `-10`, `10` — all accepted and normalized to `1`.

**This pre-rejection guard makes KCv4-compat STRICTER than KCv3/SSv1.** A user who sent `Integer(42)` to a BOOLEAN column in KCv3 got TRUE (via SSv1 coercion). With this PR, the same input is rejected — that's a behavioral regression, not parity.

**Recommendation:** Remove the `d != 0.0 && d != 1.0` guard entirely. Keep the normalization to `Boolean` (line 167) — that part is correct and necessary because SSv2 SDK only accepts `Boolean`. The correct SSv1-matching behavior is: any non-zero number → TRUE, zero → FALSE, via the existing `validateAndParseBoolean` result.

src/main/java/com/snowflake/kafka/connector/internal/validation/RowValidator.java
```diff
@@ -154,6 +180,26 @@ private Object validateAndNormalizeColumnValue(ColumnSchema col, Object value)
 
       case TEXT:
       case CHAR:
+        // SSv1/SSv2 SDKs serialize complex objects (Map, List) to JSON strings for VARCHAR
```
sfc-gh-ppopov: **This comment is factually incorrect.** SSv1 SDK does NOT serialize Map/Collection to JSON strings for VARCHAR columns.

I verified against the KC's own copy of `DataValidationUtil.validateAndParseString` (lines 704-716):
```java
} else {
  throw typeNotAllowedException(
      columnName, input.getClass(), "STRING",
      new String[] {"String", "Number", "boolean", "char"}, insertRowIndex);
}
```

SSv1 **rejects** Map/Collection for VARCHAR with `typeNotAllowedException`. The allowed types are only String, Number, boolean, and char. There is no Jackson/NDJSON serialization for VARCHAR in either SDK — Jackson is only used for semi-structured types (VARIANT/ARRAY/OBJECT). A search for "NDJSON" across the entire SSv1 SDK returns zero results.

The E2E test (`xtype_obj_str`) does show KCv3 successfully "coercing" Map→VARCHAR to `'{"key":"value"}'`, which means the serialization must happen in the **KC v3 pipeline** (record conversion layer) before the data reaches the SSv1 SDK — not inside the SDK itself.

The normalization behavior is likely correct for **KC end-to-end parity**, but the comment is misleading. Consider updating to something like:

> KC v3 serializes Map/Collection to JSON before passing to the SSv1 SDK (the SDK itself rejects these types for VARCHAR). RowValidator replicates that KC-level serialization to maintain v3 parity.

src/main/java/com/snowflake/kafka/connector/internal/validation/RowValidator.java
```diff
@@ -179,26 +227,73 @@ private Object validateAndNormalizeColumnValue(ColumnSchema col, Object value)
         break;
 
       case TIMESTAMP_NTZ:
-        validateTimestamp(col, value, insertRowIndex, true);
+        if (value instanceof Integer || value instanceof Long) {
```
sfc-gh-ppopov: Can this if be moves inside `validateTimestamp` function. The code repeats 3 times

sfc-gh-ppopov: we could keep the function below

src/main/java/com/snowflake/kafka/connector/internal/validation/RowValidator.java
```diff
@@ -138,8 +141,29 @@ private Object validateAndNormalizeColumnValue(ColumnSchema col, Object value)
...
+        //
+        // Although SSv1 SDK's DataValidationUtil.validateAndParseBoolean accepts any Number
+        // (non-zero → 1, zero → 0 via BigDecimal.compareTo), in KC v3 the value never reaches
+        // SSv1 as a Number — the record mapper (StreamingRecordMapper.getTextualValue) converts
+        // all values to Strings first. SSv1's convertStringToBoolean only accepts
+        // "0"/"1"/"true"/"false"/"yes"/"no"/"on"/"off", so numeric strings like "42" are rejected.
+        // We pre-reject non-0/1 Numbers here to match end-to-end KC v3 behavior.
+        if (value instanceof Number && !(value instanceof Boolean)) {
+          double d = ((Number) value).doubleValue();
+          if (d != 0.0 && d != 1.0) {
```
sfc-gh-ppopov: IS there a better way to check this without comparing floats?

src/main/java/com/snowflake/kafka/connector/internal/validation/RowValidator.java
sfc-gh-ppopov: The code comment has been updated to clarify. The guard is for KC v3 end-to-end parity, not SSv1 SDK behavior in isolation. In KC v3, StreamingRecordMapper.getTextualValue() stringifies all values before SSv1 sees them, and SSv1's convertStringToBoolean rejects strings like "42". We pre-reject non-0/1 Numbers to match that end-to-end behavior.

src/main/java/com/snowflake/kafka/connector/internal/validation/RowValidator.java
sfc-gh-ppopov: The code comment has been updated. It now correctly attributes the serialization to KC v3's StreamingRecordMapper.getTextualValue(), not to SSv1 SDK. The SSv1 SDK does reject Map/Collection for VARCHAR — the point is that KC v3's record mapper serializes them to JSON strings before SSv1 ever sees them, so we replicate that pipeline-level behavior.

src/main/java/com/snowflake/kafka/connector/internal/validation/RowValidator.java
sfc-gh-ppopov: Done — extracted a `validateAndNormalizeTimestamp` helper in RowValidator that handles the Integer/Long dispatch. Also collapsed TIMESTAMP_LTZ and TIMESTAMP_TZ into a single fall-through case since they're identical.

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
```diff
@@ -684,6 +747,49 @@ static TimestampWrapper validateAndParseTimestamp(
...
+   * strings, so the Snowflake backend interprets them correctly regardless of channel timezone.
+   *
+   * @param columnName Column name, used in error messages
+   * @param input Timestamp value (Integer, Long, String, or java.time.* object)
+   * @param defaultTimezone Timezone for inputs without timezone info
+   * @param trimTimezone true for TIMESTAMP_NTZ (strip timezone), false for LTZ/TZ
+   * @param insertRowIndex Row index for error messages
+   * @return ISO timestamp string (e.g., "2024-01-15T10:00" for NTZ, "2024-01-15T10:00Z" for LTZ)
+   */
+  static String validateAndFormatTimestamp(
```
sfc-gh-bmikaili, line 779: nit: This method omits the `scale` parameter that `validateAndParseTimestamp` accepts. While Integer/Long epochs don't have fractional seconds so scale doesn't matter today, this creates an API asymmetry — if a future caller passes a String timestamp with fractional seconds through this path, there's no scale validation. Consider adding a `scale` parameter for consistency with the sibling method, or a comment noting why it's intentionally omitted.

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
```diff
@@ -273,6 +273,37 @@ static String validateAndParseVariant(String columnName, Object input, long inse
...
+    int stringLength = output.getBytes(StandardCharsets.UTF_8).length;
+    if (stringLength > MAX_SEMI_STRUCTURED_LENGTH) {
+      throw valueFormatNotAllowedException(
+          columnName,
+          "VARIANT",
+          String.format(
+              "Variant too long: length=%d maxLength=%d", stringLength, MAX_SEMI_STRUCTURED_LENGTH),
+          insertRowIndex);
+    }
+    return objectMapper.convertValue(node, Object.class);
```
sfc-gh-bmikaili: nit: Triple conversion here: parse to JsonNode → serialize to String for size check → `convertValue` to Object. `treeToValue(node, Object.class)` would avoid one intermediate serialization cycle. Also, `convertValue`'s type mapping for large numbers may differ from SSv1's behavior (e.g., Long vs BigInteger for values > Long.MAX_VALUE). Low risk given typical VARIANT payloads, but worth a comment.

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
```diff
@@ -1120,7 +1226,7 @@ private static SFExceptionValidation typeNotAllowedException(
    * @param rowIndex Index of the Input row primarily for debugging purposes.
    * @return SFExceptionValidation is thrown
    */
-  private static SFExceptionValidation valueFormatNotAllowedException(
+  static SFExceptionValidation valueFormatNotAllowedException(
```
sfc-gh-bmikaili, line 1243: nit: Visibility widened from `private` to package-private so `RowValidator` can call it. Consider adding a brief comment (e.g., `/* package-private for RowValidator */`) to signal this is intentional, so future readers don't tighten it back.

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
sfc-gh-ppopov: Added a note to the Javadoc explaining that scale is intentionally omitted since this method only handles Integer/Long epoch inputs which have no fractional seconds.

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
sfc-gh-ppopov: Switched to treeToValue for both validateAndParseVariantAsObject and validateAndParseArrayAsList — thanks for catching that.

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
sfc-gh-ppopov: Added a comment above the method to signal the visibility widening is intentional.

sfc-gh-bmikaili (approved):
LGTM

sfc-gh-ppopov:
### Merge activity

* **Apr 15, 11:15 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1419).
* **Apr 15, 11:16 AM UTC**: @sfc-gh-ppopov merged this pull request with [Graphite](https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1419).

PR #1423, opened 2026-04-10
SNOW-3361266 fix race condition in channel recovery

sfc-gh-skurella:
Issue originally identified in #1397.

## Problem

When a non-retryable `SFException` is thrown by `appendRow` mid-batch, `reopenChannel()` starts an async recovery on the IO executor. The task thread continues processing the next record in the same batch. This creates a race window where `shouldProcess()` and `resetAfterRecovery()` are not synchronised, causing `processedOffset` to advance past the recovery reset and subsequent re-delivered records to be silently dropped.

```mermaid
sequenceDiagram
    participant Task as Task Thread
    participant IO as IO Thread
    Note over Task: Batch: records 6, 7, 8
    Task->>Task: insertRecord(7) -> appendRow fails
    Task->>IO: reopenChannel() (async)
    Task->>Task: insertRecord(8) -> shouldProcess(8)=true (race!)
    IO->>IO: resetAfterRecovery(5) -> processedOffset=5
    Task->>Task: getChannel().join() blocks, then unblocks
    Task->>Task: appendRow(8) on new channel -> OK
    Task->>Task: recordProcessed(8) -> processedOffset=8 (OVERWRITES 5!)
    Note over Task: Re-delivered record 6: shouldProcess(6) -> 6 < 8+1 -> SKIP -> DATA LOSS
```

With the fix, after `insertRecord(7)` returns `false`, the task thread adds `tp -> 7` to `offsetsOfFirstSkippedRecord` and skips record 8 entirely. No `recordProcessed(8)` ever runs.

## Fix

Whenever a channel fails insertion and needs to be recovered, we propagate this information up to `SnowflakeSinkServiceV2`'s `insert` method so that all remaining records within this batch are skipped.

## Cross-batch safety

The same race could theoretically occur across batch boundaries: batch N triggers recovery, and batch N+1 arrives before recovery completes. However, this case is already handled by the existing `isInitializing()` check. `SnowpipeStreamingPartitionChannel.isInitializing()` returns `!channel.isDone()` — during recovery, `reopenChannel()` replaces the channel future with a new one that won't complete until the new channel is opened and `resetAfterRecovery` has run. The batch loop in `insert(Collection)` skips all initializing partitions and rewinds their offsets, so no `shouldProcess` / `recordProcessed` can run while recovery is in flight.

- If recovery is **still in flight** when the next batch starts: the partition is skipped (initializing), offsets are rewound, no race.
- If recovery has **already completed**: `resetAfterRecovery` has set `processedOffset` to the committed offset and rewound Kafka, so the batch receives correct records and `processedOffset` is authoritative.

## Additional change

The recovery logic in `insert` was getting complex so I also simplified the backpressure recovery to simply skip subsequent records across all partitions rather than just in the current one - this behavior matches the fact that we skip records in all partitions in *subsequent* batches.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkService.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkService.java
@@ -37,8 +37,10 @@ public interface SnowflakeSinkService {
    * call pipe to insert a JSON record will not trigger time based flush
    *
    * @param record record content
+   * @return true if the record was processed successfully, false if recovery was triggered and the
+   *     caller should stop feeding records to this partition for the remainder of the batch
    */
-  void insert(final SinkRecord record);
+  boolean insert(final SinkRecord record);
 
   /**
    * retrieve offset of last loaded record for given pipe name
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
@@ -356,11 +356,11 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
 
     // If still in cooldown from a recent backpressure event, treat all partitions as
     // backpressured so we skip the entire batch and give the SDK time to drain.
-    Set<TopicPartition> backpressuredPartitions = new HashSet<>();
+    boolean skipAllPartitions = false;
     if (Instant.now().isBefore(backpressureUntil)) {
       LOGGER.debug(
           "Backpressure cooldown active until {}. Skipping entire batch.", backpressureUntil);
-      backpressuredPartitions.addAll(partitions);
+      skipAllPartitions = true;
     }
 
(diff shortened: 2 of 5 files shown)


sfc-gh-skurella:
* **#1427** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1427?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1423** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1423?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1423?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sfc-gh-ppopov:
@sfc-gh-skurella this fix is more simple way to address the problem within a single batch. I would like us to check that it also addressed the case when Task Thread finished the batch and start the next one, while recovery is happening. Let's discuss.

sfc-gh-ppopov (approved):

PR #1428, opened 2026-04-13
Add ClientRecreationException and ClientRecreator interface

sfc-gh-bmikaili:
## Summary
- `ClientRecreationException`: unchecked exception wrapping `SFException` for client-invalid error codes (`InvalidClientError`, `SfApiPipeFailedOverError`, `ClosedClientError`). Static `isClientInvalidError()` classifies without instantiation; static `wrap(Throwable)` safely casts and wraps.
- `ClientRecreator`: `@FunctionalInterface` for replacing an invalid SDK client via CAS semantics.
- 11 unit tests covering all error code detection, constructor validation, wrap helper, and negative cases.

No callers yet — these are consumed in subsequent PRs in this stack.

**JIRA**: SNOW-3360429

## Test plan
- [x] `ClientRecreationExceptionTest` — 11 tests all pass

PR #1429, opened 2026-04-13
Add CAS client recreation to StreamingClientPool

sfc-gh-bmikaili:
## Summary

- **`StreamingClientPool.recreateClient()`** — atomically replaces the `RefCountedClient` entry only if the current resolved client is identity-equal to the given invalid client (CAS semantics). Preserves task registrations via `copyTasksFrom()`. Old client is closed best-effort **outside** the `compute()` lock to avoid blocking the `ConcurrentHashMap` bucket during I/O. Failed replacement-client creation evicts the entry so the next caller retries from scratch.
- **`StreamingClientPools.getClient()`** — retry loop (3 attempts) on client-invalid errors during initial client creation. Exhausted retries wrap the underlying `SFException` via `ClientRecreationException.wrap()` so the batch loop (PR #1431) can rewind Kafka offsets instead of crashing the task.
- **`StreamingClientPools.recreateClient()`** — static delegate to the per-connector `StreamingClientPool.recreateClient()` with the same retry+wrap pattern.
- Concurrent callers with the same invalid-client reference produce exactly one replacement — validated by a latch-coordinated test.

**JIRA**: SNOW-3360429

## Test plan

- [x] `StreamingClientPoolTest` — 25 tests (8 new): CAS replaces entry + preserves task registrations, old client is closed, no-op when already replaced, `getClient` returns the new client after recreation, concurrent callers create exactly one replacement, failed replacement evicts the entry, retry exhaustion wraps as `ClientRecreationException`
- [x] `StreamingClientPoolsTest` — 2 tests (1 new): `getClient` retries on client-invalid then wraps on exhaustion

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPool.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPool.java
@@ -8,8 +8,10 @@ import com.snowflake.kafka.connector.internal.streaming.StreamingClientPropertie
 import com.snowflake.kafka.connector.internal.streaming.v2.service.ThreadPools;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * Manages clients for a single connector. Tracks which tasks use which pipes and only closes
@@ -75,6 +77,11 @@ public class StreamingClientPool {
 
     int taskCount() {
       return taskIds.size();
+    }
+
+    /** Copies all task registrations from another entry into this one. */
+    void copyTasksFrom(RefCountedClient other) {
+      taskIds.addAll(other.taskIds);
     }
 
     void close(String pipeName, String connectorName) {
@@ -158,6 +165,128 @@ public class StreamingClientPool {
             }
             return entry;
           });
+    }
+  }
+
+  /**
+   * Atomically replaces the client for a pipe if the current client matches the given invalid
+   * client. Uses compare-and-swap semantics: if another caller already replaced the entry, the
+   * existing new client is returned without creating a second one.
+   *
+   * @param pipeName the pipe whose client should be replaced
+   * @param invalidClient the client instance that the caller believes is invalid (identity check)
+   * @param config task config for creating the replacement client
+   * @param streamingClientProperties streaming client properties
+   * @param taskMetrics metrics for timing the new client creation
+   * @return the new (or already-replaced) client
+   */
+  SnowflakeStreamingIngestClient recreateClient(
+      final String pipeName,
+      final SnowflakeStreamingIngestClient invalidClient,
+      final SinkTaskConfig config,
+      final StreamingClientProperties streamingClientProperties,
+      final TaskMetrics taskMetrics) {
+
+    // AtomicReference to capture the entry chosen inside the atomic compute() block.
+    // The actual .join() happens outside the lock to avoid blocking other pipes.
+    AtomicReference<RefCountedClient> chosenEntry = new AtomicReference<>();
+
+    pipes.compute(
+        pipeName,
+        (key, current) -> {
+          if (current == null) {
+            LOGGER.warn(
+                "recreateClient called for pipe {} but no entry exists in connector {}."
+                    + " Creating a fresh entry.",
+                pipeName,
+                connectorName);
+            RefCountedClient fresh =
+                new RefCountedClient(
+                    pipeName,
+                    connectorName,
+                    config,
+                    streamingClientProperties,
+                    taskMetrics,
+                    ioExecutor);
+            chosenEntry.set(fresh);
+            return fresh;
+          }
+
+          // Check if the current entry still holds the invalid client (CAS guard).
+          // If someone already replaced it, return the current (new) entry as-is.
+          SnowflakeStreamingIngestClient currentClient;
+          try {
+            currentClient = current.clientFuture.join();
+          } catch (CompletionException e) {
+            // Current entry failed to create — replace it unconditionally.
+            LOGGER.warn(
+                "recreateClient for pipe {}: current entry has a failed client future,"
+                    + " replacing unconditionally",
+                pipeName);
+            RefCountedClient replacement =
+                new RefCountedClient(
+                    pipeName,
+                    connectorName,
+                    config,
+                    streamingClientProperties,
+                    taskMetrics,
+                    ioExecutor);
+            replacement.copyTasksFrom(current);
+            chosenEntry.set(replacement);
+            return replacement;
+          }
+
+          if (currentClient != invalidClient) {
+            LOGGER.info(
+                "recreateClient for pipe {} in connector {}: client already replaced"
+                    + " by another caller, reusing existing entry",
+                pipeName,
+                connectorName);
+            chosenEntry.set(current);
+            return current;
+          }
+
+          // CAS matches — replace with a new entry, preserving task registrations.
+          LOGGER.info(
+              "Recreating streaming client for pipe {} in connector {}."
+                  + " Old client will be closed best-effort.",
+              pipeName,
+              connectorName);
+
+          RefCountedClient replacement =
+              new RefCountedClient(
+                  pipeName,
+                  connectorName,
+                  config,
+                  streamingClientProperties,
+                  taskMetrics,
+                  ioExecutor);
+          replacement.copyTasksFrom(current);
+
+          // Best-effort close of the old (invalid) client for resource cleanup.
+          try {
+            currentClient.close();
+          } catch (Exception e) {
+            LOGGER.warn(
+                "Best-effort close of invalid client for pipe {} failed: {}",
+                pipeName,
+                e.getMessage());
+          }
+
+          chosenEntry.set(replacement);
+          return replacement;
+        });
+
+    try {
+      return chosenEntry.get().clientFuture.join();
+    } catch (CompletionException e) {
+      // If the new client also fails to create, evict the failed entry so the next caller retries.
+      RefCountedClient failedEntry = chosenEntry.get();
+      pipes.compute(pipeName, (key, current) -> current == failedEntry ? null : current);
+      if (e.getCause() instanceof RuntimeException) {
+        throw (RuntimeException) e.getCause();
+      }
+      throw e;
     }
   }
 
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPool.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPool.java
@@ -10,7 +10,10 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 
 /**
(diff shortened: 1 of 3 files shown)


sfc-gh-bmikaili:
* **#1430** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1430?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1431](https://github.com/snowflakedb/snowflake-kafka-connector/pull/1431) <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1431?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1432](https://github.com/snowflakedb/snowflake-kafka-connector/pull/1432) <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1432?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1429** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1429?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1429?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1428** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1428?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPool.java
```diff
@@ -161,6 +168,135 @@ void closeTaskClients(final String taskId) {
...
+      final StreamingClientProperties streamingClientProperties,
+      final TaskMetrics taskMetrics) {
+
+    // AtomicReference to capture the entry chosen inside the atomic compute() block.
+    // The actual .join() happens outside the lock to avoid blocking other pipes.
+    AtomicReference<RefCountedClient> chosenEntry = new AtomicReference<>();
+    // Captured inside compute() so the old client can be closed outside the lock.
+    AtomicReference<SnowflakeStreamingIngestClient> clientToClose = new AtomicReference<>();
+
+    pipes.compute(
```
sfc-gh-skurella: I think we can simplify this:

```suggestion
    RefCountedClient chosenEntry = pipes.compute(
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPool.java
```diff
@@ -161,6 +168,135 @@ void closeTaskClients(final String taskId) {
...
+            RefCountedClient fresh =
+                new RefCountedClient(
+                    pipeName,
+                    connectorName,
+                    config,
+                    streamingClientProperties,
+                    taskMetrics,
+                    ioExecutor);
+            chosenEntry.set(fresh);
+            return fresh;
```
sfc-gh-skurella: I suspect we're missing `fresh.addTask(taskId);` - otherwise, it's created with no tasks at all, which could lead to the client being prematurely evicted during reassignment. Given that this is a set, we can probably be safe and just add it in all code paths in this function.

A nice way to achieve this would be to extract at the top of this method a lambda `createClient(Optional<RefCountedClient> previousClient)` that would both invoke the constructor, optionally copy tasks from the previous client, and finally (always) add our own task ID.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPool.java
```diff
@@ -161,6 +168,135 @@ void closeTaskClients(final String taskId) {
...
+          }
+
+          // Check if the current entry still holds the invalid client (CAS guard).
+          // If someone already replaced it, return the current (new) entry as-is.
+          SnowflakeStreamingIngestClient currentClient;
+          try {
+            currentClient = current.clientFuture.join();
```
sfc-gh-skurella: We ideally shouldn't block in a `compute` supplier. Maybe we could add a timeout=0 here? If the future isn't ready, I wouldn't expect it to contain an invalid client anyway.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPool.java
```diff
@@ -161,6 +168,135 @@ void closeTaskClients(final String taskId) {
...
+      return chosenEntry.get().clientFuture.join();
+    } catch (CompletionException e) {
+      // If the new client also fails to create, evict the failed entry so the next caller retries.
+      RefCountedClient failedEntry = chosenEntry.get();
+      pipes.compute(pipeName, (key, current) -> current == failedEntry ? null : current);
+      if (e.getCause() instanceof RuntimeException) {
+        throw (RuntimeException) e.getCause();
+      }
+      throw e;
+    }
```
sfc-gh-skurella, line 315: Duplication - could you please extract this?

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPools.java
```diff
@@ -52,16 +60,42 @@ public static SnowflakeStreamingIngestClient getClient(
...
+        if (ClientRecreationException.isClientInvalidError(cause)) {
+          if (attempt < MAX_CLIENT_CREATION_RETRIES) {
+            LOGGER.warn(
+                "Client creation for pipe {} failed with client-invalid error (attempt {}/{}): {}."
+                    + " Retrying with a new client.",
+                pipeName,
+                attempt,
+                MAX_CLIENT_CREATION_RETRIES,
+                cause.getMessage());
+            continue;
```
sfc-gh-skurella: We're retrying immediately - it'd be good to have some delay.

Also, I see duplication with `recreateClient`. WDYT about using retry tooling we already use in other places in the code? E.g. https://failsafe.dev/retry/

```java
private static final RetryPolicy<SnowflakeStreamingIngestClient> CLIENT_CREATION_RETRY_POLICY =
    RetryPolicy.<SnowflakeStreamingIngestClient>builder()
        .handleIf(e -> e instanceof CompletionException
          && ClientRecreationException.isClientInvalidError(e.getCause()))
        .withMaxAttempts(3)
        .withDelay(Duration.ofSeconds(10))        // or withBackoff(...)
        .onRetry(event -> LOGGER.warn(
            "Client creation retry attempt #{}: {}",
            event.getAttemptCount(),
            event.getLastException().getMessage()))
        .build();
```

and then

```java
public static SnowflakeStreamingIngestClient getClient(...) {
    try {
        return Failsafe.with(CLIENT_CREATION_RETRY_POLICY)
            .get(() -> getClientAsync(...).join());
    } catch (CompletionException e) {
        // unwrap as before
    }
}
```

sfc-gh-skurella (changes requested):
Thank you! Approach LGTM, modulo minor correctness issue with missing `addTask`. Also really worth adding delays to the retries.

This part of the code is very tricky to reason about so let's be sure to reduce the added LoC and deduplicate as much as possible. I left some tips. They are not blocking.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPool.java
sfc-gh-bmikaili: Done — simplified to return `pipes.compute(...)` directly.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPool.java
sfc-gh-bmikaili: Done — extracted `createReplacement(taskId, pipeName, previous, ...)` helper that always calls `addTask(taskId)` in every code path (no entry, failed-future, and CAS-match paths all go through it).

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPool.java
sfc-gh-bmikaili: Done — replaced the `.join()` inside `compute()` with `future.get(0, MILLISECONDS)` and handle `TimeoutException` as "replacement already in flight" (addTask + reuse the current entry), per your reasoning that an incomplete future can't be the invalid client the caller just observed.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPool.java
sfc-gh-bmikaili: Done — extracted `joinAndEvictOnFailure(pipeName, entry)` helper, used by both `getClientAsync` (indirectly via its failure handler) and `recreateClient`.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPools.java
sfc-gh-bmikaili: Addressed in two steps:

1. **Dropped the `getClient` retry entirely** — after tracing, the SDK only throws `ClientInvalidError` from `openChannel`/`appendRow`, not from client construction (`newClientFFI` succeeds even if the server is mid-failover). So the retry loop on `getClient` was dead code. Removed + the two tests that exercised it.

2. **Converted `recreateClient` retry to Failsafe** with a 5s delay, per your suggestion:

```java
RetryPolicy.<SnowflakeStreamingIngestClient>builder()
    .handleIf(e -> e instanceof RuntimeException
        && ClientRecreationException.isClientInvalidError(e))
    .withMaxAttempts(3)
    .withDelay(Duration.ofSeconds(5))
    .onRetry(event -> LOGGER.warn(...))
    .onRetriesExceeded(event -> LOGGER.error(...))
    .build();
```

E2E (`test_client_recreation.py`) still passes with the slimmer version.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPools.java
```diff
@@ -29,6 +34,13 @@ public class StreamingClientPools {
 
   private StreamingClientPools() {}
 
+  /**
+   * Maximum retry attempts when a replacement client also fails with a client-invalid error during
+   * {@link #recreateClient}. Three attempts provide enough headroom for transient failover windows
+   * while keeping total blocking time bounded (each attempt creates a fresh SDK client).
+   */
+  private static final int MAX_CLIENT_CREATION_RETRIES = 3;
```
sfc-gh-skurella: NIT: should probably be defined next to `CLIENT_CREATION_RETRY_DELAY`

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPools.java
```diff
@@ -104,6 +116,88 @@ public static long getClientCountForTask(final String connectorName, final Strin
...
+   * @param taskId the ID of the task requesting recreation; registered on the replacement entry so
+   *     the pool does not prematurely evict it on task-local cleanup
+   * @param pipeName the pipe whose client should be replaced
+   * @param invalidClient the client instance the caller believes is invalid (identity check)
+   * @param config task config for creating the replacement client
+   * @param streamingClientProperties streaming client properties
+   * @param taskMetrics metrics for timing the new client creation
+   * @return the new (or already-replaced) client
+   */
+  public static SnowflakeStreamingIngestClient recreateClient(
```
sfc-gh-skurella, line 127: Since this can get called from multiple TopicChannelPartitions, it's possible that the client will be recreated more than every 5 seconds, i.e. the retry delay will be essentially bypassed. I think this is OK - the delay basically means that each channel will retry over the span of 15+ seconds, but let's be sure to document this in the code so that we set the correct expectations for whoever is working with logs next time.

sfc-gh-skurella (approved):
LGTM. Maybe let's try and add more retries or a longer delay or an exponential backoff (should be a one-liner) because I'd expect that in the real failover case, it might take some time for the pipe coordinator to start up on the new machine, and I'm not entirely sure what the tail latencies for this would be. 15s feels pretty conservative - I'd aim at a minute or more. But that's not a blocker for this PR, to be sure. Thank you!

sfc-gh-bmikaili:
### Merge activity

* **May 13, 12:58 PM UTC**: @sfc-gh-bmikaili merged this pull request with [Graphite](https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1429).

PR #1430, opened 2026-04-13
Add client recreation recovery with batch-level pipe skip

sfc-gh-bmikaili:
## Summary

Adds SDK pipe-failover (HTTP 410) recovery on top of #1460. The seek-overwrite race that caused silent loss is fixed in #1460; this PR is now strictly the client-recreation pieces.

## Recovery flow

```
appendRow → InvalidClientError
  → AppendRowWithFallbackPolicy invokes reopenChannel(reason, recreateClient=true)
  → StreamingClientPool.recreate (CAS: only one recreation wins)
  → close old channel → open new → resetAfterRecovery(committed)
  → sinkTaskContext.offset(partition, committed + 1)   [Kafka rewind]
  → batch loop receives RECOVERY_IN_FLIGHT, skips its own rewind   (handled in #1460)
  → task stays RUNNING; next put() replays from committed + 1
```

## What's in this PR

- `reopenChannel(reason, recreateClient)` — `recreateClient=true` swaps the shared SDK client (via `StreamingClientPool` CAS) before the existing reopen chain.
- Fallback routes `InvalidClientError` to `recreateClient=true`; other errors keep the plain channel-reopen path.
- `MAX_CONSECUTIVE_RECOVERIES` 5 → 30 + `FALLBACK_DELAY` 500ms → 5s, sized for real SSv2 pipe failovers (1-3 min propagation window).
- `BatchOffsetFetcher` skips offset fetch on client-invalid (recreation kicks in on next `put()`).
- `flush()` from `stop()` swallows client-invalid; uncommitted rows replay on task restart.
- `PartitionChannelManager` wires per-pipe `ClientRecreator` lambda into each partition channel.
- Telemetry: `clientRecreationCount` per channel, `clientRecreationSkipCount` per task.
- `ClientRecreationIT` integration test.

## Test plan

- [x] Unit tests pass (`AppendRowWithFallbackPolicyTest`, `SnowpipeStreamingPartitionChannelTest`, `SnowflakeSinkServiceV2Test`)
- [x] Integration test: `ClientRecreationIT`
- [x] E2E test (in #1432): mitmproxy injects 410 → 900/900 records land, zero Kafka-offset gaps.

## Stacked on

- #1460 (PR-A) — seek-overwrite race fix.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithFallbackPolicy.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithFallbackPolicy.java
@@ -1,5 +1,6 @@
 package com.snowflake.kafka.connector.internal.streaming.v2;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.snowflake.ingest.streaming.SFException;
 import com.snowflake.kafka.connector.internal.KCLogger;
 import dev.failsafe.Failsafe;
@@ -12,37 +13,49 @@ import java.time.Duration;
  * fallback functionality.
  *
  * <p>This class provides a clean interface to execute append row operations with automatic channel
- * recovery on non-retryable {@link SFException}. For retryable backpressure errors, it throws
- * {@link BackpressureException} to signal the batch-level insert loop to abandon the batch and
- * rewind offsets.
+ * recovery on non-retryable {@link SFException}. The fallback supplier determines the recovery
+ * strategy:
+ *
+ * <ul>
+ *   <li>Retryable backpressure errors throw {@link BackpressureException}
+ *   <li>Client-invalid errors trigger client recreation via the fallback, which throws {@link
+ *       ClientRecreationException}
+ *   <li>Other errors trigger channel-level recovery (reopen channel)
+ * </ul>
+ *
+ * Both {@link BackpressureException} and {@link ClientRecreationException} propagate out of
+ * Failsafe to signal the batch-level insert loop.
  */
 class AppendRowWithFallbackPolicy {
 
   private static final KCLogger LOGGER = new KCLogger(AppendRowWithFallbackPolicy.class.getName());
 
-  /** Delay before fallback attempt (channel reopening). */
-  private static final Duration FALLBACK_DELAY = Duration.ofMillis(500);
+  /**
+   * Initial delay before the first recovery attempt. Backoff doubles each attempt up to {@link
+   * #MAX_DELAY}. Package-private + non-final so tests can override.
+   */
+  @VisibleForTesting static Duration BASE_DELAY = Duration.ofSeconds(1);
 
-  /** Random jitter added to fallback delays to prevent retry storms. */
-  private static final Duration JITTER_DURATION = Duration.ofMillis(200);
+  /** Cap on the exponential backoff delay. */
+  @VisibleForTesting static Duration MAX_DELAY = Duration.ofSeconds(30);
+
+  /** Jitter factor (±) applied to each delay to prevent retry storms across partitions. */
+  @VisibleForTesting static double JITTER_FACTOR = 0.2;
 
   /**
-   * Executes the given action after a delay with jitter to prevent retry storms.
+   * Computes an exponential backoff delay for the given consecutive attempt number, capped at
+   * {@link #MAX_DELAY} with ±{@link #JITTER_FACTOR} jitter. Mirrors the formula used by Failsafe's
+   * {@code RetryPolicy.withBackoff(...).withJitter(...)}.
    *
-   * @param action the action to execute after the delay
-   * @param channelName the channel name for logging purposes
+   * @param attempt 1-based consecutive recovery attempt number
    */
-  private static void withDelay(CheckedRunnable action, String channelName) throws Throwable {
-    long delayMs = FALLBACK_DELAY.toMillis() + (long) (Math.random() * JITTER_DURATION.toMillis());
-    LOGGER.info("Delaying channel recovery by {}ms for channel: {}", delayMs, channelName);
-    try {
-      Thread.sleep(delayMs);
-    } catch (InterruptedException e) {
-      Thread.currentThread().interrupt();
-      return;
-    }
-    LOGGER.info("Executing channel recovery for channel: {}", channelName);
-    action.run();
+  static Duration computeBackoffDelay(int attempt) {
+    long baseMs = BASE_DELAY.toMillis();
+    long capMs = MAX_DELAY.toMillis();
+    long shift = Math.min(Math.max(attempt - 1, 0), 30); // bound to avoid long overflow
+    long delayMs = Math.min(baseMs << shift, capMs);
+    long jitterMs = (long) ((Math.random() * 2 - 1) * delayMs * JITTER_FACTOR);
+    return Duration.ofMillis(Math.max(0, delayMs + jitterMs));
   }
 
   /**
@@ -52,15 +65,12 @@ class AppendRowWithFallbackPolicy {
    * to signal the batch-level insert loop that the batch should be abandoned and offsets should be
    * rewound. The channel remains valid.
    *
-   * <p>On non-retryable {@link SFException}, it will execute the fallback supplier to reopen the
-   * channel and reset offsets after a simple blocking delay with jitter to prevent retry storms.
+   * <p>On non-retryable {@link SFException}, executes the fallback supplier to reopen the channel
+   * and reset offsets. The supplier owns its own backoff delay (see {@link #computeBackoffDelay}).
    *
    * @param appendRowAction the action to execute (typically channel.appendRow call)
-   * @param fallbackSupplier the fallback action to execute on non-retryable failure (channel
-   *     reopening logic)
+   * @param fallbackSupplier the fallback action to execute on non-retryable failure
    * @param channelName the channel name for logging purposes
-   */
-  /**
    * @return true if the append row action succeeded normally, false if the fallback was executed
    *     (meaning the record was NOT inserted). When this returns false, callers must NOT advance
    *     processedOffset — the fallback's recovery logic has already reset offset state.
@@ -84,9 +94,12 @@ class AppendRowWithFallbackPolicy {
                     throw new BackpressureException((SFException) lastException);
                   }
 
-                  // Non-retryable error: proceed with channel reopening
+                  // Non-retryable error: proceed with channel reopening. The fallback supplier is
+                  // responsible for any backoff delay before reopen — see callers'
+                  // computeBackoffDelay usage. Done here in the supplier (not the policy) because
+                  // the supplier owns the consecutive-attempt counter that drives the backoff.
                   succeeded[0] = false;
-                  withDelay(() -> fallbackSupplier.execute(lastException), channelName);
+                  fallbackSupplier.execute(lastException);
                 })
             .handle(SFException.class)
             .onFailedAttempt(
@@ -100,6 +113,11 @@ class AppendRowWithFallbackPolicy {
                   if (event.getException() instanceof BackpressureException) {
                     LOGGER.warn(
                         "Backpressure on channel {}: {}",
+                        channelName,
+                        event.getException().getMessage());
+                  } else if (event.getException() instanceof ClientRecreationException) {
+                    LOGGER.warn(
+                        "Client recreation triggered on channel {}: {}",
                         channelName,
                         event.getException().getMessage());
                   } else {
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
@@ -67,15 +67,26 @@ public class SnowpipeStreamingPartitionChannel implements TopicPartitionChannel 
   // are cumulative and don't reset when a channel is reopened.
   private long initialErrorCount = 0;
 
-  /** Max consecutive channel recoveries before giving up and letting the task fail. */
-  private static final int MAX_CONSECUTIVE_RECOVERIES = 5;
+  /**
+   * Maximum wall-clock duration we keep retrying recovery before giving up. Sized for real SSv2
+   * pipe failover propagation (observed 2-4 min). Package-private + non-final so tests can
+   * override.
+   */
+  @VisibleForTesting static Duration MAX_RECOVERY_DURATION = Duration.ofMinutes(6);
 
   /**
    * Consecutive recovery counter. Incremented each time the fallback reopens the channel, reset to
-   * zero on every successful appendRow. If this reaches {@link #MAX_CONSECUTIVE_RECOVERIES} the
-   * fallback re-throws to let the KC framework kill the task.
+   * zero on every successful appendRow. Drives the exponential backoff in {@link
+   * AppendRowWithFallbackPolicy#computeBackoffDelay(int)} — it is no longer used as the giving-up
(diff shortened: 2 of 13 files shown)


sfc-gh-bmikaili:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1430?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#1430** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1430?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1431](https://github.com/snowflakedb/snowflake-kafka-connector/pull/1431) <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1431?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1432](https://github.com/snowflakedb/snowflake-kafka-connector/pull/1432) <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1432?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1430?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1460** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1460?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

sfc-gh-bmikaili:
## Client Recreation E2E: SDK Error Propagation Gap

### Goal
When the SDK's streaming client becomes invalid (pipe failover → HTTP 410), the connector should recreate the client and recover without task crash.

### What we discovered
The SDK has two parallel invalidation paths when 410 hits `insert_rows`:

1. **Client path**: `client_error_tx` → `invalidate_client()` → `is_valid = false`
2. **Channel path**: error propagates through `snowflake_informer` → `invalidate_channel()` → `is_locally_valid = false`

When Java calls `appendRow()`, the channel check fires first: `check_channel_valid()` sees `is_locally_valid = false` and throws `SFException("InvalidChannelError")`. The client check (`check_client_valid()`) never runs because the channel check fails first.

### The problem
The connector classifies errors by `SFException.getErrorCodeName()`:
- `InvalidClientError` / `SfApiPipeFailedOverError` / `ClosedClientError` → client recreation
- `InvalidChannelError` → channel reopen

A pipe failover invalidates both, but `appendRow` only surfaces `InvalidChannelError`. The connector reopens the channel, which succeeds — but the new channel immediately fails again because the client is still invalid. This loops until the task crashes.

### Options

**A. SDK change: check client validity before channel validity in `appendRow`**
If `check_client_valid()` ran before `check_channel_valid()`, the Java side would see `InvalidClientError` and trigger client recreation directly. This is a one-line reorder in the Rust `append_bytes_internal()`. Clean, correct, no connector changes needed.

**B. SDK change: expose `isValid()` on the Java client interface**
The connector could check `streamingClient.isValid()` during the channel-reopen fallback and escalate to client recreation if the client is also invalid. Small SDK change, small connector change.

**C. Connector-only: escalate after N consecutive channel recoveries**
After `MAX_CONSECUTIVE_RECOVERIES` channel reopens fail, attempt client recreation instead of crashing. No SDK change. Pragmatic but can't distinguish "channel is genuinely broken" from "client is invalid" — though in practice, client recreation is harmless if the client is healthy (CAS miss).

---
Discovered during E2E testing with mitmproxy 410 injection. See also SNOW-3360429.

sfc-gh-bmikaili (commented):
Self-review with per-file WHY notes. Each comment below explains the reviewer-relevant reason for the change on that file.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
```diff
@@ -66,8 +71,13 @@ public class SnowpipeStreamingPartitionChannel implements TopicPartitionChannel
   // are cumulative and don't reset when a channel is reopened.
   private long initialErrorCount = 0;
 
-  /** Max consecutive channel recoveries before giving up and letting the task fail. */
-  private static final int MAX_CONSECUTIVE_RECOVERIES = 5;
+  /**
+   * Max consecutive channel recoveries before failing the task. Each recovery attempt blocks on
+   * AppendRowWithFallbackPolicy.FALLBACK_DELAY (5s + 0-500ms jitter), so 30 attempts covers ~2.5
+   * minutes — enough for real SSv2 pipe failovers (1-3 min) without hammering Snowflake during the
+   * outage.
```
sfc-gh-bmikaili: **WHY (MAX_CONSECUTIVE_RECOVERIES 5 → 30, FALLBACK_DELAY 500ms → 5s):** real SSv2 pipe failovers take 1-3 minutes to fully propagate. The old 5×500ms budget gave up after ~2.5s — far too short. 30×5s covers ~2.5min without hammering Snowflake during the outage. The 5s fallback delay is set in `AppendRowWithFallbackPolicy`.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
```diff
@@ -242,7 +260,21 @@ public CompletableFuture<Void> waitForLastProcessedRecordCommitted() {
         () -> {
           LOGGER.info("Starting flush for channel: {}", this.channelName);
 
-          streamingClient.initiateFlush();
+          try {
+            streamingClient.initiateFlush();
```
sfc-gh-bmikaili, line 266: **WHY (flush swallows client-invalid):** called from `stop()`. If the client is invalid here we can't flush; the correct behavior is to skip, since the next task instance will get a fresh client and Kafka will redeliver uncommitted rows. Without this catch, an already-invalid client would throw out of `initiateFlush` during shutdown — noisy and adds nothing.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
```diff
@@ -298,7 +336,8 @@ private boolean insertRowWithFallback(Map<String, Object> row, long offset) {
               this.channelName,
               consecutiveRecoveryCount,
               MAX_CONSECUTIVE_RECOVERIES);
-          reopenChannel("APPEND_ROW_FALLBACK");
+          boolean clientInvalid = ClientRecreationException.isClientInvalidError(ex);
```
sfc-gh-bmikaili, line 359: **WHY (route client-invalid to `reopenChannel(recreateClient=true)`):** a plain channel reopen can't fix a client-invalid state — the new channel would be opened on the same invalid client. The fallback inspects the error type and tells `reopenChannel` whether to swap the client first. This is the single entry point for both recovery modes.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
```diff
@@ -327,13 +366,23 @@ private static void closeChannelWithoutFlushing(SnowflakeStreamingIngestChannel
...
 
     if (this.snowflakeTelemetryChannelStatus != null) {
       this.snowflakeTelemetryChannelStatus.incRecoveryCount();
     }
 
+    if (recreateClient) {
+      // Client recreation is a higher-level recovery than channel reopen — the fresh client can
+      // unblock a pipe that channel-reopen alone couldn't. The pool's CAS ensures only one new
+      // client is created even if multiple channels call this concurrently. Reset the retry
+      // budget so subsequent channel-reopen attempts on the new client get their full window.
```
sfc-gh-bmikaili, line 400: **WHY (merged `reopenChannel` handles both modes):** previously there was a separate `recreateClientAndReopenChannel`. Since the only difference is one `clientRecreator.recreate()` call before the chain, merging saves a method and keeps the reopen chain in one place. The counter reset (`consecutiveRecoveryCount = 0`) is conditional — client recreation is a higher-level recovery, so subsequent channel-reopen attempts on the new client get their full retry budget.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithFallbackPolicy.java
```diff
@@ -12,19 +13,34 @@
...
  */
 class AppendRowWithFallbackPolicy {
 
   private static final KCLogger LOGGER = new KCLogger(AppendRowWithFallbackPolicy.class.getName());
 
-  /** Delay before fallback attempt (channel reopening). */
-  private static final Duration FALLBACK_DELAY = Duration.ofMillis(500);
+  /**
+   * Delay before fallback attempt (channel reopening / client recreation). Sized for SSv2 pipe
+   * failovers (1-3 min) — with MAX_CONSECUTIVE_RECOVERIES=30 upstream, 5s per attempt covers ~2.5
```
sfc-gh-bmikaili: **WHY (FALLBACK_DELAY 500ms → 5s, JITTER 200ms → 500ms):** see MAX_CONSECUTIVE_RECOVERIES note — sized for real pipe failover duration (1-3 min). The `@VisibleForTesting` on these makes unit tests overridable so they don't actually sleep 5s per retry.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithFallbackPolicy.java
```diff
@@ -109,6 +118,11 @@ static boolean executeWithFallback(
                         "Backpressure on channel {}: {}",
                         channelName,
                         event.getException().getMessage());
+                  } else if (event.getException() instanceof ClientRecreationException) {
+                    LOGGER.warn(
+                        "Client recreation triggered on channel {}: {}",
+                        channelName,
```
sfc-gh-bmikaili, line 121: **WHY (ClientRecreationException log branch):** logs client-invalid events distinctly from other errors so operators can tell at a glance whether a recovery was channel-reopen or client-recreation. Helps correlate with the new `clientRecreationCount` telemetry counter.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/service/BatchOffsetFetcher.java
```diff
@@ -108,11 +109,24 @@ private CompletableFuture<Void> fetchOffsetsAsync(
...
+              // Corner case: if the topic is idle (no new records), no appendRow will fire and
+              // the batch loop will never trigger recreation. Accepted — when traffic resumes,
+              // the first appendRow will fail → recreation kicks in → normal recovery.
+              taskMetrics.incClientRecreationSkipCount();
+              LOGGER.error(
+                  "Client is invalid for pipe: {}, skipping offset fetch for {} channel(s)."
+                      + " Client will be recreated on the next put() cycle.",
+                  pipeName,
+                  channelsByPartition.size(),
+                  e);
```
sfc-gh-bmikaili, line 122: **WHY (skip offset fetch on client-invalid):** offset fetch is background; it shouldn't crash the task. The proper recovery path is the next `put()` → appendRow → fallback → `recreateClient=true`. Until then, it's fine for the fetcher to log and move on; stale offsets won't cause loss because `shouldProcess` uses `processedOffset` (tracker state), not the fetched offset. **Corner case** noted inline: idle topic won't trigger recreation; accepted since data lands when traffic resumes.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/service/PartitionChannelManager.java
```diff
@@ -164,6 +165,18 @@ private TopicPartitionChannel buildChannel(
             taskConfig,
             streamingClientProperties,
             taskMetrics);
+
+    final ClientRecreator clientRecreator =
```
sfc-gh-bmikaili, line 169: **WHY (inject `ClientRecreator`):** `SnowpipeStreamingPartitionChannel` needs a way to trigger client recreation on the shared pool when its appendRow hits client-invalid. A lambda binding the current pipe/connector/task IDs to `StreamingClientPools.recreateClient` keeps the pool's CAS-based dedup centralized. The lambda is constructed per-partition so each channel sees consistent identifiers.

sfc-gh-bmikaili (commented):
Per-enum-value and per-case-label WHY notes.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
```diff
@@ -339,13 +357,31 @@ private static void closeChannelWithoutFlushing(SnowflakeStreamingIngestChannel
...
+      try {
```
sfc-gh-bmikaili, line 401: **WHY (try/catch around recreateClient):** `recreateClient` retries 3×5s internally. If failover is still propagating, all attempts fail and it throws. Without this catch, that exception escapes the Failsafe fallback → kills the task on the first failed attempt. With the catch, we proceed to install the CF chain (which also fails → done-but-failed future). On the next `put()`, fallback fires again, `consecutiveRecoveryCount++`, retry. Verified: real failover took 3-4 min to propagate; task stayed RUNNING throughout with this catch.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
```diff
@@ -67,8 +67,8 @@ public class SnowpipeStreamingPartitionChannel implements TopicPartitionChannel
   // are cumulative and don't reset when a channel is reopened.
   private long initialErrorCount = 0;
 
-  /** Max consecutive channel recoveries before giving up and letting the task fail. */
-  private static final int MAX_CONSECUTIVE_RECOVERIES = 5;
+  // 100 × ~5.5s per attempt ≈ 9 min — covers real pipe failovers (observed 2-4 min propagation).
+  private static final int MAX_CONSECUTIVE_RECOVERIES = 100;
```
sfc-gh-skurella: Let's explore whether we can use the Failsafe policy to perform exponential backoff to ease the pressure on the backend during failover, and maybe deal with a time unit rather than a retry count to decide when to give up

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/service/BatchOffsetFetcher.java
```diff
@@ -108,11 +109,24 @@ private CompletableFuture<Void> fetchOffsetsAsync(
...
-                "Failed to fetch committed offsets for pipe: {}, skipping {} channel(s)",
-                pipeName,
-                channelsByPartition.size(),
-                e);
+            if (ClientRecreationException.isClientInvalidError(e)) {
+              // Corner case: if the topic is idle (no new records), no appendRow will fire and
+              // the batch loop will never trigger recreation. Accepted — when traffic resumes,
+              // the first appendRow will fail → recreation kicks in → normal recovery.
+              taskMetrics.incClientRecreationSkipCount();
+              LOGGER.error(
```
sfc-gh-skurella: NIT: warning?

sfc-gh-skurella (commented):
My initial impression is good, I just need more time to go through it in more detail, especially the tests. But I don't anticipate we'll need more work on this beyond maybe using Failsafe a bit more

PR #1431, opened 2026-04-13
Add batch-level pipe skip on client recreation

sfc-gh-bmikaili:
## Summary

- **`SnowflakeSinkServiceV2`** — catches `ClientRecreationException` in the batch insert loop. Maintains a per-batch `Set<String> invalidatedPipes`. Records on partitions whose pipe is in the set are skipped via `isPartitionOnInvalidatedPipe()` (fast-path: short-circuits when the set is empty, which is the common case). Skipped-record offsets go into `offsetsOfFirstSkippedRecord` and are rewound via `sinkTaskContext::offset` at the end of the batch — the same mechanism backpressure uses, but pipe-scoped instead of task-global. Records on other pipes keep processing.
- **`BatchOffsetFetcher`** — distinct error-level logging when `getChannelStatus` fails due to client invalidation, and increments `clientRecreationSkipCount`. Client will be recreated on the next `put()` cycle (the offset-fetch path doesn't own recreation).
- **`TaskMetrics`** — new `incClientRecreationSkipCount()` counter on the interface, noop, and JMX implementation (`client-recreation-skip-count` in the task sub-domain).

**JIRA**: SNOW-3360429

## Test plan

- [x] `SnowflakeSinkServiceV2Test` — 20 tests (2 new): `ClientRecreationException` triggers offset rewind and pipe-scoped skip; a second pipe on the same connector keeps processing when the first pipe's client is invalidated
- [x] `ClientRecreationIT` — 3 integration tests with `FakeSnowflakeStreamingIngestClient`: CAS returns existing replacement on concurrent recreate calls, 10-thread concurrent recreation produces exactly one new client (latch-coordinated), error-code detection matches `InvalidClientError` / `SfApiPipeFailedOverError` / `ClosedClientError`

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
@@ -19,6 +19,7 @@ import com.snowflake.kafka.connector.internal.metrics.MetricsJmxReporter;
 import com.snowflake.kafka.connector.internal.metrics.TaskMetrics;
 import com.snowflake.kafka.connector.internal.streaming.channel.TopicPartitionChannel;
 import com.snowflake.kafka.connector.internal.streaming.v2.BackpressureException;
+import com.snowflake.kafka.connector.internal.streaming.v2.ClientRecreationException;
 import com.snowflake.kafka.connector.internal.streaming.v2.client.StreamingClientPools;
 import com.snowflake.kafka.connector.internal.streaming.v2.service.BatchOffsetFetcher;
 import com.snowflake.kafka.connector.internal.streaming.v2.service.PartitionChannelManager;
@@ -32,6 +33,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import org.apache.kafka.common.TopicPartition;
@@ -348,6 +350,7 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
     }
 
     Map<TopicPartition, Long> offsetsOfFirstSkippedRecord = new HashMap<>();
+    Set<String> invalidatedPipes = new HashSet<>();
     boolean newBackpressure = false;
     for (SinkRecord record : records) {
       // check if it needs to handle null value records
@@ -362,9 +365,31 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
         // records in this partition.
         continue;
       }
+
+      // Skip other partitions on a pipe that was invalidated earlier in THIS batch.
+      // Do NOT add to offsetsOfFirstSkippedRecord: the async reopen's
+      // `resetAfterRecovery` issues the authoritative Kafka seek; a batch-loop rewind would
+      // race with that and could override it, causing silent loss.
+      // Across batches, partitions on a still-recovering pipe will hit appendRow, see the
+      // invalid client, and each trigger their own recreate. The CAS in StreamingClientPool
+      // deduplicates to a single fresh client, so the redundant calls are tolerable.
+      if (isPartitionOnInvalidatedPipe(tp, invalidatedPipes)) {
+        continue;
+      }
+
       if (skipAllPartitions || initializingPartitions.contains(tp)) {
-        // Make sure we store the first record in each partition that we skipped so we can correctly
-        // rewind the offset.
+        // If the channel has an active recovery floor, the reset-after-recovery's seek to
+        // (committedOffset + 1) is the authoritative Kafka position — a batch-loop rewind
+        // to `record.kafkaOffset()` would override it and cause data loss. Skip without
+        // rewinding in that case.
+        if (channelManager
+            .getChannel(tp)
+            .map(TopicPartitionChannel::hasActiveRecoveryFloor)
+            .orElse(false)) {
+          continue;
+        }
+        // Normal initializing case (cold start). Rewinding to the record's offset is safe
+        // because no SDK buffer has seen these records.
         offsetsOfFirstSkippedRecord.putIfAbsent(tp, record.kafkaOffset());
         continue;
       }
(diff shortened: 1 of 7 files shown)


sfc-gh-bmikaili:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1431?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#1431** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1431?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1431?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1430** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1430?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1432](https://github.com/snowflakedb/snowflake-kafka-connector/pull/1432) <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1432?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1460** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1460?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/service/BatchOffsetFetcher.java
```diff
@@ -108,11 +109,21 @@ private CompletableFuture<Void> fetchOffsetsAsync(
           try {
             result.putAll(getCommittedOffsetsForPipe(pipeName, channelsByPartition));
           } catch (SFException e) {
-            LOGGER.error(
-                "Failed to fetch committed offsets for pipe: {}, skipping {} channel(s)",
-                pipeName,
-                channelsByPartition.size(),
-                e);
+            if (ClientRecreationException.isClientInvalidError(e)) {
```
sfc-gh-skurella, line 112: I'm wondering if we should propagate it - what if a topic isn't receiving any new records but the old ones haven't caught up? IMO this corner case is not worth the effort but it'd be good to leave a one-line comment here.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -72,6 +74,12 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
 
   // Set that keeps track of the channels that have been seen per input batch
   private final Set<String> channelsVisitedPerBatch = new HashSet<>();
+  // Pipes whose client was invalidated and is currently being recreated. Cross-batch marker:
+  // while a pipe is here, the batch loop skips all its partitions WITHOUT rewinding — the
+  // async reopen's `resetAfterRecovery` is the sole source of truth for the Kafka seek.
+  // Entries are cleared when the channel's future completes successfully (`!isInitializing()
+  // && !isChannelClosed()`).
+  private final Set<String> pipesInRecovery = ConcurrentHashMap.newKeySet();
```
sfc-gh-skurella: Do we need to keep state? State typically makes code considerably harder to reason about, and this is right in the critical path for exactly-one correctness.

We stream over all partitions in every `insert` call anyway, so it doesn't sound like a performance optimization. Is there something that requires persistence between `insert` invocations?

sfc-gh-skurella (commented):
Let's review this in person on Monday?

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/service/BatchOffsetFetcher.java
sfc-gh-bmikaili: Added a comment explaining the idle-topic corner case:

```java
// Corner case: if the topic is idle (no new records), no appendRow will fire and
// the batch loop will never trigger recreation. Accepted — when traffic resumes,
// the first appendRow will fail → recreation kicks in → normal recovery.
```

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-bmikaili: Removed. Good call — this was Guard 4a in the [four-guards recovery design](https://github.com/snowflakedb/snowflake-kafka-connector/pull/1431), and the design doc explicitly calls it out as removable:

> **Minimum safe subset**: Guard 1 + Guard 2 + Guard 4b. [...] Guard 4a (pipesInRecovery set) could be removed at the cost of slight duplicate-appendRow noise during recovery.

What I removed:
- The `pipesInRecovery` field and its reaper (`removeIf(...)` at top of batch)
- The cross-batch `recoveryGatedPartitions` snapshot
- `isPartitionInRecoveringPipe()` and `pipeHasInitializingChannel()` helpers
- The `pipesInRecovery.add(...)` in the `ClientRecreationException` catch

What I kept (in-batch, no cross-batch state):
- `invalidatedPipes` local `HashSet` in `insert(Collection)` — ensures other partitions on the same pipe in THIS batch skip without retrying the dead client
- `isPartitionOnInvalidatedPipe(tp, invalidatedPipes)` check
- The end-of-batch `hasActiveRecoveryFloor` rewind-drop guard (Guard 4b)

Without Guard 4a, partitions on a still-recovering pipe each fire one failing `appendRow` before seeing the new client. The CAS in `StreamingClientPool.recreateClient` deduplicates to a single fresh client per pipe, so it's noise, not correctness breakage.

**Validated with E2E** (`test_client_recreation.py` with mitmproxy 410 injection): all 900 records across pre-fault + during-fault + post-heal phases land in Snowflake, no data loss.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -383,6 +408,25 @@ public void insert(final Collection<SinkRecord> records) {
...
+        // Do NOT call `sinkTaskContext::offset` here. The current record's offset is
+        // unreliable as a rewind target: the SDK's appendRow is buffered, so records
+        // earlier than this one may have been batched into a flush that got the 410.
+        // Rewinding to `record.kafkaOffset()` would silently drop those earlier records.
+        // Instead, let `SnowpipeStreamingPartitionChannel.recreateClientAndReopenChannel ->
+        // reopenChannel -> PartitionOffsetTracker.resetAfterRecovery` set the correct
+        // Kafka offset (Snowflake's latest committed offset + 1) after the async reopen
+        // completes. Until then, subsequent `put()` calls on this pipe will keep hitting
+        // ClientRecreationException and skip cleanly via `isPartitionOnInvalidatedPipe`.
       }
```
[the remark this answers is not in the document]
sfc-gh-bmikaili: Good catch — added `taskMetrics.incClientRecreationSkipCount();` on the `ClientRecreationException` catch path to match `BatchOffsetFetcher`'s usage.

sfc-gh-bmikaili:
Folded into #1430 — the batch-level pipe skip and the partition-channel recovery are too interdependent to review separately. Combined PR contains all four guards with a single unified commit.

PR #1447, opened 2026-05-12
add OAuth error codes, URL extensions, and token infrastructure

sfc-gh-skurella:
## Summary

Adds OAuth error codes, the `URL` interface extensions (`sslEnabled()`, `path()`), and the token infrastructure needed by subsequent PRs in the OAuth stack.

- **Error codes**: ERROR_0026, ERROR_0027, ERROR_0029, ERROR_0033, ERROR_1004
- **`URL` interface**: added `sslEnabled()` and `path()` as abstract methods; `SnowflakeURL` implements both (with `@Override` on `sslEnabled()` and `getScheme()`)
- **`OAuthURL`**: URL parser for OAuth endpoints (non-Snowflake URLs with arbitrary paths). Scheme and host are normalized to lowercase (case-insensitive per RFC 3986), but the path is preserved verbatim so case-sensitive token endpoints (e.g. `/oauth2/v2.0/Token`) are not corrupted.
- **`OAuthAccessTokenFetcher`**: fetches OAuth access tokens via `refresh_token` or `client_credentials` grant — extracted from `Utils.getSnowflakeOAuthToken()` in KC v3. Retries with exponential backoff (via Failsafe `RetryPolicy`, matching v3's backoff schedule), but **only for genuinely transient failures**: `IOException`/timeout and HTTP 5xx/429 (rethrown as `IOException`) are retried, while terminal errors surfaced as `SnowflakeKafkaConnectorException` (bad client id/secret → 4xx, malformed request, or a 2xx response with no `access_token`) abort immediately instead of retrying. This diverges from v3, whose `backoffAndRetry` retried every failure — so a typo'd secret no longer hangs `validate()` for ~15s or repeatedly hits the IdP. Uses the JVM's default `ProxySelector`, so proxy settings from `Utils.enableJVMProxy` — including `jvm.nonProxy.hosts` bypass rules — are respected, matching KC v3 behavior.

OAuth constants (grant types, header values, etc.) are inlined as private fields in `OAuthAccessTokenFetcher` — they were only used by that class. All OAuth classes are grouped under `internal.oauth`.

## Cross-reference with KC v3 (`3.2.x`)

| v4 file | v3 equivalent | Notes |
|---------|---------------|-------|
| [`SnowflakeErrors.java`](https://github.com/snowflakedb/snowflake-kafka-connector/blob/skurella-oauth-infra/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeErrors.java) | [`SnowflakeErrors.java`](https://github.com/snowflakedb/snowflake-kafka-connector/blob/3.2.x/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeErrors.java) | ERROR_0026, 0027, 0029, 0033, 1004 — same codes as v3 |
| [`URL.java`](https://github.com/snowflakedb/snowflake-kafka-connector/blob/skurella-oauth-infra/src/main/java/com/snowflake/kafka/connector/internal/URL.java) | [`URL.java`](https://github.com/snowflakedb/snowflake-kafka-connector/blob/3.2.x/src/main/java/com/snowflake/kafka/connector/internal/URL.java) | v4 makes `sslEnabled()` and `path()` abstract (v3 had defaults) |
| [`SnowflakeURL.java`](https://github.com/snowflakedb/snowflake-kafka-connector/blob/skurella-oauth-infra/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeURL.java) | [`SnowflakeURL.java`](https://github.com/snowflakedb/snowflake-kafka-connector/blob/3.2.x/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeURL.java) | Added `path()` returning `""`, added `@Override` annotations |
| [`OAuthURL.java`](https://github.com/snowflakedb/snowflake-kafka-connector/blob/skurella-oauth-infra/src/main/java/com/snowflake/kafka/connector/internal/oauth/OAuthURL.java) | [`OAuthURL.java`](https://github.com/snowflakedb/snowflake-kafka-connector/blob/3.2.x/src/main/java/com/snowflake/kafka/connector/internal/OAuthURL.java) | Moved to `internal.oauth` package; fixes v3 bug where the whole URL was lowercased (now preserves path case) |
| [`OAuthAccessTokenFetcher.java`](https://github.com/snowflakedb/snowflake-kafka-connector/blob/skurella-oauth-infra/src/main/java/com/snowflake/kafka/connector/internal/oauth/OAuthAccessTokenFetcher.java) | [`Utils.getSnowflakeOAuthAccessToken()` / `Utils.getSnowflakeOAuthToken()`](https://github.com/snowflakedb/snowflake-kafka-connector/blob/3.2.x/src/main/java/com/snowflake/kafka/connector/Utils.java#L791-L862) | Extracted into dedicated class; adds `client_credentials` grant and Failsafe retry. Retry now aborts on terminal `SnowflakeKafkaConnectorException` (4xx/malformed/missing token) and only retries transient failures (IOException/timeout, 5xx/429) — v3 retried everything. Proxy handling unchanged from v3 (JVM default `ProxySelector`, honoring `jvm.nonProxy.hosts`) |
| *(inlined in `OAuthAccessTokenFetcher`)* | [`OAuthConstants.java`](https://github.com/snowflakedb/snowflake-kafka-connector/blob/3.2.x/src/main/java/com/snowflake/kafka/connector/internal/OAuthConstants.java) | Constants moved to private fields in the only consumer |

## Test plan

- [x] `OAuthURLTest` — parameterized URL parsing and invalid URL rejection, including mixed-case scheme/host normalization and case-sensitive path preservation
- [x] `OAuthAccessTokenFetcherTest` — unit tests covering `refresh_token` grant, `client_credentials` grant, empty refresh token fallback, HTTP error status, missing `access_token` in response, `HttpClient` exceptions, and retry-then-succeed (all using Mockito-mocked `HttpClient`)
- [x] `OAuthAccessTokenFetcherTest` — asserts a permanent 4xx (401) is not retried (`verify(httpClient, times(1))`), so terminal failures abort immediately

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/oauth/OAuthAccessTokenFetcher.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/oauth/OAuthAccessTokenFetcher.java
@@ -1,0 +1,194 @@
+package com.snowflake.kafka.connector.internal.oauth;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.snowflake.kafka.connector.internal.KCLogger;
+import com.snowflake.kafka.connector.internal.SnowflakeErrors;
+import com.snowflake.kafka.connector.internal.SnowflakeKafkaConnectorException;
+import com.snowflake.kafka.connector.internal.URL;
+import dev.failsafe.Failsafe;
+import dev.failsafe.RetryPolicy;
+import java.io.IOException;
+import java.net.ProxySelector;
+import java.net.URI;
+import java.net.URLEncoder;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.nio.charset.StandardCharsets;
+import java.time.Duration;
+import java.util.Base64;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.apache.kafka.common.config.types.Password;
+
+/**
+ * Fetches an OAuth access token from an OAuth token endpoint. Used only for JDBC connections; the
+ * streaming SDK handles its own OAuth token lifecycle.
+ *
+ * <p>Supports two grant types:
+ *
+ * <ul>
+ *   <li><b>refresh_token</b>: when a refresh token is provided
+ *   <li><b>client_credentials</b>: when no refresh token is provided
+ * </ul>
+ *
+ * @see <a
+ *     href="https://github.com/snowflakedb/snowflake/blob/4fdb96cd5849f266cda430c5d49a13c29e866af5/GlobalServices/src/main/java/com/snowflake/resources/ResourceConstants.java">ResourceConstants</a>
+ */
+public class OAuthAccessTokenFetcher {
+
+  private static final KCLogger LOGGER = new KCLogger(OAuthAccessTokenFetcher.class.getName());
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  private static final String TOKEN_REQUEST_ENDPOINT = "/oauth/token-request";
+  private static final String OAUTH_CONTENT_TYPE_HEADER = "application/x-www-form-urlencoded";
+  private static final String BASIC_AUTH_HEADER_PREFIX = "Basic ";
+  private static final String GRANT_TYPE_PARAM = "grant_type";
+  private static final String GRANT_TYPE_REFRESH_TOKEN = "refresh_token";
+  private static final String GRANT_TYPE_CLIENT_CREDENTIALS = "client_credentials";
+  private static final String REFRESH_TOKEN = "refresh_token";
+  private static final String ACCESS_TOKEN = "access_token";
+  private static final String REDIRECT_URI = "redirect_uri";
+  private static final String DEFAULT_REDIRECT_URI = "https://localhost.com/oauth";
+
+  private static final int HTTP_TOO_MANY_REQUESTS = 429;
+
+  private static final int MAX_RETRIES = 4;
+  private static final Duration INITIAL_DELAY = Duration.ofSeconds(1);
+  private static final Duration MAX_DELAY = Duration.ofSeconds(8);
+
+  /**
+   * Fetch an OAuth access token. Retries with exponential backoff on transient failures. Uses the
+   * JVM's default {@link ProxySelector} so that proxy settings from {@code Utils.enableJVMProxy}
+   * (including {@code jvm.nonProxy.hosts} bypass rules) are respected, matching KC v3 behavior.
+   *
+   * @param url OAuth server URL (scheme, host, port, path)
+   * @param clientId OAuth client ID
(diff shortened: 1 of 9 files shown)


sfc-gh-skurella:
* **#1448** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1448?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1447** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1447?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1447?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1458** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1458?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

sfc-gh-bmikaili (approved):
Approving to unblock the stack.

src/main/java/com/snowflake/kafka/connector/internal/oauth/OAuthAccessTokenFetcher.java
```diff
@@ -0,0 +1,179 @@
...
+        HttpRequest.newBuilder()
+            .uri(tokenUri)
+            .header("Content-Type", OAUTH_CONTENT_TYPE_HEADER)
+            .header("Authorization", basicAuth)
+            .POST(HttpRequest.BodyPublishers.ofString(formBody))
+            .build();
+
+    RetryPolicy<String> retryPolicy =
+        RetryPolicy.<String>builder()
+            .handle(Exception.class)
```
sfc-gh-bmikaili, line 125: The retry policy handles `Exception.class` with no `abortOn`, so non-retryable failures get retried. A 401 (bad client secret), a 400 (malformed request), or a 200 whose body has no `access_token` all throw `ERROR_1004` *inside* the Failsafe block — and since that's a `RuntimeException`, it's caught here and retried 4× with backoff (1+2+4+8 ≈ 15s) before failing with the same error. Retrying can't help: the credentials won't become valid on the next attempt.

Consequences:
- `validate()` performs a live token fetch, so a typo'd secret makes every config-validation call hang ~15s before returning the error.
- Repeated failed auth attempts against the IdP (5 per fetch) can trip rate-limiting or client lockout on some providers.
- Each attempt logs a WARN, so a permanent config error looks like a transient/network problem in the logs.

Suggest aborting on the terminal connector exception so only genuinely transient failures (IOException/timeout) retry:

```java
RetryPolicy.<String>builder()
    .handle(Exception.class)
    .abortOn(SnowflakeKafkaConnectorException.class)
    .withBackoff(INITIAL_DELAY, MAX_DELAY)
    .withMaxRetries(MAX_RETRIES)
```

Also worth adding `verify(httpClient, times(1))` to `fetchAccessToken_nonSuccessStatus_throwsError1004` — it currently passes while silently sleeping ~15s, so the wasted retries are invisible in CI.

PR #1446, opened 2026-05-12
add config for OAuth

sfc-gh-skurella:
## Summary

Adds OAuth configuration support to the Kafka Connector v4, laying the groundwork for OAuth authentication (JWT key-pair remains the default). This PR introduces only the config layer — no runtime behavior changes.

Changes:
- New `AuthenticatorType` enum (`SNOWFLAKE_JWT`, `OAUTH`) to replace raw string comparisons, addressing the [v3 TODO](https://github.com/snowflakedb/snowflake-kafka-connector/blob/3.2.x/src/main/java/com/snowflake/kafka/connector/Utils.java#L83)
- Five new config parameters in `KafkaConnectorConfigParams` and `ConnectorConfigDefinition`, matching the [v3 config names](https://github.com/snowflakedb/snowflake-kafka-connector/blob/3.2.x/src/main/java/com/snowflake/kafka/connector/Utils.java#L82-L87) exactly: `snowflake.authenticator`, `snowflake.oauth.client.id`, `snowflake.oauth.client.secret`, `snowflake.oauth.refresh.token`, `snowflake.oauth.token.endpoint`
- `SinkTaskConfig` extended with typed `AuthenticatorType getAuthenticator()` and nullable OAuth field getters
- `DefaultConnectorConfigValidator` updated to switch on `AuthenticatorType`: validates OAuth client ID/secret when `oauth`, validates private key when `snowflake_jwt`
- `refresh_token` is intentionally optional to support both `refresh_token` and `client_credentials` grant types

## v3 cross-reference

Config parameter names and authenticator values match the v3 (3.2.x) branch:
- [Config constants (`Utils.java`)](https://github.com/snowflakedb/snowflake-kafka-connector/blob/3.2.x/src/main/java/com/snowflake/kafka/connector/Utils.java#L82-L91)
- [ConfigDef entries (`SnowflakeSinkConnectorConfig.java`)](https://github.com/snowflakedb/snowflake-kafka-connector/blob/3.2.x/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkConnectorConfig.java#L64-L68)

## Test plan

- [x] `SinkTaskConfigTest`: OAuth fields parsed correctly, default authenticator is `SNOWFLAKE_JWT`, optional fields nullable
- [x] `ConnectorConfigValidatorTest`: valid OAuth config, client\_credentials grant (no refresh token), custom token endpoint, invalid authenticator rejected, missing client ID/secret rejected, private key not required for OAuth

PR #1448, opened 2026-05-12
add OAuth support to JDBC connection properties

sfc-gh-skurella:
## Summary

Adds OAuth support to JDBC connection properties. When the connector is configured with `snowflake.authenticator=oauth`, `makeJdbcDriverProperties` fetches an access token via `OAuthAccessTokenFetcher` and passes it to the JDBC driver instead of a private key.

- **`InternalUtils.makeJdbcDriverProperties`**: switches on `AuthenticatorType` from the parsed `SinkTaskConfig` — `SNOWFLAKE_JWT` uses key-pair auth (existing behavior), `OAUTH` validates client ID/secret, fetches an access token, and sets the JDBC `token` property. Includes `default` case throwing `IllegalStateException` for future-proofing.
- **`SnowflakeStreamingSinkConnector.validate()`**: maps the OAuth-specific error codes that a live token fetch can raise to per-field config errors instead of letting them escape as a generic worker error — `ERROR_0033` (invalid token endpoint URL) to `snowflake.oauth.token.endpoint`, and `ERROR_1004` (token fetch failed) to the client ID, client secret, and token endpoint fields.
- **`TestUtils.getProfile()`**: made public so tests can access `Profile` fields directly (e.g. OAuth credentials).
- **`OAuthConnectionIT`**: integration test that connects to Snowflake with OAuth credentials read from `profile.json` (skipped when OAuth fields are absent). Uses try/finally for connection cleanup.

## Config validation and live OAuth fetch

`SnowflakeStreamingSinkConnector.validate()` builds a real `SnowflakeConnectionService` to verify credentials, which now triggers a live OAuth token fetch. This is intentional — it's consistent with the existing behavior (JWT validation already makes a live JDBC connection here) and gives the customer immediate feedback if their OAuth credentials are misconfigured. KC v3 behaves the same way.

## Token refresh strategy

The access token is fetched once when JDBC properties are built — this matches KC v3 behavior exactly ([v3 `createProperties`](https://github.com/snowflakedb/snowflake-kafka-connector/blob/3.2.x/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java#L212-L216) also calls `getSnowflakeOAuthAccessToken()` once). A 6-hour soak test (`OAuthSoakIT`) with 1-hour token expiry showed no issues: JDBC DDL calls continued past expiry, and the streaming SDK handles token renewal transparently via its internal `OAuthManager`.

If the token does expire during a long-lived task, the auth failure triggers task restart via `SnowflakeSinkTaskAuthorizationExceptionTracker`, which re-fetches the token on restart.

Future work: [SNOW-3535094](https://snowflakecomputing.atlassian.net/browse/SNOW-3535094) tracks adding a headless `oauth_refresh_token` authenticator to the JDBC driver. If that lands, we can delegate token renewal to JDBC directly, but we found no evidence that complex manual refresh logic is necessary at this point.

## Cross-reference with KC v3 (`3.2.x`)

| v4 file | v3 equivalent | Notes |
|---------|---------------|-------|
| [`InternalUtils.makeJdbcDriverProperties`](https://github.com/snowflakedb/snowflake-kafka-connector/blob/skurella-oauth-jdbc/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java) | [`InternalUtils.createProperties`](https://github.com/snowflakedb/snowflake-kafka-connector/blob/3.2.x/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java#L131-L230) | v4 reads from `SinkTaskConfig` instead of raw map; uses `AuthenticatorType` enum instead of string comparison |
| [`OAuthConnectionIT`](https://github.com/snowflakedb/snowflake-kafka-connector/blob/skurella-oauth-jdbc/src/test/java/com/snowflake/kafka/connector/internal/OAuthConnectionIT.java) | [`ConnectionServiceIT.testOAuthAZ`](https://github.com/snowflakedb/snowflake-kafka-connector/blob/3.2.x/src/test/java/com/snowflake/kafka/connector/internal/ConnectionServiceIT.java) | v4 reads OAuth creds from `profile.json` fields; v3 used `TestUtils.getConfWithOAuth()` |

### Key differences from v3

- v3 `createProperties` iterates a raw `Map<String, String>` and compares authenticator strings; v4 uses the already-parsed `SinkTaskConfig` and switches on an enum
- v3 requires `refresh_token` (throws ERROR_0028 if absent); v4 allows omitting it to fall back to `client_credentials` grant
- v3 calls `Utils.getSnowflakeOAuthAccessToken()`; v4 calls `OAuthAccessTokenFetcher.fetchAccessToken()` (introduced in PR #1447)
- Invalid authenticator values are caught at `SinkTaskConfig` parse time by `AuthenticatorType.fromConfig()` rather than in `makeJdbcDriverProperties`

## Test plan

- [x] `InternalUtilsTest.testMakeJdbcDriverProperties_oauthMissingClientId_throws`
- [x] `InternalUtilsTest.testMakeJdbcDriverProperties_oauthMissingClientSecret_throws`
- [x] `InternalUtilsTest.testMakeJdbcDriverProperties_invalidAuthenticator_throws` — tests `AuthenticatorType.fromConfig()` rejects unknown authenticator values
- [x] `ConnectorIT.testValidateErrorOAuthTokenEndpointConfig` — a malformed `snowflake.oauth.token.endpoint` surfaces as a per-field config error (ERROR_0033 mapping)
- [x] `OAuthConnectionIT.oauthConnection_shouldConnectAndQuerySnowflake` — reads OAuth credentials from `profile.json`; verified locally against Snowflake

[SNOW-3535094]: https://snowflakecomputing.atlassian.net/browse/SNOW-3535094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeStreamingSinkConnector.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeStreamingSinkConnector.java
@@ -17,6 +17,7 @@
 package com.snowflake.kafka.connector;
 
 import com.snowflake.kafka.connector.Constants.KafkaConnectorConfigParams;
+import com.snowflake.kafka.connector.config.AuthenticatorType;
 import com.snowflake.kafka.connector.config.ConnectorConfigDefinition;
 import com.snowflake.kafka.connector.internal.KCLogger;
 import com.snowflake.kafka.connector.internal.SnowflakeConnectionService;
@@ -48,6 +49,10 @@ public class SnowflakeStreamingSinkConnector extends SinkConnector {
   // create logger without correlationId for now
   private static final KCLogger LOGGER =
       new KCLogger(SnowflakeStreamingSinkConnector.class.getName());
+
+  // Matches a Kafka Connect config-provider reference such as ${vault:...} or ${file:...}, whose
+  // value is resolved per-worker at task start and is therefore not available during validate().
+  private static final Pattern CONFIG_PROVIDER_PREFIX = Pattern.compile("[$][{][a-zA-Z]+:");
 
   private Map<String, String> config; // connector configuration, provided by
   // user through kafka connect framework
@@ -216,11 +221,16 @@ public class SnowflakeStreamingSinkConnector extends SinkConnector {
       Utils.updateConfigErrorMessage(result, invalidKey, invalidProxyParams.get(invalidKey));
     }
 
-    // If private key or private key passphrase is
-    // provided through a config provider, skip validation
-    if (isUsingConfigProviderForPrivateKey(connectorConfigs)) {
+    // Skip live connection validation only when a credential is supplied through a config provider
+    // (private key/passphrase or an OAuth credential); its placeholder cannot be resolved here.
+    if (isUsingConfigProviderForPrivateKey(connectorConfigs)
+        || isUsingConfigProviderForOAuth(connectorConfigs)) {
       return result;
     }
+
+    // Apply proxy settings before the test connection so the OAuth token fetch (which uses the JVM
+    // default ProxySelector) sees the same proxy/nonProxyHosts config as runtime does in start().
+    Utils.enableJVMProxy(connectorConfigs);
 
     // We don't validate name, since it is not included in the return value
     // so just put a test connector here
@@ -243,12 +253,16 @@ public class SnowflakeStreamingSinkConnector extends SinkConnector {
               ": Cannot connect to Snowflake");
           Utils.updateConfigErrorMessage(
               result,
-              KafkaConnectorConfigParams.SNOWFLAKE_PRIVATE_KEY,
-              ": Cannot connect to Snowflake");
-          Utils.updateConfigErrorMessage(
-              result,
               KafkaConnectorConfigParams.SNOWFLAKE_USER_NAME,
               ": Cannot connect to Snowflake");
+          // A 1001 for OAuth means the token fetch already succeeded (a fetch failure surfaces as
+          // 1004), so the private key -- which OAuth users do not configure -- is not the culprit.
+          if (!isUsingOAuth(connectorConfigs)) {
+            Utils.updateConfigErrorMessage(
+                result,
+                KafkaConnectorConfigParams.SNOWFLAKE_PRIVATE_KEY,
+                ": Cannot connect to Snowflake");
+          }
           break;
         case "0007":
           Utils.updateConfigErrorMessage(
@@ -271,6 +285,45 @@ public class SnowflakeStreamingSinkConnector extends SinkConnector {
               result,
               KafkaConnectorConfigParams.SNOWFLAKE_PRIVATE_KEY,
               " must be a valid PEM RSA private key");
+          break;
+        case "0026":
+          Utils.updateConfigErrorMessage(
+              result,
+              KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_CLIENT_ID,
+              " must be non-empty when using oauth authenticator");
+          break;
+        case "0027":
+          Utils.updateConfigErrorMessage(
+              result,
+              KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_CLIENT_SECRET,
+              " must be non-empty when using oauth authenticator");
+          break;
+        case "0029":
+          Utils.updateConfigErrorMessage(
+              result,
+              KafkaConnectorConfigParams.SNOWFLAKE_AUTHENTICATOR,
+              " is not a valid authenticator");
+          break;
+        case "0033":
+          Utils.updateConfigErrorMessage(
+              result,
+              KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_TOKEN_ENDPOINT,
+              " is not a valid OAuth token endpoint URL");
+          break;
+        case "1004":
+          {
+            // The token fetch can fail for a bad client id/secret, a wrong token endpoint, or a
+            // transient network/proxy issue - flag the whole OAuth credential set.
+            String tokenFetchError =
+                ": Could not fetch OAuth access token. Check the client ID, client secret, and"
+                    + " token endpoint.";
+            Utils.updateConfigErrorMessage(
+                result, KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_CLIENT_ID, tokenFetchError);
+            Utils.updateConfigErrorMessage(
+                result, KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_CLIENT_SECRET, tokenFetchError);
+            Utils.updateConfigErrorMessage(
+                result, KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_TOKEN_ENDPOINT, tokenFetchError);
+          }
           break;
         default:
           throw e; // Shouldn't reach here, so crash.
@@ -310,18 +363,38 @@ public class SnowflakeStreamingSinkConnector extends SinkConnector {
     return result;
   }
 
-  private static boolean isUsingConfigProviderForPrivateKey(Map<String, String> connectorConfigs) {
-    Pattern configProviderPrefix = Pattern.compile("[$][{][a-zA-Z]+:");
+  private static boolean isUsingOAuth(Map<String, String> connectorConfigs) {
+    try {
+      return AuthenticatorType.fromConfig(
+              connectorConfigs.getOrDefault(
+                  KafkaConnectorConfigParams.SNOWFLAKE_AUTHENTICATOR,
+                  AuthenticatorType.SNOWFLAKE_JWT.toConfigValue()))
+          == AuthenticatorType.OAUTH;
+    } catch (IllegalArgumentException e) {
+      return false;
+    }
+  }
 
-    return configProviderPrefix
-            .matcher(
-                connectorConfigs.getOrDefault(KafkaConnectorConfigParams.SNOWFLAKE_PRIVATE_KEY, ""))
-            .find()
-        || configProviderPrefix
-            .matcher(
-                connectorConfigs.getOrDefault(
-                    KafkaConnectorConfigParams.SNOWFLAKE_PRIVATE_KEY_PASSPHRASE, ""))
-            .find();
+  private static boolean isUsingConfigProviderForPrivateKey(Map<String, String> connectorConfigs) {
+    return isConfigProviderReference(
+            connectorConfigs.get(KafkaConnectorConfigParams.SNOWFLAKE_PRIVATE_KEY))
+        || isConfigProviderReference(
+            connectorConfigs.get(KafkaConnectorConfigParams.SNOWFLAKE_PRIVATE_KEY_PASSPHRASE));
+  }
+
+  private static boolean isUsingConfigProviderForOAuth(Map<String, String> connectorConfigs) {
+    return isConfigProviderReference(
+            connectorConfigs.get(KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_CLIENT_ID))
+        || isConfigProviderReference(
+            connectorConfigs.get(KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_CLIENT_SECRET))
+        || isConfigProviderReference(
+            connectorConfigs.get(KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_REFRESH_TOKEN))
+        || isConfigProviderReference(
+            connectorConfigs.get(KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_TOKEN_ENDPOINT));
+  }
+
+  private static boolean isConfigProviderReference(String value) {
+    return value != null && CONFIG_PROVIDER_PREFIX.matcher(value).find();
   }
 
   /**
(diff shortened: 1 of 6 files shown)


sfc-gh-skurella:
* **#1448** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1448?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1448?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1447** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1447?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1458** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1458?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
```diff
@@ -88,15 +91,39 @@ static Properties makeJdbcDriverProperties(SinkTaskConfig config, SnowflakeURL u
...
+      case SNOWFLAKE_JWT:
+        properties.put(JdbcPropertyKeys.AUTHENTICATOR, SNOWFLAKE_JWT);
+        String privateKey = config.getSnowflakePrivateKey();
+        if (isBlank(privateKey)) {
+          throw SnowflakeErrors.ERROR_0013.getException();
+        }
+        properties.put(
+            JDBC_PRIVATE_KEY,
+            PrivateKeyTool.parsePrivateKey(privateKey, config.getSnowflakePrivateKeyPassphrase()));
+        break;
```
sfc-gh-bmikaili, line 120: Missing `default:` — same pattern I flagged on #1446's validator switch. A future third `AuthenticatorType` value silently yields a `Properties` with no `AUTHENTICATOR` key and no credentials.

```java
default:
  throw SnowflakeErrors.ERROR_0029.getException("unhandled authenticator: " + config.getAuthenticator());
```

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
```diff
@@ -88,15 +91,39 @@ static Properties makeJdbcDriverProperties(SinkTaskConfig config, SnowflakeURL u
...
+        if (isBlank(oAuthClientId)) {
+          throw SnowflakeErrors.ERROR_0026.getException();
+        }
+        if (isBlank(oAuthClientSecret)) {
+          throw SnowflakeErrors.ERROR_0027.getException();
+        }
+        String oAuthTokenEndpoint = config.getOauthTokenEndpoint();
+        URL oauthUrl = isBlank(oAuthTokenEndpoint) ? url : OAuthURL.from(oAuthTokenEndpoint);
+
+        String accessToken =
```
sfc-gh-bmikaili: The happy OAuth path here is only exercised by `OAuthConnectionIT`, which `assumeTrue`-skips when `profile.json` lacks OAuth creds — so CI regressions in `JDBC_TOKEN` key, `AuthenticatorType.OAUTH.toConfigValue()`, or grant selection won't be caught.

`OAuthAccessTokenFetcher.fetchAccessToken(...)` is `static`, so it can't be mocked as-is. A small seam (supplier/factory with a package-private setter, matching the `StreamingClientFactory.setStreamingClientSupplier` pattern) would let you write a unit test that asserts `properties[authenticator]=oauth` and `properties[token]=<mocked>` taking the OAuth branch.

src/test/java/com/snowflake/kafka/connector/internal/OAuthConnectionIT.java
```diff
@@ -0,0 +1,59 @@
...
+      config.put(KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_REFRESH_TOKEN, refreshToken);
+    }
+    config.remove(KafkaConnectorConfigParams.SNOWFLAKE_PRIVATE_KEY);
+    config.remove(KafkaConnectorConfigParams.SNOWFLAKE_PRIVATE_KEY_PASSPHRASE);
+
+    SnowflakeConnectionService conn =
+        SnowflakeConnectionServiceFactory.builder().setProperties(config).build();
+    // databaseExists throws if it does not exist; no exception means success
+    conn.databaseExists(config.get(KafkaConnectorConfigParams.SNOWFLAKE_DATABASE_NAME));
+    conn.close();
```
sfc-gh-bmikaili: Two issues:

1. If `databaseExists` throws, `conn.close()` is skipped — connection leaks. Use a field + `@AfterEach` (matches `ConnectionServiceIT`), or try/finally.
2. The comment says "throws if it does not exist" but `databaseExists` returns a boolean — if OAuth connects to the wrong account, this passes silently. Make the assertion explicit:
   ```java
   assertThat(conn.databaseExists(...)).isTrue();
   ```

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
```diff
@@ -88,15 +91,39 @@ static Properties makeJdbcDriverProperties(SinkTaskConfig config, SnowflakeURL u
...
 
-    properties.put(JdbcPropertyKeys.AUTHENTICATOR, SNOWFLAKE_JWT);
-
-    String privateKey = config.getSnowflakePrivateKey();
-    if (isBlank(privateKey)) {
-      throw SnowflakeErrors.ERROR_0013.getException();
+    switch (config.getAuthenticator()) {
+      case OAUTH:
+        properties.put(JdbcPropertyKeys.AUTHENTICATOR, AuthenticatorType.OAUTH.toConfigValue());
+        String oAuthClientId = config.getOauthClientId();
```
sfc-gh-bmikaili: Nit: locals use `oAuthClientId` / `oAuthClientSecret` / `oAuthTokenEndpoint`, but `SinkTaskConfig` accessors are `getOauthClientId()` etc. Use `oauthClientId` to match.

sfc-gh-bmikaili (changes requested):
Mechanically clean, but one blocker: the access token is fetched once at `setProperties(...)` and burned into JDBC `Properties` for the life of the connection. Access tokens live minutes; Connect tasks live weeks. There's no 401/refresh path in `StandardSnowflakeConnectionService` or `CachingSnowflakeConnectionService` — every expiry will surface as auth failures on DDL calls.

What's the plan for expiry? Options I see: (a) pass `oauthRefreshToken`/`clientId`/`clientSecret` straight to the JDBC driver so it refreshes internally, (b) wrap the connection with a retry-on-401 interceptor that re-fetches, (c) rely on task-restart via `SnowflakeSinkTaskAuthorizationExceptionTracker` — but that assumes the expired-token error matches its substring check and is disruptive hourly.

Also: `validate()` at `SnowflakeStreamingSinkConnector.java:230-234` will now hit the customer's OAuth server on every Connect config-validation event. Probably worth short-circuiting or at least noting in the PR body.

Note: I'm not blocking on the `OAuthAccessTokenFetcher` exception-body sanitization or HttpClient timeouts — those live in the #1447 stacked base. Consistent with your scope preference on #1446, happy to file follow-up tickets for those.

See inline for smaller items.

src/test/java/com/snowflake/kafka/connector/internal/OAuthConnectionIT.java
sfc-gh-skurella: Added `conn.close()` in a finally block.

> databaseExists returns a boolean

I don't think it does

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
sfc-gh-skurella: I believe this will be covered by integration tests once the whole thing lands. LMK if you prefer that I add this test anyway.

sfc-gh-bmikaili:
Still want to hear your plan on the two open items from the top-level review before I dismiss the changes-requested:

1. **Token expiry** — the fetched token is baked into JDBC `Properties` for the life of the `Connection`. `StandardSnowflakeConnectionService.checkConnection()` only checks `isClosed()`, and I don't see any 401/refresh handler anywhere in the connection-service chain. What's the intended behavior when the token expires mid-task?

2. **`validate()` live OAuth fetch** — every Connect config-validation call (`POST /connectors/{name}/config/validate`) will now hit the customer's OAuth endpoint. Intentional, or worth short-circuiting?

Happy to unblock once there's a direction on #1 (even if the refresh PR is a follow-up, as long as it's explicitly tracked).

src/test/java/com/snowflake/kafka/connector/internal/OAuthConnectionIT.java
sfc-gh-bmikaili: You're right — `databaseExists` is `void` and throws `ERROR_2001` on SQL failure. My bad. The try/finally fully covers this.

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
sfc-gh-bmikaili: Fair — IT coverage works for the happy path. Not blocking.

sfc-gh-skurella:
@sfc-gh-bmikaili

Re. (1), I think it's safe to leave as is for now. Two reasons:
1. I did a 6h soak test where I periodically called some JDBC stuff and it worked - probably once the JDBC session is established, it has its own mechanism for maintaining a secure connection.
2. This PR mirrors what we had in KC v3.

I created [SNOW-3535094](https://snowflakecomputing.atlassian.net/browse/SNOW-3535094) to add support for the refresh token to JDBC. Then we'll be able to clean up much of our current code.

Re. (2), I think this is a legit thing to do - better UX, and that's also what KC v3 did.

I updated the PR description.

[SNOW-3535094]: https://snowflakecomputing.atlassian.net/browse/SNOW-3535094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
```diff
@@ -88,13 +93,40 @@ static Properties makeJdbcDriverProperties(SinkTaskConfig config, SnowflakeURL u
...
+            config.getOauthClientId().orElseThrow(SnowflakeErrors.ERROR_0026::getException);
+        Password oauthClientSecret =
+            config.getOauthClientSecret().orElseThrow(SnowflakeErrors.ERROR_0027::getException);
+        URL oauthUrl =
+            config.getOauthTokenEndpoint().isPresent()
+                ? OAuthURL.from(config.getOauthTokenEndpoint().get())
+                : url;
+        properties.put(
+            JDBC_TOKEN,
+            OAuthAccessTokenFetcher.fetchAccessToken(
```
sfc-gh-bmikaili, line 108: `validate()` in `SnowflakeStreamingSinkConnector` maps only error codes `1001/0007/0018/0013/0002` in its catch block; the `default:` arm re-throws. The OAuth path added here can throw `ERROR_1004` (token fetch failed) and `OAuthURL.from(...)` can throw `ERROR_0033` (invalid OAuth URL) during the live token fetch that `validate()` performs. Both codes hit `default:`, so the exception escapes `validate()` unhandled and surfaces as a Connect worker error instead of a clean per-field config error.

Consider mapping `1004`/`0033` to the relevant fields (`snowflake.oauth.client.id`/`client.secret` and the token endpoint) via `Utils.updateConfigErrorMessage`, so a bad OAuth credential or endpoint gives the user actionable feedback. Non-blocking.

sfc-gh-bmikaili (approved):
Approving. Left one non-blocking note re: validate() OAuth error mapping.

src/main/java/com/snowflake/kafka/connector/SnowflakeStreamingSinkConnector.java
```diff
@@ -272,6 +272,27 @@ public Config validate(Map<String, String> connectorConfigs) {
               KafkaConnectorConfigParams.SNOWFLAKE_PRIVATE_KEY,
               " must be a valid PEM RSA private key");
           break;
+        case "0033":
+          Utils.updateConfigErrorMessage(
+              result,
+              KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_TOKEN_ENDPOINT,
+              " is not a valid OAuth token endpoint URL");
+          break;
+        case "1004":
```
sfc-gh-bmikaili, line 313: Nice to see `0033` and `1004` mapped here now. One gap remains: `validate()` doesn't run the OAuth cross-field validator — `connectorConfigValidator.validateConfig(...)` is only called from `start()`, not from `validate()`. So a config with `authenticator=oauth` and a blank/missing `oauth_client_id` or `oauth_client_secret` passes single-field validation, reaches the live token fetch, and `makeJdbcDriverProperties` throws `ERROR_0026`/`ERROR_0027`. Those codes fall through to `default: throw e`, escaping `validate()` as an unhandled worker error instead of a clean per-field config message.

Suggest adding cases mirroring the `0033`/`1004` handling:

```java
case "0026":
  Utils.updateConfigErrorMessage(result,
      KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_CLIENT_ID,
      " must be non-empty when using oauth authenticator");
  break;
case "0027":
  Utils.updateConfigErrorMessage(result,
      KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_CLIENT_SECRET,
      " must be non-empty when using oauth authenticator");
  break;
```

PR #1449, opened 2026-05-12
add OAuth support to streaming SDK and connector validation

sfc-gh-skurella:

PR #1445, opened 2026-05-11
WIP

sfc-gh-skurella:

PR #1444, opened 2026-05-11
actually use StreamingClientProperties

sfc-gh-skurella:
Turns out, neither `clientName` nor `clientProperties` from `StreamingClientProperties::from` were in use - its only utility was for `parameterOverrides`. We had two different ways of creating `Properties` out of the `SinkTaskConfig` and one was only relevant for tests.

This PR moves SSv2 SDK property creation from `StreamingClientFactory` into `StreamingClientProperties` - the old code in `StreamingClientProperties` is effectively thrown away.

PR #1451, opened 2026-05-12
use SinkTaskConfig to configure JDBC

sfc-gh-skurella:
use SinkTaskConfig in makeJdbcDriverProperties

use SinkTaskConfig in SnowflakeConnectionServiceBuilder

use SinkTaskConfig when building JDBC/proxy Properties

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
@@ -2,7 +2,6 @@ package com.snowflake.kafka.connector.internal;
 
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
-import com.snowflake.kafka.connector.Constants.KafkaConnectorConfigParams;
 import com.snowflake.kafka.connector.Utils;
 import com.snowflake.kafka.connector.config.SinkTaskConfig;
 import java.sql.ResultSet;
@@ -10,7 +9,6 @@ import java.sql.SQLException;
 import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
-import java.util.Map;
 import java.util.Properties;
 
 public class InternalUtils {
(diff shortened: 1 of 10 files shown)


sfc-gh-skurella:
* **#1448** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1448?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1447** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1447?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1451** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1451?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1451?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1446** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1446?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1455](https://github.com/snowflakedb/snowflake-kafka-connector/pull/1455) <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1455?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1444** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1444?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1445](https://github.com/snowflakedb/snowflake-kafka-connector/pull/1445) <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1445?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`


<h2></h2>

sfc-gh-ppopov (approved):

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
```diff
@@ -75,65 +74,31 @@ static String timestampToDate(long time) {
...
-          role = entry.getValue();
-          break;
-        default:
-          // ignore unrecognized keys
-      }
-    }
+    putIfNotBlank(properties, JDBC_DATABASE, config.getSnowflakeDatabase());
+    putIfNotBlank(properties, JDBC_SCHEMA, config.getSnowflakeSchema());
+    putIfNotBlank(properties, JDBC_USER, config.getSnowflakeUser());
+    putIfNotBlank(properties, JdbcPropertyKeys.ROLE, config.getSnowflakeRole());
```
sfc-gh-bmikaili, line 89: **Silent loss of case-insensitive key lookup.**

The old `makeJdbcDriverPropertiesFromConnectorConfiguration` iterated `connectorConfiguration.entrySet()` and matched on `entry.getKey().toLowerCase()`, so a config with a mixed-case key like `"SNOWFLAKE.DATABASE.NAME"` would still populate `JDBC_DATABASE`. The new path in `SinkTaskConfig.builderFrom` reads keys with exact-case `config.get(KafkaConnectorConfigParams.SNOWFLAKE_DATABASE_NAME)`, so those mixed-case entries are now silently dropped — and since the typed getter then returns null, `makeJdbcDriverProperties` falls through to `ERROR_0014/0015/0016` ("Missed snowflake ... name in connector config"). A user who previously had a working config could now get a confusing startup failure.

My read is that this is probably dead defensive code — Kafka Connect's own `ConfigDef` normalizes keys before delivery to `SinkTask.start()`, and our `ConnectorConfigDefinition` registers keys in lowercase. But `SnowflakeStreamingSinkConnector.validate()` and the `setProperties(Map)` shim are also callers that don't necessarily go through Connect's normalization — worth one last audit.

Could you either (a) restore the case-insensitive lowercasing inside `SinkTaskConfig.builderFrom`, or (b) drop a note on the PR confirming this is intentional, plus a unit test with an uppercase key asserting the new behavior? Also note `InternalUtils.assertNotEmpty` still does `toLowerCase()` on entries, which becomes inconsistent either way.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceFactory.java
```diff
@@ -30,17 +28,21 @@ public SnowflakeConnectionServiceBuilder setTaskID(String taskID) {
     }
 
     public SnowflakeConnectionServiceBuilder setProperties(Map<String, String> conf) {
-      if (!conf.containsKey(KafkaConnectorConfigParams.SNOWFLAKE_URL_NAME)) {
+      return setProperties(SinkTaskConfig.from(conf, true));
+    }
+
+    public SnowflakeConnectionServiceBuilder setProperties(SinkTaskConfig parsedConfig) {
+      if (parsedConfig.getSnowflakeUrl() == null || parsedConfig.getSnowflakeUrl().isEmpty()) {
```
sfc-gh-bmikaili, line 35: **Error-code change for empty-string URL.**

Old code guarded on `conf.containsKey(SNOWFLAKE_URL_NAME)`, so a config containing the key with an empty-string value passed this check and then failed later inside `new SnowflakeURL("")` with whatever exception that constructor throws.

New code treats `""` and missing identically — both now throw `ERROR_0017`. I consider this a strict improvement (fail-fast-and-loud), but it's a user-visible error-code change for a config shape that may exist in the wild (e.g., an operator templating the URL from an empty env var). Worth a one-liner in the commit message so it shows up in the release notes.

src/test/java/com/snowflake/kafka/connector/internal/InternalUtilsTest.java
```diff
@@ -157,15 +163,92 @@ public void testResultSize() throws SQLException {
...
+            .snowflakePrivateKey(pemKey)
+            .snowflakeUrl(url.getFullUrl())
+            .build();
+
+    Properties props = InternalUtils.makeJdbcDriverProperties(taskConfig, url);
+
+    assertFalse(
+        props.containsKey(JdbcPropertyKeys.ROLE),
+        "JDBC properties should not contain role when role is blank");
+  }
```
sfc-gh-bmikaili, line 253: **Test gap: empty-string / whitespace role.**

`makeJdbcDriverProperties_noRole_omitsRoleProperty` only exercises the missing-role case (role absent from the builder). The old code used `isBlank(role)`, which covers `null`, `""`, and `"   "`; the new `putIfNotBlank` helper preserves that contract — good — but the test doesn't lock it in. A future change that swaps `isBlank` for `!= null` would pass this test and silently regress the whitespace-role behavior.

Please add two more cases: one where `.snowflakeRole("")` is set, and one with `"   "`. Both should still assert the role property is absent from the returned `Properties`.

sfc-gh-bmikaili (commented):
Reviewed the 3-commit migration to `SinkTaskConfig` for JDBC/proxy property plumbing. The refactor is clean and the stacked-PR hygiene is good (only the 3 advertised commits, no parent-PR leakage). `putIfNotBlank` is a nice simplification over the per-field `isBlank` blocks, and dropping the `noCaching()` mutation-after-the-fact is a genuine win.

Three correctness concerns inline below — #1 is the only one I'd block on. Credential handling looks preserved (private key / passphrase / proxy password pass straight through typed getters; no new log sites).

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
sfc-gh-skurella: The config validator would throw if the correct, i.e. lowercase, keys were not present. So the lowercasing was essentially dead code.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeConnectionServiceFactory.java
sfc-gh-skurella: This is also effectively dead code since the validator will ensure the URL is set before this ever runs.

src/test/java/com/snowflake/kafka/connector/internal/InternalUtilsTest.java
sfc-gh-skurella: Added 2 test cases.

sfc-gh-skurella:
### Merge activity

* **May 13, 4:49 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1451).
* **May 13, 4:52 PM UTC**: [Graphite](https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1451) rebased this pull request as part of a merge.
* **May 13, 5:28 PM UTC**: @sfc-gh-skurella merged this pull request with [Graphite](https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1451).

PR #1460, opened 2026-05-18
Fix recovery seek-overwrite race in batch loop

sfc-gh-bmikaili:
## Summary

Fixes a silent data-loss race in `SnowflakeSinkServiceV2.insert(Collection)` where channel recovery's Kafka seek gets overwritten by the batch loop's end-of-batch rewind. Reproducible on master with `test_invalidation_offset_consistency` once `SYSTEM\$STREAMING_CHANNEL_INVALIDATE` is deployed (Seb's recent finding).

## The race

1. `appendRow` throws `InvalidChannelError` → fallback fires `reopenChannel` → tracker's `resetAfterRecovery(committed)` issues `sinkTaskContext.offset(tp, committed+1)` (**seek #1**).
2. `insertRowWithFallback` returns false → `transformAndSend` returns false → batch loop adds `tp -> failingOffset` to `offsetsOfFirstSkippedRecord`.
3. End of `put()` → `offsetsOfFirstSkippedRecord.forEach(sinkTaskContext::offset)` → `sinkTaskContext.offset(tp, failingOffset)` (**seek #2**).
4. Last writer wins; consumer seeks to `failingOffset`, skipping records `committed+1..failingOffset-1`.

## Fix

- **`InsertResult` enum** replaces `boolean` from `insert`/`insertRecord`. New `RECOVERY_IN_FLIGHT` value tells the batch loop "tracker already issued the authoritative seek; do NOT add this partition to the rewind map."
- **Batch loop** is a single switch on `InsertResult`. `RECOVERY_IN_FLIGHT` and `PROCESSED` land in the no-rewind branch.
- **`insert(record)` rebuild gate** simplified to `existing.isEmpty()`. Master's `isChannelClosed` check would rebuild the partition when `this.channel` completed exceptionally, wiping `PartitionOffsetTracker` recovery state.
- **Max-recoveries exhaustion** throws `ConnectException` (not `TopicPartitionChannelInsertionException`, which `transformAndSend` swallows → silent loss).
- **`BackpressureException`** now caught inside `transformAndSend` and returned as `BACKPRESSURE_TRIGGERED` — flow control out of exception handling.
- **`withDelay`** restructured: native exception propagation instead of re-wrapping in `RuntimeException`.

## Test plan

- [x] Unit tests pass (`SnowflakeSinkServiceV2Test`, `SnowpipeStreamingPartitionChannelTest`).
- [x] Reproduces against `test_invalidation_offset_consistency` once `SYSTEM\$STREAMING_CHANNEL_INVALIDATE` is enabled on the account.
- Stacked above: #1430 (client recreation) + #1432 (mitmproxy E2E).

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkService.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkService.java
@@ -2,6 +2,7 @@ package com.snowflake.kafka.connector.internal;
 
 import com.codahale.metrics.MetricRegistry;
 import com.google.common.annotations.VisibleForTesting;
+import com.snowflake.kafka.connector.internal.streaming.channel.InsertResult;
 import com.snowflake.kafka.connector.internal.streaming.channel.TopicPartitionChannel;
 import java.util.Collection;
 import java.util.Map;
@@ -36,11 +37,11 @@ public interface SnowflakeSinkService {
   /**
    * call pipe to insert a JSON record will not trigger time based flush
    *
-   * @param record record content
-   * @return true if the record was processed successfully, false if recovery was triggered and the
-   *     caller should stop feeding records to this partition for the remainder of the batch
+   * @return the outcome — see {@link
+   *     com.snowflake.kafka.connector.internal.streaming.channel.InsertResult}. The batch loop uses
+   *     this to decide whether to rewind Kafka and whether to enter backpressure cooldown.
    */
-  boolean insert(final SinkRecord record);
+  InsertResult insert(final SinkRecord record);
 
   /**
    * retrieve offset of last loaded record for given pipe name
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
@@ -17,8 +17,8 @@ import com.snowflake.kafka.connector.internal.SnowflakeErrors;
 import com.snowflake.kafka.connector.internal.SnowflakeSinkService;
 import com.snowflake.kafka.connector.internal.metrics.MetricsJmxReporter;
 import com.snowflake.kafka.connector.internal.metrics.TaskMetrics;
+import com.snowflake.kafka.connector.internal.streaming.channel.InsertResult;
 import com.snowflake.kafka.connector.internal.streaming.channel.TopicPartitionChannel;
-import com.snowflake.kafka.connector.internal.streaming.v2.BackpressureException;
 import com.snowflake.kafka.connector.internal.streaming.v2.client.StreamingClientPools;
 import com.snowflake.kafka.connector.internal.streaming.v2.service.BatchOffsetFetcher;
 import com.snowflake.kafka.connector.internal.streaming.v2.service.PartitionChannelManager;
@@ -338,57 +338,47 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
           initializingPartitions);
     }
 
-    // If still in cooldown from a recent backpressure event, treat all partitions as
-    // backpressured so we skip the entire batch and give the SDK time to drain.
-    boolean skipAllPartitions = false;
-    if (Instant.now().isBefore(backpressureUntil)) {
-      LOGGER.debug(
-          "Backpressure cooldown active until {}. Skipping entire batch.", backpressureUntil);
-      skipAllPartitions = true;
-    }
-
     Map<TopicPartition, Long> offsetsOfFirstSkippedRecord = new HashMap<>();
-    boolean newBackpressure = false;
     for (SinkRecord record : records) {
-      // check if it needs to handle null value records
       if (shouldSkipNullValue(record)) {
         continue;
       }
-
       TopicPartition tp = new TopicPartition(record.topic(), record.kafkaPartition());
 
+      InsertResult result;
       if (offsetsOfFirstSkippedRecord.containsKey(tp)) {
-        // We've already skipped a record in this partition, so should also skip the remaining
-        // records in this partition.
-        continue;
-      }
-      if (skipAllPartitions || initializingPartitions.contains(tp)) {
-        // Make sure we store the first record in each partition that we skipped so we can correctly
-        // rewind the offset.
-        offsetsOfFirstSkippedRecord.putIfAbsent(tp, record.kafkaOffset());
-        continue;
+        result = InsertResult.PREVIOUSLY_SKIPPED_IN_BATCH;
+      } else if (Instant.now().isBefore(backpressureUntil)) {
+        result = InsertResult.IN_BACKPRESSURE_COOLDOWN;
+      } else if (initializingPartitions.contains(tp)) {
+        result = InsertResult.CHANNEL_INITIALIZING;
+      } else {
+        result = insert(record);
       }
 
-      try {
-        if (!insert(record)) {
+      switch (result) {
+        case PROCESSED:
+        case RECOVERY_IN_FLIGHT:
+          // PROCESSED: nothing to rewind. RECOVERY_IN_FLIGHT: the channel's resetAfterRecovery
+          // owns the authoritative seek; our rewind would clobber it.
+          break;
+
+        case PREVIOUSLY_SKIPPED_IN_BATCH:
+          // Already in the rewind map from an earlier record in this partition.
+          break;
+
+        case CHANNEL_INITIALIZING:
+        case IN_BACKPRESSURE_COOLDOWN:
           offsetsOfFirstSkippedRecord.putIfAbsent(tp, record.kafkaOffset());
-        }
-      } catch (BackpressureException e) {
-        LOGGER.warn(
-            "Backpressure on partition {}. Skipping remaining records for this partition."
-                + " Exception: {}",
-            tp,
-            e.getMessage());
-        taskMetrics.incBackpressureRewindCount();
-        offsetsOfFirstSkippedRecord.putIfAbsent(tp, record.kafkaOffset());
-        skipAllPartitions = true;
-        newBackpressure = true;
-      }
-    }
+          break;
 
-    if (newBackpressure) {
-      backpressureUntil = Instant.now().plus(BACKPRESSURE_COOLDOWN);
-      LOGGER.info("Backpressure cooldown set until {}", backpressureUntil);
+        case BACKPRESSURE_TRIGGERED:
+          LOGGER.warn("Backpressure on partition {}. Starting cooldown.", tp);
+          taskMetrics.incBackpressureRewindCount();
+          offsetsOfFirstSkippedRecord.putIfAbsent(tp, record.kafkaOffset());
+          backpressureUntil = Instant.now().plus(BACKPRESSURE_COOLDOWN);
+          break;
+      }
     }
 
     if (!offsetsOfFirstSkippedRecord.isEmpty()) {
@@ -398,21 +388,26 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
   }
 
   /**
-   * Inserts individual records into buffer. It fetches the TopicPartitionChannel from the map and
-   * then each partition(Streaming channel) calls its respective appendRows API
+   * Inserts a single record. Fetches the channel (creating one on first contact) and delegates to
+   * {@link TopicPartitionChannel#insertRecord}.
+   *
+   * <p>We never rebuild an already-registered partition from scratch here. Rebuilding would
+   * construct a fresh {@link
+   * com.snowflake.kafka.connector.internal.streaming.v2.channel.PartitionOffsetTracker} and wipe
+   * the recovery state (pending seek, processedOffset). Recovery for an existing channel goes
+   * through the appendRow fallback path ({@code reopenChannel} / {@code
+   * recreateClientAndReopenChannel}), which preserves the tracker. The manager removes entries only
+   * on explicit {@link #close}/{@link #closeAll}, so {@code isEmpty()} cleanly distinguishes "first
+   * time we see this partition" from "recovery needed".
    */
   @Override
-  public boolean insert(SinkRecord record) {
+  public InsertResult insert(SinkRecord record) {
     LOGGER.trace("Inserting record: {}", record);
 
     TopicPartition topicPartition = new TopicPartition(record.topic(), record.kafkaPartition());
 
-    // Initialize a new topic partition if it's not in the cache or if the channel is closed.
-    if (channelManager
-        .getChannel(topicPartition)
-        .map(TopicPartitionChannel::isChannelClosed)
-        .orElse(true)) {
-      LOGGER.warn("Streaming channel doesn't exist or is closed for {}", topicPartition);
+    if (channelManager.getChannel(topicPartition).isEmpty()) {
+      LOGGER.warn("Streaming channel missing for {}, creating", topicPartition);
       startPartition(topicPartition);
     }
 
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/InsertResult.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/InsertResult.java
@@ -1,0 +1,39 @@
+package com.snowflake.kafka.connector.internal.streaming.channel;
+
+/**
+ * Outcome of attempting to insert one record into a {@link TopicPartitionChannel}, or — when
+ * returned by the batch loop's pre-check — the reason the record was not even attempted.
+ *
+ * <p>Each value names a distinct reason so the batch loop can decide whether to rewind Kafka (add
+ * to the skipped-offsets map) and whether to trigger backpressure cooldown, without having to
+ * reason about two-writer races between {@code sinkTaskContext.offset} callers.
+ */
+public enum InsertResult {
+  /** appendRow succeeded and the record is in the SDK buffer. */
+  PROCESSED,
+
+  /**
+   * Tracker has a Kafka seek in flight (either from this call's fallback, or from an earlier
+   * batch's recovery that hasn't landed yet). The tracker owns the rewind; do NOT add to the skip
+   * map — doing so would overwrite the authoritative seek.
+   */
+  RECOVERY_IN_FLIGHT,
+
+  /** Channel has not finished opening. Rewind so Kafka redelivers once it's ready. */
+  CHANNEL_INITIALIZING,
+
+  /** Prior batch triggered backpressure; we're still in cooldown. Rewind to try later. */
+  IN_BACKPRESSURE_COOLDOWN,
+
+  /**
+   * Earlier record in this batch was added to the rewind map for this partition (because of {@link
+   * #CHANNEL_INITIALIZING}, {@link #IN_BACKPRESSURE_COOLDOWN}, or {@link #BACKPRESSURE_TRIGGERED});
+   * skip the rest. Does NOT apply to {@link #RECOVERY_IN_FLIGHT} — that partition is never added to
+   * the rewind map in the first place, so each record is independently classified as
+   * RECOVERY_IN_FLIGHT by the in-channel check.
+   */
+  PREVIOUSLY_SKIPPED_IN_BATCH,
+
+  /** appendRow returned a retryable backpressure error. Rewind + start cooldown. */
+  BACKPRESSURE_TRIGGERED
+}
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/TopicPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/TopicPartitionChannel.java
@@ -11,7 +11,7 @@ public interface TopicPartitionChannel {
   long NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE = -1L;
 
   /**
-   * Inserts the record into buffer
+   * Inserts the record into the SDK buffer.
    *
    * <p>Step 1: Initializes this channel by fetching the offsetToken from Snowflake for the first
    * time this channel/partition has received offset after start/restart.
@@ -19,14 +19,10 @@ public interface TopicPartitionChannel {
    * <p>Step 2: Decides whether given offset from Kafka needs to be processed and whether it
    * qualifies for being added into buffer.
    *
-   * @param kafkaSinkRecord input record from Kafka
-   * @param isFirstRowPerPartitionInBatch indicates whether the given record is the first record per
-   *     partition in a batch
-   * @return true if the record was processed (or legitimately skipped as a duplicate), false if
-   *     recovery was triggered and the caller should stop feeding records to this partition for the
-   *     remainder of the batch
+   * @return the outcome — see {@link InsertResult}. The batch loop uses this to decide whether to
+   *     rewind Kafka and whether to enter backpressure cooldown.
    */
-  boolean insertRecord(SinkRecord kafkaSinkRecord, boolean isFirstRowPerPartitionInBatch);
+  InsertResult insertRecord(SinkRecord kafkaSinkRecord, boolean isFirstRowPerPartitionInBatch);
 
   /**
    * Asynchronously closes a channel associated to this partition. Any {@link SFException} occurred
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithFallbackPolicy.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithFallbackPolicy.java
@@ -33,23 +33,16 @@ class AppendRowWithFallbackPolicy {
    * @param channelName the channel name for logging purposes
    */
   private static void withDelay(CheckedRunnable action, String channelName) throws Throwable {
+    long delayMs = FALLBACK_DELAY.toMillis() + (long) (Math.random() * JITTER_DURATION.toMillis());
+    LOGGER.info("Delaying channel recovery by {}ms for channel: {}", delayMs, channelName);
     try {
-      long delayMs =
-          FALLBACK_DELAY.toMillis() + (long) (Math.random() * JITTER_DURATION.toMillis());
-
-      LOGGER.info("Delaying channel recovery by {}ms for channel: {}", delayMs, channelName);
       Thread.sleep(delayMs);
-
-      LOGGER.info("Executing channel recovery for channel: {}", channelName);
-      action.run();
     } catch (InterruptedException e) {
       Thread.currentThread().interrupt();
-    } catch (SFException e) {
-      // Re-throw SFException unchanged so Fallback can handle it properly
-      throw e;
-    } catch (Exception e) {
-      throw new RuntimeException(e);
+      return;
     }
+    LOGGER.info("Executing channel recovery for channel: {}", channelName);
+    action.run();
   }
 
   /**
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
@@ -24,6 +24,7 @@ import com.snowflake.kafka.connector.internal.schemaevolution.SnowflakeSchemaEvo
 import com.snowflake.kafka.connector.internal.schemaevolution.ValidationResultMapper;
 import com.snowflake.kafka.connector.internal.streaming.StreamingErrorHandler;
 import com.snowflake.kafka.connector.internal.streaming.TopicPartitionChannelInsertionException;
+import com.snowflake.kafka.connector.internal.streaming.channel.InsertResult;
 import com.snowflake.kafka.connector.internal.streaming.channel.TopicPartitionChannel;
 import com.snowflake.kafka.connector.internal.streaming.telemetry.SnowflakeTelemetryChannelCreation;
 import com.snowflake.kafka.connector.internal.streaming.telemetry.SnowflakeTelemetryChannelStatus;
@@ -164,14 +165,17 @@ public class SnowpipeStreamingPartitionChannel implements TopicPartitionChannel 
   }
 
   @Override
-  public boolean insertRecord(SinkRecord kafkaSinkRecord, boolean isFirstRowPerPartitionInBatch) {
+  public InsertResult insertRecord(
+      SinkRecord kafkaSinkRecord, boolean isFirstRowPerPartitionInBatch) {
     if (offsetTracker.shouldProcess(kafkaSinkRecord.kafkaOffset(), isFirstRowPerPartitionInBatch)) {
       return transformAndSend(kafkaSinkRecord);
     }
-    return true;
+    // shouldProcess returned false — the record is either a duplicate (already processed) or
+    // skipped because needToSkipCurrentBatch is set after a mid-batch recovery. Nothing to do.
+    return InsertResult.PROCESSED;
   }
 
-  private boolean transformAndSend(SinkRecord kafkaSinkRecord) {
+  private InsertResult transformAndSend(SinkRecord kafkaSinkRecord) {
     try {
       final long kafkaOffset = kafkaSinkRecord.kafkaOffset();
       final SnowflakeSinkRecord record =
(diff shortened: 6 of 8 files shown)


sfc-gh-bmikaili:
* **#1430** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1430?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1431](https://github.com/snowflakedb/snowflake-kafka-connector/pull/1431) <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1431?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1432](https://github.com/snowflakedb/snowflake-kafka-connector/pull/1432) <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1432?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1460** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1460?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1460?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
probe

sfc-gh-bmikaili (commented):
Per-file WHY notes for the seek-overwrite race fix. Moved from PR #1430 after the split.

src/main/java/com/snowflake/kafka/connector/internal/SnowflakeSinkService.java
```diff
@@ -36,11 +37,11 @@ public interface SnowflakeSinkService {
...
    *
-   * @param record record content
-   * @return true if the record was processed successfully, false if recovery was triggered and the
-   *     caller should stop feeding records to this partition for the remainder of the batch
+   * @return the outcome — see {@link
+   *     com.snowflake.kafka.connector.internal.streaming.channel.InsertResult}. The batch loop uses
+   *     this to decide whether to rewind Kafka and whether to enter backpressure cooldown.
    */
-  boolean insert(final SinkRecord record);
+  InsertResult insert(final SinkRecord record);
```
sfc-gh-bmikaili, line 44: **WHY:** `insert` now returns `InsertResult` instead of `boolean`. The boolean couldn't distinguish 'record inserted' from 'fallback fired and owns the rewind' — which led to the batch loop overwriting a fallback-issued Kafka seek with its own. The enum carries that distinction; every caller was updated.

src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/TopicPartitionChannel.java
```diff
@@ -11,22 +11,18 @@ public interface TopicPartitionChannel {
...
-   * @param isFirstRowPerPartitionInBatch indicates whether the given record is the first record per
-   *     partition in a batch
-   * @return true if the record was processed (or legitimately skipped as a duplicate), false if
-   *     recovery was triggered and the caller should stop feeding records to this partition for the
-   *     remainder of the batch
+   * @return the outcome — see {@link InsertResult}. The batch loop uses this to decide whether to
+   *     rewind Kafka and whether to enter backpressure cooldown.
    */
-  boolean insertRecord(SinkRecord kafkaSinkRecord, boolean isFirstRowPerPartitionInBatch);
+  InsertResult insertRecord(SinkRecord kafkaSinkRecord, boolean isFirstRowPerPartitionInBatch);
```
sfc-gh-bmikaili, line 25: **WHY:** same signature change as the service interface — `insertRecord` now returns `InsertResult`. Pure plumbing.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -398,21 +388,26 @@ public void insert(final Collection<SinkRecord> records) {
...
 
     TopicPartition topicPartition = new TopicPartition(record.topic(), record.kafkaPartition());
 
-    // Initialize a new topic partition if it's not in the cache or if the channel is closed.
-    if (channelManager
-        .getChannel(topicPartition)
-        .map(TopicPartitionChannel::isChannelClosed)
-        .orElse(true)) {
-      LOGGER.warn("Streaming channel doesn't exist or is closed for {}", topicPartition);
+    if (channelManager.getChannel(topicPartition).isEmpty()) {
```
sfc-gh-bmikaili, line 409: **WHY (silent data loss fix): timeline of the bug the old gate enabled.**

### Initial state
- `PartitionChannelManager.partitionChannels[tp]` → SSPC instance `A` (opened at startup).
- `A.channel` = completed `CompletableFuture<SnowflakeStreamingIngestChannel>`.
- `A.offsetTracker.processedOffset = 149`. Snowflake committed = 99.
- Shared SDK client was just marked invalid by a background 410 (e.g. status cache refresh in `core::prod::channel_status_cache`).

### Timeline (all on the task thread unless noted)

**T=0** — Kafka Connect → `SnowflakeSinkTask.put([r150..r199])` → `SnowflakeSinkServiceV2.insert(Collection)` (`SnowflakeSinkServiceV2.java:338`).

**T=1** — loop iter r150 → `classifyAndInsert` → `insert(r150)` (`SnowflakeSinkServiceV2.java:406`) → old gate passes (`A` still present, `isChannelClosed=false`) → `A.insertRecord(r150)`.

**T=2** — `A.insertRecord` → `A.transformAndSend` → `A.insertRowWithFallback` (`SnowpipeStreamingPartitionChannel.java:307`) → `A.getChannel().appendRow(...)`.

**T=3** — SDK throws `SFException(InvalidClientError)` (Rust: `channel.rs:invalidate_channel` marked `is_locally_valid=false` when the background 410 hit).

**T=4** — `Failsafe` fallback in `AppendRowWithFallbackPolicy.executeWithFallback` → runs the fallback lambda at `SnowpipeStreamingPartitionChannel.java:315`.

**T=5** — fallback detects `ClientRecreationException.isClientInvalidError(ex)=true` → calls `A.reopenChannel("CLIENT_RECREATION", recreateClient=true)` (`SnowpipeStreamingPartitionChannel.java:369`).

**T=6** — `reopenChannel` calls `A.clientRecreator.recreate(...)` → `StreamingClientPools.recreateClient` → CAS-swaps the shared client. **The replacement client is also seeing 410s** (pipe failover still propagating — typical 1–3 min window).

**T=7** — `reopenChannel` rebinds `A.channel` to `A.channel.thenAccept(closeOld).exceptionally(noop).thenApply(open+reset)`. Previous `A.channel` is completed, so the chain runs **inline on the task thread**.

**T=8** — inside the `thenApply` body: `A.openChannelForTable` (`SnowpipeStreamingPartitionChannel.java:~460`) → SDK → `InvalidClientError` (replacement client also invalid). Body throws. `A.channel` now completes **exceptionally**.

**T=9** — `insertRowWithFallback` returns `false` → `transformAndSend` returns `InsertResult.RECOVERY_IN_FLIGHT` → batch loop switch → no rewind entry. Loop advances to r151.

**T=10** — iter r151 → `classifyAndInsert` → `insert(r151)` → **old gate runs**:
- `channelManager.getChannel(tp)` → present (entry `A` is still in the map; `PartitionChannelManager.partitionChannels.remove` is only called by `close(partitions)` / `closeAll()` on rebalance or stop).
- `A.isInitializing()` = `!A.channel.isDone()` = **false** (future IS done, just done-exceptionally).
- `A.isChannelClosed()` (`SnowpipeStreamingPartitionChannel.java:819`) calls `A.channel.join()` → throws `CompletionException` → `catch (RuntimeException)` returns **true** ("failed to initialize, treating as closed").
- Old gate `!isInitializing() && isChannelClosed()` → **true** → `SSS2.startPartition(tp)`.

**T=11** — `startPartition` → `PartitionChannelManager.startPartitions` (`PartitionChannelManager.java:123`) → constructs SSPC instance `B` and does `partitionChannels.put(channelName, B)`. Instance `A` (with `processedOffset=149`) is now unreachable. A fresh `PartitionOffsetTracker(processedOffset=-1)` is built for `B` (`PartitionChannelManager.java:178`).

**T=12** — batch loop re-fetches the channel; now gets `B`. `B.insertRecord(r151)` → `B.offsetTracker.shouldProcess(151)` → `processedOffset=-1`, `151 > -1` → returns **true**. Record accepted.

**T=13** — `B.appendRow(...)` on the (eventually) open new SDK channel succeeds for r151..r199. `B.offsetTracker.processedOffset` advances from `-1` to `199`. Records 100..150 never appended anywhere.

**T=14** — end of `put()`. No rewind issued (nothing in `offsetsOfFirstSkippedRecord` for tp).

**T=15** — Kafka Connect → `SnowflakeSinkTask.preCommit` → reports 199 for tp. Kafka commits 199. **Records 100..150 are now past the committed offset; never redelivered. Silent loss.**

### The fix

At **T=10**, new gate is `existing.isEmpty()` only (`SnowflakeSinkServiceV2.java:411`):
- `A` is still in the map → gate returns false → no `startPartition` call.
- Batch loop runs `A.insertRecord(r151)` → `A.getChannel().join()` throws → fallback at T=4 fires again.
- On this retry, the replacement client has had `FALLBACK_DELAY` (5s + jitter) to stabilize. `openChannelForTable` succeeds → `A.offsetTracker.resetAfterRecovery(99)` (`PartitionOffsetTracker.java:~160`):
  - `sinkTaskContext.offset(tp, 100)` → **pending Kafka seek to 100**.
  - `processedOffset.set(99)` (was 149, now 99).
  - `needToSkipCurrentBatch = true`.
- `insertRowWithFallback` returns false → `transformAndSend` returns `RECOVERY_IN_FLIGHT` → batch loop skips rewind for tp.
- Remaining records r152..r199 hit `shouldProcess` → `needToSkipCurrentBatch=true` → return false.
- End of `put()` → no rewind overwrite → pending seek to 100 stands.
- Next `poll()` → Kafka consumer seeks to 100 → replays r100..r199. No loss.

### Why `isEmpty()` alone is safe

`PartitionChannelManager.partitionChannels.remove` is called exclusively from `close(partitions)` (`PartitionChannelManager.java:326`) and `closeAll` (`PartitionChannelManager.java:293`), both invoked by Kafka Connect on rebalance/stop. Nothing else removes entries. So `isEmpty()=true` unambiguously means "first time the task has seen this partition," which is the only case where `startPartition` (and a fresh tracker) is correct.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
```diff
@@ -203,32 +207,34 @@ private boolean transformAndSend(SinkRecord kafkaSinkRecord) {
...
-            // recordProcessed() here — that would advance processedOffset past the
-            // recovery point and cause replayed offsets to be skipped. See SNOW-3344243.
-            return false;
+            // Fallback fired: the record was NOT inserted. The fallback's resetAfterRecovery
+            // already issued the authoritative Kafka seek to (committed + 1). Do NOT call
+            // recordProcessed() here — that would advance processedOffset past the recovery
+            // point and cause replayed offsets to be skipped. See SNOW-3344243. The batch
+            // loop receives RECOVERY_IN_FLIGHT and skips its own end-of-batch rewind for
+            // this partition, so the authoritative seek survives.
+            return InsertResult.RECOVERY_IN_FLIGHT;
```
sfc-gh-bmikaili, line 221: **WHY (don't call `recordProcessed` on fallback):** when `insertRowWithFallback` returns false, the fallback's `resetAfterRecovery(committed)` has already set `processedOffset = committed` and issued a seek. Calling `recordProcessed(kafkaOffset)` here would overwrite `committed` with the failing record's offset, which would cause `shouldProcess` on the replayed records to skip them. This comment pins down an issue that already bit us (SNOW-3344243).

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
```diff
@@ -283,13 +289,19 @@ private boolean insertRowWithFallback(Map<String, Object> row, long offset) {
...
+            // TopicPartitionChannelInsertionException — transformAndSend's catch swallows that
+            // and commits the offset without calling recordProcessed, silently losing the
+            // record). ConnectException fails the Kafka Connect task so offsets are not
+            // committed; the task retries recovery on restart.
             LOGGER.error(
                 "Channel {} exceeded max consecutive recoveries ({}), giving up",
                 this.channelName,
                 MAX_CONSECUTIVE_RECOVERIES);
-            throw new TopicPartitionChannelInsertionException(
+            throw new ConnectException(
```
sfc-gh-bmikaili, line 301: **WHY (throw ConnectException not TopicPartitionChannelInsertionException on max retries):** `transformAndSend`'s outer `catch (TopicPartitionChannelInsertionException)` swallows the exception and returns `PROCESSED`, which would then call `recordProcessed` on the failing offset and commit past it — silent loss. `ConnectException` propagates all the way to Kafka Connect, which fails the task cleanly so offsets aren't committed and the task retries recovery on restart.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithFallbackPolicy.java
```diff
@@ -33,23 +33,16 @@ class AppendRowWithFallbackPolicy {
...
       Thread.currentThread().interrupt();
-    } catch (SFException e) {
-      // Re-throw SFException unchanged so Fallback can handle it properly
-      throw e;
-    } catch (Exception e) {
-      throw new RuntimeException(e);
+      return;
     }
+    LOGGER.info("Executing channel recovery for channel: {}", channelName);
+    action.run();
```
sfc-gh-bmikaili, line 45: **WHY (`withDelay` restructuring):** previously caught `SFException` to re-throw, wrapped `Exception` in `RuntimeException`, and ran `action.run()` inside the try block. That mixed the sleep's `InterruptedException` handling with the action's exception handling — confusing and easy to break. New form: sleep in try, action outside. Thrown exceptions from `action.run()` propagate naturally for Failsafe to handle.

src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/InsertResult.java
```diff
@@ -0,0 +1,39 @@
...
+ * Outcome of attempting to insert one record into a {@link TopicPartitionChannel}, or — when
+ * returned by the batch loop's pre-check — the reason the record was not even attempted.
+ *
+ * <p>Each value names a distinct reason so the batch loop can decide whether to rewind Kafka (add
+ * to the skipped-offsets map) and whether to trigger backpressure cooldown, without having to
+ * reason about two-writer races between {@code sinkTaskContext.offset} callers.
+ */
+public enum InsertResult {
+  /** appendRow succeeded and the record is in the SDK buffer. */
+  PROCESSED,
```
sfc-gh-bmikaili, line 13: **PROCESSED** — set in `SnowpipeStreamingPartitionChannel`: `insertRecord` line 183 (`shouldProcess` returned false: duplicate or `needToSkipCurrentBatch`); `transformAndSend` line 218 (client-side validation failed, handled by DLQ/tolerance); `transformAndSend` line 235 (happy path after `appendRow`); `transformAndSend` line 245 (caught `TopicPartitionChannelInsertionException`, record swallowed, Kafka will retry).

src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/InsertResult.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Tracker has a Kafka seek in flight (either from this call's fallback, or from an earlier
+   * batch's recovery that hasn't landed yet). The tracker owns the rewind; do NOT add to the skip
+   * map — doing so would overwrite the authoritative seek.
+   */
+  RECOVERY_IN_FLIGHT,
```
sfc-gh-bmikaili, line 20: **RECOVERY_IN_FLIGHT** — set in `SnowpipeStreamingPartitionChannel.transformAndSend` line 229, when `insertRowWithFallback` returns false. That path only runs after Failsafe's fallback at `SnowpipeStreamingPartitionChannel.java:315` called `reopenChannel(..., false)` or `reopenChannel(..., true)` — both rebind `this.channel` to a chain whose terminal stage calls `offsetTracker.resetAfterRecovery(committed)` → `sinkTaskContext.offset(tp, committed+1)`. That seek is the authoritative rewind; the batch loop must NOT add this partition to its own rewind map.

src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/InsertResult.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /** Channel has not finished opening. Rewind so Kafka redelivers once it's ready. */
+  CHANNEL_INITIALIZING,
```
sfc-gh-bmikaili, line 23: **CHANNEL_INITIALIZING** — set in `SnowflakeSinkServiceV2.insert(Collection)` line 354, inline pre-check. Fires when `initializingPartitions.contains(tp)`. That set is captured once at the top of the batch via `currentlyInitializing(partitions)`, which checks `channelManager.getChannel(tp).isInitializing()` = `!this.channel.isDone()`. Channel is still opening; rewind so Kafka redelivers after it's ready.

src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/InsertResult.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /** Prior batch triggered backpressure; we're still in cooldown. Rewind to try later. */
+  IN_BACKPRESSURE_COOLDOWN,
```
sfc-gh-bmikaili, line 26: **IN_BACKPRESSURE_COOLDOWN** — set in `SnowflakeSinkServiceV2.insert(Collection)` line 352, inline pre-check. Fires when `Instant.now().isBefore(backpressureUntil)`, where `backpressureUntil` was set by the previous batch's `BACKPRESSURE_TRIGGERED` handler (line 379). Gives the SDK time to drain before retrying.

src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/InsertResult.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Earlier record in this batch was added to the rewind map for this partition (because of {@link
+   * #CHANNEL_INITIALIZING}, {@link #IN_BACKPRESSURE_COOLDOWN}, or {@link #BACKPRESSURE_TRIGGERED});
+   * skip the rest. Does NOT apply to {@link #RECOVERY_IN_FLIGHT} — that partition is never added to
+   * the rewind map in the first place, so each record is independently classified as
+   * RECOVERY_IN_FLIGHT by the in-channel check.
+   */
+  PREVIOUSLY_SKIPPED_IN_BATCH,
```
sfc-gh-bmikaili, line 35: **PREVIOUSLY_SKIPPED_IN_BATCH** — set in `SnowflakeSinkServiceV2.insert(Collection)` line 350, inline pre-check. Fires when `offsetsOfFirstSkippedRecord` already contains `tp` (earlier record in this batch was classified as `CHANNEL_INITIALIZING`, `IN_BACKPRESSURE_COOLDOWN`, or `BACKPRESSURE_TRIGGERED` — the three cases that add to the rewind map). Does NOT apply to `RECOVERY_IN_FLIGHT`; that partition never enters the rewind map, and its subsequent records are filtered by `shouldProcess` via `needToSkipCurrentBatch`.

src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/InsertResult.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /** appendRow returned a retryable backpressure error. Rewind + start cooldown. */
+  BACKPRESSURE_TRIGGERED
```
sfc-gh-bmikaili, line 38: **BACKPRESSURE_TRIGGERED** — set in `SnowpipeStreamingPartitionChannel.transformAndSend` line 238, in the `catch (BackpressureException)` block. `BackpressureException` is thrown from `AppendRowWithFallbackPolicy.executeWithFallback` when Failsafe's policy classifies the error as retryable backpressure (SDK buffer full). Before this PR the exception propagated out of `insert`; now the channel catches it and returns the enum value so the batch loop doesn't need an exception handler.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -338,57 +338,47 @@ public void insert(final Collection<SinkRecord> records) {
...
+      } else if (initializingPartitions.contains(tp)) {
+        result = InsertResult.CHANNEL_INITIALIZING;
+      } else {
+        result = insert(record);
       }
 
-      try {
-        if (!insert(record)) {
+      switch (result) {
+        case PROCESSED:
```
sfc-gh-bmikaili, line 360: **PROCESSED** — maps to old `insert(record) == true`. No rewind, nothing to do.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -338,57 +338,47 @@ public void insert(final Collection<SinkRecord> records) {
...
+        case RECOVERY_IN_FLIGHT:
```
sfc-gh-bmikaili, line 361: **RECOVERY_IN_FLIGHT** — new case. Old code returned `false` from `insert` for both 'fallback fired' and 'channel initializing,' so both landed in the rewind map. That was the bug: the fallback had already issued a seek to `committed+1`, and the batch loop's own rewind to the first-skipped offset clobbered it. Now the fallback returns `RECOVERY_IN_FLIGHT` and we deliberately skip the rewind here — the tracker owns the seek.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -338,57 +338,47 @@ public void insert(final Collection<SinkRecord> records) {
...
+          // PROCESSED: nothing to rewind. RECOVERY_IN_FLIGHT: the channel's resetAfterRecovery
+          // owns the authoritative seek; our rewind would clobber it.
+          break;
+
+        case PREVIOUSLY_SKIPPED_IN_BATCH:
```
sfc-gh-bmikaili, line 366: **PREVIOUSLY_SKIPPED_IN_BATCH** — maps to the old `if (offsetsOfFirstSkippedRecord.containsKey(tp)) continue;` short-circuit.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -338,57 +338,47 @@ public void insert(final Collection<SinkRecord> records) {
...
+          // Already in the rewind map from an earlier record in this partition.
+          break;
+
+        case CHANNEL_INITIALIZING:
```
sfc-gh-bmikaili, line 370: **CHANNEL_INITIALIZING** — maps to the old `if (skipAllPartitions || initializingPartitions.contains(tp))` → add first-skipped offset. Same behavior, just named.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -338,57 +338,47 @@ public void insert(final Collection<SinkRecord> records) {
...
+        case IN_BACKPRESSURE_COOLDOWN:
```
sfc-gh-bmikaili, line 371: **IN_BACKPRESSURE_COOLDOWN** — maps to the old `skipAllPartitions` flag set at the top of the method from `Instant.now().isBefore(backpressureUntil)`. Now checked per-record inline (the cooldown window is short, so re-checking each record is cheap and removes one piece of loop-carried state).

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -338,57 +338,47 @@ public void insert(final Collection<SinkRecord> records) {
...
-        skipAllPartitions = true;
-        newBackpressure = true;
-      }
-    }
+          break;
 
-    if (newBackpressure) {
-      backpressureUntil = Instant.now().plus(BACKPRESSURE_COOLDOWN);
-      LOGGER.info("Backpressure cooldown set until {}", backpressureUntil);
+        case BACKPRESSURE_TRIGGERED:
```
sfc-gh-bmikaili, line 375: **BACKPRESSURE_TRIGGERED** — maps to the old `catch (BackpressureException e)` block. `insertRecord` now returns the enum value instead of throwing, which keeps flow-control out of exception handling.

src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/TopicPartitionChannel.java
```diff
@@ -11,22 +11,18 @@ public interface TopicPartitionChannel {
...
    *
    * <p>Step 2: Decides whether given offset from Kafka needs to be processed and whether it
    * qualifies for being added into buffer.
    *
-   * @param kafkaSinkRecord input record from Kafka
-   * @param isFirstRowPerPartitionInBatch indicates whether the given record is the first record per
-   *     partition in a batch
-   * @return true if the record was processed (or legitimately skipped as a duplicate), false if
-   *     recovery was triggered and the caller should stop feeding records to this partition for the
-   *     remainder of the batch
```
sfc-gh-skurella, line 27: Can we please bring those back?

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/AppendRowWithFallbackPolicy.java
sfc-gh-skurella: Would it be possible to limit the scope and not include this change? If it's a good change but unrelated, let's move it to a separate PR to reduce the complaxity of this one.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -338,57 +338,47 @@ public void insert(final Collection<SinkRecord> records) {
...
-        // records in this partition.
-        continue;
-      }
-      if (skipAllPartitions || initializingPartitions.contains(tp)) {
-        // Make sure we store the first record in each partition that we skipped so we can correctly
-        // rewind the offset.
-        offsetsOfFirstSkippedRecord.putIfAbsent(tp, record.kafkaOffset());
-        continue;
+        result = InsertResult.PREVIOUSLY_SKIPPED_IN_BATCH;
+      } else if (Instant.now().isBefore(backpressureUntil)) {
```
sfc-gh-skurella, line 351: This is a race condition. If the backpressure expires in the middle of the for loop, we'll end up with a gap. Let's bring back the `skipAllPartitions` logic - we can rename the variable to `backpressureCooldown`.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -398,21 +388,26 @@ public void insert(final Collection<SinkRecord> records) {
...
+   * com.snowflake.kafka.connector.internal.streaming.v2.channel.PartitionOffsetTracker} and wipe
+   * the recovery state (pending seek, processedOffset). Recovery for an existing channel goes
+   * through the appendRow fallback path ({@code reopenChannel} / {@code
+   * recreateClientAndReopenChannel}), which preserves the tracker. The manager removes entries only
+   * on explicit {@link #close}/{@link #closeAll}, so {@code isEmpty()} cleanly distinguishes "first
+   * time we see this partition" from "recovery needed".
    */
   @Override
-  public boolean insert(SinkRecord record) {
+  public InsertResult insert(SinkRecord record) {
```
sfc-gh-skurella, line 404: TODO(seb): review more thoroughly

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
sfc-gh-skurella, line 301: TODO(seb): review more thoroughly

sfc-gh-skurella (changes requested):
I'll continue the review later. It'd be great to split off some changes (discussed in person)

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
sfc-gh-skurella: Treated more rigorously in https://github.com/snowflakedb/snowflake-kafka-connector/pull/1471

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-skurella: (AI slop warning)

The `isEmpty()` change should be reverted — keep the existing `isChannelClosed` gate.

**The race is self-healing.** The scenario in the WHY comment cannot cause data loss because the stale epoch prevents any data from being committed past the old offset. If `startPartition` replaces the SSPC and the old recovery later opens a newer epoch, `appendRow` on the stale channel will eventually be rejected (channel sequencer mismatch). That triggers recovery on the new SSPC, which resets `processedOffset` to the committed offset and seeks Kafka back. The system converges.

**`isEmpty()` loses init-failure recovery.** The existing gate handles two cases: (1) partition not in the map (`orElse(true)`), and (2) channel init failed (`isChannelClosed` catches the `RuntimeException` from `getChannel()` and returns `true`). With `isEmpty()`, a failed-init SSPC stays in the map, `insertRecord` calls `getChannel().join()` which rethrows the init exception, and since it is not `SFException` the catch in `insertRowWithFallback` does not apply — the exception propagates to `put()` and kills the task. The old gate retries by replacing the SSPC via `startPartition`.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-skurella: I think we also need one more operation. We previously used `offsetsOfFirstSkipperRecord` to "remember" that some records were skipped for a given partition.

In your fix, we no longer write to that structure because we don't want the offset to be skipped. Think of the following sequence:
- partition A, offset 1 insert fails because the channel was invalidated
- partition A, offset 2 - **what happens here?**

I think what happens is that the loop will attempt to insert this new piece of data and block on the channel being reopened. The issue with this is that we'll basically ignore the offset we got from the channel reopen - this will only take effect on the next `put` call. So I think it's both a correctness and a performance bug.

I'd recommend inserting `tp` into `initializingPartitions` so the remaining records in this partition are skipped, and we can then rely on the offset rewind / channel readiness ordering.

UPDATE: writing to `initializingPartitions` would result in the next record being added to `offsetsOfFirstSkipperRecord`, defeating the purpose of the change. We need to either introduce a separate structure or approach this differently still.

PR #1476, opened 2026-05-27
SNOW-3574225 thread-safe offset rewinds

sfc-gh-skurella:
## Problem

When a channel is invalidated, `SinkTaskContext::offset` was called from two threads:
- main thread, at the end of `put`, to rewind the partition that was skipped
- channel IO thread, after the reopen, with the offset fetched from the Snowflake channel status.

This creates a race condition - if the `put`'s offset wins, the subsequent `put` may not receive some records because we didn't rewind far enough.

For example:
- record 80 fails due to a channel invalidation
- this and each subsequent `put` that receives records for this partition rewinds the partition to offset 80
- Snowflake channel open completes with a committed offset 70 while the `put` is running for this partition
  - the partition is rewound to the "recovery" offset 71 from the IO thread
  - at the end of `put`, the offset is rewound to 80 from the main thread
- records 71–79 are permanently lost because the consumer resumes from 80 instead of 71.

## Fix

Move offset resets off the IO thread and onto the task thread via a `ConcurrentHashMap<TopicPartition, Long>` in `PartitionChannelManager`:

1. `PartitionOffsetTracker` no longer calls `SinkTaskContext::offset`. Instead it takes a `Consumer<Long> onOffsetReset` callback that writes to the pending map.
2. `PartitionChannelManager.pendingOffsetResets` buffers offset resets from init/recovery (IO threads).
3. `SSV2.insert(Collection)` drains pending resets at the top of each batch into `offsetsToRewindTo` — these partitions are then skipped for the rest of the batch. At the end, a single `sinkTaskContext.offset(Map)` call applies all rewinds atomically.
4. `close()` clears pending resets for revoked partitions.

This ensures `sinkTaskContext.offset()` is only ever called from the task thread, and recovery offsets cannot be overwritten by the skip-tracking logic.

PR #1478, opened 2026-05-27
SNOW-3574225: Add client recreation for SDK pipe failover (HTTP 410)

sfc-gh-bmikaili:
## Why

SSv2 pipe failover surfaces to the connector as an HTTP 410 / `InvalidClientError` `SFException`. Once a `SnowflakeStreamingIngestClient` is in this state, every subsequent SDK call on it fails the same way — the only path forward is to swap in a freshly-constructed client. Without in-process recovery the task either dies on the first failover or, worse, spins reopening channels against the dead client forever (observed in the e2e suite stuck at 300/900 rows).

## Recovery flow

Client recreation lives at a single seam — `openChannelWithClientRecovery` — that every channel open funnels through (initial open in the constructor and every `reopenChannel`). The flow per open is:

1. Call `openChannelForTable`.
2. If it throws an `SFException` that `ClientRecreationException.isClientInvalidError` recognises, call `clientRecreator.recreate(this.streamingClient)`. The pool does a CAS swap and runs a Failsafe policy with a wall-clock budget (multi-minute, exponential backoff + jitter) so concurrent partitions on the same pipe share a single recreation. If the budget is exhausted, `recreate` throws and `openChannelWithClientRecovery` translates it to `ConnectException` — KC then fails and restarts the task, which is the appropriate response when the fault is no longer transient.
3. Retry `openChannelForTable` once. A second client-invalid error propagates — at that point the failure is no longer transient client state either.

`appendRow` no longer makes its own recreation decision; its failure handler just calls `reopenChannel`, which goes through the same helper. The recovery decision exists in exactly one place.

Two operations don't fit the open-channel seam and handle the same error inline:

- `initiateFlush` (in SSPC) swallows client-invalid with a WARN. Uncommitted data is replayed on task restart — failing the explicit flush would surface a recoverable condition as a hard failure.
- `BatchOffsetFetcher.getCommittedOffsets` logs WARN + skip on client-invalid (vs. ERROR for other `SFException`s). The next `appendRow` on that pipe reopens its channel and picks up recreation through the seam.

## Invariants

**Process-or-rewind.** Within a single `put()`, a record is either appended to its channel or its partition is rewound (via `sinkTaskContext.offset()` from `startPartition`) — never both. `SnowflakeSinkServiceV2.insert` now `return false`s immediately after `startPartition`; previously it fell through and appended the record onto the freshly-created channel in the same call, racing with the deferred Kafka Connect seek.

**Single recreate per open.** Each individual open attempt asks the pool to recreate at most once; the pool's Failsafe loop owns the multi-attempt patience. The outer recovery loop (`appendRow` failure → `reopenChannel` → `openChannelWithClientRecovery`) will trigger another recreation on the next iteration if the first one returned a still-broken client, but no open call retries recreation in a tight loop.

## Alternatives considered

- **Keep recreation in the `appendRow` failure path.** This was the starting point and is exactly what causes the 300/900 starvation. When the client is invalid `isChannelClosed` returns true, so `insert` short-circuits to `startPartition` + `return false` and `appendRow` is never reached. The only recreation trigger never fires; new channels are continuously built on the same dead client.
- **Bridge the rewind window with `sinkTaskContext.pause()` / `resume()`.** Pursued against an earlier (wrong) hypothesis — a "pre-seek buffer race" between the deferred seek and Kafka Connect's poll buffer. The implementation didn't change the outcome and was reverted; the real issue was the invariant violation above.
- **Put the multi-minute retry budget on the `appendRow` recovery loop instead of inside `recreateClient`.** That bloated the per-task recovery time on every transient `SFException`, not just client-invalid ones, and conflated two distinct conditions (transient append failure vs. dead client). The patience now lives where the long wait is actually needed: the recreate call itself; `appendRow` recovery keeps its original short count-based circuit breaker.

## Aside: exponential backoff inside `recreateClient`

Independent of moving recreation to the open-channel seam: the pool's recreate retry policy is a wall-clock budget (multi-minute) with exponential backoff starting at 1 s and capped at 30 s, plus ±20% jitter so concurrent recreators on the same pipe don't retry in lockstep. Observed SSv2 failover propagation windows are 2–4 minutes; the budget covers that with headroom.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
@@ -421,6 +421,7 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
         .orElse(true)) {
       LOGGER.warn("Streaming channel doesn't exist or is closed for {}", topicPartition);
       startPartition(topicPartition);
+      return false;
     }
 
     TopicPartitionChannel channel =
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
@@ -58,6 +58,14 @@ public class SnowflakeTelemetryChannelStatus extends SnowflakeTelemetryBasicInfo
   // channel recovery counter (always tracked; also registered as JMX gauge if enabled)
   private final AtomicLong recoveryCount = new AtomicLong(0);
 
+  // Client recreation counters — incremented from openChannelWithClientRecovery when a
+  // client-invalid SDK error (HTTP 410 / pipe failover) triggers a swap of the streaming client.
+  // attempts is bumped on entry to the recreate branch; success/failure split records whether
+  // the underlying StreamingClientPools.recreateClient retry budget held.
+  private final AtomicLong clientRecreationAttemptCount = new AtomicLong(0);
+  private final AtomicLong clientRecreationSuccessCount = new AtomicLong(0);
+  private final AtomicLong clientRecreationFailureCount = new AtomicLong(0);
+
   // Aggregated count of client-side validation failures for this channel.
   // Reported in channel status telemetry on close, avoiding per-record telemetry overhead.
   private final AtomicLong validationFailureCount = new AtomicLong(0);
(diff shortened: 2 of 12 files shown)


sfc-gh-bmikaili:
* **#1479** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1479?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1478** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1478?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1478?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1476** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1476?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1477](https://github.com/snowflakedb/snowflake-kafka-connector/pull/1477) <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1477?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`


<h2></h2>

sfc-gh-skurella (approved):

sfc-gh-skurella:
### Merge activity

* **May 29, 10:07 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1478).
* **May 29, 10:07 AM UTC**: @sfc-gh-skurella merged this pull request with [Graphite](https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1478).

PR #1480, opened 2026-05-27
Fix duplicate-counting race in test_invalidation_offset_consistency

sfc-gh-bmikaili:
## Summary

The test was computing `duplicates = total_rows - distinct_offsets` from
**two separate `cursor.execute()` calls** (`SELECT DISTINCT ...` followed by
`SELECT count(*) ...`). Snowpipe Streaming v2 commits rows asynchronously, so
even with the producer stopped these two back-to-back queries can observe
different table states and produce phantom duplicates that don't actually
exist in the table.

## Reproduction

On master against an account with `SYSTEM$STREAMING_CHANNEL_INVALIDATE`
deployed (qa3), the test intermittently logs `Found N duplicate rows`. Same
wall-clock second:

| query | result |
|-------|--------|
| `SELECT DISTINCT off ...` (the test's first call) | 170 rows, range `[0..279]` |
| `SELECT count(*) ...` (the test's second call) | 200 |
| `SELECT count(*), count(DISTINCT off), max(off) ...` *(single snapshot)* | total=200, distinct=200, max=309 |
| `SELECT off, count(*) FROM t GROUP BY off HAVING c > 1` | `[]` |

No offset ever appears more than once. The 30-dupe count came entirely from
the first query seeing 30 fewer rows than the second one issued microseconds
later.

## Fix

Wait for the connector to commit every record we sent before measuring,
using the existing `driver.get_consumer_group_offset` helper. The connector's
`preCommit` reports SSv2's `latestCommittedOffsetToken + 1` as the safe-to-commit
Kafka offset, so when the consumer group offset reaches `producer.records_produced`
the streaming pipeline has finished landing every record we sent. Then read
`count(*)`, `count(DISTINCT off)`, and `max(off)` from a single SQL statement
so both metrics describe the same snapshot — no retry needed because the table
state is known to be stable.

Tighten the previously-tolerated `if duplicates > 0: log(...)` to
`assert total_rows == distinct_offsets`, since SSv2's offset tokens guarantee
exactly-once and the connector should never produce duplicate offsets after
channel-recovery.

## Verification

Built JAR from master HEAD (`0846f61c`) and ran the test 10 times in a row
against qa3. With the fix, **every run** waited until the connector caught
up, then read a self-consistent total/distinct that matched exactly:

```
RUN  1: total=200 distinct=200  consumer group offset=310 (target=310)
RUN  2: total=250 distinct=250  consumer group offset=360 (target=360)
RUN  3: total=250 distinct=250  consumer group offset=360 (target=360)
RUN  4: total=230 distinct=230  consumer group offset=360 (target=360)
RUN  5: total=250 distinct=250  consumer group offset=360 (target=360)
RUN  6: total=200 distinct=200  consumer group offset=310 (target=310)
RUN  7: total=200 distinct=200  consumer group offset=310 (target=310)
RUN  8: total=200 distinct=200  consumer group offset=310 (target=310)
RUN  9: total=200 distinct=200  consumer group offset=310 (target=310)
RUN 10: total=250 distinct=250  consumer group offset=360 (target=360)
```

All 10 runs fail on the **gap** assertion (unrelated, pre-existing master bug
— seek-overwrite race in `SnowflakeSinkServiceV2.insert()`). Once that bug
lands its own fix, the new `assert total_rows == distinct_offsets` will catch
any regression where the connector actually produces duplicate offsets after
recovery. Zero phantom-dup outcomes across the loop.

## Test plan

- [x] `ruff check` + `ruff format --check` clean
- [x] 10 consecutive runs against qa3: zero phantom-dup outcomes
- [ ] CI E2E run on PR

### DIFF
--- a/test/tests/test_channel_invalidation.py
+++ b/test/tests/test_channel_invalidation.py
@@ -443,7 +443,7 @@ def test_invalidation_offset_consistency(
 ):
     """Test 6: Verify offset consistency after invalidation recovery.
 
-    Checks the full offset range (0..max_offset) has no gaps. Duplicates are OK.
+    Checks the full offset range (0..max_offset) has no gaps and no duplicates.
     """
     topic = f"test_invalidation_offset_consistency{name_salt}"
     table_name = topic
@@ -476,25 +476,57 @@ def test_invalidation_offset_consistency(
     producer.stop_continuous()
     _assert_task_running(driver, connector.name)
 
-    # Verify offset integrity: the full range 0..max_offset must have no gaps.
-    # With the recordProcessed fix (SNOW-3344243), the offset rewind replays all
-    # records that were in-flight during the flush-failure window, so no data is lost.
-    cur = driver.snowflake_conn.cursor()
-    offsets = sorted(
-        row[0]
-        for row in cur.execute(
-            f"SELECT DISTINCT record_metadata:offset::int AS off "
-            f'FROM "{table_name}" ORDER BY off'
-        ).fetchall()
+    # Wait for the connector to commit every record it consumed before
+    # measuring. The connector's preCommit reports SSv2's
+    # latestCommittedOffsetToken + 1 as the Kafka offset that is safe to
+    # commit, so when the consumer group offset reaches records_produced,
+    # the streaming pipeline has finished landing every record we sent and
+    # the verification queries below see a stable table state without
+    # needing to retry.
+    target_kafka_offset = producer.records_produced  # one past max sent offset
+    deadline = time.monotonic() + 60
+    committed = None
+    while time.monotonic() < deadline:
+        committed = driver.get_consumer_group_offset(connector.name, topic, 0)
+        if committed is not None and committed >= target_kafka_offset:
+            break
+        time.sleep(1)
+    else:
+        raise AssertionError(
+            f"Connector did not commit through Kafka offset {target_kafka_offset - 1} "
+            f"within 60s; last consumer group offset={committed}"
+        )
+    logger.info(
+        f"Connector caught up: consumer group offset={committed} "
+        f"(target={target_kafka_offset})"
     )
 
-    total_rows = int(driver.select_number_of_records(table_name))
-    distinct_offsets = len(offsets)
-    max_offset = offsets[-1]
-
+    # Verify offset integrity: the full range 0..max_offset must have no gaps
+    # and no duplicates. With the recordProcessed fix (SNOW-3344243) the
+    # offset rewind replays all records that were in-flight during the flush
+    # window, so no data is lost; SSv2's offset tokens guarantee exactly-once
+    # so no record is committed twice. Read both metrics from a single SQL
+    # snapshot to avoid the inter-query race that previously made two
+    # back-to-back cursor.execute() calls observe different table states.
+    cur = driver.snowflake_conn.cursor()
+    total_rows, distinct_offsets, max_offset = cur.execute(
+        f"WITH t AS (SELECT record_metadata:offset::int AS off "
+        f'FROM "{table_name}") '
+        f"SELECT count(*), count(DISTINCT off), max(off) FROM t"
+    ).fetchone()
     logger.info(
         f"Offset check: {total_rows} total rows, {distinct_offsets} distinct offsets, "
         f"range [0..{max_offset}]"
+    )
+
+    offsets = sorted(
+        row[0]
+        for row in driver.snowflake_conn.cursor()
+        .execute(
+            f"SELECT DISTINCT record_metadata:offset::int AS off "
+            f'FROM "{table_name}" ORDER BY off'
+        )
+        .fetchall()
     )
 
     # No gaps in the full range 0..max_offset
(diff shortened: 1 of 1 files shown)


sfc-gh-bmikaili:
* **#1480** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1480?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1480?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

test/tests/test_channel_invalidation.py
```diff
@@ -443,7 +443,7 @@ def test_invalidation_offset_consistency(
 ):
     """Test 6: Verify offset consistency after invalidation recovery.
 
-    Checks the full offset range (0..max_offset) has no gaps. Duplicates are OK.
+    Checks the full offset range (0..max_offset) has no gaps and no duplicates.
```
sfc-gh-bmikaili, line 446: **Docstring update.** The old text said "Duplicates are OK" because the pre-fix test tolerated phantom duplicates from racing two queries. Now that the test waits for the pipeline to drain and reads from a single snapshot, duplicates are not OK and the docstring reflects the tightened contract.

test/tests/test_channel_invalidation.py
```diff
@@ -476,27 +476,59 @@ def test_invalidation_offset_consistency(
...
+        time.sleep(1)
+    else:
+        raise AssertionError(
+            f"Connector did not commit through Kafka offset {target_kafka_offset - 1} "
+            f"within 60s; last consumer group offset={committed}"
+        )
+    logger.info(
+        f"Connector caught up: consumer group offset={committed} "
+        f"(target={target_kafka_offset})"
     )
```
sfc-gh-bmikaili, line 502: **Wait for the connector to drain.** This is the new pre-measurement barrier.

`producer.stop_continuous()` only stops the Python producer thread — records are still in Kafka, in the connector's consumer buffer, in the SDK's batch, and in SSv2's async commit pipeline. If we measure now, the verification SQL races against in-flight ingestion.

The connector's `preCommit` reports `latestCommittedOffsetToken + 1` to the Kafka Connect framework, which gets written to the consumer group offset. So `consumer_group_offset >= producer.records_produced` is the SSv2-native signal that **every record we sent has been committed to the table**.

The `while ... else:` block raises if the connector never catches up within 60s (real bug; not a test artifact).

test/tests/test_channel_invalidation.py
```diff
@@ -476,27 +476,59 @@ def test_invalidation_offset_consistency(
...
+    cur = driver.snowflake_conn.cursor()
+    total_rows, distinct_offsets, max_offset = cur.execute(
+        f"WITH t AS (SELECT record_metadata:offset::int AS off "
+        f'FROM "{table_name}") '
+        f"SELECT count(*), count(DISTINCT off), max(off) FROM t"
+    ).fetchone()
     logger.info(
         f"Offset check: {total_rows} total rows, {distinct_offsets} distinct offsets, "
         f"range [0..{max_offset}]"
     )
```
sfc-gh-bmikaili, line 520: **Single-snapshot measurement.** The previous version computed `total_rows - distinct_offsets` from two separate `cursor.execute()` calls (`SELECT DISTINCT ...` then `SELECT count(*) ...`). Even with the producer stopped, those two queries can land on different table snapshots and produce phantom duplicates that don't exist (proven by `GROUP BY off HAVING c > 1` returning `[]` while the test reported "30 dupes").

Reading `count(*)`, `count(DISTINCT off)`, and `max(off)` in one statement guarantees both metrics describe the same row set. Combined with the wait above, no retry/poll-until-stable is needed.

test/tests/test_channel_invalidation.py
```diff
@@ -476,27 +476,59 @@ def test_invalidation_offset_consistency(
...
 
+    offsets = sorted(
+        row[0]
+        for row in driver.snowflake_conn.cursor()
+        .execute(
+            f"SELECT DISTINCT record_metadata:offset::int AS off "
+            f'FROM "{table_name}" ORDER BY off'
+        )
+        .fetchall()
+    )
```
sfc-gh-bmikaili, line 530: **Distinct offset list for the gap check.** This second query is purely to enumerate which offsets exist (so the assertion below can show *which* offsets are missing in the failure message). It only feeds the gap-set comparison — it never participates in the duplicate count, so it cannot reintroduce the inter-query race even though it's a separate `cursor.execute()` call.

sfc-gh-bmikaili (commented):
Per-hunk inline explanations.

test/tests/test_channel_invalidation.py
```diff
@@ -506,9 +538,12 @@ def test_invalidation_offset_consistency(
...
 
-    duplicates = total_rows - distinct_offsets
-    if duplicates > 0:
-        logger.info(f"Found {duplicates} duplicate rows (expected after recovery)")
+    # Snowpipe Streaming v2 provides exactly-once semantics via offset tokens, so
+    # the connector should never produce duplicate offsets after channel-recovery.
+    assert total_rows == distinct_offsets, (
+        f"Duplicate rows detected: {total_rows} total rows, "
+        f"{distinct_offsets} distinct offsets ({total_rows - distinct_offsets} duplicates)"
+    )
```
sfc-gh-bmikaili, line 546: **Tightened duplicate assertion.** Was:

```python
duplicates = total_rows - distinct_offsets
if duplicates > 0:
    logger.info(f"Found {duplicates} duplicate rows (expected after recovery)")
```

This log line was misleading ("expected after recovery") because SSv2 commits each `record_metadata:offset` exactly once via offset tokens — the connector should never produce duplicates after channel-recovery. The previous lenient handling existed only to absorb the racing-query phantom duplicates, which the wait + single-snapshot measurement above now eliminate. Asserting `total_rows == distinct_offsets` will catch any future regression where the connector actually produces duplicates.

sfc-gh-skurella (approved):

PR #1487, opened 2026-06-12
SNOW-3161094: Kafka Connector — managed-Iceberg table auto-creation + ingestion (snowflake.autocreate.table.type)

sfc-gh-bmikaili:
## Summary

Adds managed-Iceberg **table auto-creation + ingestion** to KC v4, gated by a new connector-wide `snowflake.autocreate.table.type` (`snowflake` | `iceberg` | `none`). The setting governs **auto-creation only** — an existing table is always used as-is. Schema evolution for managed Iceberg is **server-side** (GS); the end-to-end SE test matrix is a follow-up PR stacked on top of this one.

- **`snowflake.autocreate.table.type`** (default `snowflake`) controls what to create when the table is **missing**:
  - `snowflake` → auto-create an FDN table (today's behavior, unchanged).
  - `iceberg` → auto-create a managed Iceberg table (`CATALOG='SNOWFLAKE'`, structured `RECORD_METADATA`, `ENABLE_SCHEMA_EVOLUTION`, `ERROR_LOGGING`).
  - `none` → create nothing; the table **and** its pipe must already exist (fail-fast otherwise).
  - **An existing table is used as-is regardless of this setting** — no cross-type guard. The setting only affects the missing-table path.
- **`snowflake.iceberg.create.table.options`** — SQL-clause passthrough spliced after the column list (e.g. `EXTERNAL_VOLUME='v' ICEBERG_VERSION=3 CLUSTER BY (id)`); valid only when `autocreate.table.type=iceberg`. Ignored if the table already exists.
- **`RECORD_METADATA` conforming** is keyed off the **actual target table type** (`isIcebergTable`), not the config — so it's correct even when `iceberg` config points at a pre-existing table, or `none` targets an existing Iceberg table. FDN's VARIANT metadata path is unchanged.
- **Non-schematized Iceberg autocreate requires `ICEBERG_VERSION=3`.** With `schematization=false` the raw record lands in a `RECORD_CONTENT` VARIANT column, and VARIANT-in-Iceberg needs v3; this is enforced **at autocreate time** (so pre-created tables are unaffected) with a clear fail-fast message.
- **Iceberg + `client_side` validation + schematization is rejected** at config time — managed-Iceberg SE is server-side; use `validation=server_side`.

## Also in this PR
- **Fix `hasErrorLoggingEnabled`**: `SHOW TABLES` renders `error_logging` as `ON`/`OFF`, not `Y`; the old `"Y"` check always returned false → spurious warnings for both FDN and Iceberg.
- Removed the now-dead FDN-create cross-type catch and the test-only `createTableIfNotExistsForTest` wrapper.
- `google-java-format` + `ruff format` applied.

## Test plan
- **Unit tests green** (config parse/rename, `use-existing-as-is` for both directions, autocreate-time v3 guard, `client_side` rejection, `ON`/`OFF` parsing, RECORD_METADATA conform incl. a drift test) + `IcebergAutoCreateIT`. Full `mvn package` suite passes.
- **Autocreate e2e validated green on a live SUT** (managed-Iceberg feature on): auto-create at Iceberg **v2 and v3**, `ERROR_LOGGING` present, existing-higher-version-table harmless, and server-side SE add-column / json-server-side on the auto-created/used table — all via the renamed config key.
- **Stress test run**: https://github.com/snowflakedb/snowflake-kafka-connector/actions/runs/29930045427/job/88957061569

## Scope / stack
- Depends on the pytest-8 collection compat fix (below in the stack).
- The full server-side SE e2e matrix is the follow-up PR stacked on top.

Related: SNOW-3552261 (server-side SE on managed Iceberg), SNOW-3053493 (primitive-column SE), SNOW-3161094 (KC for SSv2 Iceberg).

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/Constants.java
+++ b/src/main/java/com/snowflake/kafka/connector/Constants.java
@@ -150,6 +150,19 @@ public final class Constants {
     public static final String SNOWFLAKE_ENABLE_SCHEMATIZATION = "snowflake.enable.schematization";
     public static final boolean SNOWFLAKE_ENABLE_SCHEMATIZATION_DEFAULT = true;
 
+    // Iceberg table auto-creation (SNOW-3552261)
+    public static final String SNOWFLAKE_AUTOCREATE_TABLE_TYPE = "snowflake.autocreate.table.type";
+    public static final String SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT = "snowflake";
+    // Optional SQL clauses spliced into the auto-created Iceberg table's CREATE statement, right
+    // after the column list (e.g. "EXTERNAL_VOLUME='v' ICEBERG_VERSION=3 CLUSTER BY (id)"). Lets
+    // operators set any managed-Iceberg create option without the connector exposing a knob per
+    // option. The connector always supplies the ingestion-mandatory clauses (RECORD_METADATA
+    // column, CATALOG='SNOWFLAKE', ENABLE_SCHEMA_EVOLUTION, ERROR_LOGGING); do not repeat those
+    // here. Only valid when snowflake.autocreate.table.type=iceberg.
+    public static final String SNOWFLAKE_ICEBERG_CREATE_TABLE_OPTIONS =
+        "snowflake.iceberg.create.table.options";
+    public static final String SNOWFLAKE_ICEBERG_CREATE_TABLE_OPTIONS_DEFAULT = "";
+
     // MDC logging header
     public static final String ENABLE_MDC_LOGGING_CONFIG = "enable.mdc.logging";
     public static final String ENABLE_MDC_LOGGING_DEFAULT = "false";
--- a/src/main/java/com/snowflake/kafka/connector/Constants.java
+++ b/src/main/java/com/snowflake/kafka/connector/Constants.java
@@ -152,10 +152,9 @@ public final class Constants {
 
     // Iceberg table auto-creation (SNOW-3552261)
     public static final String SNOWFLAKE_AUTOCREATE_TABLE_TYPE = "snowflake.autocreate.table.type";
-    // Use the canonical config value from TableType so the default stays in sync if the enum
-    // changes.
-    public static final String SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT =
-        com.snowflake.kafka.connector.config.TableType.SNOWFLAKE.configValue();
+    public static final com.snowflake.kafka.connector.config.TableType
+        SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT =
+            com.snowflake.kafka.connector.config.TableType.SNOWFLAKE;
     // Optional SQL clauses spliced into the auto-created Iceberg table's CREATE statement, right
     // after the column list (e.g. "EXTERNAL_VOLUME='v' ICEBERG_VERSION=3 CLUSTER BY (id)"). Lets
     // operators set any managed-Iceberg create option without the connector exposing a knob per
--- a/src/main/java/com/snowflake/kafka/connector/Constants.java
+++ b/src/main/java/com/snowflake/kafka/connector/Constants.java
@@ -90,6 +90,17 @@ public final class Constants {
         "snowflake.feature.normalize.time";
     public static final boolean SNOWFLAKE_FEATURE_NORMALIZE_TIME_DEFAULT = true;
 
+    // Kill-switch (unregistered, default on) for the managed-Iceberg structured-OBJECT
+    // RECORD_METADATA handling: probing whether a table's RECORD_METADATA is a structured OBJECT,
+    // validating an existing structured schema at startup, and conforming the metadata map for
+    // ingestion. Set to false to revert to 4.0.x behavior (RECORD_METADATA always treated as
+    // VARIANT, no probe, no conforming) if the new handling ever misbehaves for existing workloads.
+    // Note: disabling this also disables correct ingestion into managed-Iceberg v2 (structured
+    // OBJECT) tables, since their metadata will no longer be conformed.
+    public static final String SNOWFLAKE_FEATURE_STRUCTURED_RECORD_METADATA =
+        "snowflake.feature.structured.record.metadata.enabled";
+    public static final boolean SNOWFLAKE_FEATURE_STRUCTURED_RECORD_METADATA_DEFAULT = true;
+
     public static final String SNOWFLAKE_FEATURE_STRUCTURED_HEADERS =
         "snowflake.feature.structured.headers";
     public static final boolean SNOWFLAKE_FEATURE_STRUCTURED_HEADERS_DEFAULT = false;
--- a/src/main/java/com/snowflake/kafka/connector/Constants.java
+++ b/src/main/java/com/snowflake/kafka/connector/Constants.java
@@ -152,7 +152,10 @@ public final class Constants {
 
     // Iceberg table auto-creation (SNOW-3552261)
     public static final String SNOWFLAKE_AUTOCREATE_TABLE_TYPE = "snowflake.autocreate.table.type";
-    public static final String SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT = "snowflake";
+    // Use the canonical config value from TableType so the default stays in sync if the enum
+    // changes.
+    public static final String SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT =
+        com.snowflake.kafka.connector.config.TableType.SNOWFLAKE.configValue();
     // Optional SQL clauses spliced into the auto-created Iceberg table's CREATE statement, right
     // after the column list (e.g. "EXTERNAL_VOLUME='v' ICEBERG_VERSION=3 CLUSTER BY (id)"). Lets
     // operators set any managed-Iceberg create option without the connector exposing a knob per
--- a/src/main/java/com/snowflake/kafka/connector/config/ConnectorConfigDefinition.java
+++ b/src/main/java/com/snowflake/kafka/connector/config/ConnectorConfigDefinition.java
@@ -523,7 +523,7 @@ public class ConnectorConfigDefinition {
                 + " column list (e.g. \"EXTERNAL_VOLUME='v' ICEBERG_VERSION=3 CLUSTER BY (id)\")."
                 + " Lets operators set any managed-Iceberg create option without a per-option knob."
                 + " Only valid when snowflake.autocreate.table.type=iceberg. Do NOT include"
-                + " RECORD_METADATA, CATALOG, ENABLE_SCHEMA_EVOLUTION or ERROR_LOGGING -- the"
+                + " CATALOG, ENABLE_SCHEMA_EVOLUTION or ERROR_LOGGING -- the"
                 + " connector always supplies those. Ignored if the table already exists.",
             CONNECTOR_CONFIG_DOC,
             16,
--- a/src/main/java/com/snowflake/kafka/connector/config/ConnectorConfigDefinition.java
+++ b/src/main/java/com/snowflake/kafka/connector/config/ConnectorConfigDefinition.java
@@ -502,7 +502,7 @@ public class ConnectorConfigDefinition {
         .define(
             KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE,
             STRING,
-            KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT,
+            KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT.configValue(),
             ConfigDef.ValidString.in("snowflake", "iceberg", "none"),
             MEDIUM,
             "Controls table auto-creation. 'snowflake' (default): auto-create a standard Snowflake"
--- a/src/main/java/com/snowflake/kafka/connector/config/ConnectorConfigDefinition.java
+++ b/src/main/java/com/snowflake/kafka/connector/config/ConnectorConfigDefinition.java
@@ -505,11 +505,11 @@ public class ConnectorConfigDefinition {
             KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT,
             ConfigDef.ValidString.in("snowflake", "iceberg", "none"),
             MEDIUM,
-            "Controls table auto-creation. 'snowflake' (default): auto-create a standard FDN table"
-                + " if it is missing. 'iceberg': auto-create a Snowflake-managed Iceberg table if"
-                + " it is missing. 'none': never create anything; the table and its pipe must"
-                + " already exist. An existing table is always used as-is, regardless of this"
-                + " setting.",
+            "Controls table auto-creation. 'snowflake' (default): auto-create a standard Snowflake"
+                + " table if it is missing. 'iceberg': auto-create a Snowflake-managed Iceberg"
+                + " table if it is missing. 'none': never auto-create the table (fail fast if it"
+                + " is missing); the pipe is still auto-managed by the connector as usual. An"
+                + " existing table is always used as-is, regardless of this setting.",
             CONNECTOR_CONFIG_DOC,
             15,
             Width.NONE,
--- a/src/main/java/com/snowflake/kafka/connector/config/ConnectorConfigDefinition.java
+++ b/src/main/java/com/snowflake/kafka/connector/config/ConnectorConfigDefinition.java
@@ -500,6 +500,36 @@ public class ConnectorConfigDefinition {
             Width.NONE,
             KafkaConnectorConfigParams.SNOWFLAKE_ENABLE_SCHEMATIZATION)
         .define(
+            KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE,
+            STRING,
+            KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT,
+            ConfigDef.ValidString.in("snowflake", "iceberg", "none"),
+            MEDIUM,
+            "Controls table auto-creation. 'snowflake' (default): auto-create a standard FDN table"
+                + " if it is missing. 'iceberg': auto-create a Snowflake-managed Iceberg table if"
+                + " it is missing. 'none': never create anything; the table and its pipe must"
+                + " already exist. An existing table is always used as-is, regardless of this"
+                + " setting.",
+            CONNECTOR_CONFIG_DOC,
+            15,
+            Width.NONE,
+            KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE)
+        .define(
+            KafkaConnectorConfigParams.SNOWFLAKE_ICEBERG_CREATE_TABLE_OPTIONS,
+            STRING,
+            KafkaConnectorConfigParams.SNOWFLAKE_ICEBERG_CREATE_TABLE_OPTIONS_DEFAULT,
+            LOW,
+            "SQL clauses spliced into the auto-created Iceberg table's CREATE statement after the"
+                + " column list (e.g. \"EXTERNAL_VOLUME='v' ICEBERG_VERSION=3 CLUSTER BY (id)\")."
+                + " Lets operators set any managed-Iceberg create option without a per-option knob."
+                + " Only valid when snowflake.autocreate.table.type=iceberg. Do NOT include"
+                + " RECORD_METADATA, CATALOG, ENABLE_SCHEMA_EVOLUTION or ERROR_LOGGING -- the"
+                + " connector always supplies those. Ignored if the table already exists.",
+            CONNECTOR_CONFIG_DOC,
+            16,
+            Width.NONE,
+            KafkaConnectorConfigParams.SNOWFLAKE_ICEBERG_CREATE_TABLE_OPTIONS)
+        .define(
             KafkaConnectorConfigParams.CACHE_TABLE_EXISTS,
             BOOLEAN,
             KafkaConnectorConfigParams.CACHE_TABLE_EXISTS_DEFAULT,
--- a/src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
@@ -160,6 +160,16 @@ public abstract class SinkTaskConfig {
    */
   public abstract boolean isNormalizeTimeEnabled();
 
+  /**
+   * Whether the managed-Iceberg structured-OBJECT {@code RECORD_METADATA} handling is enabled:
+   * probing whether a table's {@code RECORD_METADATA} is a structured OBJECT, validating an
+   * existing structured schema at startup, and conforming the metadata map for ingestion. When
+   * {@code false}, {@code RECORD_METADATA} is always treated as VARIANT (4.0.x behavior).
+   * Kill-switch for existing workloads. See {@link
+   * KafkaConnectorConfigParams#SNOWFLAKE_FEATURE_STRUCTURED_RECORD_METADATA}.
+   */
+  public abstract boolean isStructuredRecordMetadataEnabled();
+
   /** Convenience overload that calls {@link #from(Map, boolean)} with {@code false}. */
   public static SinkTaskConfig from(Map<String, String> raw) {
     return from(raw, false);
@@ -398,6 +408,13 @@ public abstract class SinkTaskConfig {
             .map(Boolean::parseBoolean)
             .orElse(KafkaConnectorConfigParams.SNOWFLAKE_FEATURE_NORMALIZE_TIME_DEFAULT);
 
+    boolean structuredRecordMetadataEnabled =
+        Optional.ofNullable(
+                config.get(KafkaConnectorConfigParams.SNOWFLAKE_FEATURE_STRUCTURED_RECORD_METADATA))
+            .map(Boolean::parseBoolean)
+            .orElse(
+                KafkaConnectorConfigParams.SNOWFLAKE_FEATURE_STRUCTURED_RECORD_METADATA_DEFAULT);
+
     TableType autocreatedTableType =
         TableType.fromConfig(config.get(KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE))
             .orElse(KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT);
@@ -491,6 +508,7 @@ public abstract class SinkTaskConfig {
         .prometheusMetricsEnabled(prometheusMetricsEnabled)
         .validationErrorTableNameEnabled(validationErrorTableNameEnabled)
         .normalizeTimeEnabled(normalizeTimeEnabled)
+        .structuredRecordMetadataEnabled(structuredRecordMetadataEnabled)
         .autocreatedTableType(autocreatedTableType)
         .icebergCreateTableOptions(icebergCreateTableOptions);
 
@@ -627,6 +645,9 @@ public abstract class SinkTaskConfig {
         boolean validationErrorTableNameEnabled);
 
     public abstract Builder normalizeTimeEnabled(boolean normalizeTimeEnabled);
+
+    public abstract Builder structuredRecordMetadataEnabled(
+        boolean structuredRecordMetadataEnabled);
 
     public abstract SinkTaskConfig build();
   }
--- a/src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
@@ -398,7 +398,7 @@ public abstract class SinkTaskConfig {
             .map(Boolean::parseBoolean)
             .orElse(KafkaConnectorConfigParams.SNOWFLAKE_FEATURE_NORMALIZE_TIME_DEFAULT);
 
-    TableType tableType =
+    TableType autocreatedTableType =
         TableType.fromConfig(config.get(KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE))
             .orElse(TableType.SNOWFLAKE);
 
@@ -407,20 +407,21 @@ public abstract class SinkTaskConfig {
                 config.get(KafkaConnectorConfigParams.SNOWFLAKE_ICEBERG_CREATE_TABLE_OPTIONS))
             .map(String::trim)
             .filter(s -> !s.isEmpty());
-    if (icebergCreateTableOptions.isPresent() && tableType != TableType.ICEBERG) {
+    if (icebergCreateTableOptions.isPresent() && autocreatedTableType != TableType.ICEBERG) {
       throw new IllegalArgumentException(
           KafkaConnectorConfigParams.SNOWFLAKE_ICEBERG_CREATE_TABLE_OPTIONS
               + " is only valid when "
               + KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE
               + "=iceberg (got table.type="
-              + tableType.configValue()
+              + autocreatedTableType.configValue()
               + ")");
     }
 
     // client_side validation cannot model the structured RECORD_METADATA column of managed Iceberg
     // tables; client-side schema evolution also relies on ALTER ADD COLUMN, which is a server-side
     // concern for Iceberg. Reject unconditionally rather than only when schematization is enabled.
-    if (tableType == TableType.ICEBERG && validation == SnowflakeValidation.CLIENT_SIDE) {
+    if (autocreatedTableType == TableType.ICEBERG
+        && validation == SnowflakeValidation.CLIENT_SIDE) {
       throw new IllegalArgumentException(
           KafkaConnectorConfigParams.SNOWFLAKE_VALIDATION
               + "=client_side is not supported with "
@@ -431,26 +432,11 @@ public abstract class SinkTaskConfig {
               + "=server_side.");
     }
 
-    // structured headers (SNOWFLAKE_FEATURE_STRUCTURED_HEADERS=true) keep header values in their
-    // native types. The managed-Iceberg RECORD_METADATA schema declares headers
-    // MAP(VARCHAR,VARCHAR),
-    // so a non-String header value would fail the strict typed-OBJECT cast at ingest time.
-    if (tableType == TableType.ICEBERG && metadataConfig.isStructuredHeadersEnabled()) {
-      throw new IllegalArgumentException(
-          KafkaConnectorConfigParams.SNOWFLAKE_FEATURE_STRUCTURED_HEADERS
-              + "=true is not supported with "
-              + KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE
-              + "=iceberg. The managed-Iceberg RECORD_METADATA schema declares"
-              + " headers MAP(VARCHAR,VARCHAR); non-String header values would fail the strict"
-              + " typed-OBJECT cast. Set "
-              + KafkaConnectorConfigParams.SNOWFLAKE_FEATURE_STRUCTURED_HEADERS
-              + "=false (the default) when using managed Iceberg.");
-    }
-
     // Managed Iceberg RECORD_METADATA is cast to a fixed structured OBJECT schema, so every
     // metadata flag that maps to a declared schema field must be enabled. Disabling any of them
     // produces a sparse map that fails the strict typed-OBJECT cast at ingest time.
-    if (tableType == TableType.ICEBERG && !metadataConfig.isFullIcebergMetadataEnabled()) {
+    if (autocreatedTableType == TableType.ICEBERG
+        && !metadataConfig.isFullIcebergMetadataEnabled()) {
       throw new IllegalArgumentException(
           KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE
               + "=iceberg requires all record metadata to be enabled"
@@ -503,7 +489,7 @@ public abstract class SinkTaskConfig {
         .prometheusMetricsEnabled(prometheusMetricsEnabled)
         .validationErrorTableNameEnabled(validationErrorTableNameEnabled)
         .normalizeTimeEnabled(normalizeTimeEnabled)
-        .autocreatedTableType(tableType)
+        .autocreatedTableType(autocreatedTableType)
         .icebergCreateTableOptions(icebergCreateTableOptions);
 
     snowflakePrivateKey.ifPresent(b::snowflakePrivateKey);
--- a/src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
@@ -442,6 +442,8 @@ public abstract class SinkTaskConfig {
               + "=iceberg requires all record metadata to be enabled"
               + " (RECORD_METADATA is cast to a fixed structured schema for managed Iceberg)."
               + " Remove any "
+              + KafkaConnectorConfigParams.SNOWFLAKE_METADATA_ALL
+              + "=false, "
               + KafkaConnectorConfigParams.SNOWFLAKE_METADATA_TOPIC
               + "=false, "
               + KafkaConnectorConfigParams.SNOWFLAKE_METADATA_OFFSET_AND_PARTITION
--- a/src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
@@ -400,7 +400,7 @@ public abstract class SinkTaskConfig {
 
     TableType autocreatedTableType =
         TableType.fromConfig(config.get(KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE))
-            .orElse(TableType.SNOWFLAKE);
+            .orElse(KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT);
 
     Optional<String> icebergCreateTableOptions =
         Optional.ofNullable(
--- a/src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
+++ b/src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
@@ -49,7 +49,7 @@ public abstract class SinkTaskConfig {
 
   public abstract boolean isEnableSchematization();
 
-  public abstract TableType getTableType();
+  public abstract TableType getAutocreatedTableType();
 
   public abstract Optional<String> getIcebergCreateTableOptions();
 
@@ -417,23 +417,34 @@ public abstract class SinkTaskConfig {
               + ")");
     }
 
-    // Client-side validation drives client-side schema evolution (ALTER ADD COLUMN issued by the
-    // connector). For managed Iceberg, schema evolution is performed server-side; the client-side
-    // RowValidator cannot even model the structured RECORD_METADATA column. Reject the combination
-    // loudly rather than silently producing a misleading runtime failure. Server-side SE is the
-    // supported mechanism: use snowflake.validation=server_side with Iceberg + schematization.
-    if (tableType == TableType.ICEBERG
-        && enableSchematization
-        && validation == SnowflakeValidation.CLIENT_SIDE) {
+    // client_side validation cannot model the structured RECORD_METADATA column of managed Iceberg
+    // tables; client-side schema evolution also relies on ALTER ADD COLUMN, which is a server-side
+    // concern for Iceberg. Reject unconditionally rather than only when schematization is enabled.
+    if (tableType == TableType.ICEBERG && validation == SnowflakeValidation.CLIENT_SIDE) {
       throw new IllegalArgumentException(
           KafkaConnectorConfigParams.SNOWFLAKE_VALIDATION
               + "=client_side is not supported with "
               + KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE
-              + "=iceberg when schema evolution is enabled ("
-              + KafkaConnectorConfigParams.SNOWFLAKE_ENABLE_SCHEMATIZATION
-              + "=true). Managed Iceberg schema evolution is server-side; set "
+              + "=iceberg. Managed Iceberg uses a structured RECORD_METADATA column that the"
+              + " client-side RowValidator cannot model. Set "
               + KafkaConnectorConfigParams.SNOWFLAKE_VALIDATION
               + "=server_side.");
+    }
+
+    // structured headers (SNOWFLAKE_FEATURE_STRUCTURED_HEADERS=true) keep header values in their
+    // native types. The managed-Iceberg RECORD_METADATA schema declares headers
+    // MAP(VARCHAR,VARCHAR),
+    // so a non-String header value would fail the strict typed-OBJECT cast at ingest time.
+    if (tableType == TableType.ICEBERG && metadataConfig.isStructuredHeadersEnabled()) {
+      throw new IllegalArgumentException(
+          KafkaConnectorConfigParams.SNOWFLAKE_FEATURE_STRUCTURED_HEADERS
+              + "=true is not supported with "
+              + KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE
+              + "=iceberg. The managed-Iceberg RECORD_METADATA schema declares"
+              + " headers MAP(VARCHAR,VARCHAR); non-String header values would fail the strict"
+              + " typed-OBJECT cast. Set "
+              + KafkaConnectorConfigParams.SNOWFLAKE_FEATURE_STRUCTURED_HEADERS
+              + "=false (the default) when using managed Iceberg.");
     }
 
     // Managed Iceberg RECORD_METADATA is cast to a fixed structured OBJECT schema, so every
@@ -492,7 +503,7 @@ public abstract class SinkTaskConfig {
         .prometheusMetricsEnabled(prometheusMetricsEnabled)
         .validationErrorTableNameEnabled(validationErrorTableNameEnabled)
         .normalizeTimeEnabled(normalizeTimeEnabled)
-        .tableType(tableType)
+        .autocreatedTableType(tableType)
         .icebergCreateTableOptions(icebergCreateTableOptions);
 
     snowflakePrivateKey.ifPresent(b::snowflakePrivateKey);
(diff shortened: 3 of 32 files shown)


sfc-gh-bmikaili:
* **#1491** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1491?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1487** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1487?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1487?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sfc-gh-bmikaili (commented):
Author annotations explaining *why* each key change is needed, for reviewers new to the managed-Iceberg auto-creation work.

src/main/java/com/snowflake/kafka/connector/config/ConnectorConfigDefinition.java
```diff
@@ -456,6 +456,36 @@ public static ConfigDef getConfig() {
...
+        .define(
+            KafkaConnectorConfigParams.SNOWFLAKE_ICEBERG_CREATE_TABLE_OPTIONS,
+            STRING,
+            KafkaConnectorConfigParams.SNOWFLAKE_ICEBERG_CREATE_TABLE_OPTIONS_DEFAULT,
+            LOW,
+            "SQL clauses spliced into the auto-created Iceberg table's CREATE statement after the"
+                + " column list (e.g. \"EXTERNAL_VOLUME='v' ICEBERG_VERSION=3 CLUSTER BY (id)\")."
+                + " Lets operators set any managed-Iceberg create option without a per-option knob."
+                + " Only valid when snowflake.autocreate.table.type=iceberg. Do NOT include"
+                + " RECORD_METADATA, CATALOG, ENABLE_SCHEMA_EVOLUTION or ERROR_LOGGING -- the"
```
sfc-gh-bmikaili: dont include record_metadata

src/main/java/com/snowflake/kafka/connector/config/ConnectorConfigDefinition.java
```diff
@@ -456,6 +456,36 @@ public static ConfigDef getConfig() {
             10,
             Width.NONE,
             KafkaConnectorConfigParams.SNOWFLAKE_ENABLE_SCHEMATIZATION)
+        .define(
+            KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE,
+            STRING,
+            KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT,
+            ConfigDef.ValidString.in("snowflake", "iceberg", "none"),
+            MEDIUM,
+            "Controls table auto-creation. 'snowflake' (default): auto-create a standard FDN table"
```
sfc-gh-skurella: NIT: I'm not sure customers are familiar with the term "FDN"

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
```diff
@@ -324,6 +328,39 @@ public static Builder builderFrom(Map<String, String> raw, boolean skipTaskSpeci
...
+    TableType tableType =
+        TableType.fromConfig(
+            config.get(KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE));
+
+    String icebergCreateTableOptions =
+        config
+            .getOrDefault(
+                KafkaConnectorConfigParams.SNOWFLAKE_ICEBERG_CREATE_TABLE_OPTIONS,
+                KafkaConnectorConfigParams.SNOWFLAKE_ICEBERG_CREATE_TABLE_OPTIONS_DEFAULT)
+            .trim();
```
sfc-gh-skurella: NIT: we can make this value `Optional<String>` and not have a default value

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
```diff
@@ -324,6 +328,39 @@ public static Builder builderFrom(Map<String, String> raw, boolean skipTaskSpeci
...
+    if (!icebergCreateTableOptions.isEmpty() && tableType != TableType.ICEBERG) {
+      throw new IllegalArgumentException(
+          "snowflake.iceberg.create.table.options is only valid when"
+              + " snowflake.autocreate.table.type=iceberg (got table.type="
```
sfc-gh-skurella: NIT: interpolate with constants rather than copy setting names because it might go stale if they ever change

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
```diff
@@ -378,7 +417,13 @@ private static Optional<Password> optionalPassword(String value) {
...
   /** Creates a new builder. Used by {@link #from(Map)} and by tests. */
   public static Builder builder() {
-    return new AutoValue_SinkTaskConfig.Builder();
+    // Default the Iceberg auto-creation properties here so every builder() caller (including
+    // raw-builder test helpers) gets valid values; builderFrom overrides them from config. Without
+    // this, AutoValue throws "missing required properties" because these props are non-nullable.
+    return new AutoValue_SinkTaskConfig.Builder()
+        .tableType(TableType.SNOWFLAKE)
+        .icebergCreateTableOptions(
+            KafkaConnectorConfigParams.SNOWFLAKE_ICEBERG_CREATE_TABLE_OPTIONS_DEFAULT);
```
sfc-gh-skurella: Please remove those defaults from here and add them to the test builder. This is not the right place to set test defaults.

src/main/java/com/snowflake/kafka/connector/config/TableType.java
```diff
@@ -0,0 +1,35 @@
...
+/** Connector-wide target table type for auto-creation routing. */
+public enum TableType {
+  SNOWFLAKE,
+  ICEBERG,
+  NONE;
+
+  /** Config string -> enum. Null/empty defaults to SNOWFLAKE (today's behavior). */
+  public static TableType fromConfig(String value) {
+    if (value == null || value.trim().isEmpty()) {
+      return SNOWFLAKE;
```
sfc-gh-skurella: Let's either use the default value constant or return `Optional<TableType>` so we set `tableType = TableType.fromConfig(...).orElse(<default_constant>);`.

src/main/java/com/snowflake/kafka/connector/Constants.java
```diff
@@ -115,6 +115,19 @@ public static final class KafkaConnectorConfigParams {
     public static final String SNOWFLAKE_ENABLE_SCHEMATIZATION = "snowflake.enable.schematization";
     public static final boolean SNOWFLAKE_ENABLE_SCHEMATIZATION_DEFAULT = true;
 
+    // Iceberg table auto-creation (SNOW-3552261)
+    public static final String SNOWFLAKE_AUTOCREATE_TABLE_TYPE = "snowflake.autocreate.table.type";
+    public static final String SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT = "snowflake";
```
sfc-gh-skurella: NIT: could this be `TableType.SNOWFLAKE`? The convention for defaults in this file seems to be the target type and not an input string.

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
```diff
@@ -288,15 +300,65 @@ public Map<String, Object> getContent() {
...
   }
 
+  /**
+   * Reshape a Kafka metadata map to exactly the fields of {@code ICEBERG_METADATA_OBJECT_SCHEMA}.
+   * Every declared field is present (absent ones {@code null}, which the strict typed-OBJECT cast
+   * accepts because the schema fields are nullable, whereas a <em>missing</em> field is rejected);
+   * the timestamp field — emitted by Kafka under {@code TimestampType.name} — is normalized to
+   * "CreateTime"; fields outside the schema are dropped.
+   */
+  private static Map<String, Object> conformIcebergMetadata(Map<String, Object> metadata) {
```
sfc-gh-skurella: I'd like to better understand the purpose of this function.

1. Could you please add unit tests which illustrate its behavior with examples?

2. I'm assuming it's not feasible to simply make ICEBERG_METADATA_OBJECT_SCHEMA accept the "normal" metadata we'd typically emit. Is this because the cast to structured objects in MBCN pipes doesn't handle missing values for optional fields? I'm not 100% sure from the description, maybe it's worth rephrasing.

3. Would it be feasible to instead say that all metadata *must* be ingested when using Iceberg? Are there any Iceberg customers that have disabled some of the metadata? Then we wouldn't need to add any nulls here at all.
  a) Removing metadata might save some storage cost but it actually wouldn't save almost any ingestion cost since we still have to send all the keys and nulls.
  b) Requiring all metadata would significantly simplify this PR - we basically wouldn't need to adjust the metadata we're sending based on the RECORD_METADATA type.

> fields outside the schema are dropped.

4. When would you expect to have to drop fields? Should this be an assertion instead? (We're not checking against the actual table's RECORD_METADATA fields anyway since we just have a static structured object schema here in this function.)

> the timestamp field — emitted by Kafka under {@code TimestampType.name} — is normalized to "CreateTime"

5. Would it make sense to just create both "create time" and "log append time" fields and just keep one of the values "null"? I'm wondering if it could be confusing to ingest one quantity under the name of another, but maybe they're interchangeable?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -260,7 +253,33 @@ public void startPartitions(Collection<TopicPartition> partitions) {
       // When validation is enabled, reject non-default pipes (pipes whose name equals the table
       // name) because validation assumptions may not hold for user-created pipes.
       final String targetPipeName;
-      if (taskConfig.getValidation() == SnowflakeValidation.CLIENT_SIDE) {
+      if (taskConfig.getTableType() == TableType.NONE) {
+        // table.type=none: the pipe must already exist (we created nothing). Fail fast if neither
+        // the default nor a named pipe is present, and name the detected table type.
+        boolean defaultPipeExists = this.conn.pipeExist(buildDefaultPipeName(tableName));
```
sfc-gh-skurella: What's the purpose of checking whether a default pipe exists? Shouldn't we just be checking if a pipe exists or a table exists with the name we derived from the topic? I think default pipes are created dynamically when we ingest data so reasoning about this is hard for me.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-skurella, line 252: Actually, this existing code might be wrong. If a pipe already exists with the target table name, we'll use that pipe, and in that case we shouldn't even attempt to create the table. WDYT? Should we fix this first before landing the current PR?

sfc-gh-skurella (commented):
Thanks! Partial review.

The PR description and setting comment both claim the table is created if a table doesn't already exist. But wouldn't it be fair to say that if a pipe exists when it doesn't matter if a table exists or not? That's my expectation—if you agree, please make sure the writing reflects that, and if not, let's discuss how the connector should behave.

sfc-gh-bmikaili:
Thanks @sfc-gh-skurella — addressed in af8b216d. Where each point landed:

**Pipe existence (your summary + both `SnowflakeSinkServiceV2` comments)**
- **:259 (`none` default-pipe check)** — you're right, fixed. Removed the `none` fail-fast entirely. The default pipe is always connector-managed (created lazily at first ingest), so requiring it to pre-exist was wrong. `none` now resolves the pipe exactly like every other type; its only job is to *not* auto-create the table (already enforced by `createTableIfNotExists`, which throws for `none` when the table is missing). Config doc updated accordingly.
- **:250 (create-before-pipe / "fix first?")** — I don't think this is a bug: `createTableIfNotExists` short-circuits `if (tableExist) return;`, and a pipe implies its table exists, so when a pipe is present the create is a no-op (no stray DDL). No prerequisite fix needed — the `none` cleanup above makes the flow clearer regardless.

**`conformIcebergMetadata` (your 5 questions)**
- **(2) why not accept normal metadata** — exactly your hypothesis: the managed-Iceberg RECORD_METADATA cast is a *strict* typed-OBJECT cast that rejects a *missing* field (present-null is fine). Rephrased the Javadoc to say this plainly.
- **(3) require all metadata → simplify** — done: `table.type=iceberg` now **rejects** any disabled `snowflake.metadata.*` flag at config time, so the config-gated fields are guaranteed present and are no longer null-padded. Caveat: `key`/`headers` are *record-level*, not config-gated (a keyless/headerless record omits them regardless of config), so config can't guarantee them — the method still pads **only those two**. Net: the pad shrank from 7 fields to 2, with a clear reason.
- **(1) tests** — added unit examples (CreateTime passthrough, LogAppendTime→CreateTime, extra field dropped, non-String key coerced, key/headers padded when absent) + a config test that iceberg + disabled metadata is rejected.
- **(4) dropping fields / assertion** — kept as a silent drop (forward-compat: a new KC metadata field before the GS schema catches up) rather than an assert.
- **(5) timestamp naming** — kept the `CreateTime` normalization for now; doing it "properly" (separate CreateTime/LogAppendTime fields) needs the **GS-side `ICEBERG_METADATA_OBJECT_SCHEMA`** to carry both fields — cross-repo, out of scope. Flagging as a known limitation / follow-up.

**NITs** — "FDN" → "standard Snowflake" in the doc; `TableType.fromConfig` → `Optional<TableType>` (callers `.orElse(SNOWFLAKE)`); default constant is now `TableType.SNOWFLAKE`; error messages interpolate the config-key constants; test-only defaults moved into `SinkTaskConfigTestBuilder`. Skipped one: `Optional<String>` for `iceberg.create.table.options` rippled to several prod+test callers — happy to do it in a follow-up if you feel strongly.

(Unrelated: the PR shows a `master` merge conflict from drift while under review — I'll rebase that separately; this push is just the review resolutions.)

sfc-gh-bmikaili:
Follow-up on **Q4** (dropping unknown metadata fields): revised per your instinct — `conformIcebergMetadata` now **throws** `IllegalStateException` on a field outside the Iceberg schema instead of silently dropping it (commit 8778353). `buildMetadata` only ever emits schema fields, so this fires only on KC/GS-schema drift (a new metadata field added on one side but not the other) — better to surface that loudly than silently omit customer metadata. Made the method package-private and added a direct unit test for the throw.

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
```diff
@@ -383,6 +387,64 @@ public static Builder builderFrom(Map<String, String> raw, boolean skipTaskSpeci
             .orElse(
                 KafkaConnectorConfigParams.SNOWFLAKE_FEATURE_VALIDATION_ERROR_TABLE_NAME_DEFAULT);
 
+    TableType tableType =
```
sfc-gh-skurella: NIT: please rename to `autocreatedTableType` or similar because it was already a little confusing for me to see table-specific info below in the `SinkTaskConfig` until I realized it's the auto-create setting

src/main/java/com/snowflake/kafka/connector/config/ConnectorConfigDefinition.java
sfc-gh-bmikaili: Done in e0fa3a66 — reworded to "standard Snowflake" (and fixed the `none` description while there).

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
sfc-gh-bmikaili: Done in e0fa3a66 — error messages now interpolate the config-key constants instead of hardcoded setting names.

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
sfc-gh-bmikaili: Done in e0fa3a66 — moved the `tableType`/`icebergCreateTableOptions` defaults out of production `builder()` into `SinkTaskConfigTestBuilder`.

src/main/java/com/snowflake/kafka/connector/config/TableType.java
sfc-gh-bmikaili: Done in e0fa3a66 — `TableType.fromConfig` now returns `Optional<TableType>`; callers use `.orElse(TableType.SNOWFLAKE)`.

src/main/java/com/snowflake/kafka/connector/Constants.java
sfc-gh-bmikaili: Done in e0fa3a66 — `SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT` is now `TableType.SNOWFLAKE.configValue()`.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-bmikaili: Done in e0fa3a66 — removed the default-pipe check entirely. `none` now resolves the pipe like every other type (the default pipe is connector-managed / created lazily at ingest); its only job is to not auto-create the table (enforced by `createTableIfNotExists`).

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
sfc-gh-bmikaili: Addressed most of this (details in the summary comment): (2) confirmed — the strict typed-OBJECT cast rejects a *missing* field (present-null is fine); Javadoc rephrased. (3) took your suggestion: `table.type=iceberg` now rejects any disabled `snowflake.metadata.*` flag at config time, so we no longer null-pad the config-gated fields — we pad only `key`/`headers`, which are record-level (keyless/headerless records) and can't be guaranteed by config. (1) added unit tests. (4) per your instinct it now **throws** on an unknown field instead of silently dropping (8778353). Leaving open only for **(5)** — normalizing `LogAppendTime` into a `CreateTime`-named field: doing it properly (separate fields) needs the GS-side `ICEBERG_METADATA_OBJECT_SCHEMA` to carry both, which is cross-repo / out of scope here. Flagging as a follow-up.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-bmikaili: Responding rather than changing: I don't think this is a live bug. `createTableIfNotExists` short-circuits with `if (tableExist) return;`, and a pipe always implies its table exists — so when a pipe is present the create is a no-op (no stray DDL). The `none` cleanup in the thread above makes the flow clearer regardless. Leaving open in case you disagree.

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
sfc-gh-bmikaili: Now done in `02cfbbd4` — `getIcebergCreateTableOptions()` is `Optional<String>` end-to-end (accessor, builder setter, and the `from()` parse via `Optional.ofNullable(config.get(...)).map(String::trim).filter(s -> !s.isEmpty())`), so there's no `""` sentinel default. The two prod callers in `SnowflakeSinkServiceV2` use `.orElse("")`, which keeps the DDL/regex behavior identical.

src/main/java/com/snowflake/kafka/connector/config/ConnectorConfigDefinition.java
sfc-gh-bmikaili: Done in `02cfbbd4` — dropped `RECORD_METADATA` from the "Do NOT include" list in the `snowflake.iceberg.create.table.options` doc. It's a column the connector always creates, not a CREATE-option clause, so it didn't belong in that list.

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
```diff
@@ -383,6 +387,63 @@ public static Builder builderFrom(Map<String, String> raw, boolean skipTaskSpeci
...
+        TableType.fromConfig(config.get(KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE))
+            .orElse(TableType.SNOWFLAKE);
```
sfc-gh-skurella: Could we please store this under `SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT` so all config-related constants are grouped together?

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
```diff
@@ -383,6 +387,63 @@ public static Builder builderFrom(Map<String, String> raw, boolean skipTaskSpeci
...
+              + ")");
+    }
+
+    // Client-side validation drives client-side schema evolution (ALTER ADD COLUMN issued by the
+    // connector). For managed Iceberg, schema evolution is performed server-side; the client-side
+    // RowValidator cannot even model the structured RECORD_METADATA column. Reject the combination
+    // loudly rather than silently producing a misleading runtime failure. Server-side SE is the
+    // supported mechanism: use snowflake.validation=server_side with Iceberg + schematization.
+    if (tableType == TableType.ICEBERG
+        && enableSchematization
```
sfc-gh-skurella: Could we remove this line, i.e. reject Iceberg + client-side, regardless of the schematization setting?

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
```diff
@@ -441,7 +504,13 @@ private static Optional<Password> optionalPassword(String value) {
...
   }
 
-  /** Creates a new builder. Used by {@link #from(Map)} and by tests. */
+  /**
+   * Creates a new builder. Used by {@link #from(Map)} (via {@code builderFrom}) and by {@code
+   * SinkTaskConfigTestBuilder} in tests. The Iceberg-related fields ({@code tableType}, {@code
+   * icebergCreateTableOptions}) have no defaults here — {@code builderFrom} always sets them from
+   * config, and tests must set them via {@code SinkTaskConfigTestBuilder} (which provides the
+   * defaults).
+   */
```
sfc-gh-skurella, line 539: NIT: I don't see the value in explaining why we don't override specific defaults in this function. Maybe leave the sentence saying that tests must set the defaults via SinkTaskConfigTestBuilder so future agents don't put things in this method again as this already happened a few times.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
```diff
@@ -108,6 +108,13 @@ public class SnowpipeStreamingPartitionChannel implements TopicPartitionChannel
   private volatile Map<String, ColumnSchema> tableSchema;
   private final boolean shouldEvolveSchema;
 
+  // Whether the target table's RECORD_METADATA is the structured Iceberg OBJECT (vs FDN VARIANT),
+  // resolved lazily on first use and memoized for the channel's lifetime. We probe the ACTUAL
+  // table via conn.isIcebergTable rather than trusting the configured table.type, because the
+  // config only governs auto-creation: an existing table may differ from it (e.g. table.type=none
+  // targeting a pre-existing Iceberg table). null = not yet resolved.
+  private volatile Boolean recordMetadataIsIcebergObject;
```
sfc-gh-skurella: I'd prefer we called this a "structured" object rather than an "Iceberg" object because:
- structured objects are supported for FDNs, also - even though FDNs won't use them in KC
- Iceberg v3 should *not* use structured objects so this name might be a little misleading when reasoning about Iceberg v3 behavior
- it will be closer to the `isRecordMetadataStructuredObject` function on `conn`

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
```diff
@@ -297,6 +305,21 @@ public CompletableFuture<Void> waitForLastProcessedRecordCommitted() {
...
    */
+  /**
+   * Whether record metadata must be conformed to the strict structured-OBJECT RECORD_METADATA
+   * schema. True iff the actual target table's RECORD_METADATA column is a structured OBJECT --
+   * which is only the case for managed-Iceberg v2 tables (v3 and FDN use VARIANT, which accepts the
+   * natural metadata map as-is). We probe the real column type via {@code
+   * isRecordMetadataStructuredObject} once and memoize; the configured table.type only governs
+   * auto-creation, and even an auto-created Iceberg table may be VARIANT (v3) or OBJECT (v2).
+   */
+  private boolean conformMetadataToIcebergObject() {
```
sfc-gh-skurella: NIT: would it make sense to inject this from a constructor like we do with `shouldEvolveSchema`? It would be more uniform this way and this class would be a little simpler to reason about.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -279,12 +276,70 @@ public void startPartitions(Collection<TopicPartition> partitions) {
...
+        this.conn.createIcebergTableWithOnlyMetadataColumn(
+            tableName, taskConfig.getIcebergCreateTableOptions().orElse(""));
+        break;
+      case NONE:
+        // Missing table + none: fail loudly and tell the operator their two ways out.
+        throw SnowflakeErrors.ERROR_0034.getException(
+            "Table '"
+                + tableName
+                + "' does not exist and snowflake.autocreate.table.type=none. Either create the"
+                + " table and its pipe yourself, or set snowflake.autocreate.table.type=snowflake"
```
sfc-gh-skurella: This is a little misleading - customers *don't* need to create a pipe themselves. It's strictly optional and we'll use the default pipe otherwise.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -279,12 +276,70 @@ public void startPartitions(Collection<TopicPartition> partitions) {
...
+      case ICEBERG:
+        // Non-schematized Iceberg stores the raw record in a RECORD_CONTENT VARIANT column, which
+        // is only supported on Iceberg format v3. Auto-creating without ICEBERG_VERSION=3 yields a
+        // v2 table (the default) that cannot hold RECORD_CONTENT, so fail fast at startup instead
+        // of mysteriously at ingest time.
+        if (!taskConfig.isEnableSchematization()
+            && !taskConfig
+                .getIcebergCreateTableOptions()
+                .orElse("")
+                .matches("(?is).*ICEBERG_VERSION\\s*=\\s*3(?!\\d).*")) {
```
sfc-gh-skurella: NIT: extract to aux variable `icebergV3Requested` for readability

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
```diff
@@ -294,15 +311,92 @@ public Map<String, Object> getContent() {
   }
 
   public Map<String, Object> getContentWithMetadata(boolean includeMetadata) {
+    return getContentWithMetadata(includeMetadata, false);
+  }
```
sfc-gh-skurella: I'd prefer to explicitly provide the extra arg at call sites and remove this 1-arg definition - this way it will be explicit if we forgot to provide `includeMetadata` at some call site

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
```diff
@@ -294,15 +311,92 @@ public Map<String, Object> getContent() {
...
+  /**
+   * @param conformToIcebergMetadataSchema when true, the RECORD_METADATA value is normalized via
+   *     {@link #conformIcebergMetadata}: the timestamp field is renamed to "CreateTime", fields
+   *     outside {@code ICEBERG_METADATA_OBJECT_SCHEMA} are dropped, and the {@code key} field is
+   *     coerced to {@code String}. Full metadata presence is guaranteed by config-time validation
+   *     (see {@code SinkTaskConfig}). FDN tables use a VARIANT RECORD_METADATA and pass {@code
+   *     false} (behavior unchanged).
+   */
+  public Map<String, Object> getContentWithMetadata(
+      boolean includeMetadata, boolean conformToIcebergMetadataSchema) {
```
sfc-gh-skurella: NIT: I'd prefer to use the term "structured" rather than "Iceberg" because it's a structured type limitation we're overcoming here

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
```diff
@@ -294,15 +311,92 @@ public Map<String, Object> getContent() {
...
+   */
+  // Package-private for testing.
+  static Map<String, Object> conformIcebergMetadata(Map<String, Object> metadata) {
+    Map<String, Object> conformed = new HashMap<>();
+    for (Map.Entry<String, Object> entry : metadata.entrySet()) {
+      String field = entry.getKey();
+      if (TimestampType.CREATE_TIME.name.equals(field)
+          || TimestampType.LOG_APPEND_TIME.name.equals(field)) {
+        field = CREATE_TIME;
+      }
```
sfc-gh-skurella: See (5) from https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1487#comment-PRRC_kwDOCzOBLM7U_w_L - I'm not convinced they're the *same* quantity and would prefer to keep both fields in the metadata and only populate one of them so they retain the original name.

But if that's what we did for KC v3, we can probably keep doing it here.

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
```diff
@@ -294,15 +311,92 @@ public Map<String, Object> getContent() {
...
+      if (!ICEBERG_METADATA_FIELD_SET.contains(field)) {
+        throw new IllegalStateException(
+            "Unexpected metadata field '"
+                + field
+                + "' not in the managed-Iceberg RECORD_METADATA schema (ICEBERG_METADATA_FIELDS)."
+                + " KC metadata emission and the Iceberg schema have drifted: add the field to"
+                + " ICEBERG_METADATA_OBJECT_SCHEMA (GS) and ICEBERG_METADATA_FIELDS, or stop"
+                + " emitting it for managed Iceberg.");
```
sfc-gh-skurella, line 387: Thinking a little more about this, we're comparing connector-produced metadata with connector-provided schema. Would it make sense to also validate the schema of RECORD_METADATA to ensure it's in sync with what the connector would have created? Basically make sure all required fields are present. Otherwise all records will go straight into the error table because of how strict the check is :/ Not a blocker but if this is easy to do, might improve the experience if we ever need to touch this schema in the future.

src/main/java/com/snowflake/kafka/connector/Constants.java
```diff
@@ -143,6 +143,22 @@ public static final class KafkaConnectorConfigParams {
     public static final String SNOWFLAKE_ENABLE_SCHEMATIZATION = "snowflake.enable.schematization";
     public static final boolean SNOWFLAKE_ENABLE_SCHEMATIZATION_DEFAULT = true;
 
+    // Iceberg table auto-creation (SNOW-3552261)
+    public static final String SNOWFLAKE_AUTOCREATE_TABLE_TYPE = "snowflake.autocreate.table.type";
+    // Use the canonical config value from TableType so the default stays in sync if the enum
+    // changes.
+    public static final String SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT =
+        com.snowflake.kafka.connector.config.TableType.SNOWFLAKE.configValue();
```
sfc-gh-skurella: NIT: would prefer to store a well-typed default and only call `configValue` in the `ConnectorConfigDefinition`.

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
```diff
@@ -294,15 +311,92 @@ public Map<String, Object> getContent() {
...
+      }
+      Object value = entry.getValue();
+      // The schema declares `key STRING`, but the record key can convert to a non-String (e.g. an
+      // INT-keyed topic yields an Integer, a struct/Avro key a Map). Coerce to String so the
+      // strict typed-OBJECT cast accepts it.
+      if (KEY.equals(field) && value != null && !(value instanceof String)) {
+        value = String.valueOf(value);
+      }
+      conformed.put(field, value);
```
sfc-gh-skurella, line 421: Headers are a string&rarr;string map - I don't think the current code converts their values into strings, it's probably a gap - although I'm not sure if the backend can maybe already convert that correctly as a JSON or sth.

sfc-gh-skurella (commented):
Thanks! Reviewed all prod code and will continue reviewing tests later today. One "global" change I'd prefer for us to do is to refer to structured type logic using the name "structured" / "structured type" rather than Iceberg because it will be confusing when reasoning about Iceberg v3 which supports VARIANT and doesn't need any special handling of metadata etc.

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
sfc-gh-bmikaili: Added unit tests illustrating `conformIcebergMetadata` (SnowflakeSinkRecordTest: per-timestamp-type passthrough, padding of absent fields, drift-throw, key coercion). Re (2): the structured-OBJECT RECORD_METADATA can't just accept the natural map — the v2 typed-OBJECT cast **rejects a missing declared field** (verified: v2 e2e fails `Typed object schema mismatch in conversion` without padding). That's exactly why conform pads absent fields (key/headers + the unused timestamp) with null. v3/FDN use VARIANT and skip conform entirely. (927cc166)

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-bmikaili: I don't think this is a live bug: `createTableIfNotExists` short-circuits with `if (tableExist) return;`, and a pipe always implies its table exists, so when a pipe is present the create is a no-op — no stray DDL. This PR now also validates the existing table's structured RECORD_METADATA schema in that same short-circuit branch. Leaving open in case you'd still prefer an explicit pipe-first check. (927cc166)

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
sfc-gh-bmikaili: Done — renamed to `getAutocreatedTableType()` / `Builder.autocreatedTableType()`; all callers updated. (927cc166)

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
sfc-gh-bmikaili: Done — the default is now `SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT` (a well-typed `TableType`), grouped with the other config constants; `.configValue()` is called only in `ConnectorConfigDefinition`. (927cc166)

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
sfc-gh-bmikaili: Done — dropped the `enableSchematization` condition; `table.type=iceberg` + `validation=client_side` is now rejected unconditionally. (927cc166)

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
sfc-gh-bmikaili: Done — trimmed to one sentence noting tests must set defaults via `SinkTaskConfigTestBuilder` so future changes don't re-add them here. (927cc166)

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
sfc-gh-bmikaili: Done — renamed to "structured object" (field + javadoc); mirrors `conn.isRecordMetadataStructuredObject`. (927cc166)

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
sfc-gh-bmikaili: Done — the structured-object flag is now a `final` constructor param injected by `PartitionChannelManager` (which computes `conn.isRecordMetadataStructuredObject(tableName)`), same pattern as `shouldEvolveSchema`. (927cc166)

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-bmikaili: Done — removed "and its pipe yourself"; the pipe is optional (default pipe used otherwise). (927cc166)

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-bmikaili: Done — extracted `icebergV3Requested`. (927cc166)

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
sfc-gh-bmikaili: Done — removed the 1-arg overload; all callers now pass the second arg explicitly. (927cc166)

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
sfc-gh-bmikaili: Done — reworded to "structured object". (927cc166)

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
sfc-gh-bmikaili: Done — we now emit `CreateTime` and `LogAppendTime` as distinct OBJECT fields (no rename). A record carries one timestamp type; conform pads the unused one with null (the v2 strict cast needs every declared field present). (927cc166)

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
sfc-gh-bmikaili: Done — added `validateStructuredObjectMetadataSchema` at startup (existing-table branch, gated on `isRecordMetadataStructuredObject`): fails fast with ERROR_0034 if the table's structured RECORD_METADATA is missing any connector-emitted field, instead of routing every row to the error table. Extra fields are tolerated (our map is a subset). (927cc166)

src/main/java/com/snowflake/kafka/connector/Constants.java
sfc-gh-bmikaili: Done — `SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT` is now a well-typed `TableType`; `.configValue()` is called only in `ConnectorConfigDefinition`. (927cc166)

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
sfc-gh-bmikaili: Good catch — real gap: structured headers + `table.type=iceberg` is reachable, and `convertHeaders` keeps native (non-String) values while the schema declares `headers MAP(VARCHAR,VARCHAR)`, so they'd fail the cast. Added a config-time guard rejecting structured-headers + iceberg. (927cc166)

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
sfc-gh-bmikaili: Follow-up: the accessor `getTableType()` and builder setter were renamed to `autocreatedTableType` in 927cc166, but a residual local variable `tableType` in `SinkTaskConfig.from()` was left behind — which read as un-renamed in the diff. Fixed in eb4711a2; zero lowercase `tableType` identifiers remain.

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
sfc-gh-bmikaili: Follow-up: revised in eb4711a2. The config-time reject guard was over-broad — the `MAP(VARCHAR,VARCHAR)` limitation only applies to the v2 structured-OBJECT path, so it wrongly rejected structured headers on v3/VARIANT tables (which hold nested headers fine) and broke IcebergIngestionIntoVariantIT. Removed the guard; `conformIcebergMetadata` now coerces header-map values to String on the v2 structured-OBJECT path only. Validated by IcebergIngestionIntoVariantIT + a new conform unit test.

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
```diff
@@ -396,39 +396,36 @@ public static Builder builderFrom(Map<String, String> raw, boolean skipTaskSpeci
...
         && validation == SnowflakeValidation.CLIENT_SIDE) {
       throw new IllegalArgumentException(
           KafkaConnectorConfigParams.SNOWFLAKE_VALIDATION
               + "=client_side is not supported with "
               + KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE
-              + "=iceberg when schema evolution is enabled ("
-              + KafkaConnectorConfigParams.SNOWFLAKE_ENABLE_SCHEMATIZATION
-              + "=true). Managed Iceberg schema evolution is server-side; set "
+              + "=iceberg. Managed Iceberg uses a structured RECORD_METADATA column that the"
+              + " client-side RowValidator cannot model. Set "
```
sfc-gh-skurella: NIT: I don't think customers know what a RowValidator is

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
```diff
@@ -309,15 +311,11 @@ public CompletableFuture<Void> waitForLastProcessedRecordCommitted() {
...
+   * PartitionChannelManager#buildChannel}.
    */
-  private boolean conformMetadataToIcebergObject() {
-    if (recordMetadataIsIcebergObject == null) {
-      recordMetadataIsIcebergObject = conn.isRecordMetadataStructuredObject(tableName);
-    }
-    return recordMetadataIsIcebergObject;
+  private boolean conformMetadataToStructuredObject() {
+    return recordMetadataIsStructuredObject;
   }
```
sfc-gh-skurella: NIT: inline this function, we can just use the constant

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
sfc-gh-bmikaili: Fixed in 1948e4ca — `SinkTaskConfig.from()` now uses `.orElse(KafkaConnectorConfigParams.SNOWFLAKE_AUTOCREATE_TABLE_TYPE_DEFAULT)` instead of the hardcoded `.orElse(TableType.SNOWFLAKE)`. The default is a well-typed `TableType` constant in `Constants`, `.configValue()` is called only in `ConnectorConfigDefinition`, and the `tableType` local var is renamed to `autocreatedTableType`. (Your review anchored to 932f01c7, before the round-2 fixes; current head is 1948e4ca.)

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
sfc-gh-skurella: Extra fields should not be tolerated since their absence will fail ingestion

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -279,12 +281,205 @@ public void startPartitions(Collection<TopicPartition> partitions) {
...
+   * <p>Splits on top-level commas only (respects nested parentheses, e.g. {@code MAP(VARCHAR,
+   * VARCHAR)}), then extracts the first whitespace-delimited token of each clause as the field
+   * name. Returns an empty set if the string does not start with {@code OBJECT(}.
+   *
+   * <p>Type checking is intentionally omitted: Snowflake uses aliases (LONG/BIGINT, INTEGER/INT)
+   * that would require alias-normalization to compare safely. Name-presence is the critical
+   * invariant; a type mismatch produces a different, comprehensible server error.
+   */
+  // Package-private for unit testing.
+  static Set<String> parseObjectSubfieldNames(String objectTypeStr) {
```
sfc-gh-skurella: We could do it without parsing the type by using https://docs.snowflake.com/en/sql-reference/info-schema/fields.

Beyond a smaller bug surface area, that would also account for special characters in field names, although that's very unlikely in this instance. But I think it would make the code much simpler to maintain.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -279,12 +281,205 @@ public void startPartitions(Collection<TopicPartition> partitions) {
...
+   * row to the error table. Fail fast at startup with a clear message instead.
+   *
+   * <p>A table declaring <em>extra</em> fields beyond what the connector emits is fine (the
+   * connector's map is a subset; those extra fields cast to null) and is not rejected.
+   *
+   * @throws com.snowflake.kafka.connector.internal.SnowflakeKafkaConnectorException (ERROR_0034)
+   *     when any connector-emitted field is absent from the table's OBJECT schema
+   */
+  @VisibleForTesting
+  void validateStructuredObjectMetadataSchema(String tableName) {
```
sfc-gh-skurella, line 376: NIT: I'd prefer to move this logic outside this file because it's already quite large, maybe to `streaming/iceberg/IcebergDDLTypes.java`?

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
```diff
@@ -293,16 +321,102 @@ public Map<String, Object> getContent() {
...
+      // The schema declares `key STRING`, but the record key can convert to a non-String (e.g. an
+      // INT-keyed topic yields an Integer, a struct/Avro key a Map). Coerce to String so the
+      // strict typed-OBJECT cast accepts it.
+      if (KEY.equals(field) && value != null && !(value instanceof String)) {
+        value = String.valueOf(value);
+      }
+      // The schema declares `headers MAP(VARCHAR, VARCHAR)`. When structured headers are enabled
+      // convertHeaders keeps native (non-String) values; coerce each to String so the strict cast
+      // accepts them (v3/VARIANT tables never reach here and keep headers as nested VARIANT).
+      if (HEADERS.equals(field) && value instanceof Map) {
```
sfc-gh-skurella: We probably shouldn't gate by `value instanceof Map` - this should be an invariant, if anything we should assert that this is the case

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
```diff
@@ -293,16 +321,102 @@ public Map<String, Object> getContent() {
...
+      conformed.put(field, value);
+    }
+    // Pad every absent-able field with null so the strict v2 typed-OBJECT cast does not reject the
+    // row. A record carries exactly one timestamp type (the other is absent); key and headers are
+    // absent on keyless/headerless records. putIfAbsent is safe: it leaves already-populated fields
+    // (e.g. the present timestamp) untouched.
+    conformed.putIfAbsent(KEY, null);
+    conformed.putIfAbsent(HEADERS, null);
+    conformed.putIfAbsent(CREATE_TIME, null);
+    conformed.putIfAbsent(LOG_APPEND_TIME, null);
```
sfc-gh-skurella: NIT: we could just loop over `ICEBERG_METADATA_FIELDS` to not miss anything here

sfc-gh-skurella (commented):
Thanks! Looks much better. Left a few comments but we're probably close to merging this PR

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
sfc-gh-bmikaili: Fixed in 3169d736 — validation now requires an EXACT field-set match and rejects EXTRA fields, not just missing ones. You're right: an extra declared field the connector never emits is absent from every row's map, so the v2 typed-OBJECT cast would reject every row. `IcebergDDLTypes.validateRecordMetadataFields` throws ERROR_0034 naming the missing and/or extra fields.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-bmikaili: Done in 3169d736 — dropped the OBJECT-type-string parser; it now queries `INFORMATION_SCHEMA.FIELDS` (`SELECT FIELD_NAME ... WHERE TABLE_NAME=UPPER(?) AND COLUMN_NAME=UPPER(?) AND TABLE_SCHEMA=CURRENT_SCHEMA()`) via a new `conn.getStructuredObjectFieldNames`.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-bmikaili: Done in 3169d736 — moved the logic to `IcebergDDLTypes.validateRecordMetadataFields`; `SnowflakeSinkServiceV2` now only gates on `isRecordMetadataStructuredObject`, fetches the fields, and delegates.

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
sfc-gh-bmikaili: Done in 3169d736 — no longer gates on `instanceof Map`; it treats it as an invariant and throws `IllegalStateException` if HEADERS is non-null but not a Map (KC/schema drift) rather than silently skipping.

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
sfc-gh-bmikaili: Done in 3169d736 — reworded to drop "RowValidator": the message now says client-side validation isn't supported for managed Iceberg (RECORD_METADATA is a structured object handled server-side) and to set `snowflake.validation=server_side`.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
sfc-gh-bmikaili: Done in 3169d736 — inlined; call sites use the `recordMetadataIsStructuredObject` field directly and the accessor method is removed.

src/main/java/com/snowflake/kafka/connector/records/SnowflakeSinkRecord.java
sfc-gh-bmikaili: Done in 3169d736 — padding now loops over `ICEBERG_METADATA_FIELDS` (`putIfAbsent` per field), so it can't miss a field.

sfc-gh-skurella (approved):

PR #1493, opened 2026-06-20
add soak test with chaos

sfc-gh-skurella:
## Summary

Adds a long-running **soak test** (`test/tests/test_soak_with_chaos.py`) that hammers the Snowpipe Streaming v2 ingestion and channel-recovery path under continuous high-rate load while injecting several kinds of chaos, and asserts strict data-integrity invariants throughout. It runs **nightly** (and on manual dispatch), not on every PR.

The test produces a steady multi-partition JSON stream and, in parallel, continuously perturbs the connector. After the soak it drains and verifies there were **no gaps, no duplicates, and no lost records**, that **no task ever entered `FAILED`**, and that the task actually spent meaningful time in `put()`.

## Chaos types

| Chaos | Mechanism | Default cadence | What it exercises |
|---|---|---|---|
| Channel invalidation | `SYSTEM$STREAMING_CHANNEL_INVALIDATE` on a random partition's channel | every 5–120 s | reactive channel recovery (STALE_CONTINUATION_TOKEN → reopen) |
| Connector recreate | hard `DELETE` + `POST` of the connector | every 2–5 min | full task teardown and channel reopen from scratch |
| tasks.max rebalance | `PUT /config` with a new `tasks.max` ∈ {1, 2, 4} | every 60–180 s | Connect group rebalance; channels reassigned between tasks (fan-in / fan-out) |
| Pause / resume | `PUT /pause`, dwell ~10 s, `PUT /resume` | every 90–240 s | graceful task suspension + offset commit on pause, then catch-up |
| In-place restart | `POST /restart?includeTasks=true&onlyFailed=false` | every 90–240 s | graceful task restart (distinct from the hard recreate) |
| Add partitions | `createPartitions` (+2, capped at 12) | every 2–5 min | opening channels for brand-new partitions live; next rebalance spreads them |

Every source is individually toggleable (for A/B baselines) and all cadences are constants at the top of the file. A background **task-failure watchdog** polls connector status every 5 s and records any task ever seen in `FAILED`.

The **health of the chaos itself** is also asserted: each enabled source must have fired a duration-scaled minimum number of times (liveness) and must not have failed most of its attempts (error rate), so a silently-broken chaos thread can't quietly downgrade the soak into a weaker test that still passes.

## Invariants asserted

- **No duplicates:** `(partition, offset)` is globally unique.
- **No gaps / no head loss:** offsets are contiguous per partition and start at 0.
- **Exact delivery** (when the drain completes): `total_rows == records_produced`. If the drain times out, integrity is still checked on whatever landed and the timeout is re-raised so the test still fails.
- **No task failures** at any point (watchdog).
- **`put()` was actually exercised** (see JMX below).

## JMX (Jolokia)

**How:** `run_tests.sh --jmx` downloads the Jolokia JVM javaagent and attaches it to the Kafka Connect JVM via `KAFKA_OPTS`, exposing the connector's MBeans over HTTP on port 8778. The test reads the `put-duration` Codahale Timer MBean (wildcard across all task MBeans) through the Jolokia HTTP bridge and aggregates `Count` and `Count × Mean` across tasks to get total `put()` invocations and total seconds spent in `put()`.

**Why:** integrity alone doesn't prove the connector was doing real work — an idle or misconfigured task, or a mere trickle of data, could pass the row checks without ever stressing the recovery path. The `put-duration` metric is a direct, task-side signal that ingestion actually happened under load. The test asserts `put_count > 0` and that the task spent a non-trivial fraction of wall-clock inside `put()` (floor 1 %, deliberately well below the ~10–20 % steady state), so a chaos window that starves `put()` fails loudly. It's also the pacing signal used to keep the producer rate below the chaos-throttled ingestion ceiling.

## CI scheduling

The soak is far too long for the PR pipeline, so it runs out-of-band:

- **New `schedule` trigger:** nightly cron `0 20 * * *` (20:00 UTC) runs the full matrix, including the new `soak_with_chaos` job.
- **New `soak_with_chaos` job:** runs **only** on `schedule` or manual `workflow_dispatch` (never on `pull_request` / `push`), with a 600-minute timeout. It builds the Confluent connector, decrypts the **AZURE** profile, and runs the `soak` marker with `--jmx`.
- **Why AZURE:** `SYSTEM$STREAMING_CHANNEL_INVALIDATE` is not enabled on the AWS CI profile (the existing invalidation tests skip there), so the soak targets AZURE to actually exercise the recovery path it's built around.
- **`workflow_dispatch` inputs now default to `false`** (opt-in per job), and a new `run_soak_with_chaos` input was added so the soak can be triggered manually on demand.
- The `soak` pytest marker is **excluded** from all regular functional / stress jobs, so the long test never runs in the normal pipeline.

## Testing

https://github.com/snowflakedb/snowflake-kafka-connector/actions/runs/29098069518/job/86379587335

### DIFF
--- a/.github/actions/run-e2e-tests/action.yml
+++ b/.github/actions/run-e2e-tests/action.yml
@@ -31,6 +31,10 @@ inputs:
     description: "Enable mitmproxy reverse proxy for fault injection tests"
     required: false
     default: 'false'
+  jmx:
+    description: "Enable JMX metrics scraping via the Jolokia javaagent (sets KAFKA_OPTS so the connector JVM exposes MBeans on port 8778)"
+    required: false
+    default: 'false'
 
 runs:
   using: composite
@@ -55,6 +59,7 @@ runs:
         LOGS_DIR: "${{ github.workspace }}/test-logs"
         MARKER_FILTER: ${{ inputs.pressure == 'true' && 'pressure' || inputs['marker-filter'] }}
         MITMPROXY_FLAG: ${{ inputs['with-mitmproxy'] == 'true' && '--with-mitmproxy' || '' }}
+        JMX_FLAG: ${{ inputs.jmx == 'true' && '--jmx' || '' }}
       run: |
         ./run_tests.sh \
           --platform="$PLATFORM" \
@@ -63,6 +68,7 @@ runs:
           --java-version="$JAVA_VERSION" \
           --logs-dir="$LOGS_DIR" \
           $MITMPROXY_FLAG \
+          $JMX_FLAG \
           -- -m "$MARKER_FILTER"
 
     - name: Upload service logs on failure
--- a/.github/workflows/end-to-end.yaml
+++ b/.github/workflows/end-to-end.yaml
@@ -5,30 +5,38 @@ on:
     branches: [master, 'release-*']
   pull_request:
     branches: ['**']
+  schedule:
+    # Nightly run with all jobs, including the soak test.
+    - cron: '0 20 * * *'
   workflow_dispatch:
     inputs:
       run_functional:
         description: 'functional tests'
         type: boolean
-        default: true
+        default: false
       run_fault_injection:
         description: 'fault injection tests'
         type: boolean
-        default: true
+        default: false
       run_core_legacy:
         description: 'core tests on legacy platforms'
         type: boolean
-        default: true
+        default: false
       run_stress:
         description: 'stress tests'
         type: boolean
-        default: true
+        default: false
+      run_soak_with_chaos:
+        description: 'soak + chaos tests'
+        type: boolean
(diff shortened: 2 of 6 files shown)


sfc-gh-skurella:
* **#1493** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1493?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1493?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sfc-gh-bmikaili (approved):

PR #1494, opened 2026-06-21
SNOW-3675649 fix client-side validation of decimals

sfc-gh-skurella:
SNOW-3675649

We were not verifying precision and scale of NUMBER columns when using client-side validation. This accidentally wasn't ported from the SSv1 SDK because it was done within the Parquet serializer.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/validation/RowValidator.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/validation/RowValidator.java
@@ -10,6 +10,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.snowflake.kafka.connector.Utils;
 import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.time.ZoneId;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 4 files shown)


sfc-gh-skurella:
* **#1494** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1494?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1494?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

src/main/java/com/snowflake/kafka/connector/internal/validation/RowValidator.java
```diff
@@ -162,10 +163,21 @@ private Object validateAndNormalizeColumnValue(ColumnSchema col, Object value)
...
-        // if needed, but SSv1 didn't enforce it at validation time either.
-        DataValidationUtil.validateAndParseBigDecimal(col.getName(), value, insertRowIndex);
+        // DataValidationUtil.validateAndParseBigDecimal doesn't check precision/scale
+        // It just parses the value.
+        BigDecimal parsedNumber =
+            DataValidationUtil.validateAndParseBigDecimal(col.getName(), value, insertRowIndex);
+        // SSv1 enforced this during Parquet/SB16 serialization.
+        // We must range-check the value here.
+        int columnScale = col.getScale() != null ? col.getScale() : 0;
+        int columnPrecision = col.getPrecision() != null ? col.getPrecision() : 38;
```
sfc-gh-bmikaili, line 173: nit: these defaults assume `NUMBER` without explicit precision/scale maps to `NUMBER(38,0)` (Snowflake's documented default). A brief comment here would help future readers understand why 38 and 0 are safe — e.g.
```java
// ColumnSchema may return null for plain NUMBER columns; Snowflake defaults to NUMBER(38,0).
```
If `ColumnSchema` always populates these for `FIXED` columns, the null guards are defensive but harmless — a comment to that effect would also be welcome.

sfc-gh-bmikaili (approved):
Clean, well-scoped fix with excellent test coverage. The rounding-before-range-check is the right approach and mirrors server behaviour correctly. One nit inline; no blocking issues.

PR #1495, opened 2026-06-22
(bad merge) SNOW-3647384 don't rewind unassigned partitions

sfc-gh-skurella:
> [!WARNING]
> The merge commit for this PR was removed because unrelated changes slipped into this PR.
> The new PR is https://github.com/snowflakedb/snowflake-kafka-connector/pull/1498.

SNOW-3647384

## Summary

During a consumer rebalance a partition can be revoked while its `openChannel` is still in-flight. When that open completes — after `PartitionChannelManager.close()` has already removed the partition's pending offset reset — `initializeFromSnowflake` fires `onOffsetReset` and re-enqueues a reset for the now-unowned partition. That stale reset is later drained on the task thread and applied via `sinkTaskContext.offset()`, so `WorkerSinkTask.rewind()` seeks a partition the consumer no longer owns and throws `IllegalStateException: No current assignment for partition`. Kafka Connect treats this as fatal and kills the task.

The fix stops unassigned partitions from ever being rewound, in three places:

- The in-flight open skips the offset reset when its channel has been cancelled (partition revoked), so it no longer re-enqueues a stale reset.
- `PartitionChannelManager.drainPendingOffsetResets` now takes the task's current assignment and drops resets for partitions no longer assigned, instead of returning them.
- `insert()` forwards the current assignment to the drain and guards the seek with an invariant check that it never rewinds an unassigned partition. The check can be disabled via `snowflake.feature.assert.partition.assignment` (default true) if it ever misfires.

## Testing

Apart from the unit tests, I confirmed with an in-progress soak test with chaos (which now includes changing the task count) that the fix works (see [successful run](https://github.com/snowflakedb/snowflake-kafka-connector/actions/runs/27969081410/job/82770400739)) while the same test failed on `master` locally in an 8-minute run.


sfc-gh-skurella:
* **#1495** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1495?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1495?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
```diff
@@ -150,7 +150,20 @@ public SnowpipeStreamingPartitionChannel(
...
               OpenChannelResult openChannelResult = openChannelWithClientRecovery(channelName);
               long offsetRecoveredFromSnowflake = parseOrMigrateOffsetToken(openChannelResult);
-              offsetTracker.initializeFromSnowflake(offsetRecoveredFromSnowflake);
+              // If this channel was cancelled (e.g. the partition was revoked during a rebalance)
+              // while the open was in-flight, skip the offset reset. Otherwise
+              // initializeFromSnowflake fires onOffsetReset, which re-enqueues a pending offset
+              // reset for a partition the task no longer owns; that reset is later applied via
+              // sinkTaskContext.offset() and seeks a revoked partition (SNOW-3647384). The channel
+              // is still returned so closeChannelAsync can close it.
+              if (cancelled.get()) {
```
sfc-gh-bmikaili, line 159: The constructor open path now guards `initializeFromSnowflake` with `cancelled.get()`, but the symmetric recovery path is not guarded: `reopenChannel` calls `offsetTracker.resetAfterRecovery(offsetRecoveredFromSnowflake)` at line 429 with no `cancelled` check. `resetAfterRecovery` fires the same `onOffsetReset` callback that re-enqueues a pending offset reset, so a rebalance revoking the partition mid-recovery can re-add a stale reset for an unowned partition — the exact bug this PR fixes, just on the recovery path.

It's masked today by the `drainPendingOffsetResets(assignment)` filter, but the asymmetry is surprising and leans on that single backstop. Suggest adding the same `if (cancelled.get())` guard before the `resetAfterRecovery` call in `reopenChannel` so both open paths protect the offset-reset enqueue symmetrically.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -306,11 +306,29 @@ private Set<TopicPartition> currentlyInitializing(Collection<TopicPartition> par
...
+
     // Skip partitions for which the partition-channel bridge is currently being initialized.
     Set<TopicPartition> partitions =
         records.stream()
             .map(record -> new TopicPartition(record.topic(), record.kafkaPartition()))
             .collect(Collectors.toSet());
+    // Kafka Connect only delivers records for partitions currently assigned to this task. If this
+    // did not hold we would ingest data for a partition we don't own without ever rewinding it.
+    if (taskConfig.isAssertPartitionAssignmentEnabled() && !assignment.containsAll(partitions)) {
+      throw new AssertionError(
```
sfc-gh-bmikaili: `AssertionError` extends `Error`, not `Exception`, so it bypasses Kafka Connect's `WorkerSinkTask` error handling (which catches `RuntimeException` / `ConnectException`) and propagates as an ungraceful JVM-level failure with no clean connector error context — and `-da`/assertion semantics differ from a real guard. Prefer `ConnectException` (already used in this file for fail-fast invariant violations) or `IllegalStateException`; both are `RuntimeException`s that Connect surfaces and handles correctly. Keep the feature-flag gate and the message as-is. Same applies to the rewind guard at line 415.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -386,6 +406,21 @@ public void insert(final Collection<SinkRecord> records) {
     }
 
     if (!offsetsToRewindTo.isEmpty()) {
+      // Defense-in-depth against SNOW-3647384: drainPendingOffsetResets already dropped revoked
+      // partitions and the skipped-record offsets come from `partitions` (asserted assigned above),
+      // so every rewind must target a currently-assigned partition. Guards any future path that
+      // could add an unassigned partition to offsetsToRewindTo before the seek.
+      if (taskConfig.isAssertPartitionAssignmentEnabled()
+          && !assignment.containsAll(offsetsToRewindTo.keySet())) {
+        throw new AssertionError(
```
sfc-gh-bmikaili: Same as the guard above: `AssertionError` is an `Error` and bypasses Connect's exception handling. Use `ConnectException` / `IllegalStateException` here too so a misfire becomes a graceful task failure rather than an ungraceful kill.

sfc-gh-bmikaili (commented):
Reviewed the rebalance-rewind fix. The three-layer approach is correct end-to-end and unusually well documented. Two should-fix items inline before merge; everything else (flag registration in ConnectorConfigDefinition, default-value choice, test/Javadoc coverage claims) is minor and non-blocking.

sfc-gh-bmikaili (approved):

PR #1481, opened 2026-05-29
bump snowpipe-streaming to 1.6.0

sfc-gh-skurella:
## Summary

Bumps `com.snowflake:snowpipe-streaming` from 1.4.0 → 1.6.0 in `pom.xml` and `pom_confluent.xml`. The mitmproxy fault-injection E2E job stays on 1.4.0 because 1.5.0+ enforce `https_only(true)` in non-test builds, which rejects the test's `http://mitmproxy:8080` destination at the request-builder layer (`reqwest::Error { kind: Builder, source: BadScheme }`).

## How the override works

- POMs declare `<snowpipe-streaming.version>1.6.0</snowpipe-streaming.version>` and reference it via `${...}`. All builds default to 1.6.0.
- `test/build_runtime_jar.sh` reads `SNOWPIPE_STREAMING_VERSION_OVERRIDE`. When set, it appends `-Dsnowpipe-streaming.version=...` and `-Dmaven.test.skip=true` to every Maven invocation. Test compilation is skipped because the test fakes track the latest SDK API surface and don't compile against 1.4.0.
- `.github/actions/build-connector/action.yml` exposes an optional `snowpipe-streaming-version` input that flows through to the env var.
- Only the `fault_injection` job in `.github/workflows/end-to-end.yaml` sets `snowpipe-streaming-version: '1.4.0'`. Every other CI job (apache + confluent verify/package, default e2e suite, integration tests) runs against 1.6.0.

## Test fakes

`FakeSnowflakeStreamingIngestClient` and `TrackingStreamingIngestClient` (in `SnowpipeStreamingPartitionChannelTest`) gain a stub `getElasticChannel()` because 1.6.0 re-declares it on the `SnowflakeStreamingIngestClient` public interface. Stubs throw `UnsupportedOperationException`; nothing in the connector exercises elastic channels.

## Tracking & follow-up

- **SNOW-3683451** — drop the 1.4.0 pin once the SDK provides an escape hatch.

Issue #1443, opened 2026-04-29
v4 connector cannot recover from InvalidClientException


charlesdong1991:
v4 connector cannot recover from InvalidClient self-invalidation triggered by transient 404 on get_channel_status_batch

A single transient HTTP 404 from get_channel_status_batch causes the underlying cyclone_shared Rust SDK to mark the entire SnowflakeStreamingIngestClient as invalid (cyclone_shared::prod::ingest_client (827) Invalidating client …). Once that happens:

- BatchOffsetFetcher (internal/streaming/v2/service/BatchOffsetFetcher.java) silently logs the failure and continues — never escalates
- SnowpipeStreamingPartitionChannel.reopenChannel() only reopens the channel on the same final SnowflakeStreamingIngestClient streamingClient reference
- StreamingClientPools (internal/streaming/v2/client/StreamingClientPools.java) caches the client JVM-statically; even task restarts via Strimzi autoRestart reuse the dead client

Result: the connector loops forever on InvalidClientError with no path to recover except pod restart or KafkaConnector CR delete/recreate. Manual intervention is required for what was a transient Snowflake-side blip.

PR #1496, opened 2026-06-23
SNOW-3670537: recover from InvalidClientError on offset-fetch and flush paths

sfc-gh-bmikaili:
> **Stacked on #1495 (SNOW-3647384) — merge #1495 first.** This PR's preCommit-triggered reopen enqueues an offset reset that is later applied as a `sinkTaskContext.offset()` seek. #1495's revoked-partition guards (skip reset when the channel was cancelled mid-open + drop drained resets for unassigned partitions) prevent a rebalance in that window from seeking a partition the task no longer owns and crashing it (`IllegalStateException: No current assignment`). Without #1495 underneath, the kill-switch (default on) would expose that race.

## Summary

Kafka Connector v4 (Snowpipe Streaming V2) does not auto-recover from a transient **HTTP 404 ("No Route")** at the Snowflake envoy layer. A brief routing blip (<30s observed) makes the Rust SDK mark the `SnowflakeStreamingIngestClient` **permanently invalid** (`InvalidClientError`); affected pipes then stay broken until a manual restart while the task still reports `RUNNING` — silent, unmonitorable degradation.

- **Jira:** SNOW-3670537
- Also reported in #1443
- Recovery mechanism by **Seb Kurella** (`skurella-precommit-client-recovery`); this PR adds parameter protection + a fault-injection e2e, stacked on #1495.

## Root cause

The existing client-recreation machinery (`openChannelWithClientRecovery`) only fires from the **`appendRow`** failure path. After a 404, `InvalidClientError` surfaces on `getChannelStatus` — the `:bulk-channel-status` request made every `preCommit`, independently of new data. That path swallowed the error and deferred "to the next `appendRow`". For a pipe with appended-but-uncommitted records and no new traffic, that append never comes, so the client is never recreated and the task is stuck — it neither ingests the last records nor fails.

## Fix

Drive recovery from `preCommit` through the **same** `reopenChannel` / `openChannelWithClientRecovery` primitive the `appendRow` path already uses — one recovery mechanism, not a new one:

1. **`TopicPartitionChannel#triggerReopenForInvalidClient()`** — new interface method, **default no-op** (SSv1 and other channels unaffected).
2. **`SnowpipeStreamingPartitionChannel`** implements it: `reopenChannel("PRECOMMIT_CLIENT_INVALID")`, unless a reopen is already in flight (`isInitializing()` guard) so repeated `preCommit` cycles don't pile up reopens.
3. **`BatchOffsetFetcher`** — on `InvalidClientError`, signals each channel of the bad pipe to reopen, instead of recreating the pooled client itself.

The `appendRow` recovery path is unchanged. The flush (`initiateFlush`) path is intentionally **not** touched: it runs only from `SnowflakeSinkServiceV2.stop()` (teardown, immediately before `closeTaskClients`), so an invalid client there has nothing to recover.

### Why this is safe (no data loss)
Kafka offsets are committed only up to Snowflake's durably-committed offset token (`processChannelStatus` → `committedOffset + 1`). A reopened pipe never advances Kafka offsets past un-committed data, so recovery replays from the last commit — at-least-once preserved (the e2e asserts **exactly**, so duplicates would fail it too).

## Parameter protection

Unregistered kill-switch (default on): `snowflake.feature.precommit.client.recovery` — `defaultValue(true)`. When `false`, the offset-fetch path falls back to the legacy log-and-skip (defer recovery to the next `appendRow`). Not added to `ConfigDef` (internal flag).

## Testing

| Layer | Result |
|---|---|
| Unit — `BatchOffsetFetcherTest`, `SnowpipeStreamingPartitionChannelTest`, `SinkTaskConfigTest`, `PartitionChannelManagerTest` | ✅ 90/90 (on the stacked base) |
| Integration — `StreamingClientManagerIT` (live Snowflake) | ✅ 8/8 |
| E2E — `test_client_recreation_404` (mitmproxy 404 on `:bulk-channel-status`) | ✅ recovers to **exactly 600 rows** (no loss, no duplicates) |

Coverage: trigger reopens every channel of the bad pipe (healthy pipes untouched; in-flight reopen not duplicated); OFF-path test asserts the trigger never fires when the flag is disabled; e2e injects HTTP 404 on `:bulk-channel-status` (the customer's exact endpoint) and asserts exact recovery — tagged `@fault_injection`, run against **SDK 1.4.0** (1.5.0+ reject the proxy's plain-http endpoint, per #1481).

## Notes for reviewers
- Recovery mechanism is Seb's (`triggerReopenForInvalidClient`); this PR layers the kill-switch + fault-injection e2e and is stacked on #1495.
- All assertions are **exact** (no "at least N").


sfc-gh-bmikaili:
* **#1496** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1496?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1496?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `skurella-SNOW-3647384-fix-open-after-rebalance`


<h2></h2>

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/service/BatchOffsetFetcher.java
```diff
@@ -146,6 +138,31 @@ private Map<TopicPartition, Long> getCommittedOffsetsForPipe(
...
+    } catch (SFException e) {
+      if (!ClientRecreationException.isClientInvalidError(e)) {
+        throw e;
+      }
+      LOGGER.warn(
+          "Client is invalid for pipe: {} ({}); recreating. Offset fetch skipped this cycle.",
+          pipeName,
+          e.getErrorCodeName());
+      try {
+        StreamingClientPools.recreateClient(
```
sfc-gh-bmikaili: Core fix. The offset-fetch path (preCommit) now recreates the SDK client on `InvalidClientError` instead of swallowing it — so an idle pipe, which never triggers the appendRow-driven recovery, actually recovers. On recreation exhaustion we log+skip (not throw): a throw here is eaten by `SnowflakeSinkTask.preCommit`'s catch and would also discard healthy pipes' offsets for the cycle. preCommit just retries next cycle; active pipes still hard-fail via the appendRow path.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/service/BatchOffsetFetcher.java
```diff
@@ -109,20 +110,11 @@ private CompletableFuture<Void> fetchOffsetsAsync(
...
-                  e);
-            } else {
-              LOGGER.error(
-                  "Failed to fetch committed offsets for pipe: {}, skipping {} channel(s)",
-                  pipeName,
-                  channelsByPartition.size(),
-                  e);
-            }
+            LOGGER.error(
+                "Failed to fetch committed offsets for pipe: {}, skipping {} channel(s)",
```
sfc-gh-bmikaili: Dead branch removed: `InvalidClientError` is now handled inside `getCommittedOffsetsForPipe`, so this outer catch only logs genuinely-unexpected `SFException`s.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
```diff
@@ -240,15 +240,22 @@ public CompletableFuture<Void> waitForLastProcessedRecordCommitted() {
...
+                    + " reopening channel.",
+                this.channelName,
+                e.getErrorCodeName());
+            // Recreate the client AND reopen the channel now -- reopenChannel funnels through
+            // openChannelWithClientRecovery, which swaps in a fresh client and opens a new
+            // channel on it (and resets the offset tracker). Do NOT defer recovery to the next
+            // appendRow: an idle partition may never append again, which is the exact failure
+            // mode behind SNOW-3670537. This flush cycle is skipped; the reopened channel
+            // serves all subsequent work.
+            reopenChannel("CLIENT_RECREATION_FLUSH");
```
sfc-gh-bmikaili: Flush path fix. On `InvalidClientError` from `initiateFlush`, reopen the channel inline via `reopenChannel` (recreate client + open a new channel through `openChannelWithClientRecovery`). It must NOT defer to the next appendRow — an idle partition may never append again, which is the exact failure mode behind this bug. On exhaustion the channel future fails with `ConnectException`, surfaced on the next `getChannel()` to fail the task.

src/test/java/com/snowflake/kafka/connector/internal/streaming/BatchOffsetFetcherTest.java
```diff
@@ -222,9 +272,11 @@ private void registerChannelWithNoOffset(
    */
   static class CountingClientSupplier implements StreamingClientSupplier {
     private final AtomicInteger batchCallCount = new AtomicInteger(0);
+    private final AtomicInteger clientCreationCount = new AtomicInteger(0);
```
sfc-gh-bmikaili: Test-only counter so the recovery tests can assert exactly how many SDK clients were created (initial + one per recreation).

src/test/java/com/snowflake/kafka/connector/internal/streaming/BatchOffsetFetcherTest.java
```diff
@@ -247,11 +316,19 @@ public SnowflakeStreamingIngestClient get(
...
         StreamingClientProperties streamingClientProperties) {
+      clientCreationCount.incrementAndGet();
       SnowflakeStreamingIngestClient client = mock(SnowflakeStreamingIngestClient.class);
       when(client.getChannelStatus(any()))
           .thenAnswer(
               invocation -> {
                 batchCallCount.incrementAndGet();
+                if (pipeName.equals(clientInvalidPipe)) {
+                  throw new SFException(
+                      "InvalidClientError",
```
sfc-gh-bmikaili: Mock throws `InvalidClientError` (409) on `getChannelStatus` for a designated pipe, reproducing the SDK's sticky invalid state after a transient 404.

src/test/java/com/snowflake/kafka/connector/internal/streaming/BatchOffsetFetcherTest.java
```diff
@@ -135,6 +136,55 @@ void sfExceptionForOnePipeDoesNotAffectOthers() {
     assertEquals(21L, result.get(tp1));
   }
 
+  // Fix target for SNOW-3670537: BatchOffsetFetcher must recreate the client on InvalidClientError
+  // so that the next preCommit cycle recovers rather than silently dropping the pipe forever.
+  @Test
+  void clientInvalidOnBulkChannelStatus_swallowedNoRecoveryNoFailure_snow3670537() {
```
sfc-gh-bmikaili: Happy-path recovery: invalid client on cycle 1 triggers recreation; cycle 2 with the fresh client returns the offset. Asserts exactly 2 client creations — no extra churn.

src/test/java/com/snowflake/kafka/connector/internal/streaming/BatchOffsetFetcherTest.java
```diff
@@ -135,6 +136,55 @@ void sfExceptionForOnePipeDoesNotAffectOthers() {
...
+    // Two client creations confirm a new client was built during recreation.
+    assertEquals(2, clientSupplier.getClientCreationCount(), "recreation created a new client");
+  }
+
+  // SNOW-3670537: when the client stays invalid (e.g. a persistent server-side outage), the
+  // offset-fetch path must NOT fail the task or discard healthy pipes' offsets. It logs, skips the
+  // pipe, and recreates again on the next preCommit — self-healing once the outage clears. Terminal
+  // failure for actively-producing partitions is the appendRow path's responsibility, not preCommit.
+  @Test
+  void clientInvalidPersists_keepsRetryingWithoutFailing_snow3670537() {
```
sfc-gh-bmikaili: Persistent-outage guard: while the client stays invalid the offset-fetch path must keep retrying (exactly one recreation per cycle) and never throw/fail the task or drop healthy pipes.

src/test/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannelTest.java
```diff
@@ -1162,7 +1230,13 @@ public CompletableFuture<Void> close(
     }
 
     @Override
-    public void initiateFlush() {}
+    public void initiateFlush() {
+      // Mirrors the real SDK: initiate_flush() calls check_client_valid() and throws
```
sfc-gh-bmikaili: Fake now mirrors the real SDK — `initiateFlush` throws `InvalidClientError` once the client is invalid — so the flush-recovery path is actually exercised (it was previously a no-op stub).

src/test/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannelTest.java
```diff
@@ -433,6 +433,74 @@ void insertRecord_channelOnlyError_doesNotRecreateClient() {
...
+  /**
+   * SNOW-3670537 (flush path). When a transient 404 invalidates the client and the invalid state
+   * surfaces through {@code initiateFlush} during {@link
+   * SnowpipeStreamingPartitionChannel#waitForLastProcessedRecordCommitted} — rather than through
+   * {@code appendRow} — the channel must invoke the {@link ClientRecreator} to recreate the client
+   * instead of swallowing the error. Without this, a pipe that is not actively appending would
+   * never recover until a manual restart (the customer's reported symptom).
+   */
+  @Test
+  void flushPathClientInvalid_recreatesClientAndReopensChannel_snow3670537() {
```
sfc-gh-bmikaili: Asserts the flush path both recreates the client AND reopens the channel — exactly once each — without relying on a subsequent appendRow.

src/test/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannelTest.java
```diff
@@ -433,6 +433,74 @@ void insertRecord_channelOnlyError_doesNotRecreateClient() {
...
+
+    assertEquals(1, recreateCallCount.get(), "recreate must be called exactly once");
+    assertEquals(
+        1,
+        newClientSupplier.getTotalChannelsCreated(),
+        "channel must be reopened exactly once on the recreated client");
+  }
+
+  @Test
+  void flushPathClientInvalid_recreateExhausted_failsTask_snow3670537() {
```
sfc-gh-bmikaili: Asserts recreation-budget exhaustion in the flush path fails the task: the `ConnectException` surfaces on the next `getChannel()` so Kafka Connect restarts it rather than looping.

test/docker/mitmproxy/addon_410.py
```diff
@@ -88,6 +88,9 @@ def reset_counters(self):
 
 fault_state = FaultState()
 
+# Separate state for 404-on-bulk-channel-status injection (SNOW-3670537)
+fault_state_404_bcs = FaultState()
```
sfc-gh-bmikaili, line 92: Separate fault toggle for 404-on-`:bulk-channel-status`, alongside the existing 410 mode, so the new e2e reproduces the customer's exact envoy error without disturbing the 410 test.

test/docker/mitmproxy/addon_410.py
```diff
@@ -134,6 +137,16 @@ def request(self, flow: http.HTTPFlow) -> None:
             fault_state.inc_injected()
             log(f"Injected 410 on {flow.request.method} {path}")
 
+        # Inject 404 on :bulk-channel-status when that fault mode is active.
```
sfc-gh-bmikaili, line 140: Injects 404 only on `:bulk-channel-status` (the endpoint the customer's SDK kept hitting), leaving other streaming calls untouched.

test/lib/mitmproxy_control.py
```diff
@@ -51,6 +51,22 @@ def reset_counters(self):
         resp.raise_for_status()
         logger.info("Counters reset")
 
+    def enable_404_bulk_channel_status(self):
```
sfc-gh-bmikaili, line 54: Python control-API helpers to enable/disable the 404 mode and read its injected count from the addon.

test/tests/test_client_recreation_404.py
```diff
@@ -0,0 +1,153 @@
+"""
```
sfc-gh-bmikaili, line 1: New e2e: inject 404 on `:bulk-channel-status`, then assert the connector recovers to EXACTLY 600 rows (no loss, no duplicates). Tagged `@fault_injection` so CI runs it against SDK 1.4.0 — 1.5.0+ reject the proxy's plain-http endpoint (per #1481).

sfc-gh-bmikaili (commented):
Self-review: inline rationale for each change.

src/main/java/com/snowflake/kafka/connector/Constants.java
```diff
@@ -52,6 +52,15 @@ public static final class KafkaConnectorConfigParams {
...
         false;
 
+    // Feature flags
+    // Kill-switch (unregistered, default on) for recreating the SDK client when
+    // InvalidClientError surfaces on the offset-fetch (bulk-channel-status) and flush paths,
+    // rather than swallowing it and waiting for the next appendRow (SNOW-3670537). Set to
+    // false to fall back to the legacy behavior if recovery ever misbehaves.
+    public static final String SNOWFLAKE_FEATURE_RECOVER_CLIENT_ON_STATUS_FLUSH =
+        "snowflake.feature.recover.client.on.status.flush";
+    public static final boolean SNOWFLAKE_FEATURE_RECOVER_CLIENT_ON_STATUS_FLUSH_DEFAULT = true;
```
sfc-gh-bmikaili: Unregistered kill-switch, default `true`. Same pattern as #1495: if the new recovery logic ever misbehaves in prod, a customer can set `snowflake.feature.recover.client.on.status.flush=false` to fall back to the legacy swallow-and-defer-to-appendRow behavior — no rollback/redeploy needed. Not added to ConfigDef on purpose (internal flag).

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/service/BatchOffsetFetcher.java
```diff
@@ -146,6 +138,42 @@ private Map<TopicPartition, Long> getCommittedOffsetsForPipe(
     final ChannelStatusBatch batch;
     try (TaskMetrics.TimingContext ignored = taskMetrics.timeOffsetFetch()) {
       batch = client.getChannelStatus(channelNames);
+    } catch (SFException e) {
+      if (!ClientRecreationException.isClientInvalidError(e)) {
+        throw e;
+      }
+      if (!taskConfig.isRecoverClientOnStatusFlushEnabled()) {
```
sfc-gh-bmikaili: Gate for the offset-fetch fix. When disabled, rethrow the `InvalidClientError` so the outer catch logs+skips the pipe — i.e. exactly the pre-fix behavior. When enabled (default), fall through to recreation below.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
```diff
@@ -240,15 +240,32 @@ public CompletableFuture<Void> waitForLastProcessedRecordCommitted() {
           try {
             streamingClient.initiateFlush();
           } catch (SFException e) {
-            if (ClientRecreationException.isClientInvalidError(e)) {
+            if (!ClientRecreationException.isClientInvalidError(e)) {
+              throw e;
+            }
+            if (!taskConfig.isRecoverClientOnStatusFlushEnabled()) {
```
sfc-gh-bmikaili: Gate for the flush fix. When disabled, log + return (legacy swallow, defer to appendRow). When enabled (default), fall through to `reopenChannel`.

src/test/java/com/snowflake/kafka/connector/config/SinkTaskConfigTestBuilder.java
```diff
@@ -53,6 +53,8 @@ public static SinkTaskConfig.Builder builder() {
         .snowflakeDatabase("")
         .snowflakeSchema("")
         .ssv1MigrationMode(Ssv1MigrationMode.SKIP)
-        .ssv1MigrationIncludeConnectorName(false);
+        .ssv1MigrationIncludeConnectorName(false)
+        .recoverClientOnStatusFlushEnabled(
```
sfc-gh-bmikaili: Default the flag to the `_DEFAULT` constant (true) so every existing test exercises the ON path — no per-test wiring needed. OFF-path tests override this explicitly.

src/test/java/com/snowflake/kafka/connector/internal/streaming/BatchOffsetFetcherTest.java
```diff
@@ -135,6 +136,82 @@ void sfExceptionForOnePipeDoesNotAffectOthers() {
...
+    assertEquals(
+        4,
+        clientSupplier.getClientCreationCount(),
+        "expected exactly 4 client creations (1 initial + 3 recreations)");
+  }
+
+  // SNOW-3670537 kill-switch: with snowflake.feature.recover.client.on.status.flush=false the
+  // offset-fetch path falls back to legacy behavior -- swallow and skip, no client recreation.
+  @Test
+  void clientInvalid_featureDisabled_legacySwallowNoRecreation_snow3670537() {
```
sfc-gh-bmikaili: OFF-path coverage: with the flag false, an invalid client on the offset-fetch path is swallowed/skipped with no recreation (exactly 1 client = the initial one). Proves the kill-switch actually disables recovery.

src/test/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannelTest.java
```diff
@@ -445,6 +514,32 @@ private SinkRecord buildValidRecord(long offset) {
         .build();
   }
 
+  // SNOW-3670537 kill-switch: with snowflake.feature.recover.client.on.status.flush=false the
+  // flush path falls back to the legacy behavior -- swallow the InvalidClientError and defer to
+  // appendRow. No client recreation, no channel reopen, no task failure.
+  @Test
+  void flushPathClientInvalid_featureDisabled_legacySwallow_snow3670537() {
```
sfc-gh-bmikaili: OFF-path coverage: with the flag false, the flush path neither recreates the client nor fails the task (legacy swallow). Exact assertions (recreate count == 0, getChannel does not throw).

sfc-gh-bmikaili (commented):
Self-review: rationale for the parameter-protection commit (mirrors #1495's unregistered-flag pattern).

src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/TopicPartitionChannel.java
```diff
@@ -43,6 +43,22 @@ default void awaitInitialization() {}
...
+   * surfaces on the next {@code appendRow} and drives recovery from there. But if the client is
+   * invalidated while there are appended-but-uncommitted records and no new records arrive, {@code
+   * appendRow} is never called again, so recovery never starts: {@code preCommit} keeps skipping
+   * the failed offset fetch and the task is stuck (it neither ingests the last records nor fails).
+   * The {@code preCommit} offset-fetch path calls this when it detects the client is invalid so
+   * recovery starts even without new data.
+   *
+   * <p>No-op by default.
+   */
+  default void triggerReopenForInvalidClient() {}
```
sfc-gh-bmikaili, line 60: New extension point (default no-op): lets a caller drive recovery outside the appendRow path. SSv1 and other channels inherit the no-op, so no behavior change for them and no flag needed at the interface level.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
```diff
@@ -863,6 +863,25 @@ public boolean isChannelClosed() {
     }
   }
 
+  @Override
+  public void triggerReopenForInvalidClient() {
```
sfc-gh-bmikaili, line 891: Reuses the existing `reopenChannel`/`openChannelWithClientRecovery` primitive (same one appendRow uses) rather than a bespoke recreate. The `isInitializing()` guard means repeated preCommit cycles won't stack reopens on top of an in-flight one. Failsafe budget + terminal-failure-via-channel-future come for free.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/service/BatchOffsetFetcher.java
```diff
@@ -110,12 +110,31 @@ private CompletableFuture<Void> fetchOffsetsAsync(
             result.putAll(getCommittedOffsetsForPipe(pipeName, channelsByPartition));
           } catch (SFException e) {
             if (ClientRecreationException.isClientInvalidError(e)) {
-              LOGGER.warn(
-                  "Client is invalid for pipe: {}, skipping offset fetch for {} channel(s)."
-                      + " Client recreation will be triggered by the next appendRow on this pipe.",
-                  pipeName,
-                  channelsByPartition.size(),
-                  e);
+              if (taskConfig.isPrecommitClientRecoveryEnabled()) {
```
sfc-gh-bmikaili, line 113: Core fix, gated by the kill-switch. On InvalidClientError the fetcher *signals* each channel of the bad pipe to reopen (line 127) instead of recreating the pooled client itself — the channel owns its recovery. When the flag is off we fall back to the legacy log-and-skip (defer to appendRow).

src/main/java/com/snowflake/kafka/connector/Constants.java
```diff
@@ -52,6 +52,15 @@ public static final class KafkaConnectorConfigParams {
...
         false;
 
+    // Feature flags
+    // Kill-switch (unregistered, default on) for preCommit-driven client recovery: when the
+    // offset-fetch (bulk-channel-status) path detects an invalid SDK client it triggers a
+    // channel reopen so recovery starts even with no appendRow traffic (SNOW-3670537). Set to
+    // false to fall back to the legacy behavior (defer recovery to the next appendRow).
+    public static final String SNOWFLAKE_FEATURE_PRECOMMIT_CLIENT_RECOVERY =
+        "snowflake.feature.precommit.client.recovery";
+    public static final boolean SNOWFLAKE_FEATURE_PRECOMMIT_CLIENT_RECOVERY_DEFAULT = true;
```
sfc-gh-bmikaili, line 66: Unregistered kill-switch, default `true` (mirrors #1495). Set `snowflake.feature.precommit.client.recovery=false` to disable preCommit-driven recovery per-connector if it ever misbehaves — no rollback. Renamed from the earlier flush-oriented name since the flush path is no longer separately handled.

sfc-gh-bmikaili (commented):
Self-review for the reopen-trigger rewrite (recovery mechanism by @sfc-gh-skurella). Supersedes the earlier two-path approach.

src/test/java/com/snowflake/kafka/connector/internal/streaming/BatchOffsetFetcherTest.java
```diff
@@ -133,6 +135,64 @@ void sfExceptionForOnePipeDoesNotAffectOthers() {
...
+    // healthy pipe are left alone.
+    verify(channels.get(tp0)).triggerReopenForInvalidClient();
+    verify(channels.get(tp1)).triggerReopenForInvalidClient();
+    verify(channels.get(tp2), never()).triggerReopenForInvalidClient();
+  }
+
+  // SNOW-3670537 kill-switch: with snowflake.feature.precommit.client.recovery=false an invalid
+  // client must NOT trigger a preCommit reopen -- legacy behavior (recovery deferred to appendRow).
+  @Test
+  void clientInvalid_recoveryDisabled_doesNotTriggerReopen_snow3670537() {
```
sfc-gh-bmikaili, line 173: OFF-path coverage: with the flag false, an invalid client must NOT trigger a reopen (verify `never()`), while the pipe is still skipped and healthy pipes are unaffected. Proves the kill-switch actually disables the new behavior.

sfc-gh-bmikaili:
Restacked onto #1495 (SNOW-3647384). Rebased my single squashed commit on top of `skurella-SNOW-3647384-fix-open-after-rebalance` and repointed this PR's base to it. Reason: the preCommit-triggered reopen enqueues offset resets applied as seeks, and #1495's revoked-partition guards are required to keep that safe across rebalances. Earlier inline review comments are on a superseded commit; the diff is unchanged in intent (Seb's reopen-trigger mechanism + kill-switch + 404 e2e).

src/test/java/com/snowflake/kafka/connector/config/SinkTaskConfigTestBuilder.java
```diff
@@ -54,6 +54,8 @@ public static SinkTaskConfig.Builder builder() {
         .snowflakeSchema("")
         .ssv1MigrationMode(Ssv1MigrationMode.SKIP)
         .ssv1MigrationIncludeConnectorName(false)
-        .assertPartitionAssignmentEnabled(true);
+        .assertPartitionAssignmentEnabled(true)
+        .precommitClientRecoveryEnabled(
+            KafkaConnectorConfigParams.SNOWFLAKE_FEATURE_PRECOMMIT_CLIENT_RECOVERY_DEFAULT);
```
sfc-gh-skurella: NIT: could just hard-code to `true` here for consistency

sfc-gh-skurella (commented):
Prod code LGTM, I haven't gone through the tests yet. I'll do that tomorrow, but don't expect any major work on this PR. Thank you!

test/tests/test_client_recreation_404.py
```diff
@@ -0,0 +1,156 @@
...
+    # --- Setup ---
+    table = create_table(
+        unsalted_name,
+        columns='(record_metadata variant, "partition" varchar, "seq" varchar, "key" varchar)',
+    )
+    driver.createTopics(topic, partitionNum=NUM_PARTITIONS, replicationNum=1)
+    driver.createConnector(
+        name_salt=name_salt,
+        rest_request_template_filename=CONFIG_FILE,
+    )
```
sfc-gh-skurella: NIT: use `create_connector` fixture rather than the legacy `driver` APIs

sfc-gh-skurella (approved):

src/test/java/com/snowflake/kafka/connector/config/SinkTaskConfigTestBuilder.java
sfc-gh-bmikaili: Done in a5984cb4 — hard-coded `precommitClientRecoveryEnabled(true)`.

test/tests/test_client_recreation_404.py
sfc-gh-bmikaili: Done in a5984cb4 — switched to the `create_connector` fixture + `RecordProducer`. Bonus: the fixture names the connector/topic/table after the test function, so the test is inherently independent of the 410 test — which let me drop the separate `client_recreation_404.json` template I had added earlier as a collision workaround.

PR #1497, opened 2026-06-23
SNOW-3655748: add offset-gap and recovery-skip telemetry for data-loss detection

sfc-gh-bmikaili:
PM action item from PROD-538073 (KC 4.0.0 silent offset-rewind data loss).
Adds two additive, per-channel signals (JMX + `kafka_channel_usage` telemetry, already version-stamped) so offset-gap abnormalities can be trended across connector releases:

- **`offset_gap_count` / `offset_gap_missing_record_count`**: forward discontinuities observed when the connector decides to process a record (`PartitionOffsetTracker.shouldProcess`, from `insertRecord`), and the total offset positions skipped across them — an upper bound on records actually lost, since legitimate filtering/compaction gaps count too. Noisy in absolute terms; the cross-version trend is the signal.
- **`recovery_skip_conflict_count`**: batches delivered entirely past a recovering channel's rewind offset — the exact PROD-538073 interleaving (`SnowflakeSinkServiceV2.detectRecoverySkipConflicts`). Low-noise; should sit at zero in healthy operation.

Observation-only; no change to offset/recovery control flow. Baselining and alerting on these signals are owned by sibling action items.

## Testing
- Unit/integration: new `PartitionOffsetTrackerTest`, `SnowflakeSinkServiceV2RecoverySkipConflictTest`; extended `SnowflakeSinkServiceV2RecoveryOffsetTest` (asserts the conflict count in the real PROD-538073 scenario) and a `dumpTo` test for the new fields. All affected unit tests green; google-java-format (v1.24.0) clean.
- Live e2e on a SUT (apache 3.7.0, via the Snowfort KC proxy holder, SDK pinned 1.4.0 per #1481): `test_string_json[v4]` and `test_channel_invalidation_recovery[v4]` both PASS.

JIRA: SNOW-3655748

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
@@ -351,6 +351,14 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
     if (!pendingResets.isEmpty()) {
       LOGGER.info("Draining {} pending offset resets: {}", pendingResets.size(), pendingResets);
       offsetsToRewindTo.putAll(pendingResets);
+
+      // Count channels recovering while this batch carries records past their resume offset -- the
+      // PROD-538073 data-loss interleaving. Benign post-fix; tracked for cross-version trending.
+      for (TopicPartition tp : detectRecoverySkipConflicts(pendingResets, records)) {
+        channelManager
+            .getChannel(tp)
+            .ifPresent(TopicPartitionChannel::incRecoverySkipConflictCount);
+      }
     }
 
     // If still in cooldown from a recent backpressure event, treat all partitions as
@@ -424,6 +432,48 @@ public class SnowflakeSinkServiceV2 implements SnowflakeSinkService {
       LOGGER.info("Rewinding offsets for skipped partitions: {}", offsetsToRewindTo);
       sinkTaskContext.offset(offsetsToRewindTo);
     }
+  }
+
+  /**
+   * Returns the set of channels for which:
+   *
+   * <ol>
+   *   <li>we enqueued an offset rewind (a pending reset in {@code pendingResets}), and
+   *   <li>the current batch contains only records with higher offsets than that rewind offset.
+   * </ol>
+   *
+   * <p>For such channels we must take care NOT to rewind to the beginning of the current batch, or
+   * we would skip the records between the rewind offset and the batch — exactly the PROD-538073
+   * data-loss interleaving (the pre-SNOW-3574225 code rewound to a skipped record's offset instead
+   * of the recovery offset). A batch that includes a record at or below the rewind offset (e.g.
+   * {@code [51, 52, 53]} after a reset to 51, which also covers ordinary channel init) is NOT a
+   * conflict — nothing is dropped.
+   *
+   * <p>The fix preserves the recovery offset, so any conflict counted here is benign today; a
+   * rising cross-version trend means the risky interleaving is more frequent or a regression has
+   * returned.
+   */
+  static Set<TopicPartition> detectRecoverySkipConflicts(
+      Map<TopicPartition, Long> pendingResets, Collection<SinkRecord> records) {
+    if (pendingResets.isEmpty()) {
+      return Collections.emptySet();
+    }
+    Set<TopicPartition> partitionsWithRecordsAfterRewindOffset = new HashSet<>();
+    Set<TopicPartition> partitionsWithRecordsBeforeOrOnRewindOffset = new HashSet<>();
+    for (SinkRecord record : records) {
+      TopicPartition tp = new TopicPartition(record.topic(), record.kafkaPartition());
+      Long rewindOffset = pendingResets.get(tp);
+      if (rewindOffset == null) {
+        continue;
+      }
+      if (record.kafkaOffset() <= rewindOffset) {
+        partitionsWithRecordsBeforeOrOnRewindOffset.add(tp);
+      } else {
+        partitionsWithRecordsAfterRewindOffset.add(tp);
+      }
+    }
+    partitionsWithRecordsAfterRewindOffset.removeAll(partitionsWithRecordsBeforeOrOnRewindOffset);
+    return partitionsWithRecordsAfterRewindOffset;
   }
 
   /**
--- a/src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/TopicPartitionChannel.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/TopicPartitionChannel.java
@@ -91,6 +91,13 @@ public interface TopicPartitionChannel {
     return CompletableFuture.completedFuture(null);
   }
 
+  /**
+   * Records that this channel finished recovering while the current batch contained only records
+   * past its resume offset for this channel (the PROD-538073 data-loss interleaving). Default is a
+   * no-op for channel implementations that do not track it.
+   */
+  default void incRecoverySkipConflictCount() {}
+
   @VisibleForTesting
   SnowflakeTelemetryChannelStatus getSnowflakeTelemetryChannelStatus();
 }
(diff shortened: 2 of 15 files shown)


sfc-gh-bmikaili:
* **#1497** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1497?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1497?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/channel/PartitionOffsetTracker.java
```diff
@@ -99,6 +110,20 @@ public boolean shouldProcess(long kafkaOffset) {
     long currentProcessedOffset = this.processedOffset.get();
     if (currentProcessedOffset == NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE
         || kafkaOffset >= currentProcessedOffset + 1) {
+      if (currentProcessedOffset != NO_OFFSET_TOKEN_REGISTERED_IN_SNOWFLAKE
+          && kafkaOffset > currentProcessedOffset + 1) {
```
sfc-gh-skurella: NIT: this doesn't need to be nested in the outer if

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
```diff
@@ -55,9 +58,20 @@ public class SnowflakeTelemetryChannelStatus extends SnowflakeTelemetryBasicInfo
   private final AtomicLong processedOffset;
   private final AtomicLong latestConsumerOffset;
 
+  // Offset-gap counters, owned by PartitionOffsetTracker and read here by reference for telemetry
+  // and JMX. Number of forward discontinuities observed in the put() offset stream, and the upper
+  // bound on offsets skipped across them. See SNOW-3655748 / PROD-538073.
```
sfc-gh-skurella: > observed in the put() offset stream

That's not accurate - that's what we observe in the `insertRecord`, i.e. after we decide whether to process a batch or not. Worth fixing the docstring as it might trip up a future analysis.

> and the upper bound on offsets skipped across them

That's not what I see in the code - it seems we're just summing the count of records that are presumed to have been skipped. WDYT about renaming the metrics to `offset-gap-count` and `offset-gap-missing-record-count`?

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -391,6 +399,42 @@ public void insert(final Collection<SinkRecord> records) {
...
+   * that resume offset — i.e. it contains no record at or below the resume offset. That is the
+   * exact condition under which the pre-SNOW-3574225 code rewound Kafka to a skipped record's
+   * offset instead of the recovery offset, silently dropping the records in between.
+   *
+   * <p>A normal contiguous batch that includes the resume record (e.g. {@code [51, 52, 53]} after a
+   * reset to 51, which also covers ordinary channel init) is NOT a conflict: offset 51 is present,
+   * nothing is dropped. The fix preserves the recovery offset, so any conflict counted here is
+   * benign today — but a rising cross-version trend means the risky interleaving is more frequent
+   * or a regression has returned.
+   */
```
sfc-gh-skurella, line 455: I wasn't able to figure out what this does from this comment and had to spend >5 min reading the implementation. Suggested phrasing that I would find clearer:

> Returns a set of channels for which:
> 1. we enqueued an offset rewind
> 2. the current batch contains only records with higher offsets for that channel.
> 
> We need to take care to not rewind to the beginning of such a batch to avoid skipping some records.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
```diff
@@ -391,6 +399,42 @@ public void insert(final Collection<SinkRecord> records) {
...
+  static Set<TopicPartition> detectRecoverySkipConflicts(
+      Map<TopicPartition, Long> pendingResets, Collection<SinkRecord> records) {
+    if (pendingResets.isEmpty()) {
+      return Collections.emptySet();
+    }
+    Set<TopicPartition> pastResume = new HashSet<>();
+    Set<TopicPartition> reachedResume = new HashSet<>();
```
sfc-gh-skurella: NIT: rename to `hasRecordsAfterRewindOffset` to `hasRecordsBeforeOrOnRewindOffset`

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/channel/PartitionOffsetTracker.java
```diff
@@ -99,6 +110,20 @@ public boolean shouldProcess(long kafkaOffset) {
...
+        offsetGapsDetected.incrementAndGet();
+        offsetGapMissingOffsets.addAndGet(missing);
+        LOGGER.warn(
+            "Channel {} - offset gap detected: expected {} but received {} ({} offset(s) missing)."
+                + " Persisted offset in Snowflake: {}",
+            channelName,
+            currentProcessedOffset + 1,
+            kafkaOffset,
+            missing,
+            offsetPersistedInSnowflake.get());
```
sfc-gh-skurella: This will absolutely spam the logs for users which have filtering - let's make this DEBUG or configurable

sfc-gh-skurella (changes requested):
Thanks! I think those are useful metrics, left a couple comments and will be happy to approve once they're resolved.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/channel/PartitionOffsetTracker.java
sfc-gh-bmikaili: Changed the gap log to DEBUG.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/channel/PartitionOffsetTracker.java
sfc-gh-bmikaili: Inverted to an early `return false` for the duplicate/already-committed case, so the gap check is no longer nested.

src/main/java/com/snowflake/kafka/connector/internal/streaming/telemetry/SnowflakeTelemetryChannelStatus.java
sfc-gh-bmikaili: Fixed the docstring: detection is in `shouldProcess` (from `insertRecord`), not raw `put()`. Also clarified the magnitude — it's the total offset positions skipped, an upper bound on records actually lost (legitimate filtering/compaction gaps inflate it). Renamed to `offset_gap_count` / `offset_gap_missing_record_count`.

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-bmikaili: Rewrote the docstring with your phrasing (enqueued rewind + batch entirely after the rewind offset, and why we must not rewind to the batch start).

src/main/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2.java
sfc-gh-bmikaili: Renamed to `partitionsWithRecordsAfterRewindOffset` / `partitionsWithRecordsBeforeOrOnRewindOffset`.

src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/TopicPartitionChannel.java
```diff
@@ -91,6 +91,13 @@ default CompletableFuture<Void> waitForLastProcessedRecordCommitted() {
     return CompletableFuture.completedFuture(null);
   }
 
+  /**
+   * Records that this channel was recovering while the current batch carried records past its
```
sfc-gh-skurella: NIT: "finished recovering while the current batch contained only records past its resume offset for this channel"

sfc-gh-skurella (approved):
Thanks!

src/main/java/com/snowflake/kafka/connector/internal/streaming/channel/TopicPartitionChannel.java
sfc-gh-bmikaili: Adopted your phrasing.

PR #1501, opened 2026-06-24
fail when a test record is not delivered to Kafka

sfc-gh-skurella:
## Summary

The e2e Kafka producer helpers (`sendBytesData` / `sendAvroSRData`) were fire-and-forget. `confluent_kafka`'s `produce()` reports delivery failures only through the `on_delivery` callback, and `flush()` returns the number of still-unsent messages instead of raising. A dropped record was therefore **silently lost**, leaving the connector with fewer rows than were sent — which then surfaced much later as a baffling assertion far from the cause (e.g. `test_dt_structured_array[v4-ht]` failing with "got 0 rows").

## Changes

- `sendBytesData` / `sendAvroSRData` now register a delivery-error callback and run a bounded flush check (`PRODUCER_FLUSH_TIMEOUT_SECS`) that raises `NonRetryableError` if any record is still undelivered or any delivery callback reported an error.
- Count produced records while iterating instead of calling `len(value)`, so callers that pass a lazy iterator (e.g. `itertools.islice`) keep working.
- Fix a latent bug in `sendAvroSRData`: the mid-loop flush flushed the plain producer instead of the avro producer.

### DIFF
--- a/test/lib/driver.py
+++ b/test/lib/driver.py
@@ -82,6 +82,13 @@ class KafkaDriver:
         self.MAX_FLUSH_BUFFER_SIZE = (
             5000  # flush buffer when 5000 data was in the queue
         )
+        # Max seconds to block waiting for produced records to be acknowledged
+        # by the broker. librdkafka retries delivery up to message.timeout.ms
+        # (default 5 min); we cap the flush wait well below that so a stuck or
+        # lost record fails the test loudly here instead of silently leaving 0
+        # rows for the connector to consume (which surfaces as a confusing
+        # "got 0 rows" assertion much further downstream).
+        self.PRODUCER_FLUSH_TIMEOUT_SECS = 120
 
         self.kafkaConnectAddress = kafkaConnectAddress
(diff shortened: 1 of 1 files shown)


sfc-gh-skurella:
* **#1501** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1501?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1501?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sfc-gh-bmikaili (approved):
Approve. Well-scoped fix correctly diagnosing a silent record-loss bug in the e2e producer helpers. The callback-based delivery check is meaningful since acks defaults to all, the lazy-iterator counting is a real improvement, and the avro mid-loop wrong-producer flush fix is a good catch. No blocking concerns.

PR #1452, opened 2026-05-12
refactor: add TopicToTableResolver

sfc-gh-skurella:
SNOW-3417175

Pure refactor. Will allow us to add an alternative implementation and switch between them with a setting.

PR #1502, opened 2026-06-25
implement `topic2table` regex replacement

sfc-gh-skurella:
[SNOW-3417175](https://snowflakecomputing.atlassian.net/browse/SNOW-3417175)

Adds optional regex **group substitution** to `snowflake.topic2table.map`, so a capturing topic pattern can project matched groups into the resolved table name (e.g. `topic_(.*):"tbl_$1"` sends `topic_events` to `tbl_events`).

## Setting

`snowflake.topic2table.map.regex.replacement` — **defaults to `false`** (opt-in). It selects which resolver backs `topic2table.map`; the mapping syntax itself is unchanged.

## Behavior: old vs new

- **`false` (default, unchanged):** a topic that matches an entry resolves to that entry's table value **verbatim**. `$1` and other `$`/`\` characters are literal. Unquoted tables are uppercased, quoted tables preserve case.
- **`true`:** the table value is treated as a Java regex **replacement template** evaluated against the topic match. `$1`, `$2`, … expand to the topic's capture groups. Unquoted templates are uppercased **after** substitution (so captured text is uppercased too); quoted templates preserve case of both the literal text and the substituted groups.

## Pitfalls

- **Literal `$` / `\` break under `true`.** With replacement enabled, these become replacement syntax, so an existing table name like `MLOG$_TBL` (Oracle CDC style) throws at **resolution time** — config validation does not catch it because it doesn't simulate substitution. Quoting the table does **not** make `$` literal. This regression risk is the reason the setting is opt-in and defaults to `false`; existing users are unaffected unless they enable it.
- **Uppercasing applies to substituted groups.** Under `true`, an unquoted template uppercases the whole result, including captured topic text. Quote the table to preserve case.
- **Known-table listing is partial.** Templates containing a group reference (`$0`–`$9`) can't be enumerated ahead of time, so they're omitted from the statically-known table names; only fixed templates are reported.
- Topic patterns still must not overlap (rejected at config time).

[SNOW-3417175]: https://snowflakecomputing.atlassian.net/browse/SNOW-3417175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/Constants.java
+++ b/src/main/java/com/snowflake/kafka/connector/Constants.java
@@ -9,6 +9,9 @@ public final class Constants {
     public static final String NAME = "name";
     public static final String TOPICS = "topics";
     public static final String SNOWFLAKE_TOPICS2TABLE_MAP = "snowflake.topic2table.map";
+    public static final String SNOWFLAKE_TOPIC2TABLE_MAP_REGEX_REPLACEMENT =
+        "snowflake.topic2table.map.regex.replacement";
+    public static final boolean SNOWFLAKE_TOPIC2TABLE_MAP_REGEX_REPLACEMENT_DEFAULT = false;
     public static final String SNOWFLAKE_URL_NAME = "snowflake.url.name";
     public static final String SNOWFLAKE_USER_NAME = "snowflake.user.name";
     public static final String SNOWFLAKE_PRIVATE_KEY = "snowflake.private.key";
--- a/src/main/java/com/snowflake/kafka/connector/RegexTopicToTableResolver.java
+++ b/src/main/java/com/snowflake/kafka/connector/RegexTopicToTableResolver.java
@@ -1,0 +1,95 @@
+package com.snowflake.kafka.connector;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
+
+/**
+ * Resolves topic names to table names using compiled regex patterns with group substitution.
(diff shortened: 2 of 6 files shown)


sfc-gh-skurella:
* **#1502** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1502?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1502?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1452** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1452?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

sfc-gh-bmikaili (approved):
LGTM — opt-in, default-off is solid (Constants → ConfigDef → getOrDefault all resolve to false), thread-safe (fresh Matcher per resolve), substitution semantics correct.

One non-blocking nit worth a follow-up:

`RegexTopicToTableResolver.containsGroupReference` only detects `$` + digit, so it misses Java named-group refs `${name}`. A template like `${env}_sink` (with a `(?<env>...)` group in the topic pattern) is treated as a *fixed* name and included in `tableNames()`, so the startup pre-flight sweep checks Snowflake for a table literally named `${ENV}_SINK` → spurious "table does not exist" noise on every startup. Runtime `resolve()` is unaffected (replaceFirst expands it correctly), so this is operability-only, not a routing bug.

Fix is one line:
```java
if (template.charAt(i) == '$'
    && (Character.isDigit(template.charAt(i + 1)) || template.charAt(i + 1) == '{')) {
  return true;
}
```

Fine to land as-is and follow up.

PR #1505, opened 2026-06-25
end-to-end: de-flake by fixing result order

sfc-gh-skurella:
## Summary

`test_native_complex_smt[v3]` flaked in CI (release-4.0): it asserted the first row's `offset == 0` but got `13`. Root cause is in the test, not the connector — all 100 rows ingested fine.

`Table.select` issues `SELECT * FROM <table>` with no `ORDER BY`, so `select(...)[0]` returns an arbitrary row. Several e2e tests send N distinct records and then assert that the "first" row is the offset-0 record (its offset, key, or value). Snowflake does not guarantee row order without `ORDER BY`, so this is inherently flaky — it only passes when the engine happens to return offset 0 first.

This selects the offset-0 row explicitly via `WHERE record_metadata:offset::int = 0`, mirroring the existing fix already in `test_string_json.py` / `test_json_json.py`.

Tests fixed (the rest of the `select(...)[0]` call sites are safe — identical payloads, `WHERE`/`COUNT` filters, single-record sends, or already ordered):

- `test_native_complex_smt.py`
- `test_avrosr_avrosr.py`
- `test_string_avrosr.py`
- `test_native_string_json_without_schema.py`
- `test_native_string_protobuf.py` (also drops a misleading `LIMIT 1`, which was equally unordered)

### DIFF
--- a/test/tests/test_avrosr_avrosr.py
+++ b/test/tests/test_avrosr_avrosr.py
@@ -88,7 +88,9 @@ def test_avrosr_avrosr(
     wait_for_rows(table.name, RECORD_COUNT)
 
     # -- Verify first row content --
-    row = table.select("*")[0]
+    # Snowflake does not guarantee row ordering without ORDER BY, so we must
+    # select the specific record at offset 0 rather than relying on insertion order.
+    row = table.select("*", "WHERE record_metadata:offset::int = 0")[0]
 
     assert row["ID"] == 0
     assert row["FIRSTNAME"] == "abc0"
--- a/test/tests/test_native_complex_smt.py
+++ b/test/tests/test_native_complex_smt.py
@@ -41,7 +41,9 @@ def test_native_complex_smt(
     wait_for_rows(table.name, RECORD_COUNT)
 
     # -- Verify first row: key extracted, c2 dropped --
-    row = table.select("*")[0]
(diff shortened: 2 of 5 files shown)


sfc-gh-skurella:
* **#1505** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1505?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1505?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sfc-gh-bmikaili (approved):
Verified the fix against the code:
- `Table.select()` appends `extra_clauses` verbatim, so `WHERE record_metadata:offset::int = 0` is valid and matches existing prior art in `test_string_json.py`.
- All 5 touched tests use the default single-partition topic (`num_partitions=1`, `partition=0`), so exactly one row has offset 0 — fully deterministic.
- Spot-checked the untouched `select(...)[0]` sites: all safe (identical payloads, single-row `wait_for_rows(..., 1)`, or WHERE/COUNT filters).
- Nice cleanup dropping the misleading unordered `LIMIT 1` in the protobuf test.

LGTM.

PR #1506, opened 2026-06-26
fix hasErrorLoggingEnabled to check SHOW TABLES error_logging value ON

sfc-gh-anewsome:
Snowflake returns "ON" (not "Y") for enabled ERROR_LOGGING in SHOW TABLES output, so tables with error logging were incorrectly flagged as disabled.


sfc-gh-anewsome:
Replaced by https://github.com/snowflakedb/snowflake-kafka-connector/pull/1511

PR #1458, opened 2026-05-14
improve password parsing

sfc-gh-skurella:
## Summary

Reworks how authentication config is modeled in `SinkTaskConfig`. Previously sensitive fields were typed as `Password` but had to tiptoe around nullability; now all auth-related config is `Optional`, which removes null-handling at call sites and reads more cleanly. The cleanup is modest here but has a larger payoff in the OAuth PR stack that builds on top of it.

## Changes

- **`SinkTaskConfig` accessors** now return `Optional`:
  - `getSnowflakePrivateKey()` / `getSnowflakePrivateKeyPassphrase()` → `Optional<Password>`
  - `getOauthClientId()` / `getOauthTokenEndpoint()` → `Optional<String>`
  - `getOauthClientSecret()` / `getOauthRefreshToken()` → `Optional<Password>`
- **`optionalString` / `optionalPassword` helpers**: parse raw config values into `Optional`, treating `null` or blank (whitespace-only) strings as absent via `isBlank()`. Builder setters take the raw `Password`/`String`; `from()` sets them conditionally with `ifPresent(...)` so AutoValue defaults unset optionals to `Optional.empty()`.
- **`PrivateKeyTool.parsePrivateKey`**: signature changed to `(Password privateKey, Optional<Password> privateKeyPassphrase)`.
- **`InternalUtils` / `StreamingClientProperties`**: updated to the new optional accessors; JWT paths use `orElseThrow` to fail fast when a private key is required but missing.
- **`DefaultConnectorConfigValidator.validateOAuthConfig`**: now rejects blank (whitespace-only) OAuth client ID / secret using `isBlank()`, consistent with the parse-time `isBlank()` semantics — previously a whitespace-only value passed validation but was dropped at parse time, surfacing later as ERROR_0026/0027 at connection setup.

## Test plan

- [x] `SinkTaskConfigTest` — updated for `Optional` accessors; blank/whitespace handling
- [x] `ConnectorConfigValidatorTest` — added `testOAuthBlankClientId` / `testOAuthBlankClientSecret`
- [x] `InternalUtilsTest`, `StreamingClientPropertiesTest` — updated `parsePrivateKey` callers to `Password` + `Optional<Password>`

PR #1512, opened 2026-07-01
e2e: wait for invalidated partition to recover instead of global row count

sfc-gh-skurella:
## Summary

`test_invalidation_one_partition_others_healthy` intermittently fails on master with `Partition 1 did not recover: before=100, after=100` (assert `100 > 100`) — seen post-merge on today's commits #1511 and #1458, unrelated to those diffs.

Root cause is a test-design race in the recovery wait. After invalidating partition 1, the test waited on the **global** row count reaching `stalled_rows + 50`, but the drip-feed writes round-robin to all 3 partitions. The two healthy partitions (0 and 2) alone can push the total past that threshold before the invalidated partition reopens, so the wait returns *before* recovery and the per-partition assertion (`after[1] > before[1]`) then fails.

Fix: wait on the **invalidated partition's own** row count exceeding its pre-invalidation count — which is exactly the post-condition under test — via a new `_wait_for_partition_recovery` helper. The stall-detection call is kept (it still confirms invalidation took hold); only its now-unused return value is dropped.

Note: `test_invalidation_all_partitions` has the same latent global-count-vs-per-partition race but is left unchanged here.

### DIFF
--- a/test/tests/test_channel_invalidation.py
+++ b/test/tests/test_channel_invalidation.py
@@ -137,6 +137,32 @@ def _get_partition_row_counts(driver, table_name):
     return {int(r[0]): int(r[1]) for r in rows}
 
 
+def _wait_for_partition_recovery(
+    driver, table_name, partition, count_before, timeout=120
+):
+    """Wait until a specific partition's row count exceeds count_before.
+
+    Waiting on the *global* row count is racy when a drip-feed writes to
+    multiple partitions: the healthy partitions alone can push the total past
+    the threshold before the invalidated partition reopens, so the wait would
+    return before recovery actually happened. Poll the invalidated partition's
+    own count instead, which is exactly the post-condition under test.
+
+    Returns the per-partition counts observed once recovery is detected.
+    """
+    deadline = time.monotonic() + timeout
+    partition_counts = {}
+    while time.monotonic() < deadline:
+        partition_counts = _get_partition_row_counts(driver, table_name)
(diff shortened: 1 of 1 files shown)


sfc-gh-skurella:
* **#1512** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1512?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1512?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sfc-gh-bmikaili (approved):
Approving.

Reviewed the fix: the old recovery wait keyed on the **global** row count, which healthy partitions 0+2 can satisfy before the invalidated partition 1 reopens — exactly the `Partition 1 did not recover: before=100, after=100` signature. Switching to `_wait_for_partition_recovery` on partition 1's *own* count is strictly more correct and test-only; it fails loud (120s timeout) if partition 1 genuinely never recovers.

Verified locally on Azure sfctest0 (v4 connector, real `SYSTEM$STREAMING_CHANNEL_INVALIDATE`, Confluent 7.8.0): pre-fix 4/4 green, fixed 3/3 green. I could not reproduce the failing race on that account — it commits all partitions on one ~30s flush tick, coupling healthy + invalidated partitions — so this approval rests on the correctness of the new wait condition rather than a caught reproduction.

PR #1514, opened 2026-07-06
SNOW-3136119: improve Kafka Connector local observability

sfc-gh-bmikaili:
## Context
Review of the connector's local observability (SNOW-3136119). Adds performance
metrics and SDK log/metrics controls so slow or stalled ingestion can be
diagnosed from JMX (k8s) and, via an opt-in SDK Prometheus endpoint, in more
detail for benchmarking — without any change to the ingest SDK repo.

## Changes
- JMX (`TaskMetrics`): add `flush-duration` (recorded on the shutdown drain,
  success only) and `records-appended` (rows actually appended, distinct from
  the existing records-entering-`put()` meter).
- `SdkBootstrapConfig`, applied in `SnowflakeSinkTask.start()` before the SDK
  client is created (task JVM, so it also works in Kafka Connect distributed
  mode):
  - default the ingest SDK log level to WARN unless the operator set
    `SS_LOG_LEVEL` themselves (env or `-D`);
  - opt-in SDK Prometheus endpoint, enabled via **internal** (not part of the
    public `ConfigDef`) properties read through typed `SinkTaskConfig`
    accessors: `snowflake.streaming.metrics.prometheus.enable` (default
    `false`), `.port`, and `.host`. Defaults are sentinels (`port = -1`,
    `host = ""`) so a typo fails loudly; `SdkBootstrapConfig` validates the
    port/host when the endpoint is enabled.
- Log noise: downgrade per-cycle INFO lines to DEBUG in `BatchOffsetFetcher`
  and `WaitForLastOffsetCommittedPolicy`. (`PartitionChannelManager`'s
  parallel-flush line stays INFO — it fires once per task teardown.)
- Lifecycle logs: include the task id in the "starting task" line (and fix a
  pre-existing arg-count bug in the "task started" line); switch
  `SnowflakeSchemaEvolutionService` to `KCLogger` and log successful DDL.

Out of scope: no SDK-repo changes; no hot-path byte counting (the removed
per-record appendRow timer would have added per-row overhead); the existing
per-channel `ChannelStatus` log is unchanged.

## Testing
- Unit tests for the new metrics, `SdkBootstrapConfig`, and the `SinkTaskConfig`
  accessors.
- `LocalObservabilityIT` (in-process) on a real account: green.
- Python e2e (`test_local_observability.py`) on a real account: green. It uses
  the `create_custom_connector` + `create_table` fixtures and `RecordProducer`,
  and is forced to run first (`pytest_collection_modifyitems`) so its connector
  is the first SDK client in the worker and the connector's own
  `prometheus.enable` config bootstraps the SDK metrics server; it then ingests,
  verifies no regression, and scrapes the SDK Prometheus endpoint cross-container
  for the expected metric names.
- google-java-format and ruff clean.


src/main/java/com/snowflake/kafka/connector/config/ConnectorConfigDefinition.java
```diff
@@ -499,6 +499,42 @@ public static ConfigDef getConfig() {
...
             14,
             Width.NONE,
-            KafkaConnectorConfigParams.CACHE_PIPE_EXISTS_EXPIRE_MS);
+            KafkaConnectorConfigParams.CACHE_PIPE_EXISTS_EXPIRE_MS)
+        .define(
+            KafkaConnectorConfigParams.PROMETHEUS_ENABLE,
+            BOOLEAN,
+            KafkaConnectorConfigParams.PROMETHEUS_ENABLE_DEFAULT,
+            LOW,
+            "When true, enables the Snowpipe Streaming SDK's Prometheus metrics endpoint for"
```
sfc-gh-skurella: The comment mentions this is for benchmarking. Are customers expected to use this? If it's not relevant to normal operation, we can skip adding the config entries to this file.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/service/PartitionChannelManager.java
```diff
@@ -289,7 +289,7 @@ public void waitForAllChannelsToCommitData() {
       return;
     }
 
-    LOGGER.info("Starting parallel flush for {} channels", channelCount);
+    LOGGER.debug("Starting parallel flush for {} channels", channelCount);
```
sfc-gh-skurella: NIT: this can probably stay as INFO - it should only be invoked in `SnowflakeSinkServiceV2::stop`, it's good to have verbosity here as typically this is only printed once per task during rebalancing/teardown.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
```diff
@@ -294,8 +300,11 @@ public CompletableFuture<Void> waitForLastProcessedRecordCommitted() {
   private boolean insertRowWithFallback(Map<String, Object> row, long offset) {
     try {
       LOGGER.trace("Inserting transformed record: {}, offset: {}", row, offset);
-      getChannel().appendRow(row, Long.toString(offset));
+      try (TaskMetrics.TimingContext ignored = taskMetrics.timeAppendRow()) {
```
sfc-gh-skurella: Can this possibly have an effect on throughput?

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
```diff
@@ -147,6 +147,14 @@ public void start(final Map<String, String> parsedConfig) {
     this.taskStartTime = System.currentTimeMillis();
     this.taskConfigId = config.getTaskId();
 
+    // First log line now carries the task id (was logged before the id was resolved).
+    this.DYNAMIC_LOGGER.info("starting task {}...", this.taskConfigId);
+
+    // Apply SDK bootstrap knobs (log level, optional Prometheus) in the TASK JVM before any
+    // streaming client is created. Distributed mode runs the task in a different JVM than the
+    // connector, so this must happen here, not in SnowflakeStreamingSinkConnector.start().
+    SdkBootstrapConfig.apply(parsedConfig);
```
sfc-gh-skurella: Could we use `config` here instead? We should avoid parsing the raw map as much as practicable.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SdkBootstrapConfig.java
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * Applies bootstrap knobs for the current task's config.
+   *
+   * <p><b>JVM-global, first-write-wins:</b> the SDK's FFIBootstrap reads these system properties
+   * once per worker JVM on first client creation. When multiple tasks or connectors share a worker
+   * JVM, whichever task calls {@code apply()} first wins; a later task enabling Prometheus (or
+   * setting a different log level) after the SDK has already bootstrapped will have no effect.
+   */
+  public static void apply(Map<String, String> config) {
+    logLevelToSet(System.getProperty(SS_LOG_LEVEL), System.getenv(SS_LOG_LEVEL))
```
sfc-gh-skurella: NIT: IMO would be better to replace with `shouldSetDefaultLogLevel` or `isLogLevelProvided` and set to the default value here if the result is `true` / `false`. This was quite confusing to read - it looks like a coalescing function, but then why would we set the log level if it's already provided? Why are we calling it "default" in the log line? And then I read the implementation and it became clear. That shouldn't be necessary.

src/main/java/com/snowflake/kafka/connector/Constants.java
```diff
@@ -86,6 +86,14 @@ public static final class KafkaConnectorConfigParams {
     public static final String JMX_OPT = "jmx";
     public static final boolean JMX_OPT_DEFAULT = true;
 
+    // SDK Prometheus metrics endpoint (opt-in; off by default)
+    public static final String PROMETHEUS_ENABLE = "snowflake.streaming.metrics.prometheus.enable";
+    public static final boolean PROMETHEUS_ENABLE_DEFAULT = false;
+    public static final String PROMETHEUS_PORT = "snowflake.streaming.metrics.prometheus.port";
+    public static final int PROMETHEUS_PORT_DEFAULT = 50000;
+    public static final String PROMETHEUS_HOST = "snowflake.streaming.metrics.prometheus.host";
+    public static final String PROMETHEUS_HOST_DEFAULT = "127.0.0.1";
```
sfc-gh-skurella: NIT: I'd prefer to leave the defaults as empty strings / -1 to make sure the values are correctly provided - a simple typo would result in a challenging debug session

test/tests/test_local_observability.py
```diff
@@ -0,0 +1,204 @@
...
+    table = create_table(
+        unsalted_name,
+        columns="(record_metadata variant)",
+    )
+    driver.createTopics(topic, partitionNum=NUM_PARTITIONS, replicationNum=1)
+
+    driver.createConnector(
+        name_salt=name_salt,
+        rest_request_template_filename=CONFIG_FILE,
+    )
```
sfc-gh-skurella: Please don't use the legacy APIs - there should generally be little to no need to call `driver` methods in tests. We have fixtures which also take care of cleanups.

test/tests/test_local_observability.py
```diff
@@ -0,0 +1,204 @@
...
+    "append-row-duration",
+    "records-appended",
+]
+
+# Glob patterns for JMX metrics files produced by an optional Jolokia scraper.
+_JMX_FILE_GLOB = "/tmp/sf-metrics-*.jsonl"
+_JMX_FILE_ENV = "METRICS_FILE"
+
+
+def _send_records(driver, topic: str, count_per_partition: int):
```
sfc-gh-skurella: Could we use the RecordProducer here instead?

test/tests/test_local_observability.py
```diff
@@ -0,0 +1,204 @@
...
+
+    if body is None:
+        # The SDK's Prometheus server is initialized once per Kafka Connect worker JVM, when
+        # FFIBootstrap runs on the *first* SDK client creation (it reads SS_ENABLE_METRICS at
+        # that moment). In a shared-worker run (e.g. the CI functional suite), an earlier test
+        # whose connector did not enable Prometheus bootstraps the SDK first, so this connector's
+        # SS_ENABLE_METRICS=true has no effect and nothing listens on the port. That is a known
+        # limitation, not a regression, so skip rather than fail when the endpoint is unreachable.
+        # The endpoint IS asserted when this connector is the first/only SDK client in the worker
+        # (e.g. running this test in isolation: -k test_prometheus_metrics_exposed).
```
sfc-gh-skurella: This will effectively make it never run in the CI. Can we force pytest to run this test first and turn the `skip` into `fail`? Either using `pytest_collection_modifyitems` or https://pypi.org/project/pytest-order/ (please remember to fix the version if you add a package)

src/main/java/com/snowflake/kafka/connector/config/ConnectorConfigDefinition.java
sfc-gh-bmikaili: Done — made the three `prometheus.*` properties internal: removed their `.define()` entries from `ConnectorConfigDefinition`, and they're now read via typed `SinkTaskConfig` accessors. Not part of the public config surface.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/service/PartitionChannelManager.java
sfc-gh-bmikaili: Agreed — reverted to INFO. It only fires once per task teardown.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SnowpipeStreamingPartitionChannel.java
sfc-gh-bmikaili: Good catch — removed the per-record `appendRow` timer entirely to avoid hot-path overhead. Kept the `flush-duration` timer and the `records-appended` meter (both cheap / not per-row).

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTask.java
sfc-gh-bmikaili: Done — `start()` now passes typed `SinkTaskConfig` values (new `isPrometheusMetricsEnabled()/getPrometheusMetricsPort()/getPrometheusMetricsHost()` accessors) instead of the raw map.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/SdkBootstrapConfig.java
sfc-gh-bmikaili: Done — renamed to `shouldSetDefaultLogLevel(sysprop, env)` returning a boolean; `apply()` sets the default when it returns true.

src/main/java/com/snowflake/kafka/connector/Constants.java
sfc-gh-bmikaili: Done — `PROMETHEUS_PORT_DEFAULT = -1` and `PROMETHEUS_HOST_DEFAULT = ""`. `SdkBootstrapConfig.validatePrometheus()` throws (fail-loud) if the endpoint is enabled without a valid port/host.

test/tests/test_local_observability.py
sfc-gh-bmikaili: Done — the test now uses the `create_custom_connector` + `create_table` fixtures (fixture-managed teardown) and no longer calls `driver` connector methods or hand-rolls try/finally cleanup.

test/tests/test_local_observability.py
sfc-gh-bmikaili: Done — switched to `RecordProducer` (single-partition send).

test/tests/test_local_observability.py
sfc-gh-bmikaili: Done — added ordering in `conftest.pytest_collection_modifyitems` so this test runs first; its connector is then the first SDK client in the worker and the connector's own `prometheus.enable` config bootstraps the SDK metrics server. The scrape is now a hard assertion (no skip), and the worker-env workaround was removed.

src/main/java/com/snowflake/kafka/connector/Constants.java
```diff
@@ -101,6 +101,14 @@ public static final class KafkaConnectorConfigParams {
     public static final String JMX_OPT = "jmx";
     public static final boolean JMX_OPT_DEFAULT = true;
 
+    // SDK Prometheus metrics endpoint (opt-in; off by default)
+    public static final String PROMETHEUS_ENABLE = "snowflake.streaming.metrics.prometheus.enable";
+    public static final boolean PROMETHEUS_ENABLE_DEFAULT = false;
+    public static final String PROMETHEUS_PORT = "snowflake.streaming.metrics.prometheus.port";
+    public static final int PROMETHEUS_PORT_DEFAULT = -1;
+    public static final String PROMETHEUS_HOST = "snowflake.streaming.metrics.prometheus.host";
+    public static final String PROMETHEUS_HOST_DEFAULT = "";
```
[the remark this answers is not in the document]
sfc-gh-bmikaili: Updated the PR description to match the code. The `-1` / `""` sentinel defaults are intentional (per review feedback above — fail loud on a typo rather than silently using a plausible value); `SdkBootstrapConfig` validates the port/host when Prometheus is enabled.

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
```diff
@@ -296,6 +308,21 @@ public static Builder builderFrom(Map<String, String> raw, boolean skipTaskSpeci
                     KafkaConnectorConfigParams
                         .SNOWFLAKE_FEATURE_PRECOMMIT_CLIENT_RECOVERY_DEFAULT)));
 
+    boolean prometheusMetricsEnabled =
+        Boolean.parseBoolean(
+            config.getOrDefault(
+                KafkaConnectorConfigParams.PROMETHEUS_ENABLE,
+                String.valueOf(KafkaConnectorConfigParams.PROMETHEUS_ENABLE_DEFAULT)));
```
sfc-gh-skurella: NIT: no need to take the default to string and back

```
boolean prometheusMetricsEnabled =
    Optional.ofNullable(config.get(KafkaConnectorConfigParams.PROMETHEUS_ENABLE))
        .map(Boolean::parseBoolean)
        .orElse(KafkaConnectorConfigParams.PROMETHEUS_ENABLE_DEFAULT);
```

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
```diff
@@ -296,6 +308,21 @@ public static Builder builderFrom(Map<String, String> raw, boolean skipTaskSpeci
...
+    int prometheusMetricsPort =
+        Integer.parseInt(
+            config.getOrDefault(
+                KafkaConnectorConfigParams.PROMETHEUS_PORT,
+                String.valueOf(KafkaConnectorConfigParams.PROMETHEUS_PORT_DEFAULT)));
+    String prometheusMetricsHost =
+        config.getOrDefault(
+            KafkaConnectorConfigParams.PROMETHEUS_HOST,
+            KafkaConnectorConfigParams.PROMETHEUS_HOST_DEFAULT);
```
sfc-gh-skurella: NIT: non-required params can be `Optional`s, e.g. the private key is an `Optional<Password>` - and then we don't need "fake" defaults

```
Optional<Integer> prometheusMetricsPort =
    Optional.ofNullable(config.get(KafkaConnectorConfigParams.PROMETHEUS_PORT)
        .map(Integer.parseInt);
Optional<String> prometheusMetricsHost =
    optionalString(config.get(KafkaConnectorConfigParams.PROMETHEUS_HOST));

File: test/tests/test_local_observability.py [hunk 13]
```diff
@@ -0,0 +1,108 @@
...
+    Connector and table/topic teardown are handled by the fixtures.
+    """
+    unsalted_name = "local_observability"
+    table = create_table(unsalted_name, columns="(record_metadata variant)")
+    driver.createTopics(table.name, partitionNum=1, replicationNum=1)
+
+    # Load the connector config template and enable Prometheus via its config (this test runs first,
+    # so its connector is the first SDK client in the worker and bootstraps the SDK metrics server).
+    base = json.loads((TEMPLATE_DIR / CONFIG_FILE).read_text())
+    connector = create_custom_connector(unsalted_name, dict(base["config"]))
```
sfc-gh-skurella: NIT: I don't think we need `create_custom_connector` here, or to put the config in a JSON file - `create_connector(v4_config={**V4_CONFIG_TEMPLATE, "key": "value", ...})` should suffice. You only need to specify `topics` and converters and your `prometheus` settings.

sfc-gh-skurella (approved):
Thanks! Left some NITs but otherwise looks good.

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
sfc-gh-bmikaili: Done — `prometheusMetricsEnabled` is now parsed via `Optional.ofNullable(config.get(...)).map(Boolean::parseBoolean).orElse(PROMETHEUS_ENABLE_DEFAULT)` (no string round-trip).

src/main/java/com/snowflake/kafka/connector/config/SinkTaskConfig.java
sfc-gh-bmikaili: Done — `getPrometheusMetricsPort()`/`getPrometheusMetricsHost()` are now `Optional<Integer>`/`Optional<String>` (parsed via `Optional.ofNullable(...).map(Integer::parseInt)` and `optionalString(...)`); dropped the `-1`/`""` sentinel defaults. `SdkBootstrapConfig.apply` takes the Optionals and fails loud if metrics are enabled but the port/host is unset.

test/tests/test_local_observability.py
sfc-gh-bmikaili: Done — the test now builds the config inline from `V4_CONFIG_TEMPLATE` via the `create_connector` fixture (with `create_topics`/`create_table`), and I removed the `local_observability.json` template.

PR #1519, opened 2026-07-06
Expose snowflake.oauth.include.scope to control OAuth scope (default false)

sfc-gh-skurella:
## Summary

The streaming SDK (>= 1.6.1) defaults `oauth_include_scope` to **true**, which makes it derive `session:role:{role}` and send it as the `scope` on OAuth refresh requests. External OAuth / IdP refresh tokens that were minted **without** a role scope are then rejected by the IdP (`access_denied` / "Denied scopes: session:role:***"). This is what broke the OAuth E2E test on `master` right after the SDK bump 1.6.0 → 1.6.1, and it would also break external OAuth customers (e.g. JPMC-style lambda-minted tokens).

Rather than hardcoding `oauth_include_scope=false`, this exposes the behavior as connector config so it can be turned back on when a deployment actually needs a role scope.

- **`snowflake.oauth.include.scope`** (boolean, default `false`): forwarded to the streaming SDK as `oauth_include_scope`. Default `false` restores KC v3 behavior (no scope) and fixes external OAuth out of the box.
- **`snowflake.oauth.scope`** (string, optional): explicit scope, forwarded to the SDK as `oauth_scope`. Only used when `include.scope` is `true`; otherwise the SDK derives `session:role:{role}`.

For symmetry, the JDBC `OAuthAccessTokenFetcher` now honors an explicit scope resolved from the same config (no scope by default; explicit scope, else `session:role:{role}`, only when `include.scope` is enabled). This mirrors what KC v3 did (no scope) by default. `validate()` rejects setting a scope while `include.scope` is `false`.

### Why default `false`?

- Matches KC v3, which never sent a scope on OAuth refresh.
- Works for external OAuth tokens minted without a role scope (the common failure mode).
- Users who rely on Snowflake role-scoped OAuth can opt in with `snowflake.oauth.include.scope=true` (and optionally an explicit `snowflake.oauth.scope`).

## Changes

- `Constants` / `ConnectorConfigDefinition`: new `snowflake.oauth.include.scope` (default false) and `snowflake.oauth.scope` params.
- `SinkTaskConfig`: parse the two new fields (`getOauthIncludeScope()`, `getOauthScope()`).
- `StreamingClientProperties`: always forward `oauth_include_scope`; forward `oauth_scope` when scopes are enabled.
- `InternalUtils` / `OAuthAccessTokenFetcher`: resolve and pass an optional scope on the JDBC token request.
- `DefaultConnectorConfigValidator`: reject `oauth.scope` set while `oauth.include.scope` is false.

## Test plan

- [x] `SinkTaskConfigTest` — parsing + defaults (`include.scope=false`, no scope).
- [x] `StreamingClientPropertiesTest` — `oauth_include_scope` always forwarded (default `false`), `oauth_scope` forwarded only when enabled.
- [x] `OAuthAccessTokenFetcherTest` — `scope` form param present when set, absent otherwise.
- [x] `ConnectorConfigValidatorTest` — scope-without-include is invalid; scope-with-include is valid.
- [ ] Re-run the OAuth E2E ingestion test on this branch (default config, no scope) to confirm master CI is green.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
@@ -12,6 +12,7 @@ import java.sql.SQLException;
 import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
+import java.util.Optional;
 import java.util.Properties;
 import org.apache.kafka.common.config.types.Password;
 
@@ -106,7 +107,11 @@ public class InternalUtils {
         properties.put(
             JDBC_TOKEN,
             OAuthAccessTokenFetcher.fetchAccessToken(
-                oauthUrl, oauthClientId, oauthClientSecret, config.getOauthRefreshToken()));
+                oauthUrl,
+                oauthClientId,
+                oauthClientSecret,
+                config.getOauthRefreshToken(),
+                resolveOauthScope(config)));
         break;
       case SNOWFLAKE_JWT:
         properties.put(JdbcPropertyKeys.AUTHENTICATOR, SNOWFLAKE_JWT);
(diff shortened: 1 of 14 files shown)


sfc-gh-ggeng (approved):
LGTM from the SSv2 SDK side

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
```diff
@@ -148,6 +153,25 @@ private static void putIfNotBlank(Properties properties, String key, String valu
...
+  private static Optional<String> resolveOauthScope(SinkTaskConfig config) {
+    if (!config.getOauthIncludeScope()) {
+      return Optional.empty();
+    }
+    if (config.getOauthScope().isPresent()) {
+      return config.getOauthScope();
+    }
+    return Optional.ofNullable(config.getSnowflakeRole())
+        .filter(role -> !isBlank(role))
+        .map(role -> "session:role:" + role);
```
sfc-gh-ggeng, line 172: Probably you don't need this. Our SDK shall be able to resolve this as you send the role to the SDK right?

sfc-gh-skurella: This is for JDBC, which doesn't perform headless refresh token / client credentials forcing us to reimplement the entire logic here 🙈

sfc-gh-bmikaili (approved):
LGTM modulo e2e tests — please confirm the OAuth E2E ingestion test is green on master before relying on this.

PR #1521, opened 2026-07-09
add support for typed headers

sfc-gh-skurella:
SNOW-3214907

## Summary

Kafka Connect deserializes raw header bytes into typed values using the configured `header.converter` before a record reaches the connector. The connector writes those values to `RECORD_METADATA:headers`.

Today the connector flattens every header value to a string (`String.valueOf(...)`) before writing it, so type information from the `header.converter` is lost. This PR restores the ability to preserve those types, gated behind a new feature flag.

## Behavior

`snowflake.feature.structured.headers` (default `false`):

- **`false` (default, unchanged):** every header value is stringified in `RECORD_METADATA:headers`. A numeric header `42` is stored as `"42"`, a boolean as `"true"`, an object as its string rendering, etc.
- **`true`:** header values keep the type produced by the `header.converter`. Numbers, booleans, objects, and arrays are stored as well-typed VARIANT fields in `RECORD_METADATA:headers` (byte-array headers are base64-encoded). This matches how KC v3 wrote headers.

The default is `false` because changing the stored representation of existing headers is a breaking change for downstream consumers that parse `RECORD_METADATA:headers`.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/Constants.java
+++ b/src/main/java/com/snowflake/kafka/connector/Constants.java
@@ -83,6 +83,10 @@ public final class Constants {
         "snowflake.feature.validation.error.table.name";
     public static final boolean SNOWFLAKE_FEATURE_VALIDATION_ERROR_TABLE_NAME_DEFAULT = true;
 
+    public static final String SNOWFLAKE_FEATURE_STRUCTURED_HEADERS =
+        "snowflake.feature.structured.headers";
+    public static final boolean SNOWFLAKE_FEATURE_STRUCTURED_HEADERS_DEFAULT = false;
+
     // Caching
     public static final String CACHE_TABLE_EXISTS = "snowflake.cache.table.exists";
     public static final boolean CACHE_TABLE_EXISTS_DEFAULT = true;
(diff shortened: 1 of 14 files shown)


sfc-gh-skurella:
* **#1521** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1521?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1521?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sfc-gh-bmikaili (approved):
LGTM. Clean, minimal, correctly flag-gated (`snowflake.feature.structured.headers`, default `false` = no behavior change for existing users). The core switch is a single branch in `convertHeaders`, threaded cleanly through `SnowflakeMetadataConfig`/`SnowflakeSinkRecord`.

Nice touches:
- The `MAP(VARCHAR,VARCHAR)` footgun is anticipated — only the VARIANT subclass opts in, with a javadoc warning on the base.
- Strong test coverage: unit tests cover both legacy-string and structured shapes; `test_headers.py` exercises four real `header.converter`s (Simple/Json/String/ByteArray incl. base64) and asserts structured-v4 == v3 for free via the version fixture.

One optional nit (non-blocking): `double`/`float` headers were dropped from the Iceberg record and aren't in `EXPECTED_HEADERS`. They were never asserted pre-PR either, so this isn't a regression — just flagging in case you want to cover float/double round-trip in structured mode later (tolerance-based, to avoid precision flakiness).

PR #1522, opened 2026-07-14
SNOW-3766306: normalize TIME values with a UTC offset so they land (KC v3 parity)

sfc-gh-bmikaili:
SNOW-3766306

## Bug

A `TIME` column receiving an ISO time string with a UTC offset but **no fractional seconds** (e.g. `"00:00:00Z"`) passes the connector's client-side check, but SSv2 can't parse it and **silently drops the row** (committed offset advances → data lost). KC v3 accepted these and landed the offset-stripped time.

## Fix (match KC v3)

For `TIME` values, strip the timezone offset before sending to the SDK (`"00:00:00Z"` → `"00:00:00"`), so anything the client accepts also lands — exactly as v3 did. Client accept/reject is unchanged from v3.

### Behavior — `TIME` column

| Input | v3 (target) | Before | After |
|---|---|---|---|
| `"00:00:00Z"` | lands `00:00:00` | ❌ silently dropped | ✅ lands `00:00:00` |
| `"22:00:00Z"`, `"22:00Z"`, `"10:30:00+05:00"` | lands (offset dropped) | ❌ dropped | ✅ lands |
| `"07:59:59.999999Z"` | lands | ✅ lands | ✅ lands (unchanged) |
| `"00:00:00"` (no offset) | lands | ✅ lands | ✅ lands (unchanged) |

**Scope:** `TIME` columns only. No change to `DATE`, `TIMESTAMP`, or any other type.

## Changes

- `DataValidationUtil` — new `validateAndFormatTime`: parses via the existing v3 cascade and returns the offset-stripped canonical string (+63 lines).
- `RowValidator` — the `TIME` case returns the normalized string (±3 lines).
- Unit tests + one live E2E IT (`"00:00:00Z"` and `"07:59:59.999999Z"` both land).

## Out of scope (follow-up)

`DATE` support for ISO-8601 literals with a trailing `Z` (e.g. `"2017-09-15Z"`) is a *new* capability KC v3 never had; tracked separately in **SNOW-3819217**.


sfc-gh-bmikaili:
* **#1523** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1523?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1522** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1522?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1522?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sfc-gh-bmikaili:
Reworked per review, @sfc-gh-skurella — thanks, you were right.

- **Client accept/reject now stays KC v3.** Dropped the "match server acceptance" matcher and the acceptance-broadening (space-sep/`DD-MON-YYYY`/RFC), plus the REAL and VARIANT-size changes (those made the client *more permissive than v3*). Closed #1523.
- **Fix is now normalization before the SDK, not rejection** — strip the offset from TIME, strip the trailing `Z` from ISO-8601 dates — so values land instead of being dropped/DLQ'd.
- **Verified v3 (v3.5.4 / SSv1 SDK 4.4.2):** `"00:00:00Z"` *was* ingested in v3 as `00:00:00` (our earlier reject was a regression — now lands again). `"2017-09-15Z"` was *not* accepted in v3, so the trailing-`Z` DATE support is a new feature, now landing as `2017-09-15`.
- Space-separated timestamps stay client-rejected (v3 parity).

Live IT on sfctest0 confirms all four cases end-to-end. Ready for another look.

sfc-gh-bmikaili:
Adopted your deduplication, @sfc-gh-skurella — thanks, much cleaner. Folded it in with your commit preserved (`6ceee7af`).

One small tweak on top (`c8fb5055`): the centralized trailing-`Z` fallback in `inputToOffsetDateTime` was recursive (one stack frame per trailing `Z`), so I made it strip all trailing `Z`s in a single pass — recursion depth is now bounded to 1, guarding against a pathological all-`Z` input. Added a test for it.

123 unit tests green. Ready for another look.

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
```diff
@@ -675,6 +675,22 @@ private static OffsetDateTime inputToOffsetDateTime(
...
+        // SNOW-3766306: trailing-'Z' fallback for ISO-8601 values carrying a UTC 'Z' that none of
+        // the parsers above accepted — most notably a bare date like "2017-09-15Z". Strip the
+        // trailing 'Z'(s) and retry the whole cascade once. All trailing 'Z's are removed in a
+        // single pass so the recursive call cannot re-enter this branch — recursion depth is
+        // bounded to 1 (guards against a pathological all-'Z' input overflowing the stack).
+        if (stringInput.endsWith("Z")) {
+          int end = stringInput.length();
+          while (end > 0 && stringInput.charAt(end - 1) == 'Z') {
+            end--;
+          }
```
sfc-gh-skurella: What's the purpose of this loop? There can only be one `Z` at the end. If anything, we should act to disallow multiple `Z`s, and not optimize for it.

sfc-gh-bmikaili:
Scoped this PR down to just the reported TIME regression (offset-without-fraction → silently dropped), to match KC v3. Force-pushed to a single commit; behavior table is in the description.

Heads up @sfc-gh-skurella: your dedup refactor came out with the scope-down — the `DATE`/`TIMESTAMP` methods are now unchanged from master, so there's nothing left to dedup here. The `DATE` `"2017-09-15Z"` support (a new capability v3 never had) and the optional temporal-parse dedup are tracked in **SNOW-3819217** if you'd like to pursue them there.

Net change now: +63 in `DataValidationUtil` (one new `validateAndFormatTime`), ±3 in `RowValidator`, plus tests + IT.

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2XpParityIT.java
```diff
@@ -0,0 +1,177 @@
+package com.snowflake.kafka.connector.internal.streaming;
```
sfc-gh-skurella: This test should probably not be called "XP parity" since that's not the aim of client-side validation. WDYT about replacing this file with a single line in the end-to-end test which tests parity with KC v3?

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
```diff
@@ -1062,6 +1062,69 @@ static BigInteger validateAndParseTime(
...
+   * <p>Accepted Java types: {@link String}, {@link LocalTime}, {@link OffsetTime}. Any offset is
+   * removed before returning; the result is the canonical {@link LocalTime#toString()} form (e.g.
+   * {@code "00:00"}, {@code "07:59:59.999999"}).
+   *
+   * @param columnName Column name for error messages
+   * @param input TIME value (String, LocalTime, or OffsetTime)
+   * @param insertRowIndex Row index for error messages
+   * @return Canonical local-time string (offset stripped)
+   */
+  static String validateAndFormatTime(String columnName, Object input, long insertRowIndex) {
```
sfc-gh-skurella: There's significant duplication between `validateAndParseTime` and `validateAndFormatTime`. The former already returns a BigInteger that we discard and after your change it becomes dead code. Could we please *replace* `validateAndParseTime` with `validateAndFormatTime` rather than duplicate it?

src/main/java/com/snowflake/kafka/connector/internal/validation/RowValidator.java
```diff
@@ -221,9 +221,9 @@ private Object validateAndNormalizeColumnValue(ColumnSchema col, Object value)
         break;
 
       case TIME:
-        DataValidationUtil.validateAndParseTime(
-            col.getName(), value, col.getScale() != null ? col.getScale() : 9, insertRowIndex);
-        break;
+        // SNOW-3766306: normalize to canonical local-time string (offset stripped) so values that
+        // pass client validation also land on the SSv2 server, matching KC v3.
+        return DataValidationUtil.validateAndFormatTime(col.getName(), value, insertRowIndex);
```
sfc-gh-skurella: If `validateAndParseTime` and `validateAndFormatTime` perform the same validation, do you think we could add a `snowflake.feature.normalize.time` setting that defaults to `true` and:
- if it's enabled, `return` the value returned from the function
- if it's disabled, `break` so we maintain the previous behavior?

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
sfc-gh-skurella: Would it make sense to return `java.time.LocalTime` here rather than a string and let the SDK serialize it as it sees fit?

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
sfc-gh-bmikaili: That multi-`Z` strip belonged to the DATE trailing-`Z` path, which is no longer in this PR — scope was reduced to just the reported TIME regression. The DATE `Z` support (where that logic lives) is tracked separately in SNOW-3819217, so there's nothing to change here.

src/test/java/com/snowflake/kafka/connector/internal/streaming/SnowflakeSinkServiceV2XpParityIT.java
sfc-gh-bmikaili: Done — deleted this file and folded the check into `SnowflakeSinkServiceV2SchematizationIT` as `timeColumnWithUtcOffset_landsNotDropped` (schematized `TIME(9)`, `client_side` validation, `00:00:00Z` lands as `00:00:00`). Agreed the "XP parity" name was wrong — this is KC v3 parity.

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
sfc-gh-bmikaili: Done — removed `validateAndParseTime` (dead in prod after the scope-down) and collapsed the cascade into a single private `inputToLocalTime` helper; `validateAndFormatTime` is now a thin wrapper. Tests ported to the surviving method.

src/main/java/com/snowflake/kafka/connector/internal/validation/RowValidator.java
sfc-gh-bmikaili: Done — added `snowflake.feature.normalize.time` (default `true`), mirroring the existing `snowflake.feature.validation.error.table.name` flag. Enabled → RowValidator returns the normalized value; disabled → `break` to the previous raw pass-through.

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
sfc-gh-bmikaili: Done — `validateAndFormatTime` now returns `java.time.LocalTime` and RowValidator hands that to the SDK. Verified against a live account: the SDK serializes the `LocalTime` correctly and `00:00:00Z` lands as `00:00:00`.

sfc-gh-skurella (approved):

PR #1524, opened 2026-07-14
e2e: wait for all partitions to recover after channel invalidation

sfc-gh-bmikaili:
## Summary

Fixes a flaky e2e test, `test_invalidation_one_partition_others_healthy`, which
failed on master (scheduled run, `core_legacy confluent 6.2.15 AZURE`) with:

```
AssertionError: Healthy partition 2 lost data: before=100, after=100
{0: 210, 1: 210, 2: 100}
```

No data was actually lost — this is a racy assertion, not a connector bug.

## Root cause

With `tasks.max=1`, a single `SinkTask.put()` carries records for all assigned
partitions. When the invalidated partition's channel throws, the whole `put()`
fails and Kafka Connect retries the entire batch — so **no** partition commits
until the channel reopens. The CI log confirms this: after invalidation the whole
table stayed frozen at 300 rows for the full 15s stall window (the records sent to
the healthy partitions did not commit either).

On recovery the partitions flush across separate cycles. The old test waited only
for the *invalidated* partition to cross its baseline, then immediately snapshotted
per-partition counts. At that instant a healthy partition can still be one flush
cycle behind at its pre-recovery count, so the assertion spuriously reported "lost
data". The records are in Kafka and land on the next flush.

## Change (test-only; no connector code touched)

- Replace the single-partition / global-count recovery waits with
  `_wait_for_partitions_recovery`, which polls until **every** partition under test
  exceeds its baseline, and times out naming a stuck partition instead of
  misreporting loss.
- Applied to both `test_invalidation_one_partition_others_healthy` and
  `test_invalidation_all_partitions`, which shared the same latent race.
- Add two offline regression tests (scripted fake connection, no live account): one
  replays the exact CI failure sequence, one asserts the timeout path.

## Testing

- Offline regression tests: pass (red before the fix, green after).
- `ruff check` / `ruff format --check`: clean.
- Real e2e against sfctest0 (east-us-2 azure):
  - `test_invalidation_one_partition_others_healthy`: **10/10**.
  - `test_invalidation_all_partitions` + `..._one_partition_others_healthy`
    together: **3/3** (6/6 executions).
  - Every partition grew past its baseline in every run.

Follow-up to #1512, which fixed the invalidated-partition side of this test.

### DIFF
--- a/test/tests/test_channel_invalidation.py
+++ b/test/tests/test_channel_invalidation.py
@@ -137,29 +137,44 @@ def _get_partition_row_counts(driver, table_name):
     return {int(r[0]): int(r[1]) for r in rows}
 
 
-def _wait_for_partition_recovery(
-    driver, table_name, partition, count_before, timeout=120
+def _wait_for_partitions_recovery(
+    driver, table_name, counts_before, partitions, timeout=120
 ):
-    """Wait until a specific partition's row count exceeds count_before.
+    """Wait until every partition in `partitions` exceeds its own baseline.
 
-    Waiting on the *global* row count is racy when a drip-feed writes to
-    multiple partitions: the healthy partitions alone can push the total past
-    the threshold before the invalidated partition reopens, so the wait would
-    return before recovery actually happened. Poll the invalidated partition's
-    own count instead, which is exactly the post-condition under test.
+    Two independent races make a narrower wait insufficient:
 
-    Returns the per-partition counts observed once recovery is detected.
+    * Waiting on the *global* row count is racy: a drip-feed to multiple
+      partitions lets the healthy partitions alone push the total past the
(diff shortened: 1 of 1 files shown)


sfc-gh-bmikaili:
* **#1524** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1524?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1524?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sfc-gh-skurella (approved):

PR #1528, opened 2026-07-22
e2e: fix flaky test_nullable_values_after_smt ordering assertion

sfc-gh-skurella:
`test_nullable_values_after_smt` compares the query result against an expected list built in ascending offset order, but `table.select(...)` issued a bare `SELECT` with no `ORDER BY`. Snowflake makes no ordering guarantee without one, so the order-sensitive `assert parsed == expected` is flaky — the data (100 rows, correct indices/offsets) can come back unsorted and fail the comparison.

Observed in [run 29927307081](https://github.com/snowflakedb/snowflake-kafka-connector/actions/runs/29927307081/attempts/1) (`core_legacy apache 3.9.2, Java 11, GCP`), where the `[v3]` parametrization failed with `parsed[0]` starting at offset 164 while `[v4]` happened to come back sorted and passed.

Fix: add `ORDER BY offset` to the query. The schema-evolution sibling `test_se_nullable_values_after_smt.py` already orders by offset; this brings the SMT test in line.

I audited every other multi-row read in `test/tests/` (`table.select`, `select_scalar`, and direct `cursor.execute(...).fetchall()`): all others either filter to a single row (`WHERE offset = N`), already use `ORDER BY`, key results into a dict, or compare counts/sets. This was the only remaining order-sensitive comparison against an unordered SELECT.

### DIFF
--- a/test/tests/test_nullable_values_after_smt.py
+++ b/test/tests/test_nullable_values_after_smt.py
@@ -38,8 +38,11 @@ def test_nullable_values_after_smt(
     wait_for_rows(table.name, EXPECTED_ROWS)
 
     # -- Verify content --
+    # ORDER BY offset: Snowflake does not guarantee row order without it, and
+    # the assertion below compares against an offset-ordered expected list.
     rows = table.select(
         "index, from_optional_field, record_metadata:offset::number AS offset",
+        "ORDER BY offset",
     )
 
     parsed = [


sfc-gh-bmikaili (approved):
LGTM — the fix is sound. `Table.select(projections, extra_clauses="")` takes the new arg, the expected list is built from `range(0, TOTAL_EVENTS, 2)`, and every produce pins `partition=0`, so `ORDER BY offset` is a total order with no tie nondeterminism. Assertion strength unchanged.

One note on the audit claim in the description ("this was the only remaining order-sensitive comparison against an unordered SELECT"). I walked every `table.select` call site and the multi-row reads are indeed all safe — but there are four in the same class that don't fall into any of the four buckets. Each sends `RECORD_COUNT = 100`, then does `table.select("*")[0]` with no `ORDER BY`/`WHERE` and asserts a literal that varies per row:

- `test/tests/test_native_complex_smt.py:44` → asserts `"offset": 0`
- `test/tests/test_native_string_json_without_schema.py:39` → asserts `"offset": 0` and `row["VAL"] == "0"`
- `test/tests/test_avrosr_avrosr.py:91` → asserts `row["ID"] == 0`
- `test/tests/test_string_avrosr.py:53` → asserts `row["ID"] == 0` and `"offset": 0`

Same latent flake as the one fixed here, passing only because Snowflake usually returns rows in insertion order. Not this PR's problem — worth a follow-up ticket?

(For what it's worth, `test_schema_mapping.py:98` and `test_confluent_protobuf_protobuf.py:80` looked similar but are genuinely fine — identical payloads across sends, and `ANY_INT` wildcards rather than literals.)

PR #1529, opened 2026-07-22
e2e: await topic creation in create_topics fixture

sfc-gh-skurella:
`KafkaDriver.createTopics` is fire-and-forget — it calls `AdminClient.create_topics([...])` but never awaits the returned futures. The `create_topics` fixture inherited that: it submitted all creations to a thread pool and returned as soon as the requests were *issued*, not once the broker confirmed them. Tests then started producing immediately.

At high topic counts this races the broker. In `test_pressure_init[v4]` (200 topics × 12 partitions) on a single-broker stress stack, several `CreateTopicsRequest`s timed out (~60s each), so some topics didn't exist yet when the producer ran, failing with `_UNKNOWN_PARTITION` ("Unable to produce message: Local: Unknown partition") — see [run 29935154653](https://github.com/snowflakedb/snowflake-kafka-connector/actions/runs/29935154653/job/88974703914).

Fix: make the fixture block until the broker confirms each topic before returning:
- `createTopics` now returns the `{topic: Future}` dict (previously the return value was unused).
- `_create_one` awaits `future.result()`, so a creation timeout/error surfaces at creation time instead of as a confusing downstream `_UNKNOWN_PARTITION`.
- `_create` now collects the per-topic executor futures and calls `.result()` on each — the `ThreadPoolExecutor` context manager waits for tasks but otherwise swallows their exceptions, which would have masked creation failures.

This does not by itself raise the broker's creation throughput, but it converts the silent race into a deterministic wait (and a clear failure if creation genuinely times out).

### DIFF
--- a/test/lib/driver.py
+++ b/test/lib/driver.py
@@ -186,7 +186,10 @@ class KafkaDriver:
             raise NonRetryableError()
 
     def createTopics(self, topicName, partitionNum=1, replicationNum=1):
-        self.adminClient.create_topics(
+        # create_topics is asynchronous: it returns a {topic: Future} dict and
+        # the broker may not have created the topic by the time this returns.
+        # Callers that must produce immediately should await the futures.
+        return self.adminClient.create_topics(
             [NewTopic(topicName, partitionNum, replicationNum)]
         )
 
--- a/test/lib/fixtures/connector.py
+++ b/test/lib/fixtures/connector.py
@@ -19,8 +19,17 @@ def create_topics(driver: KafkaDriver, name_salt):
     def _create_one(topic, num_partitions, replication_factor, with_table):
         salted = f"{topic}{name_salt}"
         logger.info(f"Creating topic {salted}")
-        driver.createTopics(salted, num_partitions, replication_factor)
+        # Track before awaiting so a topic whose creation partially succeeded is
+        # still torn down.
         created_topics.append(salted)
+        # Block until the broker confirms creation. createTopics is otherwise
+        # fire-and-forget, so without this the caller can start producing before
+        # the topic exists and fail with _UNKNOWN_PARTITION. A creation timeout
+        # or error surfaces here instead of as a confusing downstream failure.
+        for future in driver.createTopics(
+            salted, num_partitions, replication_factor
+        ).values():
+            future.result()
         if with_table:
             driver.create_table(salted)
             created_tables.append(salted)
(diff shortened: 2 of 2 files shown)


sfc-gh-bmikaili (commented):
The diagnosis and the shape of the fix are both right — `createTopics` really was fire-and-forget, and awaiting the futures (plus surfacing the executor exceptions that `ThreadPoolExecutor.__exit__` was swallowing) is the correct change. Tracking `created_topics` before the await so partial creations still get torn down is a nice touch.

I don't think it can land as-is though: the four red `functional` jobs on this PR are caused by this change, and they've surfaced a real pre-existing bug.

Root cause — `test/tests/compatibility/test_compatibility_case_sensitivity.py:207` and `:314` are two different tests that both call `create_topics([test_case.case_name for test_case in test_cases])`, and their case lists share names (`upper_A`, `lower_b_into_upper_B`, `pair_D_d`, `pair_E_f`, `unicode`). Same `name_salt` ⇒ identical salted topic names. The fixture is function-scoped so teardown deletes them, but Kafka topic deletion is asynchronous broker-side even after `deleteTopic`'s `f.result()` returns — so the second test recreates a topic the broker still has:

```
cimpl.KafkaException: KafkaError{code=TOPIC_ALREADY_EXISTS,val=36,str="Topic 'upper_A_KCBA1T2_V3' already exists."}
/usr/local/lib/python3.13/concurrent/futures/_base.py:402: KafkaException

FAILED tests/compatibility/test_compatibility_case_sensitivity.py::test_case_sensitivity_schema_evolution[v3]
```

Previously that error was discarded, so the collision was invisible — and worse, the two tests could silently share a topic, meaning the second could read the first's records whenever deletion lagged. So this is your change doing exactly what it's supposed to do.

I'd argue against simply catching `TOPIC_ALREADY_EXISTS` to get green, since reusing a not-yet-deleted topic can carry the previous test's data into the next test. Giving the two tests distinct case-name prefixes (or adding a per-test discriminator to the salt) fixes the isolation bug properly. What do you think?

Two smaller things:

- There's no explicit timeout on `future.result()`, so it rides librdkafka's ~60s default. With `max_workers=10` and 200 topics that's potentially many minutes stalled inside the fixture with no outer backstop — worth passing an explicit `request_timeout` / `operation_timeout`?
- The `stress` lane where `test_pressure_init` runs is skipped on PRs, so there's no CI evidence here that the original failure is actually fixed — and the description notes this doesn't by itself raise creation throughput. Could you kick a `workflow_dispatch` with `run_stress: true` and link it?

And as a follow-up rather than something for this PR: `test/lib/driver.py:217` `createPartitions` has the same log-and-swallow shape (it awaits, then logs the exception and continues), and ~36 test files call `driver.createTopics(...)` directly and discard the returned futures, so those call sites stay as fire-and-forget as before.

PR #1530, opened 2026-07-22
ci: cache protobuf builder image via buildx GitHub Actions cache

sfc-gh-skurella:
The protobuf builder image (`test/docker/Dockerfile.builder`) is rebuilt from scratch on every e2e job: `run_tests.sh` runs `docker build -t protobuf-builder`, which re-runs `apt-get` (Ubuntu mirrors), a `protoc` download, a `git clone`, and two `mvn` builds each time. Those external fetches flake — e.g. a run failed when `apt-get` couldn't reach `archive.ubuntu.com` ([core_legacy apache 2.8.2](https://github.com/snowflakedb/snowflake-kafka-connector/actions/runs/29935154653/job/88974704072)).

## Change

The `run-e2e-tests` action prebuilds `protobuf-builder` with `docker/build-push-action` using a GitHub Actions layer cache (`cache-from`/`cache-to: type=gha, scope=protobuf-builder`, `load: true`), and `run_tests.sh` reuses the loaded image (gated by `PROTOBUF_BUILDER_PREBUILT` so local runs still build as before). On a warm cache the apt/protoc/git/maven layers are restored instead of re-fetched, removing those flaky external dependencies from the hot path.

Notes on behavior:
- The **first** build on a branch (cold cache) still builds the full image and populates the cache; the benefit applies to **subsequent** builds on that branch.
- Cache busts only when `Dockerfile.builder`, `sensor.proto`, `pom.xml`, or the `maven:3.9-eclipse-temurin-11` base digest change.
- A single shared `scope=protobuf-builder` is used since the image is identical across all matrix jobs/platforms.

Not a connector code change.

### DIFF
--- a/.github/actions/run-e2e-tests/action.yml
+++ b/.github/actions/run-e2e-tests/action.yml
@@ -35,10 +35,30 @@ inputs:
     description: "Enable JMX metrics scraping via the Jolokia javaagent (sets KAFKA_OPTS so the connector JVM exposes MBeans on port 8778)"
     required: false
     default: 'false'
+  dockerhub-username:
+    description: "Docker Hub username for authenticated image pulls (avoids anonymous pull rate limits on confluentinc/* and the maven base image). Login is skipped when empty."
+    required: false
+    default: ''
+  dockerhub-token:
+    description: "Docker Hub access token/password paired with dockerhub-username."
+    required: false
+    default: ''
 
 runs:
   using: composite
   steps:
+    - name: Log in to Docker Hub (avoids anonymous pull rate limits)
+      # Both platforms pull from Docker Hub (the maven base image for the
+      # protobuf builder, and confluentinc/cp-* for the Confluent stack).
+      # Anonymous pulls share the runner egress IP's rate limit and surface as
+      # "unauthorized: authentication required". Authenticating avoids that.
+      # Skipped (no-op) when the DOCKERHUB_* secrets are not configured.
+      if: inputs.dockerhub-username != ''
+      uses: docker/login-action@v3
+      with:
+        username: ${{ inputs.dockerhub-username }}
+        password: ${{ inputs.dockerhub-token }}
+
     - name: Log in to GHCR (for prebuilt Apache Kafka image)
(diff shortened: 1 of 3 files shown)


sfc-gh-bmikaili (approved):
LGTM. Verified the wiring: `setup-buildx-action` precedes `build-push-action`, `context: test` + `file: test/docker/Dockerfile.builder` matches the original `cd test/docker && docker build -f Dockerfile.builder ..`, and `protobuf-builder:latest` is what the script's `docker create protobuf-builder` resolves to. `PROTOC_VERSION=21.12` and `tags/v3.1.0` are both pinned, so the apt / protoc / git-clone / maven layers are genuinely cacheable instead of re-fetching on a warm cache. Action pinning matches the repo-wide convention.

The `build_and_test (AWS)` failure looks unrelated — different workflow (`IntegrationTest.yml`, which never calls this action), failing on a Connect REST timeout in `SchemaEvolutionJsonIT`.

A few non-blocking notes:

- The branch is named `...-cache-auth`, but there's no auth change in the diff — `grep -rn "dockerhub\|DOCKERHUB" .github/` comes back empty, and the only `docker/login-action` is GHCR-only, gated on `platform == 'apache'`. Was a Docker Hub login intentionally deferred to a follow-up? If so the anonymous-pull flake is still open.
- The cache-miss path is silent: when `PROTOBUF_BUILDER_PREBUILT=true` but the image isn't in the local store, it falls through to a normal `docker build` with the same log output as a local run. Would a `warn` there be worth it, so a permanently broken cache doesn't quietly degrade to always-rebuild with no CI signal?
- Both new steps are ungated, so they run for every matrix job in every calling workflow. Is `compile_protobuf_dependencies` actually needed for every platform / marker-filter combination, or could some jobs skip this?
- This writes the cache on every PR e2e job across all four e2e workflows, whereas the existing `apache-kafka-*` scopes only write on master pushes. Any concern about eviction pressure on the Maven caches under the shared 10 GB budget? I couldn't measure the exported layer size, so this may well be a non-issue.

PR #1537, opened 2026-07-27
Iceberg e2e: cloud-aware external volume (Azure managed storage, GCP graceful-skip)

sfc-gh-bmikaili:
## Problem
Iceberg e2e tests (`@pytest.mark.iceberg`) pass on AWS but **failed on the Azure and GCP CI accounts**. Managed-Iceberg tables need an *external volume*, and `kafka_push_e2e_volume_aws` only exists on the AWS account. The connector auto-create tests (from #1487) drive `CREATE ICEBERG TABLE ... EXTERNAL_VOLUME=...`; on Azure/GCP that volume doesn't exist so the create errored.

There was already a skip-probe (`iceberg_external_volume`), but it only gated tests using the `create_iceberg_table` fixture. The auto-create tests don't use that fixture, so they **bypassed the probe and failed instead of skipping** — that bypass is the core defect.

## Approach
Use Snowflake-provided storage (`EXTERNAL_VOLUME='SNOWFLAKE_MANAGED'` — no customer bucket/IAM/trust). Verified empirically against the real accounts: **Azure supports it; GCP does not**.

1. **`run-e2e-tests` action** — select `ICEBERG_EXTERNAL_VOLUME` per cloud: AWS keeps its volume, all others use `SNOWFLAKE_MANAGED`.
2. **`conftest.py`** — autouse `_require_iceberg_volume` gates **every** `@pytest.mark.iceberg` test, so auto-create tests can no longer bypass the probe.
3. **`table.py`** — Iceberg is skipped **deterministically** on clouds known to lack a usable external volume (`ICEBERG_UNSUPPORTED_CLOUDS` = `{GCP}`, read from `--cloud`). On every other cloud the volume MUST work: the probe creates/drops a table and any failure **fails loudly** (never swallowed into a skip), so a real regression on AWS/Azure can't hide as green-by-skip. `IcebergTable.create()` omits `BASE_LOCATION` for managed storage (which rejects it). Probe uses create/drop rather than `DESC` (role has USAGE not OWNERSHIP; `SNOWFLAKE_MANAGED` isn't describable).

Also pins `ruff==0.15.0` in the formatting workflow: an unpinned install had jumped to 0.16.0, which reddens `formatting/python` tree-wide (170+ pre-existing violations, unrelated to this change).

## Validation
Full iceberg e2e ran **green against Azure managed storage: 20 passed, 7 skipped, 0 failed** — proving connector auto-create + streaming ingest into a `SNOWFLAKE_MANAGED` table. The 7 skips are Avro tests (the `apache` test platform lacks a schema registry) + 2 pre-existing conditional skips; CI's Azure entry uses `confluent`, which exercises Avro. Reviewed by two independent agents; findings applied.

**Manual CI dispatch** (functional=AWS + core_legacy=GCP/Azure), per review: https://github.com/snowflakedb/snowflake-kafka-connector/actions/runs/30357915811

## Follow-ups
- **GCP Iceberg coverage** is tracked in **SNOW-3851561**: `SNOWFLAKE_MANAGED` isn't available on GCP and a GCS external volume is blocked by a GCP domain-restricted-sharing org policy (ProdSec-gated; see the 2024 #team-cloud thread). Until resolved, GCP Iceberg tests skip cleanly.

Related: #1487 (SNOW-3161094).

### DIFF
--- a/test/lib/fixtures/table.py
+++ b/test/lib/fixtures/table.py
@@ -84,15 +84,21 @@ class IcebergTable(Table):
 
     The iceberg-specific clauses (``EXTERNAL_VOLUME``, ``CATALOG``,
     ``BASE_LOCATION``, ``ICEBERG_VERSION``) are appended automatically.
+
+    Note: Snowflake-managed storage (``EXTERNAL_VOLUME='SNOWFLAKE_MANAGED'``)
+    rejects ``BASE_LOCATION`` ("not supported for ... Snowflake Managed
+    Storage"), so it is omitted in that case — Snowflake constructs the path.
     """
 
     def create(self, columns: str):
+        managed = ICEBERG_EXTERNAL_VOLUME.upper() == "SNOWFLAKE_MANAGED"
+        base_location = "" if managed else f"BASE_LOCATION = '{self.name}' "
         self.driver.snowflake_conn.cursor().execute(
             f"CREATE OR REPLACE ICEBERG TABLE {quote_name(self.name)} "
             f"{columns} "
             f"EXTERNAL_VOLUME = '{ICEBERG_EXTERNAL_VOLUME}' "
             f"CATALOG = 'SNOWFLAKE' "
-            f"BASE_LOCATION = '{self.name}' "
+            f"{base_location}"
             f"ICEBERG_VERSION = 3"
         )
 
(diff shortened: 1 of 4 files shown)


test/lib/fixtures/table.py
```diff
@@ -104,34 +110,55 @@ def drop(self):
...
-    )
+        pytest.fail(
+            f"Iceberg external volume '{ICEBERG_EXTERNAL_VOLUME}' probe failed "
+            f"unexpectedly (not a 'volume unavailable' error): {e}"
+        )
+    finally:
+        try:
+            probe.drop()
+        except Exception:
+            pass
```
sfc-gh-skurella, line 152: We should ideally not silently skip a test in case of exceptions. This can lead to us being blind to a real issue. Could you please instead:
- Add a `skip_if` clause to the test conditioned on the Snowflake cloud being GCP? I think this information is already passed to pytest.
- Make this fail loudly if the volume probe fails?

sfc-gh-skurella (approved):
Thanks! Approving in advance but please see my comment and let's discuss tomorrow if you have a different opinion. Also please make sure to run the tests which are failing on master via a manual dispatch and paste a link in the PR description.

test/lib/fixtures/table.py
sfc-gh-bmikaili: Done in 11568b01 — agreed, inferring the skip from the probe's error string could've masked a real volume regression on AWS/Azure as a silent skip.

Now: iceberg skips **deterministically** only on `ICEBERG_UNSUPPORTED_CLOUDS = {"GCP"}` (from the `--cloud` option you mentioned), and on every other cloud any probe failure just propagates and **fails loudly** — nothing is swallowed into a skip. I kept it in the shared fixture rather than a per-test `skipif` so all `@pytest.mark.iceberg` tests (incl. the auto-create ones that don't use `create_iceberg_table`) are covered in one place.

One trade-off worth flagging: when `SNOWFLAKE_MANAGED` eventually lands on GCP, someone has to drop `GCP` from that set — but it's explicit/greppable and tracked in SNOW-3851561. Happy to switch to a pure probe-based auto-detect instead if you'd prefer self-healing over fail-loud — let's chat tomorrow.

Manual dispatch (functional=AWS + core_legacy=GCP/Azure) is linked in the PR description: https://github.com/snowflakedb/snowflake-kafka-connector/actions/runs/30357915811. Heads-up: the Azure/GCP `core_legacy` jobs may still red on **auth** — the gpg-encrypted CI profile keys look stale (JWT only worked with the current key). That's independent of this change; I'll flag it separately.

sfc-gh-bmikaili:
Manual dispatch is **green on all clouds** — run 30357915811:

- **AZURE** (confluent 6.2.15): `SNOWFLAKE_MANAGED is usable` → **84 passed, 4 skipped** — full core suite incl. Avro ran on managed storage. ✅
- **GCP** (apache 3.9.2): iceberg tests **skipped** cleanly (15 skipped). ✅
- **AWS** (apache + confluent) and all functional groups: ✅

Also: no auth issues in CI, so please disregard my earlier stale-gpg-key heads-up — it doesn't affect CI (removed it from the description).

PR #1013, opened 2024-12-02
SNOW-1831140 E2e tests of Iceberg JSON ingestion

sfc-gh-mbobowski:
# Overview

SNOW-1831140

A solid contribution to iceberg JSON e2e's + some minor bugs fixes.

## Pre-review checklist
- [ ] This change should be part of a Behavior Change Release. See [go/behavior-change](http://go/behavior-change).
- [ ] This change has passed Merge gate tests
- [ ] Snowpipe Changes
- [ ] Snowpipe Streaming Changes
- [ ] This change is TEST-ONLY
- [ ] This change is README/Javadocs only
- [ ] This change is protected by a config parameter <PARAMETER_NAME> eg `snowflake.ingestion.method`.
    - [ ] `Yes` - Added end to end and Unit Tests. 
    - [ ] `No` - Suggest why it is not param protected
- [ ] Is his change protected by parameter <PARAMETER_NAME> on the server side?
    - [ ] The parameter/feature is not yet active in production (partial rollout or PrPr, see [Changes for Unreleased Features and Fixes](http://go/ppp-prpr)).
    - [ ] If there is an issue, it can be safely mitigated by turning the parameter off. This is also verified by a test (See [go/ppp](http://go/ppp)).

PR #1541, opened 2026-08-18
SNOW-3961901: detect authorization failures structurally; fix OAuth credential-rejection blind spot

sfc-gh-bschiefer:
**Independent of the SPCS chain (PRs 2–5).** No shared files; reviewable in either order.

## Why

`SnowflakeSinkTaskAuthorizationExceptionTracker` exists to catch a credential
rejection during `preCommit()` (where Kafka Connect swallows exceptions) and fail
the task on the next `put()`.  Without it, a task whose credential has been revoked
keeps running and silently makes no progress.

**It has not fired since Nov 2025.** #1227 removed the only dependency that
produced the string this class matched against, silently making the mechanism
inert.  The issue was not noticed because nothing points at the tracker.

This PR restores it in two commits and extends it to cover OAuth credential
rejection, which the original code predates.

## Two commits

| # | What |
|---|---|
| 1 | Walk the cause chain; match SQLState class `28`; keep message fragments as a fallback; fix a latent NPE on a null exception message |
| 2 | Also match vendor codes `390303` / `390309` — without these, commit 1 never fires for OAuth because Snowflake returns `08001`, not class `28`, for a rejected token |

Commit 2 (+28 production lines) is where the behavior change lives.

## Measured, not inferred

Real credential rejections were provoked against a live account.  Every case comes
back `08001` (generic connection class) rather than class `28`:

```
garbage token                    08001 / 390303   "Invalid OAuth access token."
empty token                      08001 / 390303   "Invalid OAuth access token."
well-formed token, bad signature 08001 / 390303   "Invalid OAuth access token."
valid token, mismatched user     08001 / 390309   "The user ... differs from the user ..."
```

The `SQLException` object from a live connection was passed directly to
`isAuthorizationFailure`; after commit 1 alone it returns **false**, after commit 2
it returns **true**.

## Deliberately narrow

Only `390303` and `390309` are fatal; a generic `08001` (transient network fault)
stays retryable.  There is a test asserting exactly that.  The default opt-in flag
(`snowflake.enable.task.fail.on.authorization.errors`) remains `false`.

## Tests

803 → **807** (net +4) across the two commits, from a clean build, zero failures.
Format-clean. Zero findings from `sf ai review run --base-ref master --reviewer agent`.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
+++ b/src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
@@ -8,6 +8,7 @@ import com.google.common.annotations.VisibleForTesting;
 import java.sql.SQLException;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Locale;
@@ -32,9 +33,12 @@ import java.util.Set;
  *   <li>Failing that, a small set of known message fragments, matched case-insensitively.
  * </ol>
  *
- * <p>The whole cause chain is inspected, because the driver and the streaming SDK both wrap
- * authorization failures inside higher-level exceptions, and the outermost message frequently says
- * nothing about authorization.
+ * <p>The whole cause chain is inspected, because the driver wraps authorization failures inside
+ * higher-level exceptions and the outermost message frequently says nothing about authorization.
+ *
+ * <p><b>SSv2 scope note:</b> for SSv2 streaming tasks, {@code BatchOffsetFetcher} catches
+ * auth-related {@code SFException} before they reach this tracker. Auth failures on the SSv2
+ * streaming path are not currently detected here.
  */
 public class SnowflakeSinkTaskAuthorizationExceptionTracker {
 
@@ -46,12 +50,41 @@ public class SnowflakeSinkTaskAuthorizationExceptionTracker {
   @VisibleForTesting static final String SQL_STATE_INVALID_AUTHORIZATION_CLASS = "28";
 
   /**
+   * Snowflake server error codes that mean the credential itself was rejected.
+   *
+   * <p>These are needed because Snowflake does <b>not</b> report OAuth token rejection under
+   * SQLState class {@code 28}. It reports SQLState {@code 08001}, the generic connection class, and
+   * distinguishes the cause only by error code. Measured against a live deployment:
+   *
+   * <pre>
+   *   garbage token                    08001 / 390303  "Invalid OAuth access token."
+   *   empty token                      08001 / 390303  "Invalid OAuth access token."
+   *   well-formed token, bad signature 08001 / 390303  "Invalid OAuth access token."
+   *   valid token, mismatched user     08001 / 390309  "The user you were trying to authenticate
+   *                                                    as differs from the user tied to the token"
+   * </pre>
+   *
+   * <p>Without these codes a task retries indefinitely against a credential the server is actively
+   * rejecting, which is the situation this tracker exists to stop.
+   *
+   * <p>390318 ({@code OAUTH_ACCESS_TOKEN_EXPIRED}) is deliberately absent: an expired token can be
+   * refreshed automatically when the connector is configured with a refresh token or client
+   * credentials flow, so expiry is potentially self-healing. 390303 and 390309 identify credentials
+   * that are permanently invalid regardless of any refresh configuration.
+   */
+  @VisibleForTesting
+  static final Set<Integer> AUTHORIZATION_FAILURE_ERROR_CODES =
+      Collections.unmodifiableSet(new HashSet<>(Arrays.asList(390303, 390309)));
+
+  /**
    * Message fragments that indicate an authorization failure, lower-cased for case-insensitive
    * matching.
    *
-   * <p>Kept deliberately short. Every entry is either present in a shipped dependency or retained
-   * for backward compatibility; nothing is speculative. Message matching is a fallback for cases
-   * where no {@code SQLState} is available, such as failures surfaced by the streaming SDK.
+   * <p>Kept deliberately short. Every entry is verified against a shipped dependency; nothing is
+   * speculative. Message matching covers JDBC-path exceptions that carry no {@code SQLState} (for
+   * example, OAuth flow errors logged by the JDBC driver before they are wrapped in a {@code
+   * SQLException}). It does <b>not</b> cover SSv2 streaming SDK failures: the streaming SDK throws
+   * {@code SFException} which {@code BatchOffsetFetcher} catches before it can reach this tracker.
    */
   @VisibleForTesting
   static final List<String> AUTHORIZATION_FAILURE_MARKERS =
@@ -60,7 +93,9 @@ public class SnowflakeSinkTaskAuthorizationExceptionTracker {
               // Historical message from the classic Snowpipe ingest SDK. Retained so behavior is
               // unchanged for anyone relying on it.
               "authorization failed after retry",
+              // Matches snowflake-jdbc SessionUtil "OAuth Access Token Expired: {}".
               "oauth access token expired",
+              // Matches snowflake-jdbc (Google OAuth) TokenVerifier "Token is expired".
               "token is expired"));
 
   private boolean authorizationTaskFailureEnabled;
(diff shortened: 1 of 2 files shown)


src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
```diff
@@ -51,4 +119,41 @@ public void throwExceptionIfAuthorizationFailed() {
...
+   * <p>Tolerates a null message at any level, and terminates on a cyclic chain rather than looping
+   * forever.
+   */
+  @VisibleForTesting
+  static boolean isAuthorizationFailure(Throwable ex) {
+    Set<Throwable> visited = Collections.newSetFromMap(new IdentityHashMap<>());
+    for (Throwable current = ex;
+        current != null && visited.add(current);
+        current = current.getCause()) {
+      if (current instanceof SQLException) {
```
sfc-gh-skurella, line 147: `preCommit` → `getCommittedOffsets` → `BatchOffsetFetcher` catches every `SFException` from `getChannelStatus`, logs it, and skips the pipe. Nothing is rethrown, so `SnowflakeSinkTask.preCommit` never calls `reportPrecommitException`.

Even if that catch were removed, the streaming SDK surfaces 401s as `SFException` (`AuthTokenError` / `SfApiAuthError`) with no `SQLException` cause — FFI construction does not attach one. The live `SQLException(08001, 390303)` objects in the PR body are the JDBC login shape, not what this `preCommit` path throws.

Did a revoked OAuth token on an SSv2 task actually reach `isAuthorizationFailure`, or only a constructed JDBC exception in unit tests? If the SDK error never leaves `BatchOffsetFetcher`, restoring the tracker cannot fail the hanging task this class exists to catch.

-- Sent by Claude

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
```diff
@@ -4,19 +4,87 @@
...
+   * where no {@code SQLState} is available, such as failures surfaced by the streaming SDK.
+   */
+  @VisibleForTesting
+  static final List<String> AUTHORIZATION_FAILURE_MARKERS =
+      Collections.unmodifiableList(
+          Arrays.asList(
+              // Historical message from the classic Snowpipe ingest SDK. Retained so behavior is
+              // unchanged for anyone relying on it.
+              "authorization failed after retry",
+              "oauth access token expired",
```
sfc-gh-skurella, line 97: The Javadoc says these fragments cover streaming-SDK failures with no `SQLState`. The SDK emits `OAuth access token has expired` (`oauth.rs`); that does not contain `oauth access token expired` (`has` sits in the middle) or `token is expired`.

`token is expired` also does not appear in this repo or the ingest SDK (closest is `token has expired` / `The provided token has expired`). Could the list be aligned with actual SDK / `SFException.getErrorCodeName()` values, or dropped in favor of structural SDK classification?

-- Sent by Claude

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
```diff
@@ -4,19 +4,87 @@
...
+   *   valid token, mismatched user     08001 / 390309  "The user you were trying to authenticate
+   *                                                    as differs from the user tied to the token"
+   * </pre>
+   *
+   * <p>Without these codes a task retries indefinitely against a credential the server is actively
+   * rejecting, which is the situation this tracker exists to stop.
+   */
+  @VisibleForTesting
+  static final Set<Integer> AUTHORIZATION_FAILURE_ERROR_CODES =
+      Collections.unmodifiableSet(new HashSet<>(Arrays.asList(390303, 390309)));
```
sfc-gh-skurella, line 77: GS splits invalid (`390303`) from expired (`390318`, `OAUTH_ACCESS_TOKEN_EXPIRED`). The PR measured garbage / mismatched tokens, not expiry. If the driver reports `08001` + `390318` with a generic connection message, would this still fail the task, or does expiry belong in the vendor set (or stay retryable because refresh should succeed)?

-- Sent by Claude

sfc-gh-skurella (approved):
Approving since this is an improvement but posting AI review findings - there are probably details that could be improved further but that's not a blocker for merging the change

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
sfc-gh-bschiefer: Confirmed correct. BatchOffsetFetcher.fetchOffsetsAsync catches SFException explicitly (line 111 of that class) and either triggers channel recovery or logs and discards it, so auth-related SFException from getChannelStatus never reaches reportPrecommitException. This is a pre-existing architectural gap in the SSv2 path, not introduced by this PR.

The tracker was originally designed for the SSv1 / JDBC path where SnowflakeSinkTask.preCommit catches SnowflakeKafkaConnectorException (line 362) and plain Exception (line 368) thrown directly from the sink. For SSv2, BatchOffsetFetcher isolates failures inside CompletableFuture lambdas before they reach the task. Fixing that would require BatchOffsetFetcher to distinguish auth errors and rethrow them as a specific type; that is a separate and larger change.

Practical scope of this PR: restores detection for the JDBC / SSv1 path and adds structural (SQLState class-28) and OAuth error-code (390303, 390309) detection for any path that does propagate through preCommit. The SSv2 streaming SDK gap is pre-existing and acknowledged.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
sfc-gh-bschiefer: The markers ARE present in shipped dependencies, but in the JDBC driver, not the streaming SDK jar.

Verified against snowflake-jdbc-4.3.2.jar:
- "oauth access token expired" is a lowercase prefix-match of "OAuth Access Token Expired: {}" emitted by SessionUtil.class (constant pool entry #1013/1014).
- "token is expired" is the exact string from net.snowflake.client.jdbc.internal.google.auth.oauth2.TokenVerifier.class (constant pool entry #197).

The Javadoc comment "fallback for cases where no SQLState is available, such as failures surfaced by the streaming SDK" is misleading: these two markers cover the JDBC driver's OAuth path, not SFException messages from the streaming SDK. The class comment can be clarified to say they cover JDBC path messages specifically. The streaming SDK path (SSv2) swallows SFException in BatchOffsetFetcher anyway (see sibling comment above), so the fallback markers are not the primary mechanism for that path.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
sfc-gh-bschiefer: 390318 (OAUTH_ACCESS_TOKEN_EXPIRED) was deliberately excluded because an expired token is potentially recoverable: if the connector is configured with a refresh token or client credentials flow, the OAuth client can obtain a new access token automatically. Treating expiry as a permanent auth failure would kill tasks that could self-heal after a token refresh.

390303 (invalid/garbage token) and 390309 (user mismatch) are fatally unrecoverable regardless of any refresh configuration: the credential itself is wrong, not just stale. Those codes mandate human intervention.

If a deployment has no OAuth refresh configured and the access token expires, the task will retry indefinitely rather than failing. That is a valid argument for adding 390318, but it changes the behaviour for refresh-capable deployments. I would prefer to track that as a separate follow-up once we have empirical data on how the driver behaves on expiry with and without refresh configured.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
sfc-gh-bschiefer: Addressed in the amended commit: added a class-level SSv2 scope note to the Javadoc documenting that BatchOffsetFetcher swallows SFException before reaching this tracker. No behavior change — the scope limitation stands as explained above.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
sfc-gh-bschiefer: Addressed in the amended commit: updated the AUTHORIZATION_FAILURE_MARKERS Javadoc to say these cover JDBC-path exceptions (not streaming SDK), and added an inline comment on each marker naming its source class in snowflake-jdbc.

src/main/java/com/snowflake/kafka/connector/SnowflakeSinkTaskAuthorizationExceptionTracker.java
sfc-gh-bschiefer: Addressed in the amended commit: added a paragraph to the AUTHORIZATION_FAILURE_ERROR_CODES Javadoc explaining that 390318 (OAUTH_ACCESS_TOKEN_EXPIRED) is deliberately absent because expiry is potentially self-healing via OAuth refresh, while 390303 and 390309 identify credentials that are permanently invalid regardless of refresh configuration.

PR #1542, opened 2026-08-18
SNOW-3961901: [1/3] SPCS ambient auth — fail-fast guard + SpcsEnvironment (inert)

sfc-gh-bschiefer:

## Stack

This is a three-PR stack.  **No configuration produces an SPCS setup until PR [3/3]**, so these first two PRs are inert by construction and nothing can enter the new code paths yet.  Start reading here, but all review effort belongs in [3/3].

|---|---|---|---|
| **[1/3] ← you are here** | `spcs-1-fail-fast`, `spcs-2-environment` | Fail-fast guard + `SpcsEnvironment` | Yes |
| [2/3] | `spcs-3-credential-paths`, `spcs-4-resolve` | Credential paths + resolution | Yes |
| [3/3] | `spcs-5-enable`, `spcs-6-concurrency` | Entry points wired up + concurrency cover | **Live** |

Chain A (PR #1541) and Chain C (PR #1545, the blank-config fix — **merged 2026-08-19**) are independent and can be reviewed in parallel.

## Why this exists

On `master` today, the connector has zero awareness of Snowpark Container Services.
A search for `SNOWFLAKE_HOST`, `SNOWFLAKE_DATABASE`, `/snowflake/session/token` returns
no matches.  Running KC inside SPCS today requires a full credential configuration,
meaning a key pair or OAuth client stored as a secret — the same setup as on bare metal.
SPCS provides a service user credential automatically; this change teaches KC to use it.

## This PR: two preparatory commits

**Commit 1 — `fail fast on unhandled authenticator in StreamingClientProperties`** (+2 prod, +36 test)

The authenticator switch in `StreamingClientProperties.from()` fell through silently
for any unknown `AuthenticatorType`.  A `default: throw` is added.  This is a guard
that makes the next commit safe to add: a new `case SPCS` can be handled explicitly
knowing that any future enum value that is not handled will immediately fail the build
via an existing exhaustiveness test.

**Commit 2 — `add SpcsEnvironment for read-only access to the SPCS service credentials`** (+120 prod, +94 test)

Adds the `SpcsEnvironment` class.  It knows three things:
- Whether the runtime is SPCS (`isInsideSpcs()`: env var `SNOWFLAKE_HOST` present AND token file readable)
- The ambient values (`host()`, `account()`, `database()`, `schema()`)
- The ambient bearer token (`readToken()`, re-read on every call — rotation requires this)

Nothing in the connector calls any of these methods yet.  The class is deliberately
isolated from connector configuration to keep environment access separate from config
policy.

## Tests

793 → **799** across these two commits, from a clean build, zero failures.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
@@ -1,0 +1,120 @@
+package com.snowflake.kafka.connector.internal.spcs;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.snowflake.kafka.connector.internal.KCLogger;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Optional;
+import java.util.function.Function;
+
+/**
+ * Access to the credentials that Snowpark Container Services (SPCS) provides to every service
+ * container, so a connector running inside SPCS needs no credential configuration.
+ *
+ * <p>Snowflake calls these <b>Snowflake-provided service user credentials</b>, or simply <i>service
+ * credentials</i>. They are also widely referred to as <b>ambient authentication</b>, the
+ * <i>ambient service identity</i>, or just <i>SPCS ambient</i>; those are informal names for the
+ * same documented mechanism. See <a
+ * href="https://docs.snowflake.com/en/developer-guide/snowpark-container-services/spcs-execute-sql">Snowpark
+ * Container Services: SQL execution</a>.
+ *
+ * <p>When a service starts, Snowflake:
+ *
+ * <ul>
+ *   <li>writes an OAuth token to {@value #DEFAULT_TOKEN_FILE}, which authenticates as the service
+ *       user and is valid only inside that service;
+ *   <li>sets {@code SNOWFLAKE_ACCOUNT} to the account identifier and {@code SNOWFLAKE_HOST} to the
+ *       hostname to connect to. The token cannot be used without {@code SNOWFLAKE_HOST};
+ *   <li>sets {@code SNOWFLAKE_DATABASE} and {@code SNOWFLAKE_SCHEMA} to the service's own database
+ *       and schema. Note that no warehouse is provided.
+ * </ul>
+ *
+ * <p>The resulting session runs as the service user, whose only roles are the service owner role
+ * and PUBLIC, with the service owner role as its default.
+ *
+ * <p>The token is deliberately <b>never cached</b>. Snowflake rewrites the file every few minutes
+ * and each token is valid for at most an hour, so a cached value would eventually be stale. Note
+ * that, per the documentation, once a connection has been established successfully the token's
+ * expiry no longer applies to that connection, exactly as for a session a user creates directly.
+ */
+public final class SpcsEnvironment {
+
+  private static final KCLogger LOGGER = new KCLogger(SpcsEnvironment.class.getName());
+
+  /** Bearer token file managed by the SPCS runtime. */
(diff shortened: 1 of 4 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
```diff
@@ -106,6 +106,8 @@ public static StreamingClientProperties from(SinkTaskConfig config) {
             Base64.getEncoder().encodeToString(privateKey.getEncoded());
         clientProperties.put("private_key", privateKeyEncoded);
         break;
+      default:
+        throw new IllegalStateException("unhandled authenticator: " + config.getAuthenticator());
```
sfc-gh-skurella, line 110: NIT: If you don't add the `default` clause, compilation will fail at the exhaustivity check for unhandled options, which may help to shift left if some logic is missing. I'd prefer listing all options and skipping the `default`.

src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+  }
+
+  private static Optional<String> env(String name) {
+    return Optional.ofNullable(envReader.apply(name)).map(String::trim).filter(v -> !v.isEmpty());
+  }
+
+  private static boolean isBlank(String value) {
+    return value == null || value.trim().isEmpty();
+  }
```
sfc-gh-skurella, line 107: Note for reviewers: this is unused here but will be used in downstream PRs

sfc-gh-skurella (approved):
Thanks! LGTM modulo one NIT.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
sfc-gh-bschiefer: The difference matters in Java 11. Switch statements (as opposed to switch expressions added in Java 14) do not produce compile errors for unhandled enum cases — an unhandled value simply falls through with no compiler complaint. Listing all cases without a default would compile cleanly today but silently produce empty ClientProperties if a new AuthenticatorType is added later.

The default: throw ensures that adding a new authenticator type fails loudly at runtime on the first connection attempt, rather than producing a misconfigured client that tries to connect without credentials. The existing AuthenticatorTypeTest (which pins the exact set of enum values that return true from suppliesAmbientIdentity) provides the shift-left signal on the enum's structure side. We think that combination is safer than relying on exhaustive listing in a Java 11 switch statement.

src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
sfc-gh-bschiefer: Acknowledged — Seb added this note for Artem's benefit. No action needed on our side; the code does become active in PR #1544 as noted.

sfc-gh-aminyaylov: I would recommend implementing a helper in the PR where it's used, so that reviewers can improve readability based on context. Without usage, I can't really comment on this function.

sfc-gh-aminyaylov (approved):
Nice code! I like the attention to detail with `map(String::trim)`, and `@EnumSource` is a good find.

src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
sfc-gh-bschiefer: Fair point, and thanks for the approval and the kind words! The isInsideSpcs/host/account/database/schema/readToken methods on SpcsEnvironment are all consumed starting in #1544 (the resolve() call sites) — happy to point you at the specific usages there when you get to that PR, or we can revisit ordering on a future stacked PR if this comes up again.

PR #1543, opened 2026-08-18
SNOW-3961901: [2/3] SPCS ambient auth — credential paths + resolution (inert)

sfc-gh-bschiefer:

## Stack

|---|---|---|---|
| [1/3] | `spcs-1-fail-fast`, `spcs-2-environment` | Fail-fast guard + `SpcsEnvironment` | Yes |
| **[2/3] ← you are here** | `spcs-3-credential-paths`, `spcs-4-resolve` | Credential paths + resolution | Yes |
| [3/3] | `spcs-5-enable`, `spcs-6-concurrency` | Entry points wired up + concurrency cover | **Live** |

**Still inert.** No configuration produces an SPCS setup until [3/3].  All review effort belongs there.

## This PR: two commits

**Commit 3 — `teach both credential paths about an ambient identity`** (+80 -4 prod, +300 -8 test)

Adds the `SPCS` enum constant to `AuthenticatorType` and a capability method:

```java
/** True when the credential itself identifies the user, so the connector must not assert one. */
public boolean suppliesAmbientIdentity() { return this == SPCS; }
```

Why a capability method rather than `== SPCS`: the condition is "this credential supplies its own identity", which today maps only to SPCS but will also be true for workload identity federation.  Enum comparisons would require finding all three call sites manually; a missed site reintroduces the live `390309` failure (wrong user in token).  `AuthenticatorTypeTest` pins the set to exactly `{SPCS}`.

Both credential paths are taught about the new authenticator:
- **JDBC** (`InternalUtils`): sets `authenticator=oauth` and reads the ambient token.  Omits `user` — the token already identifies the service user; asserting a different name is rejected by Snowflake with `390309`.
- **Streaming SDK** (`StreamingClientProperties`): sets `authorization_type=spcs` and nothing else.  The SDK defaults the token path and owns its own file read.  Setting any `oauth_*` key here would be rejected by the SDK's exclusive-field validation (documented in #1519).

Also carries the **golden non-SPCS regression test** pinning the exact property key set for every existing authenticator, so any accidental addition of an SPCS-only key to a non-SPCS path fails immediately.

**Commit 4 — `resolve the ambient SPCS values into a configuration`** (+131 -1 prod, +265 test)

Adds `SpcsEnvironment.resolve()`.  This is the single function that fills in the
Snowflake-provided ambient values — host, account, database, schema, user placeholder —
into a configuration map.  Key design properties:

- **Pure map-to-map**: takes raw config, returns an enriched copy.  No side effects.
- **No-op outside SPCS**: first line is `if (!isInsideSpcs()) return raw`.
- **`putIfBlank` only**: never overwrites a value the operator supplied.  Explicit configuration always wins.
- **Credential guard**: if any credential is present and no authenticator is set, auto-adoption is blocked with a WARN.  A deployment with a key pair that happens to run inside SPCS keeps its key-pair auth and the operator is told what is happening.
- **Explicit authenticator wins**: any `snowflake.authenticator` that is not `spcs` causes an immediate `return raw`.

Nothing calls `resolve()` in production yet.  That is commit 5.

## Tests

799 → **807** (commit 3) → **824** (commit 4), zero failures at each step.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
@@ -7,6 +7,7 @@ import com.snowflake.kafka.connector.config.AuthenticatorType;
 import com.snowflake.kafka.connector.config.SinkTaskConfig;
 import com.snowflake.kafka.connector.internal.oauth.OAuthAccessTokenFetcher;
 import com.snowflake.kafka.connector.internal.oauth.OAuthURL;
+import com.snowflake.kafka.connector.internal.spcs.SpcsEnvironment;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.sql.ResultSet;
@@ -92,7 +93,16 @@ public class InternalUtils {
 
     putIfNotBlank(properties, JDBC_DATABASE, config.getSnowflakeDatabase());
     putIfNotBlank(properties, JDBC_SCHEMA, config.getSnowflakeSchema());
-    putIfNotBlank(properties, JDBC_USER, config.getSnowflakeUser());
+    if (!config.getAuthenticator().suppliesAmbientIdentity()) {
+      // Deliberately NOT set when the credential identifies the user itself. Snowflake rejects a
+      // session whose asserted user differs from the token's subject:
+      //   The user you were trying to authenticate as differs from the user tied to the
+      //   access token.
+      // The connector has no way to know the generated service user's name, so it must not
+      // claim one. Verified against a live SPCS service: sending the placeholder user fails,
+      // omitting the user succeeds.
+      putIfNotBlank(properties, JDBC_USER, config.getSnowflakeUser());
+    }
     putIfNotBlank(properties, JdbcPropertyKeys.ROLE, config.getSnowflakeRole());
 
     switch (config.getAuthenticator()) {
@@ -125,6 +135,15 @@ public class InternalUtils {
                     .orElseThrow(SnowflakeErrors.ERROR_0013::getException),
                 config.getSnowflakePrivateKeyPassphrase()));
         break;
+      case SPCS:
+        // Snowflake-provided service user credentials ("ambient" authentication). The driver has
+        // no SPCS mode: as the SPCS documentation prescribes, the container connects with
+        // authenticator=oauth and the token from the file Snowflake manages. The token is read
+        // here, while properties are assembled, so every new connection uses the current one --
+        // Snowflake rewrites the file every few minutes.
+        properties.put(JdbcPropertyKeys.AUTHENTICATOR, AuthenticatorType.OAUTH.toConfigValue());
+        properties.put(JDBC_TOKEN, SpcsEnvironment.readToken());
+        break;
       default:
         throw new IllegalStateException("unhandled authenticator: " + config.getAuthenticator());
     }
@@ -142,7 +161,10 @@ public class InternalUtils {
     if (!properties.containsKey(JDBC_DATABASE)) {
       throw SnowflakeErrors.ERROR_0015.getException();
     }
-    if (!properties.containsKey(JDBC_USER)) {
+    // Not required when the credential supplies the identity: the user is deliberately omitted
+    // above, because the token identifies the user and asserting any other one is rejected.
+    if (!config.getAuthenticator().suppliesAmbientIdentity()
+        && !properties.containsKey(JDBC_USER)) {
       throw SnowflakeErrors.ERROR_0016.getException();
     }
 
--- a/src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
@@ -1,12 +1,16 @@
 package com.snowflake.kafka.connector.internal.spcs;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.snowflake.kafka.connector.Constants.KafkaConnectorConfigParams;
+import com.snowflake.kafka.connector.config.AuthenticatorType;
 import com.snowflake.kafka.connector.internal.KCLogger;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
 
@@ -52,6 +56,24 @@ public final class SpcsEnvironment {
   static final String ENV_DATABASE = "SNOWFLAKE_DATABASE";
   static final String ENV_SCHEMA = "SNOWFLAKE_SCHEMA";
 
+  /**
+   * Placeholder written to {@code snowflake.user.name} in ambient mode. The effective identity is
+   * carried by the token rather than by this value, but several code paths require a non-blank
+   * user.
+   *
+   * <p>Not the real service user name. Snowflake names the service user after the service itself,
+   * or {@code SF$SERVICE$<unique-id>} for services created before the 8.35 release, which this
+   * class has no reliable way to discover from inside the container.
+   *
+   * <p><b>This value must never be sent to Snowflake.</b> It exists only to satisfy the connector's
+   * own requirement that a user name be configured. The ambient token already identifies the
+   * service user, and a session that asserts a different user is rejected with "The user you were
+   * trying to authenticate as differs from the user tied to the access token". Both {@code
+   * InternalUtils.makeJdbcDriverProperties} and {@code StreamingClientProperties} therefore omit
+   * the user entirely when the authenticator is {@code spcs}. Verified against a live SPCS service.
+   */
+  public static final String AMBIENT_USER_PLACEHOLDER = "spcs-service-user";
+
   /** Environment lookup. Overridable so tests can simulate an SPCS container. */
   private static Function<String, String> envReader = System::getenv;
 
@@ -86,8 +108,9 @@ public final class SpcsEnvironment {
    * @throws IllegalStateException if the token file cannot be read
    */
   public static String readToken() {
+    String token;
     try {
-      return new String(Files.readAllBytes(tokenPath), StandardCharsets.UTF_8).trim();
+      token = new String(Files.readAllBytes(tokenPath), StandardCharsets.UTF_8).trim();
     } catch (IOException e) {
       throw new IllegalStateException(
           "Failed to read the SPCS session token at "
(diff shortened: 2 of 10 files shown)


src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
```diff
@@ -52,6 +56,24 @@ public final class SpcsEnvironment {
...
+   * class has no reliable way to discover from inside the container.
+   *
+   * <p><b>This value must never be sent to Snowflake.</b> It exists only to satisfy the connector's
+   * own requirement that a user name be configured. The ambient token already identifies the
+   * service user, and a session that asserts a different user is rejected with "The user you were
+   * trying to authenticate as differs from the user tied to the access token". Both {@code
+   * InternalUtils.makeJdbcDriverProperties} and {@code StreamingClientProperties} therefore omit
+   * the user entirely when the authenticator is {@code spcs}. Verified against a live SPCS service.
+   */
+  public static final String AMBIENT_USER_PLACEHOLDER = "spcs-service-user";
```
sfc-gh-skurella, line 75: Would adopting the code paths which currently (think they) need a user name add a lot of bloat? Treating the user name as optional and validating it together with `suppliesAmbientIdentity` might actually help understand the ambient auth flow

src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
```diff
@@ -96,6 +119,113 @@ public static String readToken() {
...
+      // credential, which is a local, inspectable condition.
+      throw new IllegalStateException(
+          "The SPCS session token at "
+              + tokenPath
+              + " is empty. Snowflake rejects an empty bearer token with '390303 Invalid OAuth"
+              + " access token', so the connection cannot succeed. The file is written by the"
+              + " SPCS runtime and is normally a JWT of several hundred bytes; an empty file"
+              + " means the runtime did not supply a credential. Check that the service"
+              + " specification sets capabilities.securityContext.enableCustomCredentials: true"
+              + " and that the token volume is mounted.");
```
sfc-gh-skurella: NIT: I'd replace

```java
              "Snowflake rejects an empty bearer token with '390303 Invalid OAuth"
              + " access token', so the connection cannot succeed. The file is written by the"
              + " SPCS runtime and is normally a JWT of several hundred bytes; an empty file"
              + " means the runtime did not supply a credential. 
```

with just "The SPCS runtime did not supply the credential." - the extra details are great in a code comment but might be TMI for an end user.

sfc-gh-skurella (approved):
Thanks! Approving in advance.

I'd ideally like to not need the `AMBIENT_USER_PLACEHOLDER` - it might improve the overall code quality but might also be too much bloat.

Rewriting the raw config is a bit hacky but I imagine doing it from within the `SinkTaskConfig` might constitute a larger refactor so I'm OK with keeping this - however that's still an avenue I'd explore; maybe turns out relatively straightforward. I tried to remove most uses of the raw config map across the codebase in favor of `SinkTaskConfig` which has stronger typing and better semantics.

src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
sfc-gh-bschiefer: We investigated removing the placeholder. The user-blank check in DefaultConnectorConfigValidator (line 130) fires after resolve() injects the placeholder, so the placeholder neatly satisfies that check with no changes to existing validation logic. Removing it would require guarding at least two locations: the validator's user-blank check and the Utils.loginPropList / isSingleFieldValid path. Making getSnowflakeUser() return Optional would touch the abstract method signature and all call sites — the larger refactor you mentioned.

We agree it would be cleaner long term, and Artem as the future owner is a natural person to drive that. Filed as a follow-up: make snowflake.user.name optional when suppliesAmbientIdentity() is true, guarding the validator and any other call sites, rather than injecting a placeholder that is never sent to Snowflake.

src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
sfc-gh-bschiefer: Done — shortened to: "The SPCS session token at <path> is empty. Check that the service specification sets capabilities.securityContext.enableCustomCredentials: true — without it the server returns 390303 Invalid OAuth access token." The two test assertions (390303 and enableCustomCredentials) still pass; the surrounding prose is removed.

src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
```diff
@@ -96,6 +119,110 @@ public static String readToken() {
...
+      //   390303 Invalid OAuth access token.
+      // which is accurate but says nothing about which file was empty or why, and arrives
+      // only after a network round trip. An empty token means the platform did not supply a
+      // credential, which is a local, inspectable condition.
+      throw new IllegalStateException(
+          "The SPCS session token at "
+              + tokenPath
+              + " is empty. Check that the service specification sets"
+              + " capabilities.securityContext.enableCustomCredentials: true"
+              + " — without it the server returns 390303 Invalid OAuth access token.");
```
sfc-gh-aminyaylov, line 133: This message is confusing to me:
1. "session token is empty" — makes sense, customer needs to provide a token
2. "spec sets `enableCustomCredentials`" — an alternative to setting a token, or in addition to?
3. "without it the server returns 390303" — customer doesn't need to know this, I would remove

src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
```diff
@@ -96,6 +119,110 @@ public static String readToken() {
               + " available to a connector running inside Snowpark Container Services.",
           e);
     }
+    if (token.isEmpty()) {
+      // Fail here rather than sending an empty bearer token. Snowflake rejects it with
+      //   390303 Invalid OAuth access token.
+      // which is accurate but says nothing about which file was empty or why, and arrives
+      // only after a network round trip. An empty token means the platform did not supply a
+      // credential, which is a local, inspectable condition.
```
sfc-gh-aminyaylov, line 127: Very wordy. I think the first sentence is perfectly sufficient.

Better yet, logs and error messages are self-documenting. Make the message below clear and remove the comment, as it's redundant.

src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
sfc-gh-aminyaylov: I think this is fine for now. IIUC, this can be changed in the future without breaking clients.

src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
```diff
@@ -96,6 +119,110 @@ public static String readToken() {
...
+  /**
+   * Fills in the values Snowflake provides to a service container, so the rest of the connector
+   * sees an ordinary, complete configuration.
+   *
+   * <p>This is the single entry point for resolving the Snowflake-provided service user
+   * credentials. It is a no-op outside SPCS and never overwrites a value the user supplied, so
+   * explicit configuration always wins. It is also idempotent, so applying it at more than one
+   * layer is safe.
+   *
+   * @param raw raw connector configuration
```
sfc-gh-aminyaylov, line 147: Nit: "raw raw" — consider renaming param `config` or `rawConfig` or `rawConfigMap` for clarity

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
```diff
@@ -106,11 +106,26 @@ public static StreamingClientProperties from(SinkTaskConfig config) {
             Base64.getEncoder().encodeToString(privateKey.getEncoded());
         clientProperties.put("private_key", privateKeyEncoded);
         break;
+      case SPCS:
+        // Snowflake-provided service user credentials ("ambient" authentication). Deliberately
+        // sets no spcs_* keys: the SDK already defaults spcs_token_path and
+        // spcs_service_token_path to the standard SPCS locations, and its exclusive-field check
+        // treats them as user-set only when they differ from those defaults. Leaving them alone
+        // also keeps the connector clear of the config-key injection path tracked in
+        // SNOW-3686036. Note no oauth_* key may be set here, including oauth_include_scope, which
```
sfc-gh-aminyaylov, line 115: This comment needs its own comment. It reminds my of my Intro to Logic course in college.

> Note no oauth_* key may be set here, including oauth_include_scope, which the SDK counts as set when false.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
```diff
@@ -106,11 +106,26 @@ public static StreamingClientProperties from(SinkTaskConfig config) {
...
+        break;
       default:
         throw new IllegalStateException("unhandled authenticator: " + config.getAuthenticator());
     }
 
-    clientProperties.put("user", config.getSnowflakeUser());
+    if (!config.getAuthenticator().suppliesAmbientIdentity()) {
+      // See the note in InternalUtils.makeJdbcDriverProperties: when the credential itself
+      // identifies the user, asserting any other user is rejected. The connector cannot know
+      // the generated service user's name, so it must not send one.
```
sfc-gh-aminyaylov, line 126: I don't love that this sends me to go hunting for another comment. I feel like this doesn't even need a comment — it's business logic.

When I read this, I understand: "Okay, so when KC uses ambient identity, we don't pass user. Got it."

If I want to learn more, I can go read the PR that introduced it, or read public docs.

src/main/java/com/snowflake/kafka/connector/internal/InternalUtils.java
```diff
@@ -92,7 +93,16 @@ static Properties makeJdbcDriverProperties(SinkTaskConfig config, SnowflakeURL u
 
     putIfNotBlank(properties, JDBC_DATABASE, config.getSnowflakeDatabase());
     putIfNotBlank(properties, JDBC_SCHEMA, config.getSnowflakeSchema());
-    putIfNotBlank(properties, JDBC_USER, config.getSnowflakeUser());
+    if (!config.getAuthenticator().suppliesAmbientIdentity()) {
+      // Deliberately NOT set when the credential identifies the user itself. Snowflake rejects a
```
sfc-gh-aminyaylov, line 97: Likewise, I recommend removing this comment. It just narrates the code and talks about how it was tested (this detail belongs in the PR description).

src/test/java/com/snowflake/kafka/connector/config/AuthenticatorTypeTest.java
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * The credential-based authenticators must keep asserting a user. Written as a parameterised test
+   * over the enum so that adding a constant without considering this property fails here rather
+   * than at a call site.
+   */
+  @ParameterizedTest
+  @EnumSource(
+      value = AuthenticatorType.class,
+      names = {"SNOWFLAKE_JWT", "OAUTH"})
+  void credentialAuthenticatorsDoNotSupplyAmbientIdentity(AuthenticatorType authenticator) {
```
sfc-gh-aminyaylov, line 58: This looks like a repeat of the above test, but it only tests two explicit types. Please remove (unless I'm missing something).

src/test/java/com/snowflake/kafka/connector/config/SnowflakeSinkConnectorConfigBuilder.java
```diff
@@ -51,6 +51,17 @@ public SnowflakeSinkConnectorConfigBuilder withUrl(String url) {
     return this;
   }
 
+  /**
+   * Removes the preset URL, so that a test can assert on a URL supplied from elsewhere. Needed by
+   * the ambient SPCS tests: {@code SpcsEnvironment.resolve()} fills the URL from {@code
+   * SNOWFLAKE_HOST} only when the configured value is blank, so a test that leaves the preset URL
+   * in place is asserting on its own input rather than on ambient resolution.
+   */
+  public SnowflakeSinkConnectorConfigBuilder withoutUrl() {
```
sfc-gh-aminyaylov, line 60: Please narrow the scope of this test method:

- Use `package-private` (no `public` keyword)
- Add`@VisibleForTesting` for clarity

src/test/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironmentTest.java
```diff
@@ -91,4 +95,265 @@ void shouldFailWithActionableMessageWhenTokenUnreadable() {
...
+  @Test
+  void shouldNotAdoptAmbientAuthWhenAKeyPairCredentialIsConfigured() throws IOException {
+    simulateSpcs("tok");
+    Map<String, String> raw = new HashMap<>();
+    raw.put(KafkaConnectorConfigParams.SNOWFLAKE_PRIVATE_KEY, "a-private-key");
+
+    Map<String, String> resolved = SpcsEnvironment.resolve(raw);
+
+    assertThat(resolved).isSameAs(raw);
+    assertThat(resolved)
```
sfc-gh-aminyaylov, line 185: This statement is a subset of the `.isSameAs()` test above. Please remove.

If you want to verify immutability of the input map, consider a separate test case.

src/test/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironmentTest.java
```diff
@@ -91,4 +95,265 @@ void shouldFailWithActionableMessageWhenTokenUnreadable() {
...
+  void shouldNotAdoptAmbientAuthWhenAnOauthCredentialIsConfigured(String credentialKey)
+      throws IOException {
+    simulateSpcs("tok");
+    Map<String, String> raw = new HashMap<>();
+    raw.put(credentialKey, "a-value");
+
+    Map<String, String> resolved = SpcsEnvironment.resolve(raw);
+
+    assertThat(resolved).isSameAs(raw);
+    assertThat(resolved).doesNotContainKey(KafkaConnectorConfigParams.SNOWFLAKE_AUTHENTICATOR);
```
sfc-gh-aminyaylov, line 209: Likewise, the `.isSameAs()` check above is sufficient.

src/test/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironmentTest.java
```diff
@@ -91,4 +95,265 @@ void shouldFailWithActionableMessageWhenTokenUnreadable() {
...
+
+  /**
+   * {@code resolve()} compares the configured authenticator with {@code equalsIgnoreCase}, so a
+   * user who writes {@code SPCS} or {@code Spcs} must get ambient resolution just as one who writes
+   * {@code spcs} does. Kafka Connect passes configuration through verbatim, so the casing is
+   * whatever the operator typed.
+   */
+  @Test
+  void shouldTreatExplicitSpcsAuthenticatorCaseInsensitively() throws IOException {
+    for (String spelling : new String[] {"spcs", "SPCS", "Spcs", " spcs "}) {
```
sfc-gh-aminyaylov, line 288: Can you do this with a parameterized test, so that `SpcsEnvironment.resetForTests()` (`@AfterEach`) is implicit?

sfc-gh-aminyaylov (approved):
Nice code. Thanks!

src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
sfc-gh-aminyaylov: Although in `PR 3/3` I'm seeing:

>  .as("message must name the Snowflake error so the operator knows what Snowflake would say")
>  .hasMessageContaining("390303")

So maybe I'm just not understanding our users...

PR #1544, opened 2026-08-18
SNOW-3961901: [3/3] SPCS ambient auth — wire entry points; concurrency cover

sfc-gh-bschiefer:

## Stack

|---|---|---|---|
| [1/3] | `spcs-1-fail-fast`, `spcs-2-environment` | Fail-fast guard + `SpcsEnvironment` | Yes |
| [2/3] | `spcs-3-credential-paths`, `spcs-4-resolve` | Credential paths + resolution | Yes |
| **[3/3] ← you are here** | `spcs-5-enable`, `spcs-6-concurrency` | Entry points wired up + concurrency cover | **This is where the behavior change lives** |

**This is the review entry point for the whole chain.** PRs [1/3] and [2/3] are prerequisites; their design rationale is in their descriptions, but they are inert and low-risk.  All behavioral change is here.

## What this does

Lets the connector run inside Snowpark Container Services with **no configured
credential** — no key pair, no OAuth client, no secret — using the Snowflake-provided
service user identity.  Set `snowflake.authenticator=spcs` (or omit it entirely
inside SPCS), and the connector authenticates as the service user automatically.

Existing deployments are not affected:
- **Outside SPCS**: `SpcsEnvironment.isInsideSpcs()` is false (requires both `SNOWFLAKE_HOST` env var and a readable `/snowflake/session/token`); `resolve()` is a no-op on its first line.
- **Inside SPCS with an explicit credential**: auto-adoption is blocked; the existing credential is kept.
- **Inside SPCS with an explicit authenticator**: `resolve()` returns the input unchanged.

## Commit 5 — `enable ambient SPCS authentication at the entry points`

Calls `resolve()` at all four places in the connector that construct or validate a configuration:
`SnowflakeStreamingSinkConnector.start()`, `SnowflakeStreamingSinkConnector.validate()`,
`DefaultConnectorConfigValidator.validateConfig()`, and `SinkTaskConfig.builderFrom()`.

Also adds `spcs` to `ConnectorConfigDefinition` so Kafka Connect surfaces it in its
REST API documentation, and lets `DefaultConnectorConfigValidator` accept it.

**One finding that only the broker run could have revealed:** Kafka Connect calls `Connector.validate()` *before* `start()`, so without the resolution call in `validate()` the feature was unusable from Connect even though every in-process test passed.  The fix is one line.

## Commit 6 — `add concurrency cover for ambient SPCS authentication` (test only)

Adds `SpcsEnvironmentConcurrencyTest`:
- A rotating token (background thread rewriting the file every 5 ms) is always read whole, never partially.
- `resolve()` applied concurrently from 8 threads produces identical results — no shared mutable state.
- `isInsideSpcs()` never disagrees across concurrent calls.

Zero production code.

## Two deployment prerequisites found by deploying, not by reading code

1. **`capabilities.securityContext.enableCustomCredentials: true`** in the service spec.
   Without it, JDBC connects fine but the streaming SDK returns `401 / 395090 "Please use OAuth
   when connecting to Snowflake"`.  The capability name is misleading — it gates the
   *platform-supplied* token, not a customer secret — and the error message points the wrong way.

2. **`snowflake.streaming.validate.compatibility.with.classic=false`** (or the five companion
   settings it requires).

## Verified end-to-end

Connector's own classes deployed as an SPCS service.  `validateConfig()` passed with no
credential.  Then a MakeBenchmark2-shaped run (Kafka Connect distributed, REST API, 4 tasks,
2000 records, 600 s):

| Check | Result |
|---|---|
| REST submit | HTTP 201 |
| All four tasks | RUNNING |
| Rows / distinct ids / duplicates | **2000 / 2000 / 0** |
| Token rotation mid-run (~10.5 min interval) | Survived; 0 auth errors |
| Worker kill + restart | Resumed cleanly; 0 duplicates |

## Tests

824 → **836** (commit 5) → **839** (commit 6), zero failures at each step, each commit built from clean.
Format-clean. `sf ai review run --base-ref master --reviewer agent` run twice: once on the original submission (4 findings, all addressed) and once on the post-rebase tip (4 new findings from the rebase merge resolution, all addressed). Zero findings on the current tip.

Full test plan with 18 identified behaviors, all unit-testable ones covered, is in Appendix A of the design document.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
@@ -10,6 +10,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
@@ -53,8 +54,16 @@ public final class SpcsEnvironment {
 
   public static final String ENV_HOST = "SNOWFLAKE_HOST";
 
-  static final String ENV_DATABASE = "SNOWFLAKE_DATABASE";
-  static final String ENV_SCHEMA = "SNOWFLAKE_SCHEMA";
+  /**
+   * Account identifier published by SPCS. It is the account locator, matching the first label of
+   * {@link #ENV_HOST} after hyphens are substituted for underscores. Using it explicitly is more
+   * robust than parsing the hostname, and is how every other SPCS client (including the Snowflake
+   * CLI) establishes its account.
+   */
+  public static final String ENV_ACCOUNT = "SNOWFLAKE_ACCOUNT";
+
+  public static final String ENV_DATABASE = "SNOWFLAKE_DATABASE";
+  public static final String ENV_SCHEMA = "SNOWFLAKE_SCHEMA";
 
   /**
    * Placeholder written to {@code snowflake.user.name} in ambient mode. The effective identity is
@@ -74,10 +83,23 @@ public final class SpcsEnvironment {
    */
   public static final String AMBIENT_USER_PLACEHOLDER = "spcs-service-user";
 
-  /** Environment lookup. Overridable so tests can simulate an SPCS container. */
+  /**
+   * Environment lookup. Overridable through {@link #overrideForTests} so tests can simulate an SPCS
+   * container.
+   *
+   * <p><b>Threading note:</b> this is a static mutable field shared across all test classes. It is
+   * safe with the default surefire configuration (no parallel test classes), because every test
+   * class that touches it resets it in an {@code @AfterEach} method, so classes run sequentially.
+   * If {@code forkCount > 1} or {@code parallel=classes} is ever enabled, two classes calling
+   * {@link #overrideForTests} in the same JVM will race. Address this by making the seam injectable
+   * rather than static before enabling parallelism.
+   */
   private static Function<String, String> envReader = System::getenv;
 
-  /** Token file location. Overridable so tests can point at a temporary file. */
+  /**
+   * Token file location. Overridable through {@link #overrideForTests} so tests can point at a
+   * temporary file. See the threading note on {@link #envReader}.
+   */
   private static Path tokenPath = Paths.get(DEFAULT_TOKEN_FILE);
 
   private SpcsEnvironment() {}
@@ -91,6 +113,14 @@ public final class SpcsEnvironment {
 
   public static Optional<String> host() {
     return env(ENV_HOST);
+  }
+
+  /**
+   * Account identifier published by SPCS, lowercased and trimmed to match the form expected by the
+   * streaming SDK and the JDBC driver.
+   */
+  public static Optional<String> account() {
+    return env(ENV_ACCOUNT).map(a -> a.toLowerCase(Locale.ROOT));
   }
 
   public static Optional<String> database() {
@@ -162,7 +192,11 @@ public final class SpcsEnvironment {
       // as the SPCS service user instead, which is a different identity with different grants.
       // So only adopt when there is no credential at all to ignore.
       if (hasConfiguredCredential(raw)) {
-        LOGGER.info(
+        // A credential is present but no authenticator was named.  Adopting ambient auth here
+        // would silently ignore that credential and connect as a different identity.  Stay out of
+        // the way, but warn loudly rather than staying silent, because the operator may not realize
+        // their credential is being bypassed inside SPCS.
+        LOGGER.warn(
             "Running inside Snowpark Container Services, but a credential is configured, so the"
                 + " existing authentication method is kept. Set '{}' to '{}' to use ambient SPCS"
                 + " authentication instead.",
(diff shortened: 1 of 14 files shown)


src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
```diff
@@ -127,7 +128,16 @@ public static StreamingClientProperties from(SinkTaskConfig config) {
       clientProperties.put("user", config.getSnowflakeUser());
     }
     clientProperties.put("role", config.getSnowflakeRole());
-    clientProperties.put("account", url.getAccount());
+    // For ambient SPCS authentication, use SNOWFLAKE_ACCOUNT from the environment
+    // rather than parsing the first label of the host URL.  The host is an undocumented
+    // format verified on one deployment; SNOWFLAKE_ACCOUNT is the canonical identifier
+    // published by every SPCS runtime and is how the Snowflake CLI establishes its account.
+    // If the environment variable is absent (not inside SPCS) the parsed value is kept.
```
sfc-gh-skurella, line 135: NIT: is it possible for `SNOWFLAKE_ACCOUNT` to be absent when running in SPCS? If not, then I'd prefer to not have this fallback and instead to fail loudly complaining that "SNOWFLAKE_ACCOUNT environment variable not found but authentication method set to SPCS - are you sure the connector is running is SPCS?" or something along those lines. Generous fallbacks are sometimes tricky to follow and can result in errors which aren't indicative of the real issue.

sfc-gh-skurella (approved):
Thanks! LGTM modulo one NIT.

src/main/java/com/snowflake/kafka/connector/internal/streaming/StreamingClientProperties.java
sfc-gh-bschiefer: SNOWFLAKE_ACCOUNT can be absent on SPCS deployments before release 8.35. The fallback to parsing the hostname was added specifically for backward compatibility with those older runtimes.

Done — replaced the silent orElse with orElseGet that emits a WARN log naming the host being parsed and recommending an upgrade to 8.35+. Operators on old runtimes still work but see an explicit message; operators on current runtimes are never affected.

src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
```diff
@@ -162,7 +192,11 @@ public static Map<String, String> resolve(Map<String, String> raw) {
       // as the SPCS service user instead, which is a different identity with different grants.
       // So only adopt when there is no credential at all to ignore.
       if (hasConfiguredCredential(raw)) {
-        LOGGER.info(
+        // A credential is present but no authenticator was named.  Adopting ambient auth here
```
sfc-gh-aminyaylov, line 195: Comment redundant with the message below. Please remove.

src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
```diff
@@ -197,11 +231,25 @@ public static Map<String, String> resolve(Map<String, String> raw) {
...
+        new String[] {
+          KafkaConnectorConfigParams.SNOWFLAKE_PRIVATE_KEY,
+          KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_CLIENT_ID,
+          KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_CLIENT_SECRET,
+          KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_REFRESH_TOKEN
+        }) {
+      if (!isBlank(resolved.get(credKey))) {
+        LOGGER.warn(
+            "'{}' is set but will be ignored: ambient SPCS authentication uses the token supplied"
+                + " by the SPCS runtime.",
```
sfc-gh-aminyaylov, line 250: This says the values will be ignored, but the code in `PR 2/3` suggests that the presence of these values actually changes which token is being used.

Something isn't adding up. Can you please fix the comment or the inconsistency?

```
// No authenticator was configured. That does NOT mean "no credential": an absent value
// makes AuthenticatorType.fromConfig return SNOWFLAKE_JWT, so a deployment that supplies
// only a private key is a key-pair deployment that never had to name its authenticator.
// Adopting ambient authentication here would silently ignore that credential and connect
// as the SPCS service user instead, which is a different identity with different grants.
// So only adopt when there is no credential at all to ignore.
if (hasConfiguredCredential(raw)) {
  LOGGER.info(
      "Running inside Snowpark Container Services, but a credential is configured, so the"
          + " existing authentication method is kept. Set '{}' to '{}' to use ambient SPCS"
          + " authentication instead.",
      KafkaConnectorConfigParams.SNOWFLAKE_AUTHENTICATOR,
      AuthenticatorType.SPCS.toConfigValue());
  return raw;
```

sfc-gh-aminyaylov (approved):
This PR adds a lot of comments on unrelated fields/methods. Consider splitting the unrelated lines out into a separate PR so that your enablement PR is as small as possible (in case we need to rollback and rework).

Otherwise, LGTM. I had to speed run the tests because I ran out of time.

PR #1545, opened 2026-08-18
SNOW-3961901: reject a blank URL, user or role instead of accepting it silently

sfc-gh-bschiefer:
**Independent of the SPCS chain (PRs #1542–#1544) and of the auth-detection fix (PR #1541).** No shared files; reviewable in any order.

## The bug

`DefaultConnectorConfigValidator` checked three required properties with `containsKey`
while its own error message claimed to check for emptiness:

```java
if (!config.containsKey(KafkaConnectorConfigParams.SNOWFLAKE_URL_NAME)) {
    invalidConfigParams.put(..., "snowflake.url.name cannot be empty.");
}
```

A property present with an empty value (`snowflake.url.name=`) passes validation.
`StreamingClientProperties.from()` then returned a property set with **zero entries** —
no authorization type, no account, no host, no user, no role, no credential.  The connector
started and failed later inside the SDK with an error naming something unrelated.  One
missing value silently suppressed six.

Measured on `master`:
```
snowflake.url.name=
validator accepted   = true
getSnowflakeUrl()    = []
clientProperties     = []   ← zero entries
```

## The fix

**1. `DefaultConnectorConfigValidator`** — replace `containsKey` with `isBlank` for the
three required properties (`snowflake.url.name`, `snowflake.user.name`, `snowflake.role.name`).
Whitespace counts as blank; a whitespace-only URL is not usable and was already producing
a silent failure.

**2. `StreamingClientProperties.from()`** — throw `ERROR_0017` on a blank URL rather than
returning an empty property set.  `SinkTaskConfigTestBuilder` (used by 8 test classes) was
updated to supply a real URL and a generated private key so those tests continue to work; one
test that specifically covers the missing-key case uses a new `builderWithoutKey()` variant.

## Scope

Ambient SPCS authentication is immune to this defect: `SpcsEnvironment.resolve()` uses
`putIfBlank`, which checks blankness rather than key presence, so an empty
`snowflake.url.name` inside SPCS is replaced by the ambient host before validation runs.
The defect bites key-pair and OAuth deployments on `master` today.

## Tests

12 new cases in `BlankRequiredConfigTest`:
- Empty, space and tab rejected for each of the three required properties (9 cases)
- An absent URL still rejected
- A populated configuration still accepted
- `StreamingClientProperties.from()` throws `ERROR_0017` on a blank URL

Mutation-verified: restoring the `containsKey` check fails exactly the three blank-URL
cases and nothing else.  Overall suite green, zero failures.
Format-clean. Zero findings from `sf ai review run --base-ref master --reviewer agent`.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/DefaultConnectorConfigValidator.java
+++ b/src/main/java/com/snowflake/kafka/connector/DefaultConnectorConfigValidator.java
@@ -105,20 +105,20 @@ public class DefaultConnectorConfigValidator implements ConnectorConfigValidator
       }
     }
 
-    if (!config.containsKey(KafkaConnectorConfigParams.SNOWFLAKE_USER_NAME)) {
+    if (isBlank(config.get(KafkaConnectorConfigParams.SNOWFLAKE_USER_NAME))) {
       invalidConfigParams.put(
           KafkaConnectorConfigParams.SNOWFLAKE_USER_NAME,
           Utils.formatString(
               "{} cannot be empty.", KafkaConnectorConfigParams.SNOWFLAKE_USER_NAME));
     }
 
-    if (!config.containsKey(KafkaConnectorConfigParams.SNOWFLAKE_URL_NAME)) {
+    if (isBlank(config.get(KafkaConnectorConfigParams.SNOWFLAKE_URL_NAME))) {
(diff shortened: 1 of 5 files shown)


sfc-gh-skurella (approved):
LGTM, thanks!

PR #1550, opened 2026-08-24
increase max LOB size in client-side validation

sfc-gh-skurella:
## Summary

Client-side validation capped VARCHAR/VARIANT/OBJECT/ARRAY at 16MB, so an account with 128MB LOBs enabled could not ingest anything larger even though the server would accept it. The ceiling is now 128MB (`DataValidationUtil.LOB_CEILING_MB`), with no new configuration.

- `LOB_CEILING_MB` drives the semi-structured length check, the string byte check, and the VARCHAR `byteLength` cap in `ColumnSchema`.
- Semi-structured values are rejected with `Variant too long` / `Object too large` / `Array too large` once the UTF-8 size exceeds `LOB_CEILING_MB - 64` (`MAX_SEMI_STRUCTURED_LENGTH`). Strings above 128MB fail with `String too long`.
- Jackson `StreamReadConstraints.maxStringLength` is `2 * LOB_CEILING_MB`. That is a parse bound, not a second ingest limit. It lets a just-oversized string be measured and rejected as too long (column + actual length), and it lets JSON wrapping around a near-ceiling string finish parsing. A 1x cap reports those as `Not a valid JSON`.
- In non-schematized mode `RowValidator` still skips `RECORD_CONTENT`, so this limit governs schematized columns and `RECORD_METADATA`.

### Unit tests

`DataValidationUtilTest` on this branch:

- `LOB_CEILING_MB` accepted, `LOB_CEILING_MB + 1` rejected as `String too long`.
- A JSON object whose string field is `LOB_CEILING_MB + 1` is rejected as `Object too large` / `Variant too long`, not `Not a valid JSON` (this is what the 2x Jackson cap is for).
- Multi-byte UTF-8 case still reports `Variant too long` against `maxLength=134217664`.

### `LargeLobIngestionIT`

Sizes a single VARIANT field to the serialized 128MB ceiling (minus the JSON wrapper and the 64-byte server-skew buffer) and to that size plus one byte. The former is ingested. The latter is routed to the DLQ with `Variant too long`.

Gated behind `SNOWFLAKE_RUN_LARGE_LOB_IT=true` with a 15-minute timeout. A ~128MB row is TRACE-logged by the connector and has twice cancelled the 6-hour AWS integration job, so it is not part of default CI.

### Manual Kind / Kafka Connect E2E (2026-09-21)

Built the connector from this branch, staged it into a local Kind cluster (public `confluentinc/cp-kafka:7.8.0` + `confluentinc/cp-kafka-connect:7.8.0` images), and ran `SnowflakeStreamingSinkConnector` with `snowflake.validation=client_side` and schematization onto a pre-created `PAYLOAD VARIANT` column.

| Case | Result |
|---|---|
| Small JSON `{"PAYLOAD":{"ok":true}}` | Ingested |
| ~24MB chunked VARIANT (above the old 16MB / Jackson 20MB cutoffs) | Ingested intact; `OCTET_LENGTH(TO_JSON(PAYLOAD))` = 25,165,908 |
| `{"PAYLOAD":"this is not json"}` | Rejected client-side; Connect log: `Not a valid JSON` |
| ~129MB chunked VARIANT | Not ingested; Connect log: `Variant too long: length=135266703 maxLength=134217664` |

The table had **2 rows** after the run (small + 24MB). The invalid-JSON and 129MB records were rejected by `RowValidator` / `DataValidationUtil` and did not land in Snowflake.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.core.StreamReadConstraints;
+import com.fasterxml.jackson.core.exc.StreamConstraintsException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectWriter;
@@ -72,7 +73,8 @@ class DataValidationUtil {
    */
   private static final long MICROSECONDS_LIMIT_FOR_EPOCH = SECONDS_LIMIT_FOR_EPOCH * 1000000L;
 
-  public static final int BYTES_8_MB = 8 * 1024 * 1024;
+  public static final int BYTES_1_MB = 1024 * 1024;
+  public static final int BYTES_8_MB = 8 * BYTES_1_MB;
   public static final int BYTES_16_MB = 2 * BYTES_8_MB;
 
   /**
@@ -81,20 +83,15 @@ class DataValidationUtil {
    * values rejected server-side, so validating against the higher tier only widens what the client
    * lets through.
    */
-  public static final int BYTES_128_MB = 8 * BYTES_16_MB;
+  public static final int LOB_CEILING_MB = 128 * BYTES_1_MB;
 
   // TODO SNOW-664249: There is a few-byte mismatch between the value sent by the user and its
   // server-side representation. Validation leaves a small buffer for this difference.
-  static final int MAX_SEMI_STRUCTURED_LENGTH = BYTES_128_MB - 64;
+  static final int MAX_SEMI_STRUCTURED_LENGTH = LOB_CEILING_MB - 64;
 
-  /**
-   * Jackson refuses to read a single string value longer than 20MB by default, well below the LOB
-   * limit. The cap is raised to twice the LOB limit so that an oversized value is reported by the
-   * size checks in this class — naming the column and the actual length — rather than surfacing as
-   * a generic "not a valid JSON" parse failure.
-   */
+  // Allow JSON construction to handle all available Snowflake object sizes.
   private static final StreamReadConstraints STREAM_READ_CONSTRAINTS =
-      StreamReadConstraints.builder().maxStringLength(2 * BYTES_128_MB).build();
+      StreamReadConstraints.builder().maxStringLength(LOB_CEILING_MB).build();
 
   private static final ObjectMapper objectMapper =
       new ObjectMapper(
@@ -222,7 +219,7 @@ class DataValidationUtil {
             generator.copyCurrentEvent(parser);
           }
         }
-      } catch (JsonParseException e) {
+      } catch (JsonParseException | StreamConstraintsException e) {
         throw valueFormatNotAllowedException(
             columnName, snowflakeType, "Not a valid JSON", insertRowIndex);
       } catch (IOException e) {
@@ -862,14 +859,13 @@ class DataValidationUtil {
     }
     byte[] utf8Bytes = output.getBytes(StandardCharsets.UTF_8);
 
-    // Strings can never be larger than 128MB
-    if (utf8Bytes.length > BYTES_128_MB) {
+    if (utf8Bytes.length > LOB_CEILING_MB) {
       throw valueFormatNotAllowedException(
           columnName,
           "STRING",
           String.format(
               "String too long: length=%d bytes maxLength=%d bytes",
-              utf8Bytes.length, BYTES_128_MB),
+              utf8Bytes.length, LOB_CEILING_MB),
           insertRowIndex);
     }
 
--- a/src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
@@ -19,6 +19,7 @@ import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.core.StreamReadConstraints;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectWriter;
@@ -74,14 +75,35 @@ class DataValidationUtil {
   public static final int BYTES_8_MB = 8 * 1024 * 1024;
   public static final int BYTES_16_MB = 2 * BYTES_8_MB;
 
+  /**
+   * Largest LOB accepted for VARCHAR, VARIANT, OBJECT and ARRAY columns. Snowflake raised the LOB
+   * ceiling from 16MB to 128MB; accounts without large LOBs enabled still get their oversized
+   * values rejected server-side, so validating against the higher tier only widens what the client
+   * lets through.
+   */
+  public static final int BYTES_128_MB = 8 * BYTES_16_MB;
+
   // TODO SNOW-664249: There is a few-byte mismatch between the value sent by the user and its
   // server-side representation. Validation leaves a small buffer for this difference.
-  static final int MAX_SEMI_STRUCTURED_LENGTH = BYTES_16_MB - 64;
+  static final int MAX_SEMI_STRUCTURED_LENGTH = BYTES_128_MB - 64;
 
-  private static final ObjectMapper objectMapper = new ObjectMapper();
+  /**
+   * Jackson refuses to read a single string value longer than 20MB by default, well below the LOB
+   * limit. The cap is raised to twice the LOB limit so that an oversized value is reported by the
+   * size checks in this class — naming the column and the actual length — rather than surfacing as
+   * a generic "not a valid JSON" parse failure.
+   */
+  private static final StreamReadConstraints STREAM_READ_CONSTRAINTS =
+      StreamReadConstraints.builder().maxStringLength(2 * BYTES_128_MB).build();
+
+  private static final ObjectMapper objectMapper =
+      new ObjectMapper(
+          JsonFactory.builder().streamReadConstraints(STREAM_READ_CONSTRAINTS).build());
 
   private static final JsonFactory factory =
-      new JsonFactory()
+      JsonFactory.builder()
+          .streamReadConstraints(STREAM_READ_CONSTRAINTS)
+          .build()
           // Handle duplicate fields in JSON objects by ourselves
           .configure(JsonGenerator.Feature.STRICT_DUPLICATE_DETECTION, false);
 
@@ -840,13 +862,14 @@ class DataValidationUtil {
     }
     byte[] utf8Bytes = output.getBytes(StandardCharsets.UTF_8);
 
-    // Strings can never be larger than 16MB
-    if (utf8Bytes.length > BYTES_16_MB) {
+    // Strings can never be larger than 128MB
+    if (utf8Bytes.length > BYTES_128_MB) {
       throw valueFormatNotAllowedException(
           columnName,
           "STRING",
           String.format(
-              "String too long: length=%d bytes maxLength=%d bytes", utf8Bytes.length, BYTES_16_MB),
+              "String too long: length=%d bytes maxLength=%d bytes",
+              utf8Bytes.length, BYTES_128_MB),
           insertRowIndex);
     }
 
--- a/src/test/java/com/snowflake/kafka/connector/internal/streaming/LargeLobIngestionIT.java
+++ b/src/test/java/com/snowflake/kafka/connector/internal/streaming/LargeLobIngestionIT.java
@@ -1,0 +1,147 @@
+package com.snowflake.kafka.connector.internal.streaming;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import com.snowflake.kafka.connector.config.SinkTaskConfig;
+import com.snowflake.kafka.connector.config.SnowflakeValidation;
+import com.snowflake.kafka.connector.dlq.InMemoryKafkaRecordErrorReporter;
+import com.snowflake.kafka.connector.internal.SnowflakeConnectionService;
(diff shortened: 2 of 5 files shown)


sfc-gh-skurella:
* **#1550** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1550?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-kafka-connector/1550?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
```diff
@@ -74,14 +75,35 @@ class DataValidationUtil {
   public static final int BYTES_8_MB = 8 * 1024 * 1024;
   public static final int BYTES_16_MB = 2 * BYTES_8_MB;
 
+  /**
+   * Largest LOB accepted for VARCHAR, VARIANT, OBJECT and ARRAY columns. Snowflake raised the LOB
+   * ceiling from 16MB to 128MB; accounts without large LOBs enabled still get their oversized
+   * values rejected server-side, so validating against the higher tier only widens what the client
+   * lets through.
+   */
+  public static final int BYTES_128_MB = 8 * BYTES_16_MB;
```
sfc-gh-aminyaylov: For readability, can you please rename or create a new constant for `LOB_CEILING_MB`, and then use it inside the `STREAM_READ_CONSTRAINTS` value?

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
```diff
@@ -74,14 +75,35 @@ class DataValidationUtil {
...
+
   // TODO SNOW-664249: There is a few-byte mismatch between the value sent by the user and its
   // server-side representation. Validation leaves a small buffer for this difference.
-  static final int MAX_SEMI_STRUCTURED_LENGTH = BYTES_16_MB - 64;
+  static final int MAX_SEMI_STRUCTURED_LENGTH = BYTES_128_MB - 64;
+
+  /**
+   * Jackson refuses to read a single string value longer than 20MB by default, well below the LOB
```
sfc-gh-aminyaylov: I don't think we should document Jackson defaults or the fact that it will output the "column name and actual length" as these can:
1. change in the future
2. be used by other callers (producer, constant in this case, should not document how callers will use it)

There's nothing wrong with being explicit about our stream constraints.

```
// Allow JSON construction to handle all available Snowflake object sizes.
```

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
```diff
@@ -74,14 +75,35 @@ class DataValidationUtil {
...
+   * limit. The cap is raised to twice the LOB limit so that an oversized value is reported by the
+   * size checks in this class — naming the column and the actual length — rather than surfacing as
+   * a generic "not a valid JSON" parse failure.
+   */
+  private static final StreamReadConstraints STREAM_READ_CONSTRAINTS =
+      StreamReadConstraints.builder().maxStringLength(2 * BYTES_128_MB).build();
```
sfc-gh-aminyaylov: I would document why you chose double. I suspect you wanted headroom, but it's unclear why it's `2x` and not `3x` or `x + 10`.

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
```diff
@@ -840,13 +862,14 @@ static String validateAndParseString(
     }
     byte[] utf8Bytes = output.getBytes(StandardCharsets.UTF_8);
 
-    // Strings can never be larger than 16MB
-    if (utf8Bytes.length > BYTES_16_MB) {
+    // Strings can never be larger than 128MB
```
sfc-gh-aminyaylov: Comment is redundant with code. Please remove.

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
```diff
@@ -840,13 +862,14 @@ static String validateAndParseString(
...
+    if (utf8Bytes.length > BYTES_128_MB) {
```
sfc-gh-aminyaylov: Can we use our new `LOB_CEILING_MB` constant here for clarity?

src/test/java/com/snowflake/kafka/connector/internal/streaming/LargeLobIngestionIT.java
```diff
@@ -0,0 +1,147 @@
...
+ * beyond the old 16MB ceiling reaches Snowflake intact, while one past 128MB is rejected locally
+ * and routed to the DLQ.
+ *
+ * <p>The payload is assembled as an array of 1MB strings rather than one huge string because
+ * Kafka's JsonConverter refuses to deserialize a single string value larger than 20MB. The record
+ * is built from a plain Java Map so that no converter sits between the test and the connector.
+ */
+public class LargeLobIngestionIT extends SnowflakeSinkServiceV2BaseIT {
+
+  private static final int ONE_MB = 1024 * 1024;
```
sfc-gh-aminyaylov: Nit: this uses different format for byte sizes. This will get ugly when mixing the two.

Should we stick with `BYTES_1_MB`?

src/test/java/com/snowflake/kafka/connector/internal/streaming/LargeLobIngestionIT.java
```diff
@@ -0,0 +1,147 @@
...
+
+  @AfterEach
+  public void afterEach() {
+    TestUtils.dropTable(table);
+    TestUtils.dropPipe(table);
+  }
+
+  @Test
+  public void variantJustUnder128Mb_isIngested() throws Exception {
+    int chunks = 127;
```
sfc-gh-aminyaylov: Why are we testing `127`? Shouldn't we be testing the boundary between `128` and `129`, the former being accepted as a LOB?

src/test/java/com/snowflake/kafka/connector/internal/streaming/LargeLobIngestionIT.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  @AfterEach
+  public void afterEach() {
+    TestUtils.dropTable(table);
+    TestUtils.dropPipe(table);
+  }
+
+  @Test
+  public void variantJustUnder128Mb_isIngested() throws Exception {
```
sfc-gh-aminyaylov, line 69: I'm a bit confused. Are these tests supposed to be testing the `BYTES_128_MB` constraint for LOBs? If so, can we please reference constants rather than magic numbers: 127, 129 (e.g. `LOB_CEILING_MB + BYTES_1_MB`)?

src/test/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtilTest.java
```diff
@@ -1048,10 +1049,31 @@ public void testTooLargeVariant() {
     expectError(ErrorCode.INVALID_VALUE_ROW, () -> validateAndParseObject("COL", m, 0));
   }
 
+  /**
+   * A value above the old 16MB LOB limit is accepted, including when it arrives as a JSON string —
```
sfc-gh-aminyaylov: I thought this code updates the limit to `128MB`. Is a `BYTES_128_MB` test insufficient?

What's the point of testing `16MB` and `20MB`?

sfc-gh-aminyaylov (approved):

src/test/java/com/snowflake/kafka/connector/internal/streaming/LargeLobIngestionIT.java
sfc-gh-aminyaylov: 127/129 are chunk counts, not serialized VARIANT size. Each chunk is 1MB of ASCII inside `{"chunks":[...]}`; JSON quotes, commas and the wrapper push a 128-chunk payload over the 128MB ceiling. Derived as `LOB_CEILING_CHUNKS ± 1` in fb6fe92b so the boundary is the named 128MB limit, not a pair of magic numbers.

src/test/java/com/snowflake/kafka/connector/internal/streaming/LargeLobIngestionIT.java
sfc-gh-aminyaylov: 🤖 Agreed — those were chunk counts, not byte sizes, so `LOB_CEILING_MB + BYTES_1_MB` would have been the wrong unit (and `DataValidationUtil` is package-private from this IT). They are now `LOB_CEILING_CHUNKS ± 1` with `LOB_CEILING_CHUNKS = 128`.

src/test/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtilTest.java
sfc-gh-aminyaylov: 🤖 The 128MB tests already cover the new ceiling. This one is a cheaper regression for the two old cutoffs that used to fail first: the previous 16MB client cap, and Jackson's default 20MB single-string limit. It sits at 24MB (`BYTES_16_MB + BYTES_8_MB`) so both are cleared without allocating another 128MB string in the unit suite.

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
```diff
@@ -71,17 +73,34 @@ class DataValidationUtil {
...
+  public static final int LOB_CEILING_MB = 128 * BYTES_1_MB;
+
   // TODO SNOW-664249: There is a few-byte mismatch between the value sent by the user and its
   // server-side representation. Validation leaves a small buffer for this difference.
-  static final int MAX_SEMI_STRUCTURED_LENGTH = BYTES_16_MB - 64;
+  static final int MAX_SEMI_STRUCTURED_LENGTH = LOB_CEILING_MB - 64;
+
+  // Allow JSON construction to handle all available Snowflake object sizes.
+  private static final StreamReadConstraints STREAM_READ_CONSTRAINTS =
+      StreamReadConstraints.builder().maxStringLength(LOB_CEILING_MB).build();
```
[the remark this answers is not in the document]
sfc-gh-aminyaylov: Intentional after review, not a bug. The 2x headroom was arbitrary (why 2x and not 3x or +10MB?), and parsing up to 256MB just to emit a slightly nicer error is a memory cost. Jackson is capped at `LOB_CEILING_MB`; `StreamConstraintsException` is caught and reported as invalid JSON. The PR description still mentioned 2x and is stale.

sfc-gh-aminyaylov:
Picking this up from Seb. I cannot push to this branch (no write on `snowflakedb/snowflake-kafka-connector`), so the remaining fixes landed on a successor: https://github.com/snowflakedb/snowflake-kafka-connector/pull/1561

What was still open here:
- Review nits on 127/129 chunk counts — now `LOB_CEILING_CHUNKS ± 1`
- Formatting (`DataValidationUtilTest` static import order)
- AWS `build_and_test` cancelled twice at the 6h cap; `LargeLobIngestionIT` is gated behind `SNOWFLAKE_RUN_LARGE_LOB_IT=true`
- Graphite's 2x Jackson comment — dismissed; Seb already capped reads at `LOB_CEILING_MB` after review

Please close this PR in favor of #1561 if that looks right.

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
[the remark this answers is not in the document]
sfc-gh-aminyaylov: 🤖 **Cursor** (per Artem) Restored `2 * LOB_CEILING_MB`. A 1x cap made a just-oversized JSON string fail as `Not a valid JSON` before the size check could run. 2x is a parse bound so we can still emit `too long` with the actual length, and so JSON wrapping around a near-ceiling string does not die in the parser. The ingest limit stays `LOB_CEILING_MB`.

src/main/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtil.java
sfc-gh-aminyaylov: 🤖 **Cursor** (per Artem) 2x is the parse bound, not a second ingest limit. Without it, an oversized string never reaches the size check and is reported as invalid JSON. Documented on `STREAM_READ_CONSTRAINTS`.

src/test/java/com/snowflake/kafka/connector/internal/validation/DataValidationUtilTest.java
sfc-gh-aminyaylov: 🤖 **Cursor** (per Artem) We do not need a 16/20/24MB case. The product pair is `LOB_CEILING_MB` accepted and `LOB_CEILING_MB + 1` rejected. Removed the 24MB test. The +1 JSON-string case now asserts `too long` / `too large`, not `Not a valid JSON`.

src/test/java/com/snowflake/kafka/connector/internal/streaming/LargeLobIngestionIT.java
sfc-gh-aminyaylov: 🤖 **Cursor** (per Artem) 127 was a chunk count, not 127MB, and 128 chunks already overflow the serialized 128MB limit. The IT now sizes a single VARIANT field to just under and just over the byte ceiling (`LOB_CEILING_BYTES` minus wrapper and the 64-byte server-skew buffer, then +1).

src/test/java/com/snowflake/kafka/connector/internal/streaming/LargeLobIngestionIT.java
sfc-gh-aminyaylov: 🤖 **Cursor** (per Artem) Dropped the chunk counts. The IT now uses the same byte pair as the unit tests: just under the ceiling, then +1.

sfc-gh-aminyaylov (approved):
🤖 **Cursor** (per Artem)

Re-approving after the format fix and the Kind 127MiB ingest / 129MiB `Variant too long` verification. Ready to merge.

sfc-gh-aminyaylov:
🤖 **Cursor** (per Artem)

CI is green (format, `build_and_test`, E2E). Squash auto-merge is on. GitHub is still waiting for a `@snowflakedb/streaming-ingest` code-owner approval — Seb is the author so he cannot stamp his own PR, and my approval does not satisfy CODEOWNERS. @sfc-gh-ggeng / @sfc-gh-bmikaili a quick approve will merge it.

sfc-gh-alhuang (approved):

PR #1552, opened 2026-08-30
Add Java 21 runtime coverage to the end-to-end matrix

sfc-gh-bschiefer:
## This changes nothing about what we ship

**The compile target is untouched.** No production Java is modified. The connector still compiles to Java 11 bytecode and the shipped artifact is unchanged.

This only adds *test* coverage: it runs the existing artifact on a newer JVM. It is positioning for a future Java move, not a Java move.

## Compile target and runtime are distinct

These two axes are easy to conflate and they move independently:

| Axis | Today | Changed here? |
|---|---|---|
| Compile target | Java 11 | **No** |
| Runtime tested | Java 11, 17 | Yes, adds 21 |
| Runtime fully supported by Apache Kafka 4.1 | Java 17, 21 | n/a |

We are already, in effect, shipping onto Java 21: customers choose the JDK their Connect worker runs on, and [Apache Kafka 4.1](https://kafka.apache.org/41/operations/java-version/) fully supports Java 17 and 21 and recommends running on the most recent LTS. Our Java 11 bytecode runs on Java 21 in the field today. This PR starts verifying what is already happening.

**Correction to an earlier version of this description:** it claimed Kafka Connect 4.x supports "17, 21 and 25". That list is [Confluent Platform's](https://docs.confluent.io/platform/current/installation/versions-interoperability.html) ("Java 11, 17, 21, and 25 are supported"), not Apache Kafka's. Apache Kafka 4.1 does not list Java 25. These lanes run Apache Kafka 4.1.1, so the Apache statement is the one that applies, and Java 21 is both fully supported and the recommended LTS. Any future Java 25 coverage belongs on a Confluent lane, where a vendor actually supports it -- not here.

Worth noting for reviewers: the existing `java_test_version: '17'` entries are the connector **runtime** JDK, not a compile target. Java 17 runtime was already covered; 21 was the untested step.

The compile target should stay at 11 while Java 11 Connect workers (Confluent 7.x, Apache 3.x) are supported, and when it moves it should move to **17**, the Connect 4.x floor. There is no reason to target 21.

## What this adds

One lane: Apache 4.1.1 on Java 21, `core` group only, to bound CI cost. The Java 17 lanes remain the full-coverage matrix. Plus the GHCR image that backs it.

## `run_tests.sh` could not express this

The 4.x branch hard-set `JAVA_VERSION="17"` *after* argument parsing, silently discarding `--java-version`. A lane declaring Java 21 would have run Java 17 and **reported success for a version it never exercised**. Confirmed against `master` before fixing.

The platform floor is now a default only. Two guards were added:

1. An explicit version below the Kafka 4.x floor fails fast rather than being quietly substituted. This matters because `end-to-end.yaml` defaults to `'11'`, so a future 4.x lane that forgets `java_test_version` now fails loudly instead of silently passing.
2. `--java-version` is validated as a bare major version. It feeds the image tag, so `11.0.2` or `17x` previously passed through and produced an unresolvable tag that failed later with a confusing error.

## Reviewer notes

- `test_group` is `core_java21`, not `core`, because the failure-log artifact name derives from it and `upload-artifact@v4` rejects duplicate names within a run. Verified no duplicate names across any job.
- Test resource collisions are not a concern: the java17 and java21 core lanes run identical tests concurrently against the same account, but `session_name_salt` generates a random 7-character salt per session and CI never pins `--name-salt`.
- Until `build-apache-kafka-images.yml` publishes `apache-kafka:4.1.1-java21`, `run_tests.sh` falls back to a local image build, so the lane degrades to slower rather than failing.

## Verification

Version resolution, all cases:

| Platform | `--java-version` | Result |
|---|---|---|
| 4.1.1 | unset / 17 / 21 | 17 / 17 / **21** |
| 4.1.1 | 11, 8 | rejected, below floor |
| 4.1.1, 3.9.2 | `17x`, `11.0.2`, `abc`, empty | rejected, malformed |
| 3.9.2 | unset / 11 / 21 | 11 / 11 / 21 |
| confluent 7.9.3 | n/a | unaffected |

Shell and YAML syntax valid.

The Java 21 runtime itself was validated locally under Podman, which does not need Snowflake credentials:

- `apache-kafka:4.1.1-java21` builds clean; Temurin 21.0.12 inside.
- Kafka 4.1.1 broker (KRaft) and Kafka Connect both start on Java 21.
- The **JDK 11-built** connector plugin loads and registers: `SnowflakeStreamingSinkConnector` v4.1.0.
- `PUT /config/validate` exercises the connector's own validation and RSA key parsing on Java 21. It fails only with `Cannot connect to Snowflake` against a deliberately bogus host, i.e. our code ran correctly.
- Logs contain **zero** `InaccessibleObjectException`, `UnsupportedClassVersionError`, `NoSuchMethodError`, `NoClassDefFoundError`, illegal-reflective-access warnings or `sun.misc.Unsafe` warnings.

Separately, compiling all 110 production sources at `release 21` produces exactly **one** JDK deprecation, `new URL(String)` in `Utils.java:116` (deprecated in Java 20, not marked for removal). `release 11` and `release 17` produce none. That is the entire Java 11 to 21 code delta.

## Not verified

This lane has never executed in GitHub Actions, and data flow into Snowflake on Java 21 is untested, since that needs credentials and cloud resources. That is precisely the gap this lane exists to close, and the first CI run is the test.

### DIFF
--- a/.github/actions/run-e2e-tests/action.yml
+++ b/.github/actions/run-e2e-tests/action.yml
@@ -12,7 +12,7 @@ inputs:
     description: "Snowflake cloud provider: 'AWS', 'GCP', or 'AZURE'"
     required: true
   java-version:
-    description: "Java version for Apache Kafka (e.g. '11', '17'). Ignored for Confluent."
+    description: "Java version for Apache Kafka (e.g. '11', '17', '21'). Kafka 4.x requires 17 or newer. Ignored for Confluent."
     required: false
     default: '11'
   marker-filter:
--- a/.github/workflows/build-apache-kafka-images.yml
+++ b/.github/workflows/build-apache-kafka-images.yml
@@ -28,6 +28,13 @@ jobs:
           - kafka_version: '4.1.1'
             scala_version: '2.13'
             java_version: '17'
+          # Java 21 runtime coverage for the connector. Kafka 4.x supports 17, 21 and 25;
+          # this image backs the Java 21 end-to-end lane. run_tests.sh falls back to a
+          # local build if the tag is absent, so the e2e lane degrades to a slower build
+          # rather than failing outright.
+          - kafka_version: '4.1.1'
+            scala_version: '2.13'
+            java_version: '21'
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/build-apache-kafka-images.yml
+++ b/.github/workflows/build-apache-kafka-images.yml
@@ -28,8 +28,9 @@ jobs:
           - kafka_version: '4.1.1'
             scala_version: '2.13'
             java_version: '17'
-          # Java 21 runtime coverage for the connector. Kafka 4.x supports 17, 21 and 25;
-          # this image backs the Java 21 end-to-end lane. run_tests.sh falls back to a
+          # Java 21 runtime coverage for the connector. Apache Kafka 4.1 fully
+          # supports Java 17 and 21 and recommends the latest LTS; this image backs
+          # the Java 21 end-to-end lane. run_tests.sh falls back to a
           # local build if the tag is absent, so the e2e lane degrades to a slower build
           # rather than failing outright.
           - kafka_version: '4.1.1'
--- a/.github/workflows/end-to-end.yaml
+++ b/.github/workflows/end-to-end.yaml
@@ -71,6 +71,22 @@ jobs:
             test_group: core
             marker_filter: 'not compatibility and not schema_evolution and not correctness and not pressure and not soak'
 
+          # ── Java 21 runtime coverage ──────────────────────────────────────
+          # The connector is compiled to Java 11 bytecode but is increasingly run on
+          # newer Connect workers. Kafka 4.x supports Java 17, 21 and 25, so this lane
+          # exercises the shipped artifact on a Java 21 JVM without changing the compile
+          # target. Only the core group runs here to bound CI cost; the 17 lanes above
+          # remain the full-coverage matrix.
+          # test_group must stay distinct from the Java 17 'core' lane above, because the
+          # failure-log artifact name is derived from it and upload-artifact rejects
+          # duplicate names.
+          - platform: apache
+            platform_version: '4.1.1'
+            snowflake_cloud: 'AWS'
+            java_test_version: '21'
+            test_group: core_java21
+            marker_filter: 'not compatibility and not schema_evolution and not correctness and not pressure and not soak'
+
           - platform: confluent
             platform_version: '8.2.0'
             snowflake_cloud: 'AWS'
--- a/.github/workflows/end-to-end.yaml
+++ b/.github/workflows/end-to-end.yaml
@@ -73,7 +73,8 @@ jobs:
 
           # ── Java 21 runtime coverage ──────────────────────────────────────
           # The connector is compiled to Java 11 bytecode but is increasingly run on
-          # newer Connect workers. Kafka 4.x supports Java 17, 21 and 25, so this lane
+          # newer Connect workers. Apache Kafka 4.1 fully supports Java 17 and 21 and
+          # recommends running on the latest LTS, so this lane
           # exercises the shipped artifact on a Java 21 JVM without changing the compile
           # target. Only the core group runs here to bound CI cost; the 17 lanes above
           # remain the full-coverage matrix.
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -65,7 +65,7 @@ usage() {
     echo ""
     echo "Options:"
     echo "  --cloud=CLOUD        Snowflake cloud platform: AWS, GCP, or AZURE"
-    echo "  --java-version=VER   Java version for Apache Kafka (default: 11)"
+    echo "  --java-version=VER   Java version for Apache Kafka (default: 11; 17 for Kafka 4.x)"
     echo "  --jmx                Enable JMX metrics scraping via Jolokia"
     echo "  --profile            Enable JVM profiling (JFR, GC logs, JMX, async-profiler)"
     echo "  --with-mitmproxy     Enable mitmproxy for fault injection tests (410 injection)"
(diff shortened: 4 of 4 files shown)


sfc-gh-bschiefer:
Re-verified against current `master` (`61c2867`), which is eight days and four merges newer than
when this PR was written — including #1551, #1554 and #1555. Summary: **nothing has changed, and
the zero-code-changes claim still holds.**

cc @sfc-gh-aminyaylov — Berzan asked that you weigh in as the new owner. (I could not add you as a
formal reviewer: GitHub rejects the request with *"Reviews may only be requested from
collaborators"* because your permission on this repo is `read` and you are not in the
`streaming-ingest` team, even though you reviewed #1542-#1544. Worth fixing separately.)

## Berzan's concern: does this force code changes?

**No. Zero production source files are touched by this PR**, and zero are required by it.

```
.github/actions/run-e2e-tests/action.yml      +1/-1
.github/workflows/build-apache-kafka-images.yml +7/-0
.github/workflows/end-to-end.yaml            +16/-0
test/run_tests.sh                            +29/-4
```

Nothing under `src/`. The compile target stays at Java 11 and the shipped artifact is unchanged.

I re-ran the compile matrix over all **110** production sources on today's `master`:

| `release` | Result | JDK deprecations | Pre-existing library deprecations |
|---|---|---|---|
| 11 | BUILD SUCCESS | 0 | 5 |
| 17 | BUILD SUCCESS | 0 | 5 |
| 21 | BUILD SUCCESS | **1** | 5 |

The single JDK-level finding at `release 21` is unchanged from the original run:

```
Utils.java:[116,37] URL(java.lang.String) in java.net.URL has been deprecated
```

That is a **warning, not an error**, deprecated in Java 20 and *not* marked for removal. It does
not fail the build and requires no change to merge this PR.

To be precise about the five library deprecations, since the original description only counted JDK
ones: they are `CacheBuilder.expireAfterWrite` (x3, Guava) and `JsonNode.fields()` (x2, Jackson).
They are **identical at 11, 17 and 21**, so they pre-date this work and are unrelated to Java
version. Nothing here introduces them.

Unit suite on current `master`: **870 tests, 0 failures.** (40 errors locally, all
`SNOWFLAKE_CREDENTIAL_FILE environment variable is not set` — a known local-environment gap, not
a code failure. It was 868 before #1555 added two tests, so the delta is fully accounted for.)

## Why do this now rather than later

**1. We already ship onto Java 21; we just don't test it.** Kafka Connect 4.x supports Java 17, 21
and 25, and the customer picks the JDK their Connect worker runs on. Our Java 11 bytecode is
running on Java 21 in the field today. The gap this closes is between what customers already run
and what CI verifies -- not between two future states.

**2. It has already found a real defect, independent of Java 21.** The 4.x branch hard-set
`JAVA_VERSION="17"` *after* argument parsing, silently discarding `--java-version`. Any lane
declaring another version would have run 17 and **reported success for a version it never
exercised**. That is a false green in CI, and it is fixed here with two guards: a version below
the Kafka 4.x floor now fails fast instead of being quietly substituted, and `--java-version` is
validated as a bare major version. That value is banked whatever we decide about Java 21.

**3. It de-risks the compile-target move with evidence instead of assumption.** When the target
moves it should go to **17**, the Connect 4.x floor -- not 21. Knowing the runtime is clean on 21
means the eventual 11 -> 17 move is a small, well-understood step rather than a leap.

**4. The cost is bounded and the timing is cheap.** One lane, Apache 4.1.1 on Java 21, `core`
group only; the Java 17 lanes remain the full matrix. Adding it now, while `master` is quiet, is
far cheaper than discovering a Java 21 problem during a release.

**5. The asymmetry favours doing it.** If it passes we gain a signal we currently lack, for one CI
lane. If it fails, we have found a real problem that customers could already hit today, and we
find it in CI rather than in a support case.

## What is still unverified

Honestly stated: **this lane has never run in GitHub Actions.** Data flow into Snowflake on Java 21
needs credentials and cloud resources, so it could not be exercised locally. Local Podman
validation confirmed the JDK-11-built plugin loads and registers on a Java 21 Connect worker with
zero `InaccessibleObjectException`, `UnsupportedClassVersionError`, `NoSuchMethodError`,
`NoClassDefFoundError`, illegal-reflective-access or `sun.misc.Unsafe` warnings -- but the first CI
run is the real test. That is the point of the lane.

sfc-gh-bmikaili (approved):

sfc-gh-bschiefer:
Thanks for the review @sfc-gh-bmikaili, and for coming straight to it after leave. Merging.

Recapping the scope question you raised, since it is the thing most likely to be asked again
later: **this touches no production source.** Four files, all CI and test harness:

```
.github/actions/run-e2e-tests/action.yml         +1/-1
.github/workflows/build-apache-kafka-images.yml  +7/-0
.github/workflows/end-to-end.yaml               +16/-0
test/run_tests.sh                               +29/-4
```

The compile target stays at Java 11 and the shipped artifact is unchanged. Re-verified against
`master` at `61c2867`: all 110 production sources build clean at `release` 11, 17 and 21, with a
single JDK-level deprecation at 21 (`new URL(String)` in `Utils.java:116`, deprecated in Java 20
and not marked for removal). Unit suite 870 tests, 0 failures.

Two things worth carrying forward:

**The support framing was corrected before merge.** An earlier description claimed Kafka Connect
4.x supports "17, 21 and 25". That is Confluent Platform's matrix; Apache Kafka 4.1 fully supports
17 and 21 and recommends the latest LTS. These lanes run Apache Kafka 4.1.1, so Java 21 is not a
speculative future step -- it is the version upstream actively recommends, which is a better
reason to test it than the original one.

**This PR fixed a false green independent of Java 21.** `run_tests.sh` hard-set
`JAVA_VERSION="17"` after argument parsing, silently discarding `--java-version`, so a lane
declaring another version would have run 17 and reported success for a version it never
exercised. That is fixed here with two guards, and it is worth remembering next time a new lane
is added.

**Still unverified, deliberately:** this lane has never executed in GitHub Actions -- it needs
credentials and cloud resources that were not available locally. The first `end-to-end` run on
`master` after this merge is its real first test, and worth watching rather than assuming.

PR #1554, opened 2026-09-06
SNOW-3961901: use the documented term "workload-identity" in SPCS log messages

sfc-gh-bschiefer:
## Summary

The docs call this **SPCS workload-identity authentication**. The connector's log and exception
messages said **"ambient"**, which appears nowhere in the documentation, so an operator who reads a
warning and searches the docs for the term finds nothing.

Four string literals in `SpcsEnvironment`, no behaviour change:

- the token-read failure message
- the warning when a credential is configured but no authenticator is named
- the info line when SPCS auth is adopted implicitly
- the warning that a superfluous credential is ignored

```
- 'snowflake.private.key' is set but will be ignored: ambient SPCS authentication uses the token supplied by the SPCS runtime.
+ 'snowflake.private.key' is set but will be ignored: SPCS workload-identity authentication uses the token supplied by the SPCS runtime.
```

Internal identifiers (`suppliesAmbientIdentity`, `AMBIENT_USER_PLACEHOLDER`) are unchanged.

## Test plan

- [x] `mvn test` -- 868 tests, 0 failures, 0 errors; same count as master
- [x] google-java-format check clean
- [x] No `ambient` left in any string literal under `src/main/java`

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
@@ -145,8 +145,8 @@ public final class SpcsEnvironment {
       throw new IllegalStateException(
           "Failed to read the SPCS session token at "
               + tokenPath
-              + ". This file is provided by the SPCS runtime; ambient authentication is only"
-              + " available to a connector running inside Snowpark Container Services.",
+              + ". This file is provided by the SPCS runtime; workload-identity authentication is"
+              + " only available to a connector running inside Snowpark Container Services.",
           e);
     }
     if (token.isEmpty()) {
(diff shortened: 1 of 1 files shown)


sfc-gh-bschiefer:
Assigning to @sfc-gh-skurella.

**What this is:** a text-only change. Four string literals in `SpcsEnvironment` said
"ambient", which is the term we use informally inside the team but which appears nowhere in
the public documentation. An operator who reads one of these warnings and searches the docs
for "ambient" finds nothing. This replaces it with the official user-facing term,
**workload-identity authentication**.

No behaviour change, no config change, no dependency change. The internal identifiers
(`suppliesAmbientIdentity`, `AMBIENT_USER_PLACEHOLDER`) are deliberately left alone — this is
about what we say to operators, not what we call things in code.

**Ticket association:** this belongs to
[SNOW-3961901](https://snowflakecomputing.atlassian.net/browse/SNOW-3961901) — "Enable SPCS
support for Kafka Connector v4 without requiring OAuth" — which is the ticket that added SPCS
authentication in the first place. That ticket already carries a "Terminology note
(documentation alignment)" section recording exactly this decision: informally "ambient
authentication", officially "workload-identity authentication", with `snowflake.authenticator`
set to `spcs`. This PR is the implementation of that note on the connector side, so no new
ticket is needed.

Deliberately **not** filed under the documentation ticket (DOC-16688). That one tracks the
docs-side cleanup; this is a connector source change and belongs with the feature work.

The matching documentation is in
[snowflake-prod-docs#24396](https://github.com/snowflake-eng/snowflake-prod-docs/pull/24396),
so after both land the log messages and the docs will use the same term.


sfc-gh-skurella (approved):

sfc-gh-bschiefer:

To be explicit about the risk on this one: **it is a logging change and nothing else.** Eight
lines in a single file, all of them string literals in log and exception messages, replacing
the informal internal word "ambient" with the documented user-facing term
**workload-identity**.

There is no change to behaviour, control flow, configuration, validation, or dependencies. The
internal identifiers (`suppliesAmbientIdentity`, `AMBIENT_USER_PLACEHOLDER`) are untouched, so
nothing that any code reads or branches on has moved — only what an operator sees.

The point is discoverability: an operator who hit one of these warnings and searched the docs
for "ambient" found nothing, because the public documentation only ever calls this
workload-identity authentication. After this and
[snowflake-prod-docs#24396](https://github.com/snowflake-eng/snowflake-prod-docs/pull/24396)
land, the log messages and the documentation use the same term.

Tracked under [SNOW-3961901](https://snowflakecomputing.atlassian.net/browse/SNOW-3961901),
which already records this terminology decision, rather than under the docs ticket.

PR #1555, opened 2026-09-08
SNOW-3961901: warn when snowflake.role.name is overridden under SPCS auth

sfc-gh-bschiefer:
## Summary

Under SPCS workload-identity authentication the session runs as the service user, whose default role is the service owner role. A configured `snowflake.role.name` is therefore accepted and then overridden — but silently.

That asymmetry is the point: an ignored **credential** already warns, an ignored **role** did not. The role is the one that changes which tables are visible, so when it's wrong the only symptom is a misleading `ERR_TABLE_DOES_NOT_EXIST_NOT_AUTHORIZED` (404) for a table that exists and is queryable by the operator.

One warning, using the existing wording convention in this file and pointing at `CREATE SERVICE` to match the documentation:

```
WARN 'snowflake.role.name' is set but will be overridden: SPCS workload-identity
     authentication uses the service owner's role. Assign the role when you run CREATE SERVICE.
```

"overridden" rather than "ignored" is deliberate — it mirrors the documented behaviour ("accepted but overridden") and distinguishes this from the credential warnings, which really are ignored.

No behaviour change: log output only.

## Test plan

- [x] `SpcsEnvironmentTest` — 2 new tests: warns when a role is configured, stays quiet when none is. **28 pass.**
- [x] Full unit suite: **870 pass / 0 fail** (868 before this change).
- [x] Formatter clean: `fmt-maven-plugin:check`, 224 files, 0 non-complying.
- [x] Behaviour verified on live SPCS services across AWS, Azure and GCP: `snowflake.role.name` is accepted and the effective role is the service owner's.

## Note

Reusing the SNOW-3961901 ticket from #1554 as this is the same SPCS documentation-alignment follow-up. Retitle if a separate ticket is preferred.

### DIFF
--- a/src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
+++ b/src/main/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironment.java
@@ -252,6 +252,15 @@ public final class SpcsEnvironment {
       }
     }
 
+    // The role is overridden rather than ignored: the session runs as the service user, whose
+    // default role is the service owner role.
+    if (!isBlank(resolved.get(KafkaConnectorConfigParams.SNOWFLAKE_ROLE_NAME))) {
+      LOGGER.warn(
+          "'{}' is set but will be overridden: SPCS workload-identity authentication uses the"
+              + " service owner's role. Assign the role when you run CREATE SERVICE.",
+          KafkaConnectorConfigParams.SNOWFLAKE_ROLE_NAME);
+    }
+
     return resolved;
   }
 
--- a/src/test/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironmentTest.java
+++ b/src/test/java/com/snowflake/kafka/connector/internal/spcs/SpcsEnvironmentTest.java
@@ -430,6 +430,55 @@ public class SpcsEnvironmentTest {
         .noneMatch(msg -> msg.contains(secret));
   }
 
+  /**
+   * The configured role is silently overridden by the service owner role, so it must warn. An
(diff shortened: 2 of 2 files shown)


sfc-gh-skurella (approved):

sfc-gh-bschiefer:

**This is a self-contained diagnostics change.** Under SPCS workload-identity authentication the
session runs as the service user, so a configured `snowflake.role.name` is accepted and then
overridden by the service owner's role. Previously that happened silently.

The asymmetry is what made it worth fixing: an ignored *credential* already warned, an ignored
*role* did not -- and the role is the one that decides which tables are visible. When it was
wrong, the only symptom was a misleading `ERR_TABLE_DOES_NOT_EXIST_NOT_AUTHORIZED` for a table
that exists and that the operator can query themselves. That is an expensive thing to debug from
a 404.

One `WARN` line plus two unit tests, +58/-0 across two files. No change to behaviour, control
flow, configuration or dependencies -- the role was already being overridden; the only
difference is that we now say so. "Overridden" rather than "ignored" is deliberate, to mirror
the documented behaviour and to distinguish this from the credential warnings, which really are
ignored.

Verified on live SPCS services across AWS, Azure and GCP; unit suite 870 pass / 0 fail.

On scope, since this shares a ticket with recently merged work: it is filed under
[SNOW-3961901](https://snowflakecomputing.atlassian.net/browse/SNOW-3961901) as another SPCS
documentation-alignment follow-up, but it is **functionally independent** of #1554 (log
terminology) and of #1551 (compiler flags). Nothing here depends on either, and neither depends
on this. Happy to move it to its own ticket if the shared one is misleading in hindsight.

PR #1558, opened 2026-09-14
SNOW-4104678 quote pipe names in pipeExist()

sfc-gh-aminyaylov:
## Summary
- `pipeExist()` binds `quoteIdentifier(pipeName)` into `DESC PIPE IDENTIFIER(?)`, matching table methods from https://github.com/snowflakedb/snowflake-kafka-connector/pull/1379.
- `migrateSsv1ChannelOffset()` binds `quoteIdentifier(pipeName)` as the fourth argument of `SYSTEM$MIGRATE_SSV1_CHANNEL_OFFSET`. That was the other production JDBC bind of a pipe name.
- Quoted identifiers may contain `.` and `-`. Unquoted, a hyphen is 001003 and a dot is a two-part name, so `pipeExist()` returned false and HP mode fell through to `{tableName}-STREAMING`.
- Relates https://snowflakecomputing.atlassian.net/browse/SNOW-4058238. Default-pipe fallback logic in `SnowflakeSinkServiceV2` is unchanged.

## Test plan
- [x] `mvn -Dtest=StandardSnowflakeConnectionServicePipeExistTest test` (quoted binds for `pipeExist` and migrate, `.` / `-` / embedded `"`)
- [ ] `ConnectionServiceIT#testPipeExist_dottedAndHyphenNames` on snowflakedb CI (this PR is from a fork). Creates a quoted pipe with `.` and `-`. Asserts that pipe exists. Asserts a separate never-created name with only `_` does not exist.


sfc-gh-alhuang (approved):

sfc-gh-aminyaylov:
🤖 **Cursor** (per Artem)

Superseded by https://github.com/snowflakedb/snowflake-kafka-connector/pull/1566, which carries the same change on a branch in this repo. The code is identical, rebased onto current `master` and squashed into one commit.

Fork PRs do not receive `secrets.SNOWFLAKE_TEST_PROFILE_SECRET`, so every Snowflake-touching job here failed inside `decrypt_secret.sh` before running a single test. On the new PR the decrypt step succeeds and the integration tests actually run.

PR #1560, opened 2026-09-17
SNOW-4104676: Treat body-less HTTP 404s as client-invalid on KCv4 create and recreate

yelgabra:
## Test Plan


src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPools.java
```diff
@@ -145,8 +150,6 @@ public static SnowflakeStreamingIngestClient recreateClient(
                           streamingClientProperties,
                           taskMetrics));
     } catch (FailsafeException e) {
-      // Retries exhausted — wrap as ClientRecreationException so the batch
```
yelgabra: this comment was false on main, removing it for clarity

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPools.java
```diff
@@ -166,54 +177,71 @@ public static SnowflakeStreamingIngestClient recreateClient(
   private static final double CLIENT_CREATION_JITTER_FACTOR = 0.2;
 
   /**
-   * Total wall-clock budget for recreate retries before giving up. Sized for real SSv2
-   * pipe-failover propagation windows (observed 2–4 min) plus headroom. {@link #recreateClient}
-   * throws {@link ClientRecreationException} when exhausted; callers are expected to convert that
-   * to a {@link ConnectException} so Kafka Connect fails and restarts the task.
+   * Wall-clock budget for first-time client creation ({@link #getClient} / {@link
+   * #getClientAsync}). Sized to observed T1 NR windows (tens of seconds, longest timed ~6 min is an
```
sfc-gh-aminyaylov, line 181: Why does this comment reference the 6 minute outlier? Meanwhile the `RECREATE` constant below makes no mention of the outlier. I think these comment got spliced incorrectly.

We should also explicitly and concisely explain why these values different between creation and recreation. We don't need to leak internal implementation details (as this is a public repo); rather, this comment should focus on the difference in expectations between initial client create and the recreate case — explain the usability tradeoff that produced these durations.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPools.java
```diff
@@ -166,54 +177,71 @@ public static SnowflakeStreamingIngestClient recreateClient(
...
+   */
+  private static RetryPolicy<SnowflakeStreamingIngestClient> createClientRetryPolicy(
+      String pipeName) {
+    return clientRetryPolicy(
+        pipeName, CLIENT_CREATE_MAX_DURATION, ClientRecreationException::isUnenvelopedNr404);
+  }
+
+  /**
+   * Recreation retries client-invalid errors (409 / 410) only. An unenveloped NR 404 is terminal:
+   * the account already resolved, so a no-route hostname is not the first-lookup race {@link
```
sfc-gh-aminyaylov, line 236: Is this correct? I thought we don't retry 404 on recreate because SDK already handles 404 retry after the client is created.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPools.java
```diff
@@ -166,54 +177,71 @@ public static SnowflakeStreamingIngestClient recreateClient(
...
                     pipeName,
                     event.getAttemptCount(),
                     event.getElapsedTime().toSeconds(),
                     event.getException().getMessage()))
         .build();
   }
 
+  /**
+   * Create retries unenveloped NR 404s only. Client-invalid errors (409 / 410) are not retried:
+   * there is no client yet.
```
sfc-gh-aminyaylov, line 226: I'm confused by this comment.

Are you saying we can't get 409/410 because there is no client yet? If so, let's explicitly say that: `client-invalid errors are not possible until after the client is created`.

Or are you saying that 409/410 during create should be treated differently? If so, the comment should explain why we are treating 409/410 differently on create.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/ClientRecreationException.java
```diff
@@ -32,6 +32,9 @@ public class ClientRecreationException extends RuntimeException {
           // Client was closed
           "ClosedClientError");
 
+  /** Needle in {@link SFException#getDetailMessage()} for an Envoy NR 404 (empty GS envelope). */
+  private static final String UNENVELOPED_NR_404_DETAIL = "HTTP 404, error_code=, message=,";
```
sfc-gh-aminyaylov, line 36: **[Blocker]** This is clearly a human-readable string and not programmable. This is akin to coding for the presence of a log line. It's brittle and not a recommended in production. Let's find an alternative.

It looks like KC does not get the original error_code and message. I think the best approach would be for the SDK to categorize this as an "404 no body" or similar, and allow clients to program against it. See the `CLIENT_INVALID_ERROR_CODE_NAMES` list above, which programs against `SFException.getErrorCodeName()` codes.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPools.java
```diff
@@ -145,10 +151,15 @@ public static SnowflakeStreamingIngestClient recreateClient(
...
       Throwable cause = e.getCause() != null ? e.getCause() : e;
-      throw ClientRecreationException.wrap(cause);
+      if (ClientRecreationException.isClientInvalidError(cause)) {
+        throw ClientRecreationException.wrap(cause);
+      }
+      if (cause instanceof RuntimeException) {
+        throw (RuntimeException) cause;
+      }
+      throw new ConnectException(
+          "Unexpected error recreating streaming client for pipe: " + pipeName, cause);
```
sfc-gh-aminyaylov, line 162: Is this an unrelated improvement to the code? If so, I would recommend doing so in a separate PR. This PR should be focused on the create (non-recreate) path.

sfc-gh-aminyaylov (changes requested):
I have a concern about programming against a human-readable string. Let's resolve this and then I'll take another look at the tests.

src/main/java/com/snowflake/kafka/connector/internal/streaming/v2/client/StreamingClientPools.java
sfc-gh-yelgabra: this is leftover from when i had recreate handle 404, calling wrap on 404 was returning illegalargument, ill revert it

PR #1562, opened 2026-09-18
SNOW-3683451: drop the 1.4 SDK pin on fault-injection e2e

yelgabra:
## Summary
- Points the fault-injection e2e at `https://mitmproxy:8080` and trusts a test-only CA via `SSL_CERT_FILE`, which SSv2 1.7+ supports
- Drops the `1.4.0` compile pin from the fault-injection job so that lane builds against the POM SDK (1.8.0)
- Keeps `SNOWPIPE_STREAMING_VERSION_OVERRIDE` as a generic hatch and removes the pin-specific comments

## Test plan
- [ ] `fault_injection` e2e is green on this PR (410 and 404 client-recreation paths)
- [ ] Other e2e jobs still use the POM SDK and are unchanged
- [ ] Confirm the Connect worker can verify the mitmproxy leaf (`SSL_CERT_FILE` points at `mitmproxy-ca.crt`)

Made with [Cursor](https://cursor.com)


test/docker/docker-compose.mitmproxy.yml
```diff
@@ -34,6 +33,16 @@ services:
       timeout: 5s
       retries: 15
 
+  # fault_injection runs on Apache; the SDK process lives in this container.
```
sfc-gh-aminyaylov: This comment is confusing to me. Apache is ambiguous, and should instead refer to the Apache kafka service in the base template. This should also hint at the trust mechanism, and why these certs are being attached to the kafka service.

```
# Overlay on Apache's kafka service (already defined in docker-compose.apache.yml).
# The connector/SDK runs in that container, so it must trust the test CA.
```

sfc-gh-aminyaylov (approved):
LGTM, thanks!

PR #1565, opened 2026-09-21
SNOW-4046695: allowlist kafka_start connector config keys

sfc-gh-aschuck:
## Summary
- `kafka_start` telemetry now copies only an explicit allowlist of connector config keys instead of sending every non-secret key. Unlisted config is omitted by default.
- Client-copied keys are a subset of the server persist allowlist and only those KCv4 actually uses (`tasks.max`, converters, error handling, topic/table map, cache flags, etc.). Service-owned fields (`app_name`, `task_id`, `start_time`, JDK/Kafka versions, ingestion method) are written separately and are not in the copy-set so colliding config cannot overwrite them.
- Adds a regression test that secrets (`snowflake.jdbc.map`, SSL keystore password, OAuth tokens) stay out of the payload, and that attacker-supplied service-owned keys do not win.

## Why this is not a 1:1 copy of the server allowlist

`KAFKA_START_ALLOWED_DATA_KEYS` is the set of **user-provided connector config keys the connector may copy** into the payload. It is deliberately a strict subset of the 27-key server persist allowlist: a key belongs here only if a KCv4 operator can actually set it and it is not already written by the connector itself. Ten server keys are therefore absent, for two distinct reasons.

### Still sent, but written by the service rather than copied from config (7)

These are populated directly in `SnowflakeTelemetryService` (`getObjectNode` and `reportKafkaConnectStart`), so they are present in every `kafka_start` payload. Listing them in the copy-set would let a connector config with a colliding key name overwrite a trusted value, since `ObjectNode.put` overwrites and the copy runs after those fields are stamped. That is not hypothetical: `task_id` is injected into every task config by `taskConfigs()`, and a config carried over from an older install can still contain `snowflake.ingestion.method`.

| Key | Source |
| --- | --- |
| `app_name` | connector name held by the telemetry service |
| `task_id` | task id held by the telemetry service |
| `start_time` | `startTime` argument to `reportKafkaConnectStart` |
| `kafka_version` | `AppInfoParser.getVersion()` |
| `jdk_version` | `java.version` system property |
| `jdk_distribution` | `java.vendor` system property |
| `snowflake.ingestion.method` | `IngestionMethodConfig.SNOWPIPE_STREAMING` |

`testReportKafkaConnectStart_serviceOwnedKeysNotOverwrittenByConfig` pins this: it seeds the connector config with hostile values for all seven and asserts the service values survive.

### Not collected at all, because the key does not exist in KCv4 (3)

| Key | Why absent |
| --- | --- |
| `max_tasks` | Legacy telemetry field name from KC v3. KCv4 reads the Kafka Connect property `tasks.max`, which **is** in the copy-set. The unused `MAX_TASKS` constant is removed in this PR. |
| `snowflake.streaming.max.client.lag` | Not a KCv4 configuration parameter. Client lag is managed by the Snowpipe Streaming v2 SDK; the connector exposes no such knob. |
| `snowflake.streaming.enable.single.buffer` | Snowpipe Streaming Classic (SSv1) buffering option. SSv1 is removed in KCv4 (`IngestionMethodConfig` has only `SNOWPIPE_STREAMING`), so nothing can set it. |

Grep confirms none of these three names appear anywhere in `src/`. Leaving them on the server side is harmless, and keeping them out of the client avoids implying the connector still collects them. If any is reintroduced as a real KCv4 config later, it can be added to the copy-set without a server-side change, since the server already persists it.

### Direction of the invariant

The rule is one-way and is documented on the constant: a key may be added to the client copy-set **only after** the server persist allowlist already contains it, and if the connector stops collecting a key it should be removed from both. Server ⊇ client, never the reverse.

## Test plan
- [x] `mvn test -DskipIntegrationTests=true -Dtest=SnowflakeTelemetryServiceTest` — 8/8 pass


src/test/java/com/snowflake/kafka/connector/internal/telemetry/SnowflakeTelemetryServiceTest.java
```diff
@@ -89,17 +111,65 @@ public void testReportKafkaConnectStart(IngestionMethodConfig ingestionMethodCon
...
 
-    // All non-sensitive config keys from the map should be present
-    assertTrue(dataNode.has(KafkaConnectorConfigParams.SNOWFLAKE_DATABASE_NAME));
-    assertTrue(dataNode.has(KafkaConnectorConfigParams.SNOWFLAKE_SCHEMA_NAME));
-    assertTrue(dataNode.has(KafkaConnectorConfigParams.SNOWFLAKE_URL_NAME));
-    assertTrue(dataNode.has(KafkaConnectorConfigParams.SNOWFLAKE_ROLE_NAME));
-
-    // Sensitive keys must NOT be present
+    // Allowlisted user-config keys must be copied through
+    assertTrue(dataNode.has(KafkaConnectorConfigParams.TOPICS));
```
sfc-gh-aminyaylov: Can you please make this more data-driven to remove the `assertTrue/False` boilerplate?

```
  List<String> allowedKeys =
      List.of(
          KafkaConnectorConfigParams.TOPICS,
          TASKS_MAX_CONFIG,
          KafkaConnectorConfigParams.VALUE_CONVERTER_SCHEMAS_ENABLE,
          KafkaConnectorConfigParams.ERRORS_TOLERANCE_CONFIG,
          KafkaConnectorConfigParams.ERRORS_LOG_ENABLE_CONFIG,
          KafkaConnectorConfigParams.ERRORS_DEAD_LETTER_QUEUE_TOPIC_NAME_CONFIG,
          KafkaConnectorConfigParams.BEHAVIOR_ON_NULL_VALUES,
          KafkaConnectorConfigParams.SNOWFLAKE_TOPICS2TABLE_MAP,
          KafkaConnectorConfigParams.SNOWFLAKE_METADATA_ALL,
          KafkaConnectorConfigParams.JMX_OPT,
          KafkaConnectorConfigParams.ENABLE_MDC_LOGGING_CONFIG,
          KafkaConnectorConfigParams.ENABLE_TASK_FAIL_ON_AUTHORIZATION_ERRORS,
          KafkaConnectorConfigParams.SNOWFLAKE_VALIDATION,
          KafkaConnectorConfigParams.CACHE_TABLE_EXISTS,
          KafkaConnectorConfigParams.CACHE_PIPE_EXISTS);
  List<String> omittedKeys =
      List.of(
          KafkaConnectorConfigParams.SNOWFLAKE_DATABASE_NAME,
          KafkaConnectorConfigParams.SNOWFLAKE_SCHEMA_NAME,
          KafkaConnectorConfigParams.SNOWFLAKE_URL_NAME,
          KafkaConnectorConfigParams.SNOWFLAKE_ROLE_NAME,
          KafkaConnectorConfigParams.SNOWFLAKE_USER_NAME,
          KafkaConnectorConfigParams.SNOWFLAKE_PRIVATE_KEY,
          KafkaConnectorConfigParams.SNOWFLAKE_PRIVATE_KEY_PASSPHRASE,
          KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_CLIENT_SECRET,
          KafkaConnectorConfigParams.SNOWFLAKE_OAUTH_REFRESH_TOKEN,
          KafkaConnectorConfigParams.SNOWFLAKE_JDBC_MAP,
          "ssl.keystore.password");
  for (String key : allowedKeys) {
    assertTrue(dataNode.has(key), () -> "expected allowlisted key to be copied through: " + key);
  }
  for (String key : omittedKeys) {
    assertFalse(dataNode.has(key), () -> "expected sensitive/untrusted key to be omitted: " + key);
  }
```

sfc-gh-aschuck:
Superseded by in-repo PR https://github.com/snowflakedb/snowflake-kafka-connector/pull/1567 so CI can receive org secrets. Closing this fork PR.

